PR #22, opened 2016-01-25
Moving from comfychair to py.test.

andy-maier:
This pull request uses `pytest` for all unit test cases, and drops use of `comfychair`. the transition has been done in such a way that it is mainly compatible with `unittest`, and `pytest` specifics are used only where needed.

I have looked at `nose` and `nose2` and have found that it is not really actively worked on, and that some projects are moving on from nose/nose2 to pytest.

This is a first step towards using standard Python tools for testing.
The next step (not yet in in this pull request) is to use `tox`, instead of setting up virtual environments manually.

Please review and post any comments here.


KSchopmeyer:
+1

andy-maier:
Karl and I discussed it and decided to merge this one.

PR #31, opened 2016-02-08
pylint changes to tupleparser

KSchopmeyer:
Cleaned up tupleparser so only a few issues remain in pylint output.  Note that I cleaned up a number
of things that andy actually disabled for now in pylint.rc so there are a lot of changes.

Issues in current pylint.log after chages are:
W:405, 0: TODO: Currently disabled, figure out how to reenable (fixme)
W:561, 0: TODO: Verify these errno numbers on Windows vs. Linux (fixme)
W:243,16: Access to a protected member _path of a client class (protected-access)
W:246,12: Access to a protected member _path of a client class (protected-access)
W:247,11: Access to a protected member _path of a client class (protected-access)
W:250, 0: Dangerous default value [] as argument (dangerous-default-value)
E:335,15: Instance of 'HTTPBaseConnection' has no 'sock' member (no-member)
E:336,19: Instance of 'HTTPBaseConnection' has no 'auto_open' member (no-member)
E:337,20: Instance of 'HTTPBaseConnection' has no 'connect' member (no-member)
E:340,15: Instance of 'HTTPBaseConnection' has no 'debuglevel' member (no-member)
E:342,12: Instance of 'HTTPBaseConnection' has no 'sock' member (no-member)
W:405,16: Using a conditional statement with a constant value (using-constant-test)
R:394,16: Redefinition of self.sock type from socket._socketobject to M2Crypto.SSL.Connection.Connection (redefined-variable-type)
E:522,34: Non-iterable value local_auth_header is used in an iterating context (not-an-iterable)
W:536,29: Used builtin function 'map'. Using a list comprehension can be clearer. (bad-builtin)
W:536,29: map/filter on lambda could be replaced by comprehension (deprecated-lambda)
W:536,33: Lambda may not be necessary (unnecessary-lambda)

### DIFF
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -21,6 +21,7 @@
 # Author: Bart Whiteley <bwhiteley@suse.de>
 #
 
+# pylint: disable=too-many-lines
 '''Tuple parser for the XML schema representing CIM messages.
 
 This framework is meant to add some value to the tuple-tree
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1 voting to commit this one

KSchopmeyer:
Note that there are several epydoc warnings now, primarily unexpected indent from where we
included the XSD fragment that is the target for a function into the doc string.  This is very inconsistent
in that in many places, the fragments are just comments and other places part of doc string but
these generally methods not included in the __all__

andy-maier:
+1

PR #36, opened 2016-02-09
pylint cleanup of wbemcli, test_cim_http, test_cim_xml

KSchopmeyer:
Cleaned up wbemcli to better understand it.  Note that this is really incomplete in that the doc strings are minimal but there are other issues behind that.  Primarily pylints and docstrings.

Cleaned up the test_* simply because there were so many trivial issues, primarly formatting, etc. No code changes at all. I did pylint disable invalid_names which seemed logical since a lot of the variables are single character and these are test programs.

### DIFF
--- a/pywbem/wbemcli.py
+++ b/pywbem/wbemcli.py
@@ -68,10 +68,15 @@ import pywbem
 
 __all__ = []
 
+# Connection global variable. Set by remote_connection and use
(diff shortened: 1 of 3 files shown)


andy-maier:
Did not do a deep review, but looks good to me.
Also contains some docstring and comment improvements.

Issue #41, opened 2016-02-11
Test failure in make all does not stop tests


KSchopmeyer:
We need to sort out what the rules are but right now, a failure in testing (I had one that was failing in the cimxml test with an exception that was my problem) did not stop the make.  You needed to sort through the result in detail to determine if there was an error.

This may be the way we want to run in some conditions but in general, for local testing I think it would be better to stop the test at the first failure.

This is open for comment.

KSchopmeyer:
The test.log sort of works but it is easy for a tester to ignore errors when the tests go to completion.

andy-maier:
Given that the intentional exceptions in the test cases that ahppen to be reported, do not count, my take is that PR #43 solves this issue.

Issue #21, opened 2016-01-21
Exceptions during test_client on Ubuntu


andy-maier:
On the Travis CI system (Ubuntu 12.04) and on Ubuntu 14.04, the test_client unit tests report exceptions, but pass. On RHEL 6.7, they pass without showing exceptions.

Analyze what is happening, and fix, if needed.

The exceptions are (from Travis CI):
(Note that the report line for GetInstance1 gets interrupted by the exceptions)

    Processing YAML file: getinstance.yaml
    Processing test case: GetInstance1: GetInstance sucException in thread Thread-96:
    Traceback (most recent call last):
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 532, in __bootstrap_inner
        self.run()
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 484, in run
        self.__target(*self.__args, **self.__kwargs)
      File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/httpretty/core.py", line 637, in fill_filekind
        headers
      File "test_client.py", line 121, in socket_104
        raise socket.error(104, "Connection reset by peer.")
    error: [Errno 104] Connection reset by peer.
    Exception in thread Thread-98:
    Traceback (most recent call last):
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 532, in __bootstrap_inner
        self.run()
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 484, in run
        self.__target(*self.__args, **self.__kwargs)
      File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/httpretty/core.py", line 637, in fill_filekind
        headers
      File "test_client.py", line 126, in socket_32
        raise socket.error(32, "Broken pipe.")
    error: [Errno 32] Broken pipe.
    Exception in thread Thread-100:
    Traceback (most recent call last):
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 532, in __bootstrap_inner
        self.run()
      File "/opt/python/2.6.9/lib/python2.6/threading.py", line 484, in run
        self.__target(*self.__args, **self.__kwargs)
      File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/httpretty/core.py", line 637, in fill_filekind
        headers
      File "test_client.py", line 116, in socket_ssl
        raise socket.sslerror(1, "Arbitrary SSL error.")
[...]

PR #43, opened 2016-02-13
Fixes and improvements in makefile

andy-maier:
This addresses issue #41 and some other issues.

Details, from the commit message:
- Added dependents for Lex/Yacc table files (mofparsetab.py/moflextab.py).
- Made sure that mofparsetab.py and moflextab.py are generated in the right
  directory (pywbem package directory) even when mof_compiler.py is invoked
  standalone.
- Bad return codes from py.test and epydoc are now surfaced as make errors
  (solves issue #41).
- Changed meaning of 'clean' target to be cleanup before other make targets.
- As a result, 'clean' target now removes anyhting that is dependent on
  Python version but has same file name.
- test.log now has a Python version dependent name (e.g. test_27.log).
- Made 'all' target consistent with tox and Travis invocations (i.e. removed
  build+install, because that is done by the 'develop' target already, and
  moved 'clean' target to the front).

### DIFF
--- a/makefile
+++ b/makefile
@@ -5,8 +5,18 @@
 #   Windows
 #   Linux
 #
+# Basic prerequisites for running this makefile:
+#   bash, sh
+#   sudo
+#   zip, unzip
+#   rm, find, xargs, grep, sed
+#   python (Some active Python version, virtualenv is supported)
+#   pip (in the active Python environment)
+#
+# Additional prerequisites for development and for running some parts of this
+# makefile will be installed by 'make develop'.
+#
 # Prerequisites for usage will be installed by 'make install'.
-# Additional prerequisites for development will be installed by 'make develop'.
 # ------------------------------------------------------------------------------
 
 # Name of this Python package
@@ -21,6 +31,9 @@ package_version := $(shell sh -c "echo $(package_specified_version) |sed 's/[.-]
 # Final version of this package (M.N.U)
 package_final_version := $(shell sh -c "echo $(package_version) |sed 's/rc[0-9]\+$$//' |sed 's/\.dev0$$//'")
 
+# Python version for use in file names
+python_version_fn := $(shell python -c "import sys; sys.stdout.write('%s%s'%(sys.version_info[0],sys.version_info[1]))")
+
 # Directory for the generated distribution files
 dist_dir := dist/$(package_name)-$(package_final_version)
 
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
Let me look this one over for an hour or so.  My only real question is on the level of dependents for lex and yacc table files.

andy-maier:
On the dependents: I added all modules that are used by mof_compiler, on the argument that if they are used, they could influence the output. So the table files are now re-built more often.

KSchopmeyer:
Not sure if necesssary but it is certainly safer.  Would have to dig into all dependencies in mof_compiler to be sure.

andy-maier:
I have added all modules that mof_compiler uses, recursively.

andy-maier:
Rebased it on master.

andy-maier:
+1

KSchopmeyer:
I sent andy some questions that appeared to me but it was after the commit. Primarily why the heavy detail output for tests now and second, we are getting an exception on a test but it may be just that we are seeing it now because of the detail test processing output.

KSchopmeyer:
Whoops.  Error in the part of comment above about commit.  It is not committed.  I sent andy a detailed display of the test output in the new branch including the test that generated the exceptions

KSchopmeyer:
Other than the question about the change to test output level and understand if the failure shown is new or just more visible I am ready to approve this pull request

andy-maier:
* On the heavy detail output for tests: I added the `-s` option to `py.test` which causes stdout/stderr of the unit test modules to be displayed instead of being captured and lost. The only test module that produces output is test_client, which has only one testcase as known to py.test, so its report about yaml files and testcases within those gives some more information. Also, any debug prints added to the test cases are shown that way. We can turn that back by simply removing the `-s` option again. Let me know.
* On the exceptions that are shown: The ones I can see are the exceptions that result from the testcases intentionally raising them. See also issue #21 describing these same exceptions. I do not yet understand why that is shown by py.test, when they are expected. Maybe it has to do with multi-threading. For now, I think their display does not hurt, and the tests that have them rightfully count as passing.

KSchopmeyer:
+1  approved
Lets leave the -s option in for now since it gives us more info. At some point it should probably become a formal makefile variable (verbose) since once this settles down not everybody will want to see the details but we can leave that for the future.

Yes. It is strange in that the exceptions actually tear output apart (inserted right in the middle of the word succeed in the output terminal indicating that a it is a threading thing, b)somewhere we are not providing enough multithread control on output.

andy-maier:
Ok. Merging it.

PR #46, opened 2016-02-13
Added support for Python 3, and some more fixes.

andy-maier:
This PR adds support for Python 3, and fixes some more functional issues that came up during that.

Please review (and test on your favorite distro using your favorite version of Python 3).

Kudos go to Ross Peoples, who did the initial changes on the way to support Python 3. See https://github.com/deejross/python3-pywbem.

* Changes that were caused directly by Python 3 rules:
      - Changed print statements to `print()` function.
      - Instead of using backticks for print arguments, used `%s` or `%r`, as needed.
      - Instead of using global functions of `string` module, using methods on string objects instead (e.g. `join()`, `split()`).
      - Changed use of `has_key()` on dicts to using `in`, because dicts in Python 3 don't have `has_key()` anymore.
      - Removed `has_key()` from `NoCaseDict` for the same reason. Note that `NoCaseDict` was never meant to be a class useable by consumers of pywbem. CIM object classes still have `has_key()`, for compatibility.
      - Changed use of comma notation for `except` clause to use of `as`.
      - Replaced all uses of map+lambda with list comprehensions.
      - Made reference to names in pywbem namespace explicit, because implicit relative module search has been removed in Python 3.
      - Added explicit declaration and initialization of class variable `ownerDocument` in class `CIMElement`, because in Python 3, class variables are no longer implicitly created on first assignment.
      - Changed CIM object types from implementing `__cmp__()` to implementing rich comparison methods (i.e. `__eq__()`). Added a mixin class `_CIMComparisonMixin` that implements the rich comparison functions based on a `_cmp()` function provided by the CIM object classes. This is possible because the CIM object classes can (and do) implement total ordering.
      - Changed use of `<>` to use `!=` instead.
      - Instead of using print with trailing comma to avoid trailing newline, used explicit `end` argument.
      - In loops on literally defined lists, made the list explicit instead of relying on implicit conversion (e.g. `for x in 1,2` became `for x in [1,2]`).
      - Changed literal long values to be added dynamically only for Python 2 (this affected some testcases in `test_cim_obj.py`).
      - In `test_client.py`, removed the `encoding` attribute in a possibly present XML declaration, before it is compared for equality, because the `etree.XMLParser` in Python 3 rejects XML with an `encoding` attribute when passed as a Unicode string.
* Use of the `six` module, for common code between Python 2 and 3:
      - Changed `isinstance()` tests on `StringTypes`, `basestring`, `str`, and `unicode` to
        `six.string_types`, `six.text_type`, `six.binary_type`.
      - Changed to six-provided portable verisons of: `next()`, `urllib`, `urlopen()`,
        `httplib`, ...
    
* Changes in build and test environment:
      - In `setup.py`, made `develop_requires` dependencies on Python header files
        release-specific.
      - Made test commands in tox and Travis consistent with make all.
      - Added Python 3.4 and 3.5 to test environments.
    
* Cleanup:
      - In `pywbem` package, used relative imports for sibling modules.
      - In `testsuite` directory, used imports referencing the pywbem package for any pywbem modules.
      - In `cim_obj`, added `_ensure_bytes()`, `_ensure_unicode()`, `_convert_unicode()`, `_convert_bytes()` to encapsulate conversions between Unicode strings and UTF-8 encoded byte strings.
    
* Functional changes:
      - For Python 3, using built-in SSL module again, instead of M2Crypto.
      - Added explicit `u` and `b` prefixes to literal strings, to qualify them as binary or Unicode in both Python 2 and 3.
      - As a consequence of adding these prefixes, made some regular expressions used in test modules tolerant against their presence.
      - In imethodcall() and methodcall() of cim_operations, fixed incorrect ignoring of typical Python exceptions such as TypeError. For now, they are just re-raised, and a TODO was added to handle that better.
      - In imethodcall() and methodcall() of cim_operations, added a check to prevent an IndexError during handling of ExpatError that was caused by an incorrect line number in the ExpatError exception data.
      - Using argparse.ArgumentParser instead of optparse.OptionParser, if available (i.e. for Python 2.7 and 3.x).
      - Added some forgotten conversions between byte strings and Unicode strings that went by undetected in Python 2.
      - Documented the interface to the httpretty callback functions in the 'Callback' class in test_client.py.
      - Improved some error messages in test_client.py.
      - In invalidresponseerror.yaml, removed testcases InvalidResponse33 and InvalidResponse34 because they use illegal Unicode characters that are 'half' surrogates. This was tolerated in Python 2, and is no longer tolerated in Python 3. A WBEM server of course still could return such illegal characters, so this reduces the test coverage at the moment. Created an according issue to address this in the future.
      - In networkerror.yaml, removed the testcases InvalidHost and HostIgnoresPost; they were attempting to test a timeout in a way that did not work. Created an according issue to address this in the future.


andy-maier:
Rebased it on master.

andy-maier:
I added two more commits to this PR:
* 296bc24 was needed because `make develop` ran `sudo python setup.py develop_os` first, which created the `__pycache__` directory under root owner/group, causing subsequent runs without sudo to fail. A similar issue existed with the `six` package.
* 58de645 was needed because Epydoc and Pylint apparently do not support running under Python 3. Because the mof table modules are now build more often, they are sometimes re-built under Python 3, even after a run under Python 2 has already generated them. This re-run then triggered a re-build of the documentation under Python 3. Same issue for Pylint.

andy-maier:
One more commit:
* edf3ad6 was needed mainly because the package names for the Python development package did not work. This caused a number of other things to be fixed in turn. For details, see the commit message.

KSchopmeyer:
I notice that at least on some of the python 3 tests it makes MANY passes through the LALR build sequence, first in the beginning of the test and then as much as 5+ passes at the end of the test.  I think I was looking at 3.5 test output and notice this.

andy-maier:
Rebased on master.

andy-maier:
As of yesterday, the python3 branch runs fine against all of my CI environments.
+1

andy-maier:
Karl votes +1 as well in our call today.

Issue #33, opened 2016-02-09
wbemcli has unused input arg for EnumerateInstances


KSchopmeyer:
on the EnumerateInstances function the optional input argument LocalOnly is in the _init_ but never used.  Need to sort out if this is an issue with the function in wbmcli or the called function. Note that, in general, this argument is deprecated.

andy-maier:
Good catch, Karl.
It is missing. (I know LocalOnly is deprecated in the CIM-XML standard, but that does not mean it went away, and we are supporting it in PyWBEM).

Also, please note the different way the arguments of most other functions in `wbemcli.py` are defined: Generically with `*args, **kwargs`. It seems to me that we could change `EnumerateInstances()`, `EnumerateInstanceNames()`, and `GetInstance()` to also be generically defined, and that that would be a backwards compatible change.

andy-maier:
PR #48 was merged. Closing this issue

PR #48, opened 2016-02-13
Added missing LocalOnly parameter to EnumerateInstances() of wbemcli.

andy-maier:
This fixes issue #33.


andy-maier:
+1

KSchopmeyer:
+1

My only concern was to be sure that this was carried through the whole process. Since most of the
request flow is actually blind to actual arguments and there are several cases of localonly in the
testsuite I am happy.  Thanks for doing this for me.

PR #50, opened 2016-02-15
Changed M2Crypto install source to master branch of its upstream repo.

andy-maier:
As of today, the master branch of the [upstream M2Crypto repo](https://gitlab.com/m2crypto/m2crypto) has all fixes that we had in our [pywbem/m2crypto fork](https://github.com/pywbem/m2crypto), so the next version of M2Crypto after its current 0.23.0 will have them all.

The change in this PR changes PyWBEM's `setup.py` to install M2Crypto from the master branch of its upstream repo, instead of using our own fork.

If the official next version of M2Crypto gets released before PyWBEM 0.8.0, we should change that again to install that version from PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -319,18 +319,25 @@ def main():
             'pywbem/wbemcli.py',
             'pywbem/mof_compiler.py',
         ],
-        # TODO: Finalize this temporary fix: Use our own fork of M2Crypto with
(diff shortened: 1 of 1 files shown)


andy-maier:
Also works on my mini-CI at home (Distros: Arch, Ubuntu 14, Debian 8, CentOS 6+7, Fedora 22).
+1

PR #51, opened 2016-02-16
Fixpylintrc so usable with geany editor 

KSchopmeyer:
Set output-format parseable, commented out template and made note about requirements of this editor. Not absolutely required but probably at least leave out the template.  alternative would be for each editor user to override output-format and set template if required.


KSchopmeyer:
+1

andy-maier:
Karl,
When using this fix, I get the pylint message:

    /home/maiera/virtualenvs/pywbem27/local/lib/python2.7/site-packages/pylint/reporters/text.py:79:
    UserWarning: parseable output format is deprecated. This is equivalent to
    --msg-template={path}:{line}: [{msg_id}({symbol}), {obj}] {msg}
    'to --msg-template=%s' % (self.name, self.line_format))

This is probably due to the change of the `output-format` parameter (from `text` to `parseable`).
This is with:

    pylint 1.3.1
    astroid 1.2.1, common 1.1.0
    Python 2.7.6 (default, Jun 22 2015, 17:58:13)

andy-maier:
One more question, Karl:
* How does one use the Pylint report with Geany? Does it require some plugin?

KSchopmeyer:
This is strange in that I use pylint 1.5.1 and do not get the same message.  However, I have no problem backing off that but would like to keep the other change, commenting out the template.

To use geany with pylint,  and created a new build command.  I sent you the config file that creates.

KSchopmeyer:
+1 voting yes on this latest fix.  The only thing this does now is to comment the template line since that is messy to do in the geany plugin.

andy-maier:
+1 for this latest fix, as well.

PR #53, opened 2016-02-17
Fixes #52: The install_os and develop_os setup commands prompt for password

### LINKED ISSUE #52: The install_os and develop_os setup commands prompt for password
On Linux, even if the userid is set up for running `sudo` without password prompt, using the `install_os` or `develop_os` commands of `setup.py` will prompt for a password on some distributions (e.g. CentOS 7.1, Fedora 22, Debian 8.2).

Investigation shows that this is likely caused by the use of `sudo -v` in the `OSInstaller.authorized()` method in `os_setup.py`. The intention was to just determine whether the current user is authorized for `sudo`, but its `-v` option seems to behave specially w.r.t. password prompting.

*Proposed solution:*
Change `sudo -v` to executing some no-op command with `sudo' without using the `-v` option.

andy-maier:


andy-maier:
Merging this to enable interruption-free testing in my mini-CI at home.

Issue #26, opened 2016-02-06
Presence of pbr package in py26 causes 'TypeError: dist must be a Distribution instance'


andy-maier:
The Travis CI run of the m2crypto_rebased branch on Python 2.6 shows an exception 'TypeError: dist must be a Distribution instance' on. For example at the end of this log: https://travis-ci.org/pywbem/pywbem/jobs/107007558.

The exception does not happen on Python 2.7 with this branch, and it does not happen with the master branch (on any of the two Python versions).

It seems the presence of the Python "pbr" package also influences the occurrence of the exception. On Travis CI, pbr 1.8.1 is installed by default. If it gets removed, the error disappears.
On a local system (forgot which one), I verified this and had the same effect.

PyWBEM or its dependencies do not need the pbr package, but that is not a solution, because the pbr package enjoys broad use.

There are various reports on this error on the Internet: One patch that appeared to be promising is https://bugs.launchpad.net/pbr/+bug/1324784, which made it into pbr. I verified that pbr 1.8.1 has the code of the patch, so that patch did not solve this problem.

One problem with pbr is that it monkey patches setuptools, and there seem to be dependencies of the behavior on the order in which things get done, which make it difficult to debug or even reproduce the problem. At least that was the case before the patch for bug 1324784; not sure whether these dependencies have been eliminated by the patch.

This needs to be further analyzed and resolved.

andy-maier:
PR #67 removes the pbr package in Travis CI runs only for Python 2.6, in order to verify that its presence in the Travis CI runs for Python 2.7, 3.4 and 3.5 is not an issue.

andy-maier:
PR #857 (starting to use pbr) for some reason solved this issue. It no longer shows up on py26, and the uninstall script could be removed. Considering this one solved.

PR #67, opened 2016-02-29
Restrict removal of 'pbr' package during Travis CI runs to Python 2.6 only

andy-maier:
Currently, the 'pbr' package is uninstalled during Travis CI runs for all Python versions, in order to not get into the issue described in #26.

However, the issue should only occur on Python 2.6. This PR therefore performs the uninstall of the 'pbr' package only for Python 2.6, in order to verify that there is really no issue on the other Python versions.


andy-maier:
+1

andy-maier:
+1

KSchopmeyer:
+1

PR #68, opened 2016-02-29
Fixes #8: Use lex+yacc from ply package; Fixes #55: Avoid rebuilds of MOF compiler tables

### LINKED ISSUE #8: The version of PLY used in pywbem is about 7 years and 8 versions old
pywbem uses version 3.0 of this code and version 3.8 was released in October 2015.
We should at least explore the changes and issues with moving pywbem to the newer version.

### LINKED ISSUE #55: Rebuilds MOF parser tables when it shouldn't
Branch Python 3 causes a rebuild of the mof and lex table files.  When compiling with python 3 it rebuilds the tables into the table python files multiple times at the end of the build and even in 2.7
it rebuilds the tables when the table files are dated later than the files from which they originate.

This may result from the extra makefile dependencies we inserted.

andy-maier:
This PR removes the very old copies of `lex.py` and `yacc.py` from the pywbem repo, and uses them from the `ply` package, instead. The setup script is updated in order to specify the additional dependency on the `ply` package.

In addition, the PR changes the approach of keeping the table modules generated by LEX+YACC in the git repo. Because the generated tables contain the LEX+YACC versions, they now need to be generated using whatever version the pywbem user has installed. They are now generated during the 'install' command of the PyWBEM setup script, into the directory of the `pywbem.mof_compiler` module, and are installed along with the other pywbem modules.

In addition, the PR removes the backup solution for the directory into which the MOF Compiler table files are generated. They are now always generated into the directory of the `mof_compiler` module that executes. The backup solution was to generate into the current directory, if the `__file__` attribute does not exist on the module object. However, that attribute exists on all Python versions, so a backup solution seems to just complicate things.

In addition, the PR makes the invocation of `lex.lex()` and `yacc.yacc()` consistent between the `_build()` function and their invocations in the `MOFCompiler` constructor (they were not consistent before, as far as error messages and debug log are concerned).

In addition, the PR changes the error messages of LEX and YACC to go to stdout instead of stderr. The generation of the YACC debug log file (`parser.out`) was disabled.

Some more fixes in this PR (unrelated to LEX/YACC but partly necessary to test this PR successfully):
* Fixed an issue that was introduced in the Python 3 work in `pywbem/wbemcli.py`, by importing
  `argparse` instead of `optparse` if available, without doing the other changes needed for such
  a migration. The `optparse` package is now again used for all Python versions, and a ToDo was
  added to resume the migration to `argparse' lateron.
* Changed `pywbem/mof_compiler.py` and `pywbem/wbemcli.py` to be useable as modules again,
  by adding `mof_compiler` and `wbemcli` scripts that import these modules.
* Added hash bang to `setup.py` to make it executable as a command.

As a result of all of that, issue #55 is now resolved. This was verified with debug code in the `yacc` and `lex` modules of the `ply` package: The YACC and LEX table modules are now generated only once during PyWBEM install, and the generated modules are reused in subsequent test runs.

Here is a patch ([ply_debug.patch.txt](https://github.com/pywbem/pywbem/files/153154/ply_debug.patch.txt)) for the `ply`package (v3.8) that adds the debug code, and the log files of a fresh `setup.py install` run ([setup_install.log.txt](https://github.com/pywbem/pywbem/files/153155/setup_install.log.txt)) (i.e. after removing any pywbem artefacts) and a subsequent `make test` run ([make_test.log.txt](https://github.com/pywbem/pywbem/files/153156/make_test.log.txt)). Look for lines starting with `Debug:`.


andy-maier:
+1

andy-maier:
+1

KSchopmeyer:
+1  Karl will run a complete comparison of created entities later.

PR #72, opened 2016-03-01
Fixes #70: Define external API of the WBEM client library

### LINKED ISSUE #70: Definition of external API
The external API of the PyWBEM Client is only vaguely defined.
The documentation is currently generated for areas that are definitely not meant to be part of the external API, and the module wildcard export definitins in `__all__` also have too much in them.

andy-maier:
Defined the external API of the WBEM client library, by defining the `__all__` variables of each sub-module accordingly. The API documentation now only shows the external API.


andy-maier:
+1

PR #73, opened 2016-03-01
Fixes #71: Two classes ParseError defined

### LINKED ISSUE #71: Two classes ParseError defined
There are two classes named `ParseError` defined currently, and both can surface at the external API:
* In `pywbem/tupleparse.py`. This class is part of the external API and is for example imported into `cim_operations.py`.
* In `pywbem/cimxml_parse.py`. This class is not used outside of that module and is not "exported" from that module via `__all__`.

The proposed solution is to remove class `ParseError` from `cimxml_parse.py`, and to import the class from `tupleparse.py`. Be aware of dependency circles.

andy-maier:
Replaced the definition of class `ParseError` in sub-module `cimxml_parse.py` by an import of `ParseError` from `tupleparse`.

### DIFF
--- a/pywbem/cimxml_parse.py
+++ b/pywbem/cimxml_parse.py
@@ -35,16 +35,17 @@
 from xml.dom import pulldom
 import six
 
-from pywbem import cim_obj
-from pywbem.cim_obj import CIMInstance, CIMInstanceName, CIMQualifier, \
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
Good catch +1

andy-maier:
+1

Issue #75, opened 2016-03-02
Add mock (test_client) end-end tests for the class and qualifier type operations


KSchopmeyer:
These operations need end-end tests added. Today their are none.

KSchopmeyer:
Done in earlier Pr and this one got left in place. the following tests exist

getclass
getqualifier
enumerateclasses
enumerateclassnames
enumeratequalifiers

Lets close this one.

andy-maier:
Agreed. Closed.

PR #76, opened 2016-03-02
Remove the unsupported comments issue #75

KSchopmeyer:
Remove the unsupported statements in the docstrings for the class and qualifier operations in cim_operations.py.  While there are no end-end tests, general testing is sufficient to assure that they do generally work.


KSchopmeyer:
+1

andy-maier:
+1

PR #79, opened 2016-03-11
Changes to test_client.py for list in result: and yaml for enum insts.

KSchopmeyer:
This change modifies test_client to allow lists for the result: yaml so that we can list multiple instances for the response.

Then it adds enumerateinstances.yaml as a new test.

Passes tests with python 27 and 3

Note that as with other of these tests:
1. The data is somewhat artifical, in particular on responses and in some cases simpler than what an
actual server returns.
2. We are testing only a subset of the current possible alternatives.

But this does test the basic paths of enumerateInstances with both multiple, single, and no instance returns and a number of the possible error paths.

Needs:
1. More options on the in put parameters, in particular the propertylist
2. more of the possible data in the responses (ex. include class origin, etc.)

Changed the overall coverage by about +1%

### DIFF
--- a/testsuite/test_client.py
+++ b/testsuite/test_client.py
@@ -16,7 +16,7 @@ Approach:
 * The testcases then performs the request at the PyWBEM client API, and
   verifies the response also at the level of the PyWBEM client API.
 
-* HTTPretty remembers the lest request that was sent, so after issuing the
+* HTTPretty remembers the last request that was sent, so after issuing the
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Issue defined in issue 61. This is partial response to that issue. Other tests must be added for
associators, references

KSchopmeyer:
+1 since this passed all tests

andy-maier:
+1 after reviewing the code

Issue #80, opened 2016-03-11
the mof compiler test uses a very old version of the dmtf mof schema.


KSchopmeyer:
The test currently uses cim schema 2.20.  Today cim is up to 2.45 so we are not really testing against
current dmtf mof.

andy-maier:
Fixed with PR #81.

PR #81, opened 2016-03-11
Modifications to compiler test to use later version of schema, issue #80

KSchopmeyer:
Issue #80 notes that we are using a very old version of the cim schema as part of our tests.

This pull requests updates the cim schema to be downloaded to the current version of the schema
and also changes it from the experimental to the final version since there is no real guarantee that the experimental compiles.

Note that it makes a couple of other changes to eliminate pylint messages in imports, constant names, and indenting.


KSchopmeyer:
+1 passed tests with the fix.

andy-maier:
+1

PR #83, opened 2016-03-12
Added tool pycmp.py to compare public interface between versions

andy-maier:
This PR adds a tool `pycmp.py` in a new `tools` directory, which compares two versions of PyWBEM w.r.t. their public interfaces. It generates a report on stdout. Invoke with --help for usage help.

It also adds a shell script `tools/pycmp_0.7.0.sh` which creates the files from PyWBEM 0.7.0 in a temp directory, removes setup.py (not part of the package) and wbemcli.py (cannot be imported in 0.7.0), and invokes `pycmp.py` to produce a report in the current directory. The shell script must be invoked in the root directory of the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ swig-2.0*.tar.gz
 /setupswig-2.0*/
 setuptools-*.egg
 setuptools-*.zip
+/tmp_*/
--- a/makefile
+++ b/makefile
@@ -187,6 +187,7 @@ clobber: clean
 clean:
 	find . -name "*.pyc" -delete
 	sh -c "find . -name \"__pycache__\" |xargs rm -Rf"
+	sh -c "ls -d tmp_* |xargs rm -Rf"
 	rm -f MANIFEST parser.out .coverage $(package_name)/parser.out $(test_tmp_file) $(package_name)/mofparsetab.py $(package_name)/moflextab.py
 	rm -Rf build tmp_install testtmp testsuite/testtmp .cache $(package_name).egg-info
 	@echo '$@ done.'
--- a/tools/pycmp.py
+++ b/tools/pycmp.py
@@ -1,0 +1,381 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import sys
+import argparse
+import os
+import fnmatch
+import importlib
+import re
+import types
+from pprint import pprint
+
+ARGS = None
+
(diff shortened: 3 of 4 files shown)


KSchopmeyer:
I have not figured out how to use this since it generated very little for me.

KSchopmeyer:
Since it breaks nothing, we should probably check it in but probably needs a readme

andy-maier:
The last commit improves the help text of the Python script and adds checks to the shell script.

KSchopmeyer:
-1  There is still something I am not catching here:

(pywbem27)kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem$ tools/pycmp_0.7.0.sh Generating report about symbol changes since PyWBEM 0.7.0
  File "tools/pycmp.py", line 406, in <module>
    sys.exit(main())
  File "tools/pycmp.py", line 65, in main
    rc = compare_package_dirs(ARGS.olddir, ARGS.newdir)
  File "tools/pycmp.py", line 90, in compare_package_dirs
    old_modinfos = get_modinfos(old_pkgdir)
  File "tools/pycmp.py", line 245, in get_modinfos
    module_import(abs_modname)
  File "tools/pycmp.py", line 357, in module_import
    exec(_wci_stmt)
  File "<string>", line 1, in <module>
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/tmp_pycmp_0.7.0/pywbem/twisted_client.py", line 28, in <module>
    from twisted.internet import reactor, protocol, defer
ImportError: No module named twisted.internet
Success: Generated report: pycmp_0.7.0.log
(pywbem27)kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem$ cat pycmp_0.7.0.log 
Comparing public symbols of Python package: pywbem
    Old package directory: tmp_pycmp_0.7.0/pywbem
    New package directory: pywbem

This report covers the following kinds of symbols:
    Non-public symbols:                                                      False
    Symbols of added or removed modules (-f option):                         False
    Symbols imported from other packages (-o option):                        False
    Symbols imported from other modules of this package (-i option):         False
    Symbols exported from this package (in __all__) (-e option):             True
(pywbem27)kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem$

andy-maier:
Merged by accident.
Will fix this in a separate PR.

PR #85, opened 2016-03-12
Fixes #77: Removed unnecessary functions in cim_obj/cim_types from external API

### LINKED ISSUE #77: Full compatibility of public symbols of WBEM client library with v0.7.0
We moved things around in an attempt to cleanup, and reduced the official external API to remove internal stuff. However, that may break users of v0.7.0 when they move to v0.8.0.

Proposal:
* Move the definitions to those sub-modules where it makes sense (e.g. ParseError).
* Import the symbols into those modules they have been in v0.7.0, for compatibility.

This results in an external API definition that is more cleaned up than currently, and it results in compatibility for v0.7.0 users.

andy-maier:
The following functions have been removed from the external API (i.e. they have been removed from `__all__`):

    pywbem.cim_obj.tocimobj
    pywbem.cim_obj.tocimxml
    pywbem.cim_types.atomic_to_cim_xml
    pywbem.cim_types.cimtype
    pywbem.cim_types.type_from_name

None of them should be needed by normal client code. For example, client code should not have to deal with CIM-XML representations.

Note that most CIM object classes have a `tomof()` method.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -51,7 +51,7 @@ from .cim_types import _CIMComparisonMixin, type_from_name, \
         
 __all__ = ['CIMClassName', 'CIMProperty', 'CIMInstanceName', 'CIMInstance',
            'CIMClass', 'CIMMethod', 'CIMParameter', 'CIMQualifier',
-           'CIMQualifierDeclaration', 'tocimxml', 'tocimobj']
+           'CIMQualifierDeclaration']
 
 # pylint: disable=too-many-lines
 class NocaseDict(object):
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
-1

I think that tocimxml or some replacement at least should be part of the public api in that this serves as a good display tool.  The 3 display tools for cim objects are 1. a tostring b. tomof, c. xml and we found with other cli tools that the users actually use all of them.

I agree with the others.

andy-maier:
Karl, I agree to leave the module-global `tocimxml()` function in. I thought it is redundant with the `tocimxml()` methods we have on every CIM object class, but I did not notice that the module-global function is needed for built-in types we use, such as int, or string types, or lists of objects.

andy-maier:
The last commit adds `tocimxml()` and `tocimobj()` back to external API. The argument for `tocimobj()` is that it may be useful if Python values need to be converted into the CIM object types used by PyWBEM.

Please re-vote.

andy-maier:
+1

andy-maier:
+1

KSchopmeyer:
Agree with the tocimobj and was going to argue for that one also.

+1

PR #88, opened 2016-03-13
Fixes #87: Fix run_operations.py issues

### LINKED ISSUE #87: run_cim_operations.py badly broken
Trying to test pywbem using this python testsuite code indicates that it is broken in several areas including

1. The startup hangs with some cmdline input
2. The exception code fails
3. Some tests incomplete
4. The log code is commented out because there is no log function.
5. The arg help defines options that are not in code.

I propose to fix enough to run tests before 0.8.0 release. Expanding tests can come later.

KSchopmeyer:
This fixes a number of issues with this code including:
1. exception handling syntax
2. The cmd line argument processing. Allso added new options.
3. A number of the tests
4. Lots of pylints.

However, I am not yet complete.  In particular, I think it will fail tests (the internal_test) with python 3 and there are issues with the classVerifier class and its verifiers.

Note that it does pass just about all of the tests in openpegasus which is a validation that most things work.  However, there are a number of unimplemented tests and the execquery fails for because we do not implement the query mechanism for classes in the repository.

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -16,9 +16,9 @@ from getpass import getpass
 import six
 
 from pywbem.cim_constants import *
-from pywbem import CIMInstance, CIMInstanceName, CIMClass, CIMProperty, \
-                   CIMQualifier, CIMQualifierDeclaration, CIMMethod, \
-                   WBEMConnection, CIMError, \
+from pywbem import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
+                   CIMProperty, CIMQualifier, CIMQualifierDeclaration, \
+                   CIMMethod, WBEMConnection, CIMError, \
                    Uint8, Uint16, Uint32, Uint64, \
                    Sint8, Sint16, Sint32, Sint64, \
                    Real32, Real64, CIMDateTime, ParseError
@@ -32,7 +32,7 @@ UNIMPLEMENTED = unittest.skip("test not implemented")
 TEST_CLASS = 'CIM_ComputerSystem'
 
 class ClientTest(unittest.TestCase):
-    """A base class that creates a pywbem.WBEMConnection for
+    """Base class that creates a pywbem.WBEMConnection for
     subclasses to use."""
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
The reason it passed is that the run... python module does not really run.  When I run it there is an error in python 3 with the check_Utf8...  internal_test.  Would like to fix that in the morning.

KSchopmeyer:
Still have one significant issue: the classverifier is completely broken.

andy-maier:
-1
I agree with everything, except that I would prefer to keep the decorator in all-lower case. That seems to be the convention; all-upper case is used for constant values, but not for every global definition.

andy-maier:
See discussion in issue #87.

KSchopmeyer:
The errors we are getting above are:

Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='pypi.python.org', port=443): Read timed out. (read timeout=15)",)': /pypi

/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:120: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.

  InsecurePlatformWarning

HTTP error 502 while getting https://pypi.python.org/pypi
error: None

make: *** [develop] Error 1

Looks like communication error with pypi but not sure how to redo the tests yet

KSchopmeyer:
Note: I only did the upcase UNIMPLEMENTED because pylint bitched at the lower case one.

andy-maier:
I still struggle with the upper case name for a decorator. If I think about it, the main issue I see is not so much the lexical case, but the fact that it is no longer clear from its name that this decorator causes the testcase to be skipped.

How about:

    UNIMPLEMENTED = "test not implemented"
    
    @skip(UNIMPLEMENTED)
    def test_xxx(self):

KSchopmeyer:
Sounds good to me.  Argh I just checked in fix for my GetQualifier booboo

KSchopmeyer:
-1 let me fix this.

KSchopmeyer:
+1. changes UNIMPLEMENTED and added tests for getinstance/enumerateinstances with properties.
Hopefully this is the final change and we can pull this

andy-maier:
+1

KSchopmeyer:
+1.  There are only 6 operations not in the test now (createclass, deleteclass, modifyclass, createqualifier, deletequalifier, executequery (execquery is tested).  Not all of the input parameters are being tested but we are picking up more and more.  Mostly it is the assocation options like role, etc. and the includclassorigin, etc. that are not tested.  Today this test is more through I think than the mock but we need to bring the mock into line more with this.

PR #92, opened 2016-03-15
Add test for functions in cim_operations.py module

KSchopmeyer:
This file currently adds only one new file and one test, the test on check_utf_8...
function in cim_operations.  This was originally in
run_cim_operations.py but is not really an end-end test so more logical
here in test_cim_operations.py to run under tests. Note that this
code bypasses a number of the tests with python3 since they fail with
message about finding an int. These tests run in python2 but fail in
python3

Put temporary bypass for bunch of tests with python 3. They fail but pass with python 2

### DIFF
--- a/testsuite/test_cim_operations.py
+++ b/testsuite/test_cim_operations.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+#
+
+"""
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

PR #97, opened 2016-03-16
Fixes #84: Moved DEFAULT_NAMESPACE from cim_operations to cim_constants module.

### LINKED ISSUE #84: why is DEFAULT_NAMESPACE defined in cim_operations.py
To me this is a core setting and finding it buried in one of the std components seems strange.  At the same time I have not found out where it would actually be logical yet

andy-maier:
Please review.

### DIFF
--- a/pywbem/NEWS
+++ b/pywbem/NEWS
@@ -1,6 +1,6 @@
 
 
-pywbem-0.8.0rc4  2016-03-01
+pywbem-0.8.0rc4  2016-03-16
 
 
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
With the comment below +1
I would reword the comment "the WBEM client library, and is supposed to be sufficient. If you find
    that you need something you were used to, please think twice as to"

to:
"If you find that you require symbols to make pywbem effective...", 

i.e. avoid the think twice.

Karl

PR #100, opened 2016-03-16
Fixes #95: Removed sudo test invocation in check for install authorization.

### LINKED ISSUE #95: in virtualenv when I move to python2 env that already has pkgs installed setup forces sudo
When I move between python3 and python2.7 envs, each time I go back to 2.7 it requires that i input
the sudo password.  This happens when I do make clobber && make all. This happens to
karl but not to Andy.

andy-maier:


KSchopmeyer:
+1  works for me.  Note that I ran the test against a bunch of virtual environments and they are all in
my workspace but no sudo requests with the root environment.

PR #109, opened 2016-03-18
Fixes #103: Embedded instance mof causes exception

### LINKED ISSUE #103: Problems with processing embedded Instances on input.
Started by will but I can duplicate. The embedded instance attribute is found, used, but for some reason the property is still a string when things like tomof is called.

Luckly I can create easily from openpegasus so this is easily solvable but I have gone blind at this point for the day.

Will upload some samples in the morning.

KSchopmeyer:
Any generation of tomof() for instance with included embedded instance causes exception. The code
did not account for properties that were embedded instances.

Fixed so it generates the mof for the indented instance. and a test for this case added to the unit tests


KSchopmeyer:
+1 passed all tests

andy-maier:
+1 based on rough code review.
Please resolve the merge conflict.

PR #114, opened 2016-03-19
Started develoment of version 0.8.2

andy-maier:
Updated the versions to `0.8.2dev`.


KSchopmeyer:
+1, Busy saturday for you.  Guess that is how we learn about this type of release.

Issue #113, opened 2016-03-19
pywbem 0.8.1 breaking Openstack Ansible builds


logan2211:
I'm seeing local failures trying to build 0.8.1 wheels: http://cdn.pasteraw.com/f3su2r6wanlc9feocqaf95btf370c33

The Openstack Ansible gate checks also seem to be affected:
http://logs.openstack.org/99/277199/16/check/gate-openstack-ansible-dsvm-commit/51de785/console.html#_2016-03-18_21_09_37_014
http://logs.openstack.org/65/294765/1/check/gate-openstack-ansible-dsvm-commit/03b96b6/console.html#_2016-03-18_19_50_33_660

Bug filed with OSA at:
https://bugs.launchpad.net/openstack-ansible/+bug/1559380

andy-maier:
I can see two issues:

1. The break of the OpenStack Ansible gate checks is caused by the dependency of pywbem on the "six" package, already during pywbem installation.

   Proposed solution: pywbem project eliminates the pre-dependency on "six". Is already under way.

2. The local build failures may be caused by the M2Crypto dependency being established in the build script after pywbem (see first line of the log you quoted). 

   Proposed solution: Please move M2Crypto before pywbem and let us know whether that solves it.

andy-maier:
I released 0.8.2 to PyPI. That version eliminates the dependency on the "six" package it had during its own installation.

Please try it out to see whether it resolves the issue in the OpenStack Ansible gate checks.

Please also let me know whether you can move the M2Crypto installation before the pywbem installation in your local build, and if so, whether that helps. Note that pywbem does state its dependency on M2Crypto in the 'install_requires' argument of its setup script.

Logan2211:
Hi Andy,

pip controls the ordering. OSA is building wheels using pip requirements files:
https://github.com/openstack/openstack-ansible/blob/liberty/playbooks/roles/repo_build/tasks/repo_build.yml#L33-L48
and an example requirements file: http://rpc-repo.rackspace.com/os-releases/12.0.5/requirements_local_filtered.txt

So OSA itself does not pull in m2crypto, but pywbem does.
From my build log:
Collecting M2Crypto (from pywbem->-r /var/www/repo/os-releases/12.0.9/requirements_local_filtered.txt (line 34))
  Downloading M2Crypto-0.23.0.tar.gz (183kB)

[...]

PR #115, opened 2016-03-19
Avoid dependency on six package during pywbem installation

andy-maier:
This PR fixes one of the two build break situations reported in issue #113.


KSchopmeyer:
+1

PR #119, opened 2016-03-20
Fixes #111: Moved unused pywbem modules into a new attic directory.

### LINKED ISSUE #111: cimxml_parse.py does not seem to be used anywhere except its own test program
It appears everything is done through tuple_parse and I do not understand the purpose for cimxml_parse.py

The question is:  is there a reason for it either now or in the future or should we eliminate it.

If it is significantly faster or more complete that tupleparse that would make sense.  Otherwise it might be best to kill it and its test.

andy-maier:
Moved the following unused files from the `pywbem` package directory into a new `attic` directory, in order to clean up the `pywbem` package:

* `cim_provider.py`
* `cim_provider2.py`
* `cimxml_parse.py`
* `twisted_client.py`


KSchopmeyer:
I was about to commit branch to move cimxml... but just passed the issue to you.

You also have to update one of the test files I did makefile to ignore attic for py.text

andy-maier:
Sorry for getting in between your work. I think it should be complete now.

KSchopmeyer:
+1 but we should have a short readme in the attic defining the purpose of that directory. Also would not help if we identified why we put particular things in that repository (i.e. not used but not sure would not be used in future, etc.)  Would give you what I wrote but I already deleted the branch

andy-maier:
Added a README file to the attic directory.

PR #123, opened 2016-03-21
Moved definition of package version into a single place: pywbem/_version.py

andy-maier:
This reduces the two definitions to one, in order to simplify the version definition (for this git repo; the pywbem.github.io repo will be dealt with separately).

The new place is a new file `pywbem/_version.py`, which gets:
- imported into by `pywbem/__init__.py` for pywbem users (a new import).
- read as Python code without importing by `setup.py` (new code in setup.py).

Already existing usages of the version, just to be complete:
- grepped for `__version__` by makefile, and then used in various ways, e.g. for distribution directory and file name (that existed before, and was just adjusted to use the new file).
- used by the Epydoc-built documentation (that existed before and did not needed to be adjusted).

### DIFF
--- a/makefile
+++ b/makefile
@@ -20,8 +20,8 @@
 # Name of this Python package
 package_name := pywbem
 
-# Package version as specified in pywbem/__init__.py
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
+1 Reviewed code but did not pull this one.

PR #124, opened 2016-03-21
Changed NEWS file to be in Markdown; Removed attic files from pkg description

andy-maier:
This PR is mainly about the change of the NEWS file to be in Markdown syntax, and bout its rename to NEWS.md.

It also includes a small change in the package description that was forgotten when the unused files were moved to the attic directory, namely to remove their description section in the package description.


KSchopmeyer:
+1

PR #125, opened 2016-03-21
Fixes #122: Enforced use of absolute imports.

### LINKED ISSUE #122: Enforce use of absolute imports
In Python 3, that is automatically enforced. For Python 2, that can be enforced by adding:

    `from __future__ import absolute_import`

The imports should already be absolute, at this point, so this is only about making sure that stays that way, and about documenting it via the statement above.

andy-maier:


KSchopmeyer:
+1

PR #129, opened 2016-03-22
Started development of 0.8.3.

andy-maier:


andy-maier:
Screwed up this PR because it should be targeting stable_0.8, not master.

PR #134, opened 2016-03-22
Fixes #40: Replaced dangerous default values `[]` and `{}` with None; Replaced list/map with list comprehension.

### LINKED ISSUE #40: Use of [] and {} as default values in function argument definitions.
Dangerous default value [] as argument (dangerous-default-value) ex variable=[]. This occurs typically in the arguments for functions, especially _init_.  It has the possibility of being a real problem in some cases in that the point in the lifecycle where defaults for arugments is set differes from where values are put.  The proposed solution is to change these to None.

andy-maier:


KSchopmeyer:
+1 Only reviewed code, did not download and run

andy-maier:
Rebased.

PR #135, opened 2016-03-23
Fixes #69: Migrated mof_compiler and wbemcli to argparse and improved help messages.

### LINKED ISSUE #69: optparse vs argparse
At this point in the python history, optparse has been officially deprecated with argparse as its replacement.  pywbem imports optparse and masks it to look like argparse.  We should really update to argparse as the import keeping this pywbem away from deprecated packages. Hopefully this is an almost free update.

andy-maier:
This PR migrated the `mof_compiler.py` and `wbemcli.py` modules to use argparse instead or optparse. It assumes that argparse is available also on Python 2.6 (it was added only in Python 2.7, but the package is available on most Linux distros).

Interesting review points:
- The help message; Invoke: `./mof_compiler -h` and `./wbemcli -h`
- Is argparse in Python 2.6 available on distro xyz.

### DIFF
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -41,6 +41,7 @@ from __future__ import print_function, absolute_import
 import sys
 import os
 from getpass import getpass
+import argparse
 
 import six
 from ply import yacc, lex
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
I only reviewed code on this one.   
+1
But couple of suggestions for future.  If we are going to use things like the argparse help folder code, shouldn't we put it one place and ref it?
Was wondering if the group arg is really necessary rather than just separate arguments.   That is only question.  

In any case, lets get this in and we can make any improvements in future

Issue #93, opened 2016-03-15
Many of the cim objects have only partial _repr_ implementations


KSchopmeyer:
Many _repr_ implementations only represent the first few components of the object and ignore the
remainder.  This makes it difficult to use this for diagnostics, etc.

andy-maier:
The PR #127 addresses this issue, but it also changes `str()` on a number of CIM object classes.
A particular discussion point is backwards compatibility. Basically, we are assuming thet the return value of `repr()` can be changed because a client should not rely on its format. For the changed `str()` functions, we also assume that, and that is probably ok, because they were previously equal to `repr()`. Except for the three mentioned changes, their return value is unchanged compared to before this PR.

PR #137, opened 2016-03-24
Resolved Epydoc warnings and improved docstrings.

andy-maier:
This PR was triggered by Karl's comment 2 on issue #93, and then evolved in some more docstring changes.

Details, from the combined commit messages:

- Resolved nearly all Eydoc warnings.
- Added class names to all docstrings of `cim_obj` classes (see Karl's comment 2 on issue #93).
- Moved CIM-XML DTD definitions in `tupleparse.py` from Python comments into the docstrings.
- In `mof_compiler._print_logger()`, changed name of `str_` argument to `msg`, to avoid Epydoc warning about unresolved link target. This is an internal function so this should not be a problem for users.

### DIFF
--- a/pywbem/__init__.py
+++ b/pywbem/__init__.py
@@ -66,7 +66,8 @@ of this package.
 
 The Epydoc documentation tool that is used to produce these web pages does
 not list all symbols that are folded into the package namespace, unfortunately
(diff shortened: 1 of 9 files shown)


andy-maier:
I am merging this without additional review after chatting with Karl.

PR #139, opened 2016-03-24
Issue # 138; Remove the code that outputs TOINSTANCE flavor from the qualifier tom…

KSchopmeyer:
…of function.

This is not part of the DMTF spec for mof so we should not be outputing
it with the mof. Since the compiler cannot handle it on input, we have a
conflict.


andy-maier:
+1 after code review.

PR #142, opened 2016-03-25
Fix issue with bad variable elf -> self

KSchopmeyer:
Changed variable name that was causing pylint undefined variable error


andy-maier:
+1

PR #143, opened 2016-03-25
Solves #136 for 0.9.0: Prerequisite issues for LEX/YACC table build for MOF compiler.

### LINKED ISSUE #136: setup.py install has dependency on packages needed for pywbem import
In a fresh Python virtual environment, and when in a clean pywbem source directory (either after `make clobber`, or when the distribution archive from PyPI has been expanded into it), the command:
    `python setup.py install`
fails because it builds the MOF lex/yacc parser tables, and to do that, it imports pywbem, causing it to stumble across missing packages such as six, M2Crypto, etc.

**Proposal:** Move the MOF parser table generation into its own module that has minimal dependencies, and import that module in setup.py.

andy-maier:
This required a lot of experiments...

This PR addresses issue #136 for v0.9.0. Another PR will address that issue for v0.8.3.

The PR is ready for review. There is not so much a need for review of the code, but for testing in all kinds of variations:
* Installer: `pip install` e.g. from local dist archive / `setup.py install` from repo work dir.
* MOF table files: missing / present with non-matching versions / present with correct versions.
* Prerequisite packages for building MOF table files (`six`, `ply`, `M2Crypto`): installed / not installed.
* Python env: system python with sudo, virtual python without sudo.


KSchopmeyer:
I have run only the standard tests.  The question here is whether we think we need more tests before committing this one.  Discuss in morning.

andy-maier:
Agreed in the call that what remains is a proper installation test, which we should document and perform before every release. To be done in a separate issue. Based on that, I'm merging this PR.

PR #145, opened 2016-03-26
Created _cliutils module with utility stuff for CLI scripts.

andy-maier:
We talked about this in one of the calls. Ready for review.

- Created `_cliutils.py`. It will be picked up automatically by build.
- Moved the SmartFormatter class from `wbemcli.py` and `mof_compiler.py` into `_cliutils.py`.
- Changed `import pywbem` in wbemcli.py to `from . import WBEMConnection` because wbemcli is a pywbem module.


KSchopmeyer:
+1

Issue #147, opened 2016-03-26
tomof for instances does not handle mof output of embedded instances


KSchopmeyer:
This actually causes an exception.

I thought I had documented this one but could not find it in the list.

andy-maier:
Is this completely addressed by PR #148?

KSchopmeyer:
Since we do not have a complete round trip test (mof back to mof) I cannot be sure.  See issue #146 where mofstr cannot handle all strings provided.  Leave this one open until we get the round trip test done.

I cannot believe that I wrote and incomplete bug description likt that one above.

andy-maier:
**Discussion point:**
* What is left for this one?

KSchopmeyer:
Found bug in tomof for property values on instances.   The spec calls for output in mof format but as a string.  The code outputs xml. At the same time, if we output as mof the formatting is horrible

instance of Test_CLITestEmbeddedClass {
    embeddedInst = "instance of Test_CLITestEmbedded1 {\n    comment = \"Instance created by "
    "provider.\";\n    Id = \"100\";\n    name = \"Ronald\";\n};\n";
};

I want to make the single change to output as mof and then move any further work to 0.10.0 since whe also have issues with the mof compiler and compiling embedded mof.  See the testmofs where the properties for embedded instance and object are commented out.

Issue #110, opened 2016-03-18
mof output of properties with arrays does not deal with 80 column width limit.  


KSchopmeyer:
Need to expand code so that we fold array outputs gt than 80 col in length.  This means two folds,

a. total of all items is gt 80 col
b. Any single item is gt 80 col.

andy-maier:
Is this completely addressed by PR #148?

KSchopmeyer:
Not sure since we do not have all test cases.   There is at least one more test case to add, the case of an array where at least one single entry is gt than 80 characters so that the entry must be folded before the array is folded.  Since there is not a single case of this in the cim model, I put it off.

There is another issue in general in that the compiler has never folded arrays of qualifier data.  I have left the fix for this out of the previous code (one problem at a time) since it is a different part of the code. Also, this occurs very frequently and in class output whereas the instance folding is less frequently seen so for this one we want to fold the array when output hits 80 col for compactness rather than the approach of one array entry per line we took with instance array output.

KSchopmeyer:
The last part of the fix for this is in pr #263 and #260

andy-maier:
Karl: Please verify whether these PRs fix the issue completely.

KSchopmeyer:
We apparently still have one problem that does not show up normally in schema mof output but does in instance mof output.  We are not folding property text values on instances.  See example:

The only problem is that the first line is too long before the fold.

instance of PG_ComputerSystem {
    Status = "OK";
    ElementName = "Computer System";
    PowerState = 1;
    Description = "Linux version 3.13.0-86-generic (buildd@lgw01-51) (gcc versio
n 4.8.2 "
    "(Ubuntu 4.8.2-19ubuntu1) ) #131-Ubuntu SMP Thu May 12 23:33:13 UTC 2016";
    PowerManagementSupported = False;

I am closing this one and creating new issue with this single problem

Marking this one closed since the originally defined issues are closed by pr #263 and #260

KSchopmeyer:
mark closed

PR #148, opened 2016-03-26
Save in process work to clean up mof indent issues #110 and #147

KSchopmeyer:
Adds code for embedded instance tomof, generalizes mof indent, sets mof indent
from 7 to 4, fixes several places where line overflows 80 col.
Adds constants to control indent and uses them to generate
indented mof.
Handles properties with arrays of values so that they do
not overflow max line length.
Add more tests to test_cim_obj.py for mof output of classes
and instances.

Remove special tests added to run_cim_operations.py that should be part of this commit/pr. Will be in separate branch
Adds one optional test to run_cim_operations.py

NOTE: I think that I had another embedinstance branch but this branch covers the error in embedded instance

Somehow I got too much into one branch, both the embedded instance fix and the clean up of mof output formatting.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -46,7 +46,7 @@ if six.PY2:
     from __builtin__ import type as builtin_type
 else:
     # pylint: disable=wrong-import-order
-    from builtins import type as builtin_type
+    from builtins import type as builtin_type  # pylint: disable=import-error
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
+1

andy-maier:
Did not review code in depth. Merged anyway.

Issue #12, opened 2016-01-20
Convert API documentation to Sphinx


andy-maier:
Convert API documentation of PyWBEM client to Sphinx. It must continue to work from the docstrings.
- Phase 1:
  - Try out the docstring support of Sphinx.
  - Decide on one syntax flavor in the docstrings.
  - Checkpoint about feasibility.
- Phase 2:
  - Change build process to either continue publishing on pywbem.github.io, or on readthedocs (just the API portion, the rest of the web site can stay on github.io and link to readthedocs).
  - Change pywbem setup script (develop/develop_os) to install Sphinx instead of Epydoc.
  - Change the actual docstrings.

andy-maier:
PR #149 addresses Phase 1 and parts of phase 2.
Still open:
* Establish automatic generation on RTD
* Update the pywbem.github.io page to use the docs from RTD
* docstrings that still need to be adjusted:
  - mof_compiler
  - docstrings of some internal elements.

andy-maier:
Update on RTD: The build there fails with "swig" not available.
See [rtd issue 2094](https://github.com/rtfd/readthedocs.org/issues/2094).

andy-maier:
Remaining work for 0.9.0, after accepting PR #149:

In the sources:
* Improve docstrings in "MOF compiler API" section.
* Document the wbemcli and mof_compiler scripts with their help and some more words.
* Change version-agnostic intersphinx links to standard Python elements to reference py3.5, for consistency with auto-generated version-agnostic links.

Infrastructure:
* Establish automatic generation on RTD.
  - The build there currently fails with "swig" not available. See [rtd issue 2094](https://github.com/rtfd/readthedocs.org/issues/2094).
* Update the pywbem.github.io page to use the docs from RTD.

andy-maier:
The wbemcli and mof_compiler scripts were documented with PR #164.
The issue with the version-agnostic intersphinx links was addressed in PR #168.

That leaves the following still open:

In the sources:
* Improve docstrings in "MOF compiler API" section.

Infrastructure:
* Establish automatic generation on RTD.
  - The build there currently fails with "swig" not available. See [rtd issue 2094](https://github.com/rtfd/readthedocs.org/issues/2094).
* Update the pywbem.github.io page to use the docs from RTD.

andy-maier:
The MOF compiler API docstrings were converted to Sphinx and improved in a new version of PR #168.
[...]

PR #149, opened 2016-03-27
Migrated documentation to Sphinx and improved it.

andy-maier:
- Established infrastructure for Sphinx documentation with autodoc and napoleon for Google docstrings. This included updating the makefile and adding a few new make targets in addition to changing the existing `builddoc` target to now use Sphinx instead of Epydoc. Removed Epydoc support from the makefile and patch function from setup script.
- Updated the docstrings if all elements that are part of the public API of the WBEM client library, to have Google docstring format.
- Fixed some errors in the docstrings.
- Added descriptions for some undocumented methods.
- Reordered methods in cim_operations and cim_obj modules to be in the desired order for the documentation.
- Updated NEWS.md and showed NEWS file in change log section, by having make builddoc copy it there.

See issue #12 for remaining tasks to do.

To produce the Sphinx documentation locally, check out the `go_sphinx` branch and run:
* `make develop` - to establish the new prereqs
* `make builddocs` - to build the docs, the main file is: `build_doc/html/docs/index.html`


KSchopmeyer:
Somewhere in all of this we don't point out that this is python 2 a

KSchopmeyer:
We should include that this is python 2 and 3 compatible in this documentation.

KSchopmeyer:
What is the future target, read-the-docs, the github webpage, or both for documentation?

Future: All this needs examples

andy-maier:
Karl, we point the supported Platforms out in section Compatibility, including the Python versions.
Some links to Python base stuff however that have no need or intention to be version dependent needed to make a decision which version to link to. In some cases that is controlled by sphinx interlink, but I believe there are still some cases where the linking happened automatically and I don' understand yet how it can be controlled if at all.

So let's first discuss the strategy: What would the goal be for version independent links?

andy-maier:
Future target: I think the RTD content would just replace the Epydoc stuff on the github.io page, which would keep the role of the general landing page. But I'm open to other suggestions. Theoretically we could move completely to Sphinx, but then we need to figure out how to have more than one page.

KSchopmeyer:
On 03/28/2016 08:50 AM, Andreas Maier wrote:
>
> Future target: I think the RTD content would just replace the Epydoc 
> stuff on the github.io page, which would keep the role of the general 
> landing page. But I'm open to other suggestions. Theoretically we 
> could move completely to Sphinx, but then we need to figure out how to 
> have more than one page.
>

Sounds good for now.
>
> —
> You are receiving this because you commented.
> <https://github.com/pywbem/pywbem/pull/149#issuecomment-202402253>
>


-- 
================
Karl Schopmeyer
tel: 972-814-5581

KSchopmeyer:
My initial python version comment was just that you have to go to compatibility to see the overview idea that this is a python2 and python3 package.  Note that in the readme on the code page we do not point this out either.  As a new developer, I was thinking that python major version would be one of the first questions and should be part of the summary wherever the developer/user looks.  i.e

"The pywbem PyPI package provides a WBEM client, written in pure Python for Python 2 and 3"

I still argue that methodcall and imethodcall should probably not be part of this public interface. Just tempts people to use them directly( there is a note in each indicating that they should probably not be used).  Also wish that the operations.

Getting picky here but we call them cim operations.  Are they not WBEM operations? 

I still say that this is enough better that we could probably commit and then clean up.

andy-maier:
Let's do these final changes now, before merging:

* "operations" -> "WBEM operations"
* More prominent mentioning of Python 2 and 3 support
* imethodcall / methodcall: That would allow users to write support for new operations. However, we don't add new operations like crazy (rather in half-decade cycles or so ;-). So let's agree to make them internal.

To be discussed and agreed in our call tomorrow.

KSchopmeyer:
+1 when we resolve these conflicts.  Lets not put more into this pr. Create new prs.

andy-maier:
Described the remaining tasks for 0ö9.0 in issue #12.
Merging this PR, to have a common base again.

PR #155, opened 2016-03-29
Fixes #34, fixes #154: Various improvements to wbemcli + None parameter fix

### LINKED ISSUE #34: wbemcli is really very undecipherable
Trying to use this utility to really do something is difficult. Some of the issues include:
1. There is no list of the client functions available. You need the python source for that.
2. There is no clear documentation on the set of parameters for any given function.  Note that some of 
   the client functions defined in wbemcli do include specific arguments but many of them just use
   _args and *_kwargs. Since the functions they call do not really include argument documentation it is a long trail for the user to find out what arguments apply to any function.
3. The display possibilities are not really documented.  Thus, how could you display mof or xml for a result, how do you display the various types of results (class, instance path qualifierType, list of classes, list of classes, instances, paths, qualifiertypes; and finally how would you input and get a response for an invoke method.

While the general direction should be to move to a real set of client browsers (both cli and gui) there should be some way to make this more usable now.

Thus, for example, you can display a class in mof format:
     cl = gc('CIM_ManagedElement')
    print cl.tomof()

or display the xml with:
    print(cl.tocimxml().toprettyxml(indent = '  '))

Just as simple examples.

### LINKED ISSUE #154: Optional keyword params in WBEMConnection operation methods cannot be specified with None
The documentation states for some optional keyword parameters in WBEMConnection operation methods that their default would be `None`.

However, if a caller specifies the parameter with `None`, the CIM-XML creation in `imethodcall()` fails with:

    File "/media/share/Projects/pywbem/repos/pywbem/pywbem/cim_operations.py", line 609, in imethodcall
      for x in params.items()]
    File "/media/share/Projects/pywbem/repos/pywbem/pywbem/cim_obj.py", line 2915, in tocimxml
      (value, builtin_type(value)))
    ValueError: Can't convert None (<type 'NoneType'>) to CIM XML

Reproduction:

    conn.EnumerateClassNames(ClassName=None)

andy-maier:
I think this PR completely addresses the issues described in #34. Please review to that extent.

Also, make sure to run operations that have the IncludeClassOrigin parameter against SFCB and OpenPegasus, as well as EnumerateClasses.

The improvements in wbemcli required a fix in `cim_operations.imethodcall()` where any parameters that are `None` are now omitted from the request. They were running into issue #154, which is hereby also fixed. I think this change is safe to make, because if it is not made, None in any case leads to the ValueError exception described in #154.

* Fixed issue that passing param=None to imethodcall() causes ValueError.

* Added help, added missing functionality.
    - Added an interactive help function `Help()` / `h()` and hinted to it in banner message.
    - Improved docstrings of convenience operations functions, to improve the result of `help(ei)`, etc.
    - Made the detailed description of the operations (WBEMConnection methods) available as `help(conn.GetInstance)` etc.
    - Made pywbem namespace content available, so that `help(CIMInstance)` works.
    - Renamed the connection object to `CONN` (was `_CONN`).
    - Renamed all symbols that are not needed in the interactive environment, to become private, by adding an underscore.
    - Added PropertyList support to EnumerateInstances and GetInstance.
    - Renamed the arguments of the convenience operation functions to be short (e.g. `cn` for class name), for easier typing.
    - Changed all convenience operation functions that had `*args` and `**kwargs`  arguments to have specific arguments.
    - Changed some defaults of some boolean arguments (compared to the WBEMConnection methods) to be more convenient (e.g. LocalOnly = False, IncludeQualifiers on classes = True).
    - Added support for debugging via pdb (see help).

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -623,7 +623,7 @@ class WBEMConnection(object):
         # Create parameter list
 
         plist = [cim_xml.IPARAMVALUE(x[0], tocimxml(x[1])) \
-                 for x in params.items()]
+                 for x in params.items() if x[1] is not None]
 
         # Build XML request
 
--- a/pywbem/wbemcli.py
+++ b/pywbem/wbemcli.py
@@ -85,7 +85,7 @@ def _remote_connection(server, opts):
     if server[0] == '/':
         url = server
     else:
-        roto = 'https'
+        proto = 'https'
 
         if opts.no_ssl:
             proto = 'http'
--- a/pywbem/wbemcli.py
+++ b/pywbem/wbemcli.py
@@ -48,43 +48,44 @@ function. For example:
 
 from __future__ import absolute_import
 
-import os
-import getpass
-import errno
-import code
-import argparse
+import sys as _sys
+import os as _os
+import getpass as _getpass
+import errno as _errno
+import code as _code
+import argparse as _argparse
 
 # Additional symbols for use in the interactive session
 from pprint import pprint as pp # pylint: disable=unused-import
 
 # Conditional support of readline module
 try:
-    import readline
+    import readline as _readline
     _HAVE_READLINE = True
 except ImportError as arg:
     _HAVE_READLINE = False
 
-from . import WBEMConnection
-from ._cliutils import SmartFormatter
+from . import *
+from ._cliutils import SmartFormatter as _SmartFormatter
 
 __all__ = []
 
 # Connection global variable. Set by remote_connection and use
 # by all functions that execute operations.
-_CONN = None
+CONN = None
 
-def remote_connection(server, opts):
+def _remote_connection(server, opts):
     """Initiate a remote connection, via PyWBEM. Arguments for
        the request are part of the command line arguments and include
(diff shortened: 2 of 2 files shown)


KSchopmeyer:
I have the following issues:
1. Once you do a help(gi) it is a guessing game to get out of it and back to the main prompt.  I figured out it was q.
2. Under printing support, should probably change the notes a bit, for example to show print(obj.tomof()) which will format correctly.

3. Think it is repr.<obj> and not obj.repr()

Note that I entered this comment maybe in process of your last checkin since integration checks are in process now so you may have fixed these issue.

Other than that, +1

by the way, pp(cl) where cl is a class sucks but that is not our problem.

KSchopmeyer:
Whoops messed up one comment above should have been:

3. The help should say repr(<obj>) I think and not obj.repr().  

I am also guessing that we will have to look more at all pp results since the results are not so hot  from what I just saw but make that separate issue.

This is a big step forward so, +1 with comments above

andy-maier:
Addressed the issues mentioned above.
+1 myself.

KSchopmeyer:
+1

PR #158, opened 2016-03-29
Fixes #157: Zip tar stable 0.8

### LINKED ISSUE #157: Setup script still requires zip/unzip but should require tar
At some point before releasing 0.8.1, the distribution archive was changed to be tar.gz again, like it was in 0.7.

However, it was forgotten to remove the os-level dependencies to zip/unzip in the setup file, and to add one for tar instead.

andy-maier:
This PR changes the os-level dependency 'zip' and 'unzip' to 'tar' to match the type of dist archive that is being built. This was forgotten when moving back to .tar.gz archive format.

I have verified that a 'tar' package exists on the distros supported for os-level dependencies, and have tested `python setup.py develop_os` on CentOS 6.7 and Ubuntu 14.04.


andy-maier:
Picked wrong target branch - closing.

PR #160, opened 2016-03-29
Fixes #157: Changed os-level dependency 'zip' and 'unzip' to 'tar' (master)

### LINKED ISSUE #157: Setup script still requires zip/unzip but should require tar

andy-maier:
This PR contains the same change as PR #157, but targets the master branch, instead.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -428,8 +428,7 @@ def main():
                     "libxslt-devel",        # for installing Python lxml pkg
                     "libyaml-devel",        # for installing Python pyyaml pkg
                     "make",                 # PyWBEM has a makefile
-                    "zip",                  # for building distribution archive
-                    "unzip",                # for installing distrib. archive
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
What are we going to do about windows since that is largely zip based?

andy-maier:
v0.7 had an installer .exe for Windows as the release vehicle. It can be built on Windows with `make buildwin`, but I have not tested it recently. We should probably have an issue for testing it on Windows, but this change here does not affect iIndows, IMO.

KSchopmeyer:
+1

andy-maier:
+1

PR #168, opened 2016-03-31
Fixes #12: Improved intersphinx links, datetime related docs, and MOF compiler API docs.

### LINKED ISSUE #12: Convert API documentation to Sphinx

andy-maier:
Ready to be merged - please review.

* Intersphinx links:
  - Fixed the mixing of version-agnostic and version-specific intersphinx links (their URLs have to differ, so a trailing slash was added on one).
  - Made any intersphinx links version-agnostic except for those that intentionally need to point to a specific version.
  - Note that this PR still leaves the version-agnostic intersphinx links to reference Python 2 documentation, because that is still our main release. In other words, the inconsistency between the links to py35 targeted by the auto-generated links for the method argument types, and user-specified links to standard Python elements still exists, but I suggest to simply accept that inconsistency. There are not too many auto-generated different types, and they are only things like bool, int, dict.
* Datetime related docs:
  - Improved the description of the `CIMDateTime` and `MinutesFromUTC` classes.
* MOF compiler API docs:
  - Reordered the classes to be more logical.
  - Converted docstring format to Google format for Sphinx napoleon.
  - Described `MOFWBEMConnection.getns()`/`setns()` as internal methods.
  - Expanded and improved all descriptions of the external elements in `mof_compiler`.
  - Clarified that in `MOFCompiler.find_mof()`, the class name is looked up by the base file names only (not by the classes defined inside).
  - Clarified that `rollback()` does not currently roll back changes to the qualifier types.
  - Described `MOFWBEMConnection` as being both the interface definition class and the implementation for WBEM server based repositories.


andy-maier:
The Travis CI failures do not seem to having to do anything with this change (gzip fails unpacking the downloaded swig tarball). I saw another network error today causing wget to fail.
Let's restart the jobs at some point.

KSchopmeyer:
I restarted the jobs and we wait.

KSchopmeyer:
They passed this time

KSchopmeyer:
"
+    Each object of this class represents a MOF compiler.

This statement in the mof compiler makes no sense to me.  Other than that I am OK but it is just code review. I have not tested this one.

andy-maier:
Well, before that I had "A MOF compiler" and to me that sounded somewhat ambiguous as to whether the class was the compiler or the instance of the class (which in Python is most of the time called "object" instead). It is actually the instance, so I made the text more specific.

It is also true that in other class descriptions, we don't use that verbiage. I would be OK with going back to "A MOF compiler."

andy-maier:
As far as "testing" goes: I made no code changes in this PR, and left those for issue #170.
So the best way to review this PR is via `make builddoc` and then using your browser. The changes are in the datetime part of "CIM data types" and in "MOF compiler API". And the links to elements in the standard Python library.

andy-maier:
Commit 171af18 goes back to "A MOF compiler.".
Waiting for your final review.

andy-maier:
+1

andy-maier:
Yeahhh - the first RTD build that worked:

https://pywbem.readthedocs.org/en/latest/

I worked around the issue that RTD does not have Swig, preventing M2Crypto installation, by not using it if on RTD.

andy-maier:
+1

KSchopmeyer:
Think we should probably not make statements about future directions in the code.  Ex.

+    At the moment, this class also acts as an interface definition class for
+    repository connection class, defining the interactions between the MOF
+    compiler and the CIM repository used by the MOF compiler.

No reason for at the moment.

There is one other in wbemcli about future better cli tool.  

Will vote +1 but would prefer to keep future promises in future doc.

andy-maier:
You are basically right, but for the interface class, I already have PR #173 up, which replaces the future direction with what it hinted at.

andy-maier:
Commit b3f4341 removes any statements about the future.
It still says "currently" in a few spots.
Please double check.

andy-maier:
+1

KSchopmeyer:
I found also an "at the moment"  and "at this point" (rollback)  in the comments.  But lets make those separate issues where we try to keep the inline doc to what exists and use other documentation, issues, etc. to define proposed changes.  Thus the rollback should really be just the statement that it covers only ...

+1

PR #172, opened 2016-04-01
Initial stab at testing all property types incl. emb.inst.

andy-maier:
This PR is now ready for merge. Please review.

It adds the YAML file `getalltypes.yaml` which contains a simple test case: A GetInstance operation returning one instance that has one property of every possible CIM datatype, including char16 and an embedded instance.

The PR needed to fix an error in the test_client.py test driver, whereby the dict of properties was not correctly converted into CIM objects.


andy-maier:
+1

KSchopmeyer:
+1

PR #173, opened 2016-04-01
Fixes #170: Some MOF compiler reengineering

### LINKED ISSUE #170: Some light MOF compiler and exception reengineering
I'd like to propose the following changes, on top of the documentation updates in PR #168:
- The current `MOFWBEMConnection` class acts as both the interface definition and the implementation for a WBEM server repository. I'd like to separate out the interface role into a new base class (say, `BaseRepositoryConnection`) that defines the interface and describes the requirements for an implementation.
- The `MOFWBEMConnection` class has two externally visible ways for getting and setting its repository namespace:
  - a pair of `getns()` / `setns()` methods,
  - a property `default_namespace` that can be read and written, and which in turn uses `getns()` / `setns()`
    I'd like to remove the getter/setter methods from the description, but keep their names, so that they still exist for compatibility.
- In the `mof_compiler` script, catch not just CIMError from the MOFCompiler methods, but its base class pywbem.Error, because a lot more can be raised than just CIMError.
- In `tupleparse`, exception `ParseError` is not derived from `pywbem.Error` yet. I'd like to do that, but that means the exceptions need to be shuffled around a bit. One option would be to move all of them into a new sub-module. We can leave alias symbols in the modules that define them today, for compatibility.

andy-maier:
This PR addresses the MOF compiler related parts of issue #170, but it needs further improvements in the area of the `default_namespace` attribute.

It is ready for review, but not for merge yet.

It is based on PR #168, so when reviewing it, look at just the commits starting with ce3f4a6.

Details:
- Added an abstract base class `BaseRepositoryConnection`, to separate interface definition from implementation.
- Clarified in description of `MOFWBEMConnection` that it does not currently update the WBEM server.
- Added clarifying comments to the code regarding removal and rollback.
- In the mof_compiler main routine, expanded the catched exceptions from just `CIMError` to `pywbem.Error`.


andy-maier:
With the previous commit, the PR is now ready for review and merge.
It addresses the MOF compiler related parts of issue #170.

andy-maier:
+1

KSchopmeyer:
+1

andy-maier:
Rebased (and squashed the two commits into one).

PR #177, opened 2016-04-02
Fixes #176: Makefile cleanup and optimization; Ensure Travis + Tox clean start

### LINKED ISSUE #176: Make all on any local branch updates the dist file
Not sure if this is what you wanted or not but any general make on any branch does a update to the file 

dist/pywbem-0.9/pywbem-0.9.0.dev0.tar.gz

Which means that we are continually committing or removing this file.

More of an annoyance than big issue

andy-maier:
Please review and vote.

This PR is a major overhaul of the makefile and in addition it ensures that Travis CI and Tox have a clean start:

- In the Tox control file, used `make clobber` instead of `make clean`. This ensures that everything is gone when Tox runs the next Python version in the same directory.
- In the Travis control file, used `make clobber` instead of `make clean` for consistency with Tox, even though Travis runs each Python versionin a fresh directory already.
- In the makefile, fixed issue #176 (`make all` was always rebuilding the dist archive) by moving the removal of the LEX/YACC table modules from `clean` to `clobber`.
- In order to get to a delta build for `make all`, removed `clean` from the dependency list of `all`.
- Removed some Epydoc remainders.
- Moved the detailed completion messages to the targets that actually produce the files, and kept the completion messages of the virtual targets short ("xyz done.").
- Added a proper dependency list to the Pylint rule.
- Moved control about excluding LEX/YACC table modules from Pylint run from its `--ignore` optoin up into the makefile.
- Added the makefile as a dependent to some more rules whose processing could be affected by changes in the makefile.


andy-maier:
There is one slight misbehavior still: In a freshly cloned repo, the LEX/YACC table files do not exist, so make all triggers a rebuild of the dist archive. The only way I found to avoid that is to check the mof*tab.py files in again. I think that would be fine, I have seen no case where their existence even with outdated tabversions would cause any confusion to lex and yacc. I'll add a commit that does that.

andy-maier:
There is still one caveat which I cannot fix: On a fresh repo clone, the files all have the current date. So the first `make build` will rebuild the dist archive needlessly. But from then on, it should work as expected now.

andy-maier:
Apart from the caveat I don't know how to fix, +1.

KSchopmeyer:
Lets make a separate issue for the dist archive and bypass it here.

+1

andy-maier:
You mean an issue for the needless initial rebuild after a fresh repo clone? I would not know how to address that with a make-based build. Other build tools look at signatures of file content (e.g. scons), and they would solve the problem. Not sure if that warrants the migration effort. Also, if git provided a way to check out keeping the original date, it would also be solved, but git does not have such an option today.

Of course, we can always carry an issue around even if we cannot solve it, but we could also leave it... I'll leave it up to you to decide, and I'm fine either way.

Merging this PR.

Issue #150, opened 2016-03-28
pywbem.cim_http.ConnectionError: Socket error: check_hostname requires server_hostname


skolsuper:
Hey,

Firstly, so glad to see this package come back to life, and really like the direction you're going with the changes.  I inherited a project that was using a copypasted local version of this package with python3 fixes, and I'm now trying to migrate it back to using the upstream version.  However I'm running into an issue I can't figure out.

**Repro:**

```python
import pywbem

conn = pywbem.WBEMConnection('https://192.168.1.1', creds=('someuser', 'password'))

conn.EnumerateInstances('OMC_Fan')
```

**Expected result:**  Information about the hardware (the `"OMC_Fan"` above is just an example, I get the same result whatever string I pass to `EnumerateInstances`)

**Actual result:** The exception in the title.  Stepping through it with the debugger, it looks like the request is never made as a sanity check in the python `ssl` module fails.

Environment:  python 3.5 virtualenv on MacOSX El Capitan

Any help would be greatly appreciated.

andy-maier:
Hello. Glad to hear you like the direction.
From what I could quickly gather, a SSL socket user (pywbem) has to pass in a "server_hostname" option somewhere. The SSL in Python was fixed in 2014 to always accept server_hostname even if the underlying OpenSSL does not support SNI yet. I have taken a look into the cim_http module of pywbem, and it does not pass any server_hostname option.

I have very little experience with dealing with SSL in Python 3, I'm afraid, so I could use your help:

-> If your locally patched version works for this case, would you mind pasting its cim_http.py file into a comment of this issue?

skolsuper:
Good news, after asking around, I found the history!  Somebody that actually understands git was the one that made the changes, and documented them.  I think [this](https://github.com/cloud-hm/pywbem3/commit/ca5c726daba4258b76fab35243b3d7e01619a322) is the relevant commit, but there are only a few commits in that repo, so hopefully the fix won't be difficult to find.  I am also looking but it's the end of the day in my timezone soon so I will resume tomorrow.  If I fix it I will submit a PR.

KSchopmeyer:
Assigned to Karl to see if he can set up test.

andy-maier:
Discussed today with Karl. Karl will try to investigate this using OP.

skolsuper:
Possibly relevant, I happened across this in an unrelated repo: https://github.com/rethinkdb/rethinkdb/issues/5470

[...]

PR #179, opened 2016-04-03
 Fixed issues in SSL support for Python 3.

KSchopmeyer:
This is currently untested since all tests fail in hotel once you use ip address.  Think they are blocking everything.

Even if it passes, it is not tested

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -182,6 +182,10 @@ Bug fixes
 * Fixed an IndexError in cim_http.wbem_request() that occurred during handling
   of another exception.
 
+* fixed issue with python 3 and https that was causing connect() to fail. 
+  This completely separates connect() code for python 3 ssl module from
+  python 2 m2cyrpto.
+
 
 pywbem v0.8.2
 -------------
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -63,7 +63,7 @@ if six.PY2 and not _ON_RTD:  # RTD has no swig to install M2Crypto
     SocketErrors = (socket.error, socket.sslerror)
 else:
     import ssl as SSL                  # pylint: disable=wrong-import-position
-    from ssl import SSLError           # pylint: disable=wrong-import-position
+    from ssl import SSLError, CertificateError # pylint: disable=wrong-import-position
     _HAVE_M2CRYPTO = False
     #pylint: disable=invalid-name
     SocketErrors = (socket.error,)
@@ -254,7 +254,7 @@ def get_default_ca_certs():
     returned. If no path is found out, None is returned.
     """
     if not hasattr(get_default_ca_certs, '_path'):
-        for path in (get_default_ca_cert_paths()):
+        for path in get_default_ca_cert_paths():
             if os.path.exists(path):
                 get_default_ca_certs._path = path
                 break
@@ -342,7 +342,6 @@ def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
             the send method
         """
         # pylint: disable=old-style-class,too-few-public-methods
-
         def send(self, strng):
             """
             A copy of httplib.HTTPConnection.send(), with these fixes:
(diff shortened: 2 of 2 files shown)


KSchopmeyer:
Note that I have not been able to test this one because of issues here in hotel.  Will test when I get to plugfest.

andy-maier:
You need to be careful with inheriting from object when the rest of the tree does not.

KSchopmeyer:
I agree. That was really just a test to see about removing some pylint disables.  Of course, passing the tests means nothing since this change is not really tested.  Get a chance to test against a number of real servers today.

andy-maier:
This also needs to go into 0.8.3.

KSchopmeyer:
Commited working update but python 3 still broken.

KSchopmeyer:
With the changes in the pr update of 30 april, we can now communicate with an ssl host successfully.  However, we have not tested with valid certificates yet (tested with N0-verification) and we have to add one more piece of code to validate the hostname when validation is true.  Since that is not in all versions of the ssl python 3 library we will have to de it externally.  Other than that, the only additonal thing is to test against a host where us and the host have certificates with a common ca.

andy-maier:
This PR fixes issue #274 (and issue #150 once we roll it back to v0.8.3).

KSchopmeyer:
At this point the code allows https connections with python3 with -no-verification.  I am having problems with the verification phase simply because of trying to get certs to match.

Isuggest that we bypass the check_hostname feature for this patch unless it is built(in python3 theey gradually added parts of the check_hostname feature bewteen 3.2 and 3.5 so we would either have to do the backport path or live with whatever the python release offers.

andy-maier:
In Python 3, we only support 3.4 and upwards. So let's figure out what is in 3.4 out of what we need.

Is the check_hostname feature verifying the hostname stated in the server certificate against the hostname the client used to connect to the server?

KSchopmeyer:
Python 3 only:We still have issues with the match-hostname in that different versions of python ssl module offer different levels of support for this concept.  See the following site:

https://pypi.python.org/pypi/backports.ssl_match_hostname

The question is whether we try to maximimze security by insuring that this function works.

KSchopmeyer:
We agreed that we can live with the limitations on the match_hostname between python 3.4 and 3.5. But we need to document them.

Karl needs to complete the test to confirm that the verifcation works with the known limitations on 2.4 adn 3.5.
We also need to add an installation check to assure that pywbem is being installed on 3.4 or 3.5.

andy-maier:
On match_hostname and the backports module:

If I understand this correctly, then Python 3.4 and 3.5 have everything we need ot of the box, and the backports module is not needed. If we want to have hostname matching for Python 2.6 and 2.7, we need the backports module.

Correct?

If so, I suggest that we use the backports module for 2.6 and 2.7.

andy-maier:
Do you want to cherry-pick the change from #293 into this PR?

    git cherry-pick <short-commit-id>

KSchopmeyer:
For python 2.7, no.  Not sure how but the trace I showed you displayed m2Crypto doing the match_hostname.

Lets commit 293 and its corresponding 0.8.4 and then do this one and its corresponding 0.8.4.  I have not prepared the 0.8.4 version but will tonight.

More straight forward that cherry-pick.  Actually I just merged them to do the test.

andy-maier:
Karl,
I have merged PR #293 and its coresponding 0.8 PR #290.

Please rebase this PR here on master to get it.

KSchopmeyer:
Integrated with pr 289, index error on exception. Note that I made one other change that I marked for discussion, changed the  SSL context setup from sslv23 to the function create_default_context(). This appears to be the general recomendation since it sets up what is best default between ssl module and openssl.  However, that is not as well tested as the remainder.
DISCUSSION: keep this change or revert.

KSchopmeyer:
+1 except for the discussion items above.

KSchopmeyer:
issue of default_context.  Andy proposes that we set a minimum level TLS as TLS1.0 and that becomes the standard.  We will do another isssue for this.

We will review the tunnel question as a separate issue.

andy-maier:
We agreed in the call today to merge this one.

PR #180, opened 2016-04-03
Fixes #152 : Eliminated keyword parameters of WBEMConnection methods, and more streamlining.

### LINKED ISSUE #152: API consistency
The Operation apis are inconsistent in that most generalize to \* and *\* but a couple (see CIMQualifierDeclaration are detailed.

This is more a question for discussion that a firm issue

andy-maier:
Ready for review.

- Changed the documented keyword parameters to become named parameters.
- Adjusted invocation of `imethodcall()` accordingly.
- Added an `**extra` keyword parameter to all intrinsic methods again, in order to be able to generate incorrect parameters (one test_client testcase for GetInstance exercised that and expected CIM error with invalid parameter, where the operations with specific parameter names and without `**extra` already fail upon their invocation.
- Replaced helper methods `_add_objectname_param()` and `_map_association_params()`  that are association specific and modify a parms dict, to `_iparam_*()` methods that simply return the converted value and are more generally useable.
- Used the new helper methods consistently for all methods.
- This caused a few additional cases where `CIMClassName` is now supported for class names, and where the path of an instance or class, or the namespace component in a path is ignored (in all cases, that was ignored before, just not documented and not removed from what was put into the CIM-XML payload.
- Fixed incorrect parameter name Qualifier to QualifierName in `GetQualifier()`.
- Clarified which defaults are server-implemented defaults, and that `None` causes parameters not to be sent to the server.
- Adjusted test_client testcase IncorrectArgType in clienterror.yaml to expect `TypeError` and not `AttributeError` which is a more sensible exception anyway for this case. This change was caused by the new helper methods mentioned above.


andy-maier:
+1

KSchopmeyer:
I understand the reasoning behind extra but it may be difficult to justify things in the user interface for our failure tests.

Removing all the 'optional' words on the description.  Doesn't that make it just a bit more difficult for the reader to realize that this is optional unless they look at the signature itself. The information is there but the user has to look further down in the documentation. Just asking at this point

why the _ after the doc name 

Note that `DSP0200`_ does not define any additional parameters for
            this operation.

I assume that this is to make hyperlink work.

Not our comment but the sentence "Name of the class to be enumerated, in any lexical case" is the long way around to 'Case-insensitive name of class to be enumerated'

Only part way through this morning. Will continue review later today.

andy-maier:
* On `**extra`: If you and I agree and nobody else comments (opinions welcome!), I think it is as justified as it can get ;-)
* On removal of 'Optional': It was just there because there is no other way to define optionality and the default in case of keyword parameters. For named parameters, optionality and default are specified in the function header. Of course we could decide to repeat that in the description text, but then we should do it consistently (and the removed 'optional' and 'default' descriptions existed only for keyword parameters but not for any optional named parameters. My take is, Pythonians are used to look at the function header for defaults.
* The trailing '_': This is link syntax in RST.

KSchopmeyer:
You convinced me.  the extra is documented so that is enough

+1

PR #181, opened 2016-04-04
Fixes #175: Restructured the Sphinx documentation to be on multiple pages.

### LINKED ISSUE #175: Link to change log on RTD docs is broken
The main page on RTD contains a change log section. The link is broken, even though the NEWS.md file is there.

andy-maier:
This PR shows how a multi-page version of the documentation could look like. It is also one possible solution to issue #175.

The PR is ready for merging; please review.

The change can be reviewed best by building the documentation locally (`make builddoc`) using this branch and looking at the result with a browser (top file: `build_doc/html/docs/index.html`).

Work still left to be done:
- Convert change log file `docs/changes.rst` from MarkDown syntax to RST.


andy-maier:
I'm now done with it, except for the conversion of changes.rst from MarkDown to RST, which I'll do as soon as we agree that we want to go to a multi-page documentation.

I'm all for going multi-page, so +1 from me.

KSchopmeyer:
The only downside is that you do not see the complete toc in the left side.  However, as we add more and more, that may become an advantage.

+1

One issue may be that we show __str__ and __rep__ as part of the operations display but that is not part of this pr.  We can talk about that later.

andy-maier:
On not seeing the complete TOC on left side: That's how it is in the standard Python documentation, and I don't know how to change it. Being consistent with the standard Python docs is also not a bad idea in the first place.

On `__str__()` and `__repr__()`: I assume you refer to the functions from `WBEMConnection`? If so, then we don't show them in the operations table in the intro, but obviously they are members of the class and should therefore be shown where the class members are shown, IMO.

So I take it the decision is to go forward with multi-page, and I'll convert the changes.rst, then.

andy-maier:
Updated:

* Converted changes.rst to RST syntax
* Reestablished conditional display of last commit for development versions that went away by mistake, and moved it into the change log.

PR #183, opened 2016-04-05
Eliminated dead code added by mistake in commit 66896b6

andy-maier:
Please review.

This PR eliminates dead code between `methodcall()` and `_iparam_namespace_from()`, that was added by mistake in commit 66896b6.


KSchopmeyer:
+1

PR #186, opened 2016-04-06
Fixes #185: Fixed KeyError when iterating over CIMInstance or CIMInstanceName.

### LINKED ISSUE #185: Some CIM objects stated to behave like dicts, but are not iterable
The `CIMInstanceName` and `CIMInstance` classes are stated to behave like dictionaries of their keybindings or properties, but iteration across them invokes the `__iter__()` method inherited from `NocaseDict` which does not work and fails with `KeyError`.

**Solution:** Add `__iter__()` methods to these classes that do it correctly.

andy-maier:
Fixed the problem and extended the testcases in the `DictTest` class in `test_cim_obj.py` so that the problem would have been found. Also, changed from hard-coding the expected dict values to passing them in.

Ready for merging, please review.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -65,6 +65,11 @@ Enhancements
   exception, so that now all pywbem specific exceptions are derived from
   `Error`.
 
+Bug fixes:
+
+* In `CIMInstance` and `CIMInstanceName`, fixed KeyError when iterating
+  over the objects.
+
 pywbem v0.8.2
 -------------
 
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -845,6 +845,9 @@ class CIMInstanceName(_CIMComparisonMixin):
     def __len__(self):
         return len(self.keybindings)
 
+    def __iter__(self):
+        return six.iterkeys(self.keybindings)
+
     def copy(self):
         """Return a copy of the :class:`~pywbem.CIMInstanceName` object.
         """
@@ -1180,6 +1183,9 @@ class CIMInstance(_CIMComparisonMixin):
 
     def __len__(self):
         return len(self.properties)
+
+    def __iter__(self):
+        return six.iterkeys(self.properties)
 
     def copy(self):
         """Return copy of the :class:`~pywbem.CIMInstance` object."""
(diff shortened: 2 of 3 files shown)


andy-maier:
+1

KSchopmeyer:
One change and you are OK.  Please docstring the new functions.

KSchopmeyer:
+1 with the new docstrings

andy-maier:
The new `__iter__()` functions are not docstringed for the same reason the existing `__getitem__()`, etc. functions are also not docstringed: We say that the class behaves like a dict, and the undocumented functions are exactly what makes up a dict. The idea is that saying it behaves like a dict is sufficient.

If you prefer having these 5 or so dict-supporting functions docstringed, that would also be fine with me.

KSchopmeyer:
Sounds goot to me. Then why not set the pylint disable to avoid pylint messages.

+1

KSchopmeyer:
On 04/06/2016 04:45 PM, Andreas Maier wrote:
>
> The new |__iter__()| functions are not docstringed for the same reason 
> the existing |__getitem__()|, etc. functions are also not docstringed: 
> We say that the class behaves like a dict, and the undocumented 
> functions are exactly what makes up a dict. The idea is that saying it 
> behaves like a dict is sufficient.
>
> If you prefer having these 5 or so dict-supporting functions 
> docstringed, that would also be fine with me.
>
> —
> You are receiving this because you commented.
> <https://github.com/pywbem/pywbem/pull/186#issuecomment-206623830>
>
In this case I just want to keep cutting down on the pylint messages.

-- 
================
Karl Schopmeyer
tel: 972-814-5581

KSchopmeyer:
just disable pylint would be fine.

andy-maier:
The no docstring message is automatically disabled in the pylint.rc file for special methods such as `__iter__()`, so no need to state it explicitly. I verified that the no docstring message does not show up.

KSchopmeyer:
Ahha. I missed that one. Sorry, this moving from place to place every few hours and now starting back to Dallas means I am not as careful as I should be in checking

+1

andy-maier:
Rebased, and fixed a merge conflict in the changes.rst file.

PR #190, opened 2016-04-06
Fixes #189: Fixed issues with MOF escaping in mofstr().

### LINKED ISSUE #189: Issues with MOF escaping in mofstr()
The `mofstr()` function in `cim_obj.py` has two issues in the area of MOF escaping:
- It attempts to recognize MOF escape sequences and will pass them through unchanged, to the output. A comment states that the motivation for that is to handle strings returned by WBEM servers that use MOF escape sequences (instead of XML rules), so because they already are MOF escape sequences, they are passed through to the MOF output.
  
  This motivation is moot however, because the `mofstr()` function is called only from two places, both of which are in `tomof()` methods of CIM objects (`CIMInstance` and `CIMQualifierDeclaration`). Any CIM string typed values in CIM objects are supposed to contain the proper strings as a Python unicode string object, without any escaping or remainders from protocol encoding.
  
  **Proposed solution:** Remove the special casing for MOF escape sequences.
- DSP0004 at some point clarified that MOF string constants may contain the whole character repertoire for CIM string typed values, which is the whole Unicode repertoire minus U+0000. The control characters (U+0001...U+001F) are not nicely handled by most editors, and after all MOF is supposed to be editable. The `mofstr()` function does not do anything with them (except for the few special ones in that range).
  
  **Proposed solution:** Add support for escaping control characters using the generic escaping, e.g. U+001B becomes to `"\x001B"`.

andy-maier:
The problem is described in issue #189. This PR contains both proposed solutions.

A main review point should be whether my understanding that the `mofstr()` function does not deal at all with string representations returned by a WBEM server (which is the justification for removing the special-casing of already escaped MOF sequences).

Please review.


andy-maier:
+1

KSchopmeyer:
Still have an issue somewhere.  See the line in the description that starts with "\"multi
The original has \\WINDOWS
The result file has \WINDOWS

Why it is reporting illegal characters in positions 11,12,13 I am not sure.

On the other hand, the cleanup is really good and I want to get in.

I am also having prolems getting rid of the ToInstance flavor for some reason but not sure why yet since it pops up in the output even if I remove it from the compiler.  I have to build a much smaller test to get this sorted out and want to get back to the ssl problem now.

Original mof:

class CIM_BootSourceSetting : CIM_SettingData {

      [Description ( 
          "A string identifying the boot source. It is typically "
          "used by the instrumentation to pass to the boot manager "
          "as a selection string. This could be a string "
          "identifying the bootable device, such as \"CDROM 1\", or "
          "could be an implementation specific address of a "
          "bootable partition, such as the following. \"fd(64)unix "
          "root=hd(40) swap=hd(41)\", or "
          "\"multi(0)disk(0)rdisk(0)partition(1)\\WINDOWS=\"Microsoft "
          "Windows XP Professional\"\"." )]
   string BootString;

rebuilt MOF
class CIM_BootSourceSetting : CIM_SettingData {

      [Description ("A string identifying the boot source. It is typically used by the "
       "instrumentation to pass to the boot manager as a selection string. "
       "This could be a string identifying the bootable device, such as "
       "\"CDROM 1\", or could be an implementation specific address of a "
       "bootable partition, such as the following. \"fd(64)unix root=hd(40) "
       "swap=hd(41)\", or "
       "\"multi(0)disk(0)rdisk(0)partition(1)\WINDOWS=\"Microsoft Windows XP "
       "Professional\"".")]
   string BootString;

Compiler error:
Illegal character '"' Line 7950, col 9
Illegal character '\' Line 7950, col 10
Illegal character '"' Line 7950, col 11
Syntax error:
test_tomofRoundTripOutput.mof:7950:
       "This could be a string identifying the bootable device, such as "
       "\"CDROM 1\", or could be an implementation specific address of a "
       "bootable partition, such as the following. \"fd(64)unix root=hd(40) "
       "swap=hd(41)\", or "
       "\"multi(0)disk(0)rdisk(0)partition(1)\WINDOWS=\"Microsoft Windows XP "
          ^^^^^
  File "./mytest_mof_roundtrip.py", line 66, in <module>
    class TestCompileMof(unittest.TestCase):
  File "./mytest_mof_roundtrip.py", line 98, in TestCompileMof
    mofc.compile_file(TMP_FILE, NAME_SPACE)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/mof_compiler.py", line 2065, in compile_file
    return self.compile_string(mof, ns, filename=filename)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/mof_compiler.py", line 2013, in compile_string
    rv = self.parser.parse(mof, lexer=lexer)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/ply-3.8-py2.7.egg/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/ply-3.8-py2.7.egg/ply/yacc.py", line 1181, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/ply-3.8-py2.7.egg/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/mof_compiler.py", line 226, in p_error
    raise ex
pywbem.mof_compiler.MOFParseError

andy-maier:
I agree that is a problem. However, it is a problem when parsing MOF. `mofstr()` is used only when producing MOF. I can see that you want to first solve the parsing issue in order to be able to validate mofstr() via the roundtrip. I'm fine with that.

KSchopmeyer:
Yes there is a parsing issue but the mofstr output is the problem in that in the pass through the system it is still not outputting the same as the original value of the string in that the original description had \\WINDOW and the reproduced mof has \WINDOW, a single escape instead of the double.  That may be the parser causing the error so I will create test to do just this one component and further id where the issue is (i.e. display the description after the initial parse).

I will vote to put this one in now even if this issue exists since that is a good cleanup and even if the problem is still there, we want to make this change.

However, that is actually the problem I started with, that particular pair of backslashes.

You get to chose.  get this work in and create second bug or try t o sort that out in this bug.

andy-maier:
There seem to be two issues here:

1. The incorrect unescaped backslash in `\WINDOWS`. That is not explicable with the updated `mofstr()`; it has several unit test cases verifying that a backslash gets escaped properly. Still, it happens on output.
2. The MOFParseError indicating illegal characters. That should be separated out.

I think it is overall simpler if we merge this `mofstr()` change, and address the other issues separately. Could you please open new issue(s) for the other problems, along with how to reproduce them based on the code level that includes this PR content?

KSchopmeyer:
agreed to commt and then sort out issues  +1

andy-maier:
Karl, I think I found significant issues in the MOF lexer rules. I have started setting up standalone tests for the lexer. I'll create separate issues.

andy-maier:
Rebased, to resolve merge conflict in changes.rst.

KSchopmeyer:
Note. Be sure to add the one about what should be keywords.  ex. is indication supposed to be a MOF keyword?

KSchopmeyer:
I knew that this mof-roundtrip test was going to be useful. :-)

PR #192, opened 2016-04-06
Fixes #178: Moved exceptions into new module 'exceptions' and derived all from 'Error'

### LINKED ISSUE #178: Inconsistency: ParseError is not derived from Error
Exception `tupleparse.ParseError` is the only externally visible exception that is not yet derived from `pywbem.Error`. I suggest to change `ParseError` to be derived from `Error`.

That means the exceptions need to be shuffled around a bit. One option would be to move all of them into a new sub-module. We can leave alias symbols in the modules that define them today, for compatibility.

Needs discussion on how to do that.

andy-maier:
Ready for merge; please review.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -61,6 +61,10 @@ Enhancements
 
 * Added CIM status codes 20 to 28.
 
+* Changed the `ParseError` exception to be derived from the `Error` base
+  exception, so that now all pywbem specific exceptions are derived from
+  `Error`.
+
 pywbem v0.8.2
 -------------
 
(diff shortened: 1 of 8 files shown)


andy-maier:
+1

KSchopmeyer:
I want to take a little more time with this one.

andy-maier:
Rebased it to resolve merge conflict in changes.rst.

KSchopmeyer:
Reviewed, tested and I don't even have any comments this time. +1. Only thing I did not do was to go through pylint for new issues.

+1

PR #201, opened 2016-04-08
Fixes #200: Added testcases for MOF lexer, and fixed some lexer issues.

### LINKED ISSUE #200: Issues with use of PLY
In the course of migrating to the latest version of PLY, some things were not done accurately enough:

1. The comment warning about the docstring use by PLY should actually apply to all `t_*()`and to all `p_*()` functions.
2. the `t_error()` and `t_IDENTIFIER()` functions got docstrings with descriptions, but should not have one. If a documentation is needed at all, it should be done with block comments before the function.

andy-maier:
Ready for merge. The testcases are just a first stab, bt I suggest to merge this now and to expand them lateron.

Please review.

Details from the commit log:

- Extended the scope of the note about the PLY processing of docstrings in mof_compiler to all `t_*()` functions and variables (for LEX) and to all `p_*()` functions (for YACC).
- Removed human readable docstrings from `t_IDENTIFIER()`, `t_error()`, and `p_error()`, because PLY looks at them. They did not seem to have caused a problem, but I wanted to be on the safe side. Changed the docstrings into block comments.
- Enabled `MOFCompiler` to work without an associated repository. This ability is used in the standalone testcases for the lexer.
- Added standalone testcases for the lexer to `test_mof_compiler.py`, by adding base class `BaseTestLexer` and test classes `TestLexerSimple` for a few very simple cases, and `TestLexerNumber` for numbers.
- Changed `t_*` variables into functions, because PLY honors the order for functions, but sorts the symbols by decreasing regexp string length. Moved `t_decimalValue()` after all other numeric formats, because it is the least specific of all number types (binary has the trailing B/b, octal has the leading zero, float has the dot). Having decimal in front if binary e.g. causes '1b' to be treated as decimal '1' followed by invalid identifier 'b', etc. Or octal '07' is treated as decimal '0' followed by decimal '7' if decimal is in front of octal. Such things can still be handled at the YACC level, but it seems more logical to fix the lexical analysis, if only for better error messages.
- In `t_binaryValue()` and `t_octalValue()`, expanded the regexp to match all digits 0-9, so that the entire token for the number is processed, and the error is now handled in the match routine.


andy-maier:
Cases that were fixed:

- binary numbers were treated as decimal number "0" for each leading zero,
  followed by the remaining digits as one decimal number, followed by
  identifier "b":
  - "0b" -> "0" + "b"
  - "01b" -> "0" + "1" + "b"
  - "0011b" -> "0" + "0" + "11" + "b"
  - "11b" -> "11" + "b"
- octal numbers were treated as decimal number "0" for each leading zero,
  followed by the remaining digits as one decimal number:
  - "07" -> "0" + "7"
  - "0012" -> "0" + "0" + "12"
- invalid binary and octal numbers with digits out of range were treated
  as decimal numbers:
  - "2b" -> "2" + "b"
  - "08" -> "0" + "8"

KSchopmeyer:
+1   agree. We can also use these tests to expand the coverage.

andy-maier:
+1. merged.

PR #206, opened 2016-04-11
Fixes #171: Fixed errors and added tests for MOF lexer in mof_compiler.

### LINKED ISSUE #171: char16 typed properties not handled in response
The `tocimobj()` function raises ValueError if it gets called with a type name 'char16'.
It is not clear why this was done that way, I think it can just be implemented to return a unicode string, like we do for the CIM datatype string.

andy-maier:
Ready for merge.
Please review.

Details from the commit log:

- In `tocimobj()`, values with CIM type name 'string' were just returned. This did not step up to the promise of returning the CIM-XML representation, which must be a unicode string. Added code that ensures that the returned value is a unicode string, even when providing a byte string as input.
- CIM datatype char16 was not supported in `tocimobj()` and in the MOF compiler. Added support for char16 datatype.
- If the MOF compiler encounters an unknown alias name, it raises a CIMError with status code CIM_ERR_FAILED and status description `"Unknown alias: <alias-name>"`. The alias name was taken from the wrong tuple item, causing another exception to be raised. Fixed this bug.
- Added a MOF unit test class `TestTypes` to `test_mof_compiler.py`, and a corresponding test MOF `testsuite/testmofs/test_types.mof`, which test a CIM class that has properties of all data types.
- Added testcases for hex numbers to test class `TestLexerNumber` in `test_mof_compiler.py`.
- Added test classes `TestLexerString` for CIM string data types, and `TestLexerChar` for CIM char16 data types.
- Extended `testsuite/unittest_extensions.py` with support functions that assert equality of CIM classes, CIM instances, properties, methods, parameters, qualifiers and qualifier types. This is used by the new test class `TestTypes`.


andy-maier:
+1

andy-maier:
Rebased to resolve merge conflicts.

andy-maier:
Something went wrong when rebasing this. This PR is closed (merged), contains no commits, and the master branch does not have the changes it ought to have according to this PR. I saw that I still have the change locally. Trying to re-establish this correctly.

andy-maier:
PR #221 re-established this change.

PR #212, opened 2016-04-13
Fixes #208, fixes #217: Added tocimxmlstr() to all CIM object classes and as a global function.

### LINKED ISSUE #208: Add tocimxmlstr() to all CIM object classes
As a result of the discussion in issue #195, I suggest to add a new method to all CIM object classes, that returns a unicode string of the CIM-XML representation of the CIM object. It basically invokes the extsting `tocimxml()` method on the object, followed by `toxml()`, or `toprettyxml()`, depending on the value of its boolean `pretty` parameter.

The existing `tocimxml()` methods would remain part of the external interface, in case someone needs to do more postprocessing on the output (like aggregating multiple results; this is done much better at the DOM level).

### LINKED ISSUE #217: wbemcli add the function for prettyprint of xml
We have the wrong function definition for the prettyprint of xml in the h() definitions.

if we are to use the prettyprint the function is toprettyxml() whereas the single line is toxml()

However, it will probably all be resolved by the the new tocimxmlstr so maybe the alternative is to
just supply that that as the definition in the help.

andy-maier:
Ready for merge. Contains test code.

Please review, particularly the interface to the new `tocimxmlstr()` (on each CIM object class, and as a global function).


andy-maier:
+1

KSchopmeyer:
-1

I hit a problem.  using wbemcli I did:

cl = gc('CIM_ManagedElement')
print cl.tocimxmlstr()
displays single line 

HOWEVER

print cl.tocimxmlstr(intent=4)

>>> cl.tocimxmlstr(indent=4)

causes crash:

  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_obj.py", line 1840, in tocimxmlstr
    return tocimxmlstr(self, indent)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_obj.py", line 3549, in tocimxmlstr
    xml_str = xml_elem.toprettyxml(indent=indent)
  File "/usr/lib/python2.7/xml/dom/minidom.py", line 60, in toprettyxml
    self.writexml(writer, "", indent, newl)
  File "/usr/lib/python2.7/xml/dom/minidom.py", line 814, in writexml
    node.writexml(writer, indent+addindent, addindent, newl)
TypeError: cannot concatenate 'str' and 'int' objects

KSchopmeyer:
I did not dig into why this is happening yet and maybe I am doing something wrong.

KSchopmeyer:
This does work with the original toprettyxml():

here is a piece
	<PROPERTY NAME="ElementName" TYPE="string">
		<QUALIFIER NAME="Description" TRANSLATABLE="true" TYPE="string">
			<VALUE>A user-friendly name for the object. This property allows each instance to define a user-friendly name in addition to its key properties, identity data, and description information. 
Note that the Name property of ManagedSystemElement is also defined as a user-friendly name. But, it is often subclassed to be a Key. It is not reasonable that the same property can convey both identity and a user-friendly name, without inconsistencies. Where Name exists and is not a Key (such as for instances of LogicalDevice), the same information can be present in both the Name and ElementName properties. Note that if there is an associated instance of CIM_EnabledLogicalElementCapabilities, restrictions on this properties may exist as defined in ElementNameMask and MaxElementNameLen properties defined in that class.</VALUE>
		</QUALIFIER>
	</PROPERTY>
	<PROPERTY NAME="Description" TYPE="string">
		<QUALIFIER NAME="Description" TRANSLATABLE="true" TYPE="string">
			<VALUE>The Description property provides a textual description of the object.</VALUE>
		</QUALIFIER>
	</PROPERTY>
</CLASS>

KSchopmeyer:
Whoops. I tried to copy in an example of formatted output but it did not transfer too well.

andy-maier:
I inserted 4 spaces on each line, that made it show reasonably.

KSchopmeyer:
Note that I tried cl.tocimxml().toprettyxml().

That produces good output.

andy-maier:
Karl, the `indent` argument must be a string. See the description. It's not my invention, it's how the underlying `toprettyxml()` works.

    >>> from pywbem import *
    >>> c1 = CIMClass('EX_C1')
    >>> p1 = CIMProperty('p1',value=42,type='uint32')
    >>> c1.properties['p1'] = p1
    >>> print(c1.tocimxml().toxml())
    <CLASS NAME="EX_C1"><PROPERTY NAME="p1" TYPE="uint32"><VALUE>42</VALUE></PROPERTY></CLASS>
    >>> print(c1.tocimxmlstr())
    <CLASS NAME="EX_C1"><PROPERTY NAME="p1" TYPE="uint32"><VALUE>42</VALUE></PROPERTY></CLASS>
    >>> print(c1.tocimxmlstr(indent='    '))
    <CLASS NAME="EX_C1">
        <PROPERTY NAME="p1" TYPE="uint32">
            <VALUE>42</VALUE>
        </PROPERTY>
    </CLASS>
    >>> print(c1.tocimxmlstr(indent=4))
      File "<stdin>", line 1, in <module>
      File "pywbem/cim_obj.py", line 1840, in tocimxmlstr
        return tocimxmlstr(self, indent)
      File "pywbem/cim_obj.py", line 3549, in tocimxmlstr
        xml_str = xml_elem.toprettyxml(indent=indent)
      File "/usr/lib64/python2.7/xml/dom/minidom.py", line 60, in toprettyxml
        self.writexml(writer, "", indent, newl)
      File "/usr/lib64/python2.7/xml/dom/minidom.py", line 817, in writexml
        node.writexml(writer, indent+addindent, addindent, newl)
    TypeError: cannot concatenate 'str' and 'int' objects

andy-maier:
You probably used the default values for the `toprettyxml()` arguments, which causes a tab string `"\t"` to be used for `indent`.

andy-maier:
Rebased.

andy-maier:
Addressed issue #217 in last commit.

KSchopmeyer:
you are right.   I should have dug further.  If we were to do this correctly we would probably make in an integer or string and if was an integer, convert to spaces, in particular because we do not like tabs anywhere any more.  But this is primarily just a display feature for us and apparently the whole toprettyxml has a really  bad reputation anyway so lets just go ahead.

+1

andy-maier:
Karl, I think specifying the indentation as an integer is a quite natural expectation, and I added support for that, as well as test code.

Please review that addition.

Also, the last commit updates the change log.

andy-maier:
+1 from me for this last change.

KSchopmeyer:
+1. Yes this is clearly more logical with the integer. I can apply the idea elsewhere to clean up some code.

PR #218, opened 2016-04-14
Fixes #198: Changes for CIMError exception.

### LINKED ISSUE #198: Issues with CIMError description
The `CIMError` description has these issues:
1. it states that there is a third argument and tuple element: `exception_obj`. That argument is not used anywhere in the code. I added that in order to have a mechanism for including lower level exceptions that caused `CIMError` to be raised, but this mechanism was never used in the code.
2. Instead of using this mechanism, I think the whole direction is moot: We should rather reduce the use of `CIMError` for failed operations, and add a new exception class for errors detected in the client. Currently, the client raises CIMError also for self-detected errors.
3. The description is vague as to where exactly the tuple value is in the exception instance. Update the description to say it is in the 'args` instance variable.
4. The fact that we supply a tuple when creating instances of `CIMError` causes the standard instance variable `message` to be empty, but that is not stated anywhere (also not in the descriptions of the standard `BaseException` or `Exception` classes). Because that is not obvious, and because `message` is a standard instance variable of all exceptions, that should be stated explicitly.
5. For the same reason, `str()` returns the tuple (e.g. `"(5, 'instance not found')"`) and not the message.
   
   **Discussion:** Should we implement `str()` such that it returns a string that describes the error, e.g. `"<error_code_name>: <description>"`? That requires a new external function that translates numeric CIM status codes into a string that is their symbolic name. Something similar for messages already exists with `_errcode2string()` in `mof_compiler`.

andy-maier:

Details:

- Removed unused third tuple item `exception_obj`.
- Added properties to `CIMError` for accessing CIM status code and CIM status description.
- Moved `_errorcode2string()` from mof_compiler.py to cim_constants, as `_statuscode2string()`.
- Added `_statuscode2name()` to cim_constants.


KSchopmeyer:
+1 for change.  This one needs entry in change log

andy-maier:
Updated the PR with the change log update.

PR #221, opened 2016-04-14
Fixes #171 (try 2): Fixed errors and added tests for MOF lexer in mof_compiler.

### LINKED ISSUE #171: char16 typed properties not handled in response

andy-maier:
This PR re-establishes the change that was lost by accident in PR #206.

Ready for merge. Please review.

Details, from the commit log:

- In `tocimobj()`, values with CIM type name 'string' were just returned. This did not step up to the promise of returning the CIM-XML representation, which must be a unicode string. Added code that ensures that the returned value is a unicode string, even when providing a byte string as input.
- CIM datatype char16 was not supported in `tocimobj()` and in the MOF compiler. Added support for 'char16'.
- If the MOF compiler encounters an unknown alias name, it raises a CIMError with status code CIM_ERR_FAILED and status description 'Unknown alias: <alias-name>'. The alias name was taken from the wrong tuple member, causing another exception to be raised. Fixed this bug.
- Added a MOF unit test class 'TestTypes` to `test_mof_compiler.py`, and a corresponding test MOF `testsuite/testmofs/test_types.mof`, which test a CIM class that has properties of all data types.
- Added testcases for hex numbers to test class `TestLexerNumber` in `test_mof_compiler.py`.
- Added test classes `TestLexerString` for CIM string data types, and `TestLexerChar` for CIM char16 data types.
- Extended `testsuite/unittest_extensions.py` with functions that compare CIM classes, CIM instances, properties, methods, parameters, qualifiers.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -3519,7 +3519,7 @@ def tocimobj(type_, value):
     # String type
 
     if type_ == 'string':
-        return value
+        return _ensure_unicode(value)
 
     # Integer types
 
@@ -3558,7 +3558,7 @@ def tocimobj(type_, value):
     # Char16
 
     if type_ == 'char16':
-        raise ValueError('CIMType char16 not handled')
+        return _ensure_unicode(value)
 
     # Datetime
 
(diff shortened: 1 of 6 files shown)


andy-maier:
+1

KSchopmeyer:
+1.  I have one comment.  Would you put a TODO on the section of the test_types.mof with the instance definitions that is commented out, for our future memory.

andy-maier:
Done. Merging.

PR #224, opened 2016-04-15
Make HTTP request method configurable.

skolsuper:
Allow injection to WBEMConnection of a class that handles the nuts
and bolts of an http request.

I have included the minimum of required changes for this PR, although github seems to have messed up the diff pretty badly so it looks like a big change, all I have done in `cim_http` is indent the `wbem_request` function to make it into a method, and add a class declaration.

Possible additional changes in the future are to hoist the `*HTTPConnection` class definitions that are nested inside the `wbem_request` function for no apparent reason, and to split the `PyWBEMRequester` into `HTTPRequester` and `FileRequester` to simplify some logic, at the expense of the user needing to know which kind of url they will use ahead of time.

### DIFF
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -249,458 +249,475 @@ def get_default_ca_certs():
             get_default_ca_certs._path = None
     return get_default_ca_certs._path
 
-# pylint: disable=too-many-branches,too-many-statements,too-many-arguments
-def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
-                 verify_callback=None, ca_certs=None,
-                 no_verification=False, timeout=None):
-    # pylint: disable=too-many-arguments,unused-argument
-    # pylint: disable=too-many-locals
+
+class BasePyWBEMRequester(object):
     """
-    Send an HTTP or HTTPS request to a WBEM server and return the response.
-
-    This function uses Python's built-in `httplib` module.
-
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
We feel that we are trying to  create a controlled working environment
and would rather not interject user options such as this at this level of the code
now but to get this level of the code clean, correct, and thoroughly
tested so that there is no need for these options. The problem we would
have with this is that it introduces more options at a level of the code
we are trying to stabilize.

We really appreciate your help in this and encourage you to continue to
contribute and critique.  However, at this point we are uncomfortable
taking the approach you propose.

At the same time, we do agree that we have a real issue with the security functionality:

1. different libraries for different versions of python
2. The libraries themselves offer different levels of security with different versions if python. Thus, for example, the ssl module has security level changes depending on your python 3 level.

Our top priority right now is to finish up the key things we had to do for 0.9.0 (bring pywbem functionality up to the spec (add push operations and indication subscription/listen support and its testing) and the ssl/tls cleanup is right after that.

skolsuper:
Hi Karl,

Thank you for taking the time to review this PR and provide feedback.  Very glad to see the care you're taking with this package and the improvements you are adding.

I am excited to hear of push operations/subscription support in the pipeline, I didn't realise something like this was possible, currently we are relying on short polling to check for changes :(.  I look forward to those features landing.

James

PR #227, opened 2016-04-15
Fixes #210: Deprecate ordering comparisons for NocaseDict and CIM objects

### LINKED ISSUE #210: Ordering of NocaseDict objects
The NocaseDict class has ordering operators, since v0.8.0 based upon the rich comparison functions.

The ordering behavior is quite unconventional, though:

    >>> from pywbem.cim_obj import NocaseDict as nd
    >>> d1=nd(a=1,b=2)
    >>> d2=nd(a=1,c=3)
    >>> d1==d2
    False
    >>> d1!=d2
    True
    >>> d1<d2
    True
    >>> d1>d2
    True
    >>> d1<=d2
    True
    >>> d1>=d2
    True
    >>> d2>=d1
    True
    >>> d2>d1
    True

The NocaseDict description says about the ordering (in `__lt__()`):

`self` is less than `other`, if:
* a key in `self` is not in `other`, or
* the value for a key in `self` is less than the value for that key in `other`, or
* `self` has less key/value pairs than `other`.

The implementation of `__lt__()` has the comparison wrong for the first rule above. There may be more issues.

The question is what should be implemented: Total ordering, or subset (like for sets).

andy-maier:
Fix issue #210.

Work in progress, not ready to be merged.

At this point, this PR contains improved testcases for dictionary ordering (ordering between two dictionaries, that is), that verify the ordering behavior between two standard Python dictionaries. The corresponding test runs for NocaseDict are disabled.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -24,6 +24,10 @@ Deprecations
   of the `CIMParameter` class, because that class represents CIM parameter
   declarations, which do not have a default value.
 
+* Deprecated ordering comparisons for NocaseDict, CIMInstance, CIMInstanceName,
+  and CIMClass objects. This affects the ordering comparisons between two such
+  objects, not the ordering of the items within such a dictionary.
+
 Clean Code
 ^^^^^^^^^^
 
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -56,6 +56,9 @@ in a method).
 Except for the case-insensitivity of its keys, it behaves like the built-in
 :class:`py:dict`. Therefore, :class:`~pywbem.NocaseDict` is not described
 in detail in this documentation.
+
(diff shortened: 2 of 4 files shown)


andy-maier:
Interesting: Python 3 refuses ordering comparisons between dictionaries with a `TypeError:  unorderable types`. See the logs of the Travis check runs for Python 3.

andy-maier:
This PR is now reaedy to be merged.
Please review.

andy-maier:
+1

andy-maier:
ping

KSchopmeyer:
strange word (bt) in one comment in __lt__
        # a case sensitive comparison, bt that will be better than the faulty

Other than that +1

andy-maier:
Fixed the comment.

Issue #223, opened 2016-04-14
Class mof output does not output default property values


KSchopmeyer:
Apparently they never put the code in to output any default values for generation of class level MOF output.

We will fix this by a moving the instance property mof output to the CIMProperty class (see another issue and pr) and then extending it so it will output mof for either instances or classes (there are some minor differences in the output.

andy-maier:
Fixed by PR #260.

PR #230, opened 2016-04-15
Fixes #151: Add tomof to cimproperty and connect to ciminstance

### LINKED ISSUE #151: CIMProperty does not have its own tomof function.
Even CIMParameter has its own mof.  However, CIMProperty tomof is actually buried in CIMClass and CIMInstance.   For clarification and usability the tomof should be pulled out to be placed on in CIMProperty

KSchopmeyer:
Adds function tomof() to CIMProperty instead of burying code with CIMInstance.  Will connect this to CIMClass in future (issue filed).

Ready for review.


andy-maier:
-1

Comments:

* I don't agree with the comment "# this should never happen" on line 2499: `None` is the normal representation of NULL.
* The docstring of `tomof()` should be adjusted to the Google format. See e.g. `tomof()` in `CIMParameter`.
* Update the change log (`docs/changes.rst`).

KSchopmeyer:
1. You are right.  I reorged the code several times and at one point I tested null before the call.  Removed.

2. Docstring.  Corrected.

3. Done.

Note that the isInstance variable is not used yet.  Will finish that in separate pr for issue #223 or new issue.

PR #234, opened 2016-04-18
Fixes #233: Correct error in cimclass.tomof() where array attribute ([]) not output

### LINKED ISSUE #233: cim_obj.py, CIMProperty init forgets is_array flag when true
Input of a property definition in a class with the is_array definition = True does not show up in mof output for the class.   

Thus, the input 
        cl = CIMClass(
            'CIM_Foo',
            properties={'Uint32Array': CIMProperty('Uint32Array', None,
                                                   type='uint32',
                                                   is_array=True)})

produces mof output of

class CIM_Foo {
    uint32 Uint32Array;

This is an error in the mof output itself since the repr of the class show it is persisted in the class.

KSchopmeyer:
Corrects error and adds tests to test_cim_obj.py and test_mof_compiler.py.


KSchopmeyer:
Note that this is issue #233

andy-maier:
Reviewed the code.
+1

KSchopmeyer:
+1

PR #235, opened 2016-04-18
Fixes #232: Fixed copy&paste error in assertRegexpMatches().

### LINKED ISSUE #232: unittestextensions.py.  Two functions do exactly the same thing
The functions assertRegexpContains and assertRegexpMatches do the same thing. They both call re.search.  This means that they are looking for the pattern in a multiline string.

Note that unittest uses assertRegexpMatches but could not find assertRegexpContains.

We actually use the re.matches in some of the unittests where it is looking for a specific pattern on the first line, not just anywhere in the file. The mof output tests.  Actually there are probably much better ways of doing these tests but they exist for the moment.

andy-maier:
Ready to be merged.
Please review.

### DIFF
--- a/testsuite/unittest_extensions.py
+++ b/testsuite/unittest_extensions.py
@@ -47,7 +47,7 @@ class RegexpMixin(object):
         Raises:
             AssertionError if not matched
           """
-        if not re.search(pattern, text):
+        if not re.match(pattern, text):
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

KSchopmeyer:
+1. I do love a major change :-)

PR #236, opened 2016-04-18
Fixes #156: Moved script-ish parts of wbemcli and mof_compiler modules into the corresponding scripts.

### LINKED ISSUE #156: Merge scriptish parts of wbemcli.py and mof_compiler.py into their scripts?
For both wbemcli and mof_compiler, there is a Python script with that name in the root directory, and a module in the pywbem directory.

In case of wbemcli, the module is really completely a script, because it does not have an API. We could move all of the module into the current script, and remove the module.

In case of mof_compiler, the module has some parts that are related to its use as a script, and some others that make sense as a module. We could move the scriptish parts into the current script.

The advantage would be that we can generate the API documentation of the remaining mof_compiler module more cleanly, and depending on how we come up with a way to document the scripts, we could do that consistently from their script files.

andy-maier:
Ready to be merged.
Please review.

Details from the commit log:

- Moved all of the `pywbem/wbemcli.py` module into the `wbemcli' script. This caused the `pywbem/wbemcli.py` module to be empty, so it was removed.
- Moved the `main()` function from the `pywbem/mof_compiler.py` module into the `mof_compiler` script.
- Updated the module docstring of both `wbemcli` and `mof_compiler` scripts to be a one-line description, plus hinting at the `--help` option.
- Added an interactive usage example to the `h()` help function of the `wbemcli` script.
- Updated the `makefile` to reflect that the `pywbem/wbemcli.py` module was removed.


andy-maier:
+1

KSchopmeyer:
+1. 

Intellectually I am unsure what we gain or lose over moving all of the wbemcli into a "non-python" sufixed file but I I do not have any hard arguments except the possibility that somewhere, someone will forget that this is really python and that is a real fuzzy argument.

Somewhere before we do many more utilities, we probably need to clarify our rules and pythonic rules for this.

andy-maier:
The hash-bang in the first line states it is python. At least the shell is clear about it ;-)

I agree that we need to come up with a strategy for this when we do more utilities.

Right now, the strategy I was using is (and we can debate that):

- Anything that could be useful as a module goes into a module
- Command line argument processing, plus all of the rest related to the script (i.e. that is not useful as a module) goes into the extension-less script file, whose hash-bang states it is Python.
- There is a .bat file for Windows that invokes python with the suffix-less script of the same name.

That strategy provides for:

- commands with a clear command name without being tied to a suffix revealing the implementation language
- same command name whether on Linux or Windows
- as a result, the same single documentation applies to both
- nicely supported by Python setup scripts (both the suffix-less file and the Windows .bat file get installed into the script directory and thus are available in the PATH).

I do realize that these benefits can also be achieved by what we had before this PR, but it allowed running the module as a script to have a command, and now users are forced to use the suffix-less command. Maybe that is the real reason.

andy-maier:
I have created issue #241 for the tool strategy, so I am merging this PR.

PR #237, opened 2016-04-18
Fixes #66: WBEM Listener / Server API

### LINKED ISSUE #66: Get CIM Listener officially supported
* Review the API, and change/extend as needed
  - Any Python packages for registration that we could/should use?
* Add API documentation
* Add/extend unit test concept
* How to package
   - Move from sibling irecv package into pywbem package namespace?
* Test with real WBEM servers

andy-maier:
This is an initial stab at addressing issue #66.

Please review and post comments here.

The docs for this `listener` branch are built automatically on RTD, so that the WBEM Listener API can better be reviewed:

https://pywbem.readthedocs.org/en/listener

Note there are now two new sections:

* WBEM server API
* WBEM listener API

In addition to the documentation, there is now a tool that demonstrates how to use the WBEM server API: `tools/explore.py`. Invoke without arguments for help.


andy-maier:
Removed the "WIP incomplete" from the title, to make it clear that this is ready for review.

KSchopmeyer:
Initial review comments without testing.

1. Lets pull the interop find functions out of here and put them into pywbem as a higher level interface component. They should be standard for the client.

This also ties to the wbemServer class.  As you describe it it is things relavent to the listener but the
only relevance is the interop namespace. We could generalize it a bit and include, for example, all namespaces.  This is really the beginning of the other thing I am getting started, the infrastructure part of the next level client API.

2. Getting old. Did not understand the followiing:

The scheme determines whether the WBEM server uses HTTP or HTTPS


So far, I like it.  I need to sort through some of the functionality more (that usually means trying to build one) but this is on the right course.

andy-maier:
On Karl's comments:

* On 1: Good catch. WIll do that.
* On WBEMServer class: I'd like this to be just what the listener needs to know about the server. This means it would not be extended because other uses of the class outside of the listener context could benefit, but instead some other class would emerge somewhere. I'm just afraid of function creep here, and would like to start small with just what we need.
* On the second 1: "scheme" means the URI scheme, e.g. "http" or "https". Will change text to "URI scheme".

andy-maier:
I established automatic docs build for this `listener` branch, so that the WBEM Listener API can better be reviewed:
https://pywbem.readthedocs.org/en/listener/

andy-maier:
It is now much more complete, and includes a usage example.

Please review closely and comment here.

Best way to review the external API is to read the documentation at https://pywbem.readthedocs.org/en/listener/listener.html#wbem-listener-api

KSchopmeyer:
I really do think that we are going to duplicate here. To do the next level of components, we would start with roughly the same thing, the wbemserver and add methods to that to acquire information about the characteristics of the server and execute against the server.

Thus, are we not going to have wbem server classes in a lot of places.

Thus, the concepts of getting the interop namespace is core to a bunch of things, not just the listener but this means that either we go to this code to get the this namespace and other code to get all namespaces or we effectively duplicate code.

Consider that I want to do profiles from the client. Does that mean that we add another wbemserver somewhere else??  Maybe for profiles we add a DMTF profile class with the get profiles, etc. but does that stay separate or just sit on its own.

The reason that am bringing this up now is that we are really moving from the very low level interfaces to the next level of support apis and it is worth getting it right the first time.

andy-maier:
We agreed in the call on:

* Expand the WBEMServer class into a generally useable thing, move to `pywbem.server` namespace.
* A built-in listener thread and a demon/service based implementation are two separate implementations. Try to find a common subscription mgmt interface, and move commonly useable functionality on the WBEMServer if it makes sense there.

andy-maier:
Will restructure the listener API, based upon the previous comment, and based upon the Python namespaces described in PR #253.

andy-maier:
The current level of this PR expanded the 'WBEMServer' class into a generally useable thing in a new `pywbem.server` namespace. It also adds a `ValueMapping` class in that namespace, that helps translating between Values and ValueMap qualifier entries.

Review focus for this level of the PR should be on the new "WBEM server API" section in https://pywbem.readthedocs.org/en/listener

KSchopmeyer:
Some questions:
1. Why is all the code for WBEMserver in the server/__init__.py class. Took some looking to find it.  I assume that this is because this was prototype.
2. How did we ever let George build that crap he did into the RegisteredProfile.mof(his pseudo pull thing). What a mess.  But that is not our problem, just our frustration.
3. The good news is that it actually runs explore.py with pegasus 2.15. Have not tested with ssl since right now I am vpn'd to the smi lab so can only access local server through localhost (back door)
4. Probably should do "OpenPegasus" for OpenPegasus. That is a nit.
5. We will probably want to expand the profiles to provide for autonomous profiles, subprofiles, etc. at requestor option but we can do that after initial work is set.
6. Wonder if something like find_interop_ns would be better than determine_interop_ns? Shorter to write.  same for the profiles, namespaces, etc.
7. I have not been able to go through the ValueMapping in detail yet.  Generally I like it but I need to create some tests and play to be comfortable.  I would be willing to check in as it stands and we can do more later.
8. Will start subscriptions now and have more comments.

KSchopmeyer:
The comment in intro

 +* WBEM server
+
+  The WBEM server represents a general access point for basic server
+  functionality to a client (i.e. it does not run as a WBEM server).

The directory is server but we call it wbem server here.

Should we be calling these things APIs.???

Suggest the following wording
The server api is a WBEM Client access point to basic WBEM server functionality such as
namespace information, profile information, and indication subscription information. It sits on
top of the pywbem connection APIs.

KSchopmeyer:
listener documentation.  The documentation for the WBEM Listener API states that:

 +The `WBEM listener API`_ is provided by the :mod:`pywbem.listener` module.
+
+It provides functionality for managing subscriptions for indications from
+one or more WBEM servers, and implements a thread-based WBEM listener service.

I believe that the managing subscriptions has been moved to the WBEM server.

Same question as server, why code in __init__.py???

I have not stepped through the whole listener but am I correct in that you have subscription code in both the listener and server???

We are going to have to explain dynamic filter I bet.

remove_dynamic_filter

Remove a dynamic indication filter from a WBEM server, and forget it for
+        cleanup.
+
+        This function deletes an indication filter instance in the WBEM server.

Drop the first sentence/ The second one explains it all except that it should probably say
sends request to the server to delete an indication filter instance in the WBEM server.

For the moment the callback works.  I presume you can add multiple callbacks and the indications will be delivered to all of them.

At some point we are going to have to add filtering to that so that indications can be directed by content.

That is as far as I have gotten but this is really very complete and we should either create an experimental branch so we can work on it or do some cleanup and commit soon.

I think this is all on the right course.

andy-maier:
Karl, on your comments:

Your first comment:

* On 1. > Why is all the code for WBEMserver in the server/init.py class.
  Using submodules in the server submodule creates one more sub-namespace. So the only other option would be to move `pywbem/server/__init__.py` into `pywbem/server.py`. Which is a transparent change for users, and we can do that if the module stays as simple as it is right now.
* On 2. > pseudo pull thing in RegisteredProfile.mof.
  This was for cases where there are too many central instances, so we went there for the same reason we introduced pull at the operation level. I insisted back then not to mix operation-level pull and method-level pull, that's why we have a consistent set of pull methods on the class, that are modeled after the operation level pull operations. So I have my share in it ;-)
So the pull portion of it has its motivation even when it does not look nice.
When looking at it, one question for me is the part about selecting central vs scoping instances. Also, these methods have not been embraced by an update of the PRP.
* On 4. > Probably should do "OpenPegasus" for OpenPegasus.
  We can do that. The `server` attribute is intended to be a machine readable string. Should we have a human readable string in addition?
* On 5. > We will probably want to expand the profiles to provide for autonomous profiles, subprofiles, etc. at requestor option...
  Agreed. It is not easy to find out, so there is value in providing an easy to use function for that.
It should also cover wrapper specs. I looked at the RP/RS classes and there is no property available for distinguishing autonomous and component. There is `SpecificationType` for distinguishing initiative specs from profiles.
* On 6. > Wonder if something like find_interop_ns would be better than determine_interop_ns?
  Before we discuss their name, there is a more basic question on them: These methods are called under the covers when the various instance properties are accessed. So normally, they don't need to be called by users at all. Except if they want to trigger a refreshing of the corresponding property data.
  I see these options:
  - Remove them from the public API, not providing a measn to refresh the data.
  - Remove them from the public API, and add some function that refreshes the whole server instance (I don't think it is worthwhile to be able to refresh with a higher granularity).

Your second comment:

* On intro for WBEM server API:
  Agreed, will use your text for an update.

Your third comment:

* On: > managing subscriptions has been moved to the WBEM server:
  Both have their share in it. The WBEM server API provides low level functions for creation of instances, that are not for the user. The WBEM listener API provides the rest of it, at the level the user wants to see. I'm not happy with that split either.
  I see these options:
  - Leave as is, maybe document better that the WBEM server creation functions are low level functions used by the listener api.
  - Make the low level functions non-public but leave on server.
  - Make the low level functions non-public and move to listener.
  - Move to listener.
* On need for explaining dynamic filters:
  I thought I did that. Let me know what exactly does not work.
* On > Remove a dynamic indication filter from a WBEM server...
  Agreed it needs to be simplified.
* On > I presume you can add multiple callbacks and the indications will be delivered to all of them.
  Correct.
* On > At some point we are going to have to add filtering to that so that indications can be directed by content.
  DIfferent indications to a different callback? I thought about that already and that is not easy. That basically means we would introduce a concept of a subscriber that is below the listener. Listeners subscribe to servers, and subscribers subscribe to the listener. The whole filter approach already does that at the level of listener to server. It would take something similar complex to have a distribution mechanism for the level subscriber to listener.
  If you can think of something easy there, let me know. But otherwise, I'd like to avoid the complexity.
  A client host can have multiple listeners against the same server, each of which would listen on a different port and thus be considered a different listener destination for the server. Then, each of these listeners can have different filters and subscriptions, so at that level there is selection possible.

I'll start working on an update.

andy-maier:
On comment about intro for WBEM server API:

Re-reading the intro, I found that it discusses *components* (e.g. "WBEM server"). Some of them have APIs (e.g. "WBEM server API"). The APIs are mentioned and linked in the description.

So your comment basically means to drop the notion of a component and only talk about APIs. The two commands do not have APIs, so they would still need to be listed separately. The further chapters in the spec are descriptions of the APIs.

Let's discuss whether to drop the notion of components in the intro, replacing it with APIs.

KSchopmeyer:
I think our commenting process is getting out of hand. :-) this last round is several pages long

On 1. My guess is that it will not stay simple.  There is more to this class that we are not covering and not even thinking about covering so far.

On 2. It just always bothers me that somehting that should be simple like a profile class is pages long. Nothing we can do now.  But our mantra should be simple.

On 4:  Think we probably should see what many users do before we go to machine readable string.  My only reason for OpenPegasus is that generally we got beat up because Pegasus is used by so many people and  copywrited elsewhere so when in public we tried to stick to OpenPegasus.
More lawyers.

on 5:  (I looked at the RP/RS classes and there is no property available for distinguishing autonomous and component. ) Now we know why we should have done this years ago :-_). For me never again, models without implementation.

Lets leave as is for now but maybe option to get both profiles and subprofiles. (i.e get_subprofiles= False)

on 6)  I recognized that part way through.  That makes things like my suggestion on 5 more difficult.
Would like to discuss the remove: refresh thing
DISCUSSION: Should we have possibility to directly call these APIs, a general refresh, or specific refresh.  Note that while right now, the instance count is small, I wonder what happens when we get subscriptions, possibly jobs, etc. into the mix (wonder, not sure)

on subscripions apis in the server.
DISCUSSION ITEM: Friday

Dynamic filters: No problem, I just always worry each time we bring in new term.

indication routing by callback.  While I know it is not easy and I don't propose right now, we know that we will probably be driven there.  The alternate is separate ports which is our current solution.

Discussion: the issue of APIs vs components.

KSchopmeyer:
on 1 - Leave in init class for now.

on 2 - Ignore pull in profiles since no servers support them

on 4) -  For now we are not guaranteeing the data, just providing information from the fields in the wbemserver profile.  We need further sort out if there is a human readable thing. We also agreed to stay with the term brand.

5) autonomous is only identified through associations.  For now we return a list of autonomous profiles and later extend this to allow an optional list of component profiles (in the context of their encompassing profiles.

andy-maier:
With commit 91f5d45, this PR is ready to be merged.
Please do a final review.
Both the WBEM server API and WBEM listener API are still marked as experimental.

andy-maier:
+1

KSchopmeyer:
We agreed that we will eliminate the server and listener sub namespaces.

andy-maier:
Reoved the subnamespaces `pywbem.server` and `pywbem.listener`. Their symbols are now available in `pywbem`, and their module names are now non-public.

Also, rebased on master (unintentionally, but does not matter).

andy-maier:
+1

KSchopmeyer:
1. It is commented out but there is at least one commented logging calls in _listener.py. What are we going to do with them, add logging as an option or remove them.  Actually for a listener, logging as a option is logical.

2. Why is stop not implemented.

3. Is there any advantage to identifying as part of the names for filters and destination instances the fact that these are pybem identified filters, etc. or even more detailed, some id that the user controls?

4. Probably should explain what the dodgy certificate is in _HTTPSServerContextfactory
--- _Server.py

    def _determine_interop_ns(self):
        """
        Determine the name of the Interop namespace of the WBEM server, by
        communicating with it and trying a number of possible Interop

suggest ... by trying to communicate with it on a number of possible..

5. This function use CIM_Error_Not_found to communcate that there is no interop.  Probably that should be noted in the Raises: specifically as the way to indicate that there is no interop namespace.
Other exceptions indicate other possible errors I assume.

6. I think that after this commit we may want to expand the validate_interop but not sure yet exactly what should be the minimum to insure that this is an interop ns.

7. determine_namespaces.  The algorithm for __namespaces namespace names is completely different than for CIM_Namespace.  CIM_Namespace carrys the namespace in the name property.

I have to look up the __namespaces but it builds up the components of the namespace class. I know since I did it wrong a long time ago and had to rewrite it for Pegasus cimcli.

8 Suggest we make getting the namespaces from _Namespace a separate function so it can be tested separately. Algorithm from Pegasus
/*
    Use the __namespace class to attempt to get namespace names.  Returns
    an array containing namespaces found. Used by _getNameSpaceNames(...)
*/
Array<CIMNamespaceName> _getNameSpacesWith__namespace(Options& opts)
{
    Array<CIMNamespaceName> namespaceNames;
    CIMName nsClassName = CIMName("__namespace");

    // TODO Determine if we really need this statement
    opts.nameSpace = PEGASUS_NAMESPACENAME_INTEROP.getString();

    // Build the namespaces incrementally starting at the root
    // ATTN: 20030319 KS today we start with the "root" directory but
    // this is wrong. We should be
    // starting with null (no directory) but today we get an xml error
    // return in Pegasus
    // returned for this call. Note that the specification requires
    // that the root namespace be used
    // when __namespace is defined but does not require that it be
    // the root for all namespaces. That  is a hole is the spec,
    // not in our code.

    // Determine why we need the following statement
    namespaceNames.append(opts.nameSpace);

    Uint32 start = 0;
    Uint32 end = namespaceNames.size();

    do
    {
        // for all new elements in the output array
        for (Uint32 range = start; range < end; range ++)
        {
            // Get the next increment in naming for all a name element
            // in the array
            Array<CIMInstance> instances = opts.client.enumerateInstances(
                namespaceNames[range],nsClassName);
            for (Uint32 i = 0 ; i < instances.size(); i++)
            {
                Uint32 pos;
                // if we find the property and it is a string, use it.
                if ((pos = instances[i].findProperty("name"))
                        != PEG_NOT_FOUND)
                {
                    CIMValue value;
                    String namespaceComponent;
                    value = instances[i].getProperty(pos).getValue();
                    if (value.getType() == CIMTYPE_STRING)
                    {
                        value.get(namespaceComponent);

                        String ns = namespaceNames[range].getString();
                        ns.append("/");
                        ns.append(namespaceComponent);
                        namespaceNames.append(ns);
                    }
                }
            }
            start = end;
            end = namespaceNames.size();
        }
    }
    while (start != end);

    return namespaceNames;
}


9. Minor thing but if we get namespaces the old way, could we set a flag somewhere since we will want to document somewhere this obsolete mechanism.

andy-maier:
It turns out that the listener had several flaws: No reasonable HTTPS certificaste and key file support, export messages not parsed, no proper export message response, etc.
Added all that in the latest commit.
This addresses Karl's latest issues 2 and 4.

Karl's other comments are still to be addressed.

andy-maier:
I added code to check for several things (HTTP headers and methods) to make the listener DSP0200 conformant.

Please review (in `_listener.py.do_POST()`).

KSchopmeyer:
explore: added the following just to reduce pylints:

--- /tmp/meld-tmpkWoKn2
+++ /home/kschopmeyer/pywbem/githubpywbem/pywbem/examples/explore.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+from __future__ import print_function, absolute_import
 import sys
 
 from pywbem import WBEMConnection, WBEMServer, ValueMapping
@@ -21,7 +22,8 @@
 
     print("Advertised management profiles:")
     org_vm = ValueMapping.for_property(server, server.interop_ns,
-        'CIM_RegisteredProfile', 'RegisteredOrganization')
+                                       'CIM_RegisteredProfile',
+                                       'RegisteredOrganization')
     indications_profile = None
     server_profile = None
     for inst in server.profiles:

Am hitting the following problem:

  DMTF Profile Registration Profile 1.0.0
  DMTF Indications Profile 1.1.0
Central instances for DMTF Indications profile (component):
Error: global name 'CIMInstanceName' is not defined
Central instances for SNIA Server profile (autonomous):
Error: global name 'CIMInstanceName' is not defined

Since this is not in explore.py I assume coming from _server.py but did not explore futher

KSchopmeyer:
Another question.  Why does the add_dynamic_filter use the server url as a parameter instead of the
server object that was defined and attached to the listener? The first thing the function does is try to lookup the server from the url. But then it uses the url to append the filter.

trying to create a pegasus create instance and listener example and may have some more questions.

KSchopmeyer:
Here is an update on the patches for the example:

--- /tmp/meld-tmp_Hve2E
+++ /home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_listener.py
@@ -24,12 +24,13 @@
         print("Received CIM indication: {0!r}".format(indication))
 
     def main():
-
-        conn1 = WBEMConnection('http://server1')
+        url1 = 'http://server1'
+        conn1 = WBEMConnection(url1)
         server1 = WBEMServer(conn1)
-        server1.determine_interop_ns()
-
-        conn2 = WBEMConnection('http://server2')
+        # NOT USED and determine is private I thinkserver1.determine_interop_ns()
+        
+        url2 = 'http://server2'
+        conn2 = WBEMConnection(url2)
         server2 = WBEMServer(conn2)
         server2.validate_interop_ns('root/PG_InterOp')
 
@@ -47,12 +48,12 @@
                 break
 
         # Create a dynamic alert indication filter and subscribe for it
-        filter2_path = listener.add_dynamic_filter(
+        filter2_path = listener.add_dynamic_filter(url2
             query_language="DMTF:CQL"
             query="SELECT * FROM CIM_AlertIndication " \\
                   "WHERE OwningEntity = 'DMTF' " \\
                   "AND MessageID LIKE 'SVPC0123|SVPC0124|SVPC0125'")
-        listener.add_subscription(filter2_path)
+        listener.add_subscription(url2, filter2_path)

Looks like you dropped the url parameter in a couple of places.

KSchopmeyer:
Sorry. putting in patches like that creates a mess.
`+++ /home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_listener.py
@@ -24,12 +24,13 @@
         print("Received CIM indication: {0!r}".format(indication))
 
     def main():
-
-        conn1 = WBEMConnection('http://server1')
+        url1 = 'http://server1'
+        conn1 = WBEMConnection(url1)
         server1 = WBEMServer(conn1)
-        server1.determine_interop_ns()
-
-        conn2 = WBEMConnection('http://server2')
+        # NOT USED and determine is private I thinkserver1.determine_interop_ns()
+        
+        url2 = 'http://server2'
+        conn2 = WBEMConnection(url2)
         server2 = WBEMServer(conn2)
         server2.validate_interop_ns('root/PG_InterOp')
 
@@ -47,12 +48,12 @@
                 break
 
         # Create a dynamic alert indication filter and subscribe for it
-        filter2_path = listener.add_dynamic_filter(
+        filter2_path = listener.add_dynamic_filter(url2
             query_language="DMTF:CQL"
             query="SELECT * FROM CIM_AlertIndication " \\
                   "WHERE OwningEntity = 'DMTF' " \\
                   "AND MessageID LIKE 'SVPC0123|SVPC0124|SVPC0125'")
-        listener.add_subscription(filter2_path)
+        listener.add_subscription(url2, filter2_path)`

KSchopmeyer:
And now, some more:  
1. Here I patched in source_namespace.  We need it because it tells us where the class in the select statement is. Pegasus checks that.
2. Second is in add_dynamic_filter where what is returned is a path, not an instance.

At this point I am probably through the creation but not sure.  Have to clean up pegasus and try again.

    --- /tmp/meld-tmpucQJjJ
    +++ /home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_listener.py
    @@ -377,7 +377,7 @@
             self.log('%s: HTTP status %s; CIM error response: %s: %s',
                      (self._get_log_prefix(), http_code,
                       _statuscode2name(status_code), status_desc),
    -                  logging.WARNING)
    +                 logging.WARNING)
 
         def send_success_response(self, msgid, methodname):
             """Send a CIM-XML response message back to the WBEM server that
    @@ -882,7 +882,7 @@
             # Remove server from this listener
             del self._servers[server_url]
 
    -    def add_dynamic_filter(self, server_url, query,
    +    def add_dynamic_filter(self, server_url, query, source_namespace,
                                query_language=DEFAULT_QUERY_LANGUAGE):
             """
             Add a dynamic indication filter to a WBEM server, by creating an
    @@ -918,9 +918,11 @@
                 raise ValueError("WBEM server not known by listener: %s" % \
                                  server_url)
             server = self._servers[server_url]
    -        filter_inst = _create_filter(server, query, query_language)
    -        self._dynamic_filters[server_url].append(filter_inst)
    -        return filter_inst.path
    +        filter_path = _create_filter(server, query, query_language, source_namespace)
    +        # TODO KS the above returns a path, not the instance
    +        self._dynamic_filters[server_url].append(filter_path)
    +        print('filter_path=%s' % repr(filter_path))
    +        return filter_path
     
         def remove_dynamic_filter(self, server_url, filter_path):
             """
    @@ -1186,7 +1188,7 @@
         return dest_path
     
    -def _create_filter(server, query, query_language):
    +def _create_filter(server, query, query_language, source_namespace):
         """
         Create a dynamic indication filter instance in the Interop namespace
         of a WBEM server and return its instance path.
    @@ -1228,7 +1230,8 @@
         filter_inst['Name'] = 'cimfilter%d' % time.time()
         filter_inst['Query'] = query
         filter_inst['QueryLanguage'] = query_language
    -
    +    filter_inst['SourceNamespace'] = source_namespace
    +    print('Created Filter %s' % filter_inst.tomof())
         filter_path = server.conn.CreateInstance(filter_inst)
         return filter_path

KSchopmeyer:
Here is one more and I have not traced it down yet:

It occurs apparently around when I create the subscription

No handlers could be found for logger "pywbem.listener.140703984508432"

KSchopmeyer:
One more issue.  In remove_dynamic_filter in the last search:  

  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/pywbem-0.9.0.dev0-py2.7.egg/pywbem/_listener.py", line 956, in remove_dynamic_filter
    if inst.path == filter_path:
AttributeError: 'CIMInstanceName' object has no attribute 'path'

Not sure about this one yet since appears that the dynamic.. function returns instances.

Issue with export methods

2016-05-11 15:49:48,001 pywbem.listener.140691001211472 WARNING  HTTP/1.1 POST from 127.0.0.1: HTTP status 200; CIM error response: CIM_ERR_INVALID_PARAMETER: Expecting one parameter NewInstance, got NewIndication

I believe that the spec defines this as NewIndication

andy-maier:
Fixed all issues reported by Karl in previous comment.

andy-maier:
The last commit fixes all issues reported by Karl in the previous comments.

andy-maier:
One more commit to fix a documentation flaw and to add some comments in the code.

andy-maier:
The last commit improves the situation for logging. The user no longer has to do anything special to logging, unless they want to enable it. For details, see the `WBEMListener.logger` property.

KSchopmeyer:
Tested against my test script and it gets through the script now that I realize you have to remove in order (subscription and then filter).
I have a problem with the test script in that we have a method call that initiates an indication and that is pretty flaky (often does exception with message about not being able to convert all parameters) except that there are no return parameters.
We are able to create subscription, create indications, remove subscription and filter against pegasus now.  Have not tested with https so far.

Question now is whether we are willing to commit without much further testing.  I vote yes since
   a. marked experimental
   b. Easier to do fixes than have it in separate branch now.

andy-maier:
I'm also for merging.

PR #238, opened 2016-04-18
Fixed DEVELOP.md readme file for v0.8.x.

andy-maier:
Adjusted to current state of 0.8.x release.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -12,7 +12,8 @@ Git workflow
   - Based upon an earlier topic branch, in case of dependencies
   - It is valid to rebase topic branches and force-push them.
 * We use pull requests to review the branches.
-  - Make sure the correct long-lived branch is used as a merge target!
+  - Use the correct long-lived branch (e.g. `master` or `stable_0.8`) as a
+    merge target!
   - Review happens as comments on the pull requests.
   - At least two +1 are required for merging.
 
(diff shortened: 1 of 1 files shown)


andy-maier:
Verified during v0.8.3 release.
+1

KSchopmeyer:
+1

Looks really good.


Couple of minor comments if you want to consider them.

1. Replace "Make sure" with "Verify".  Makes it just a bit more official

2. The statement:
     +1.  Make sure the `stable_0.8` branch is checked out, and that it is in a
    Shouldn't it indicate that the stable_0.x' branch is checked out where x is the version so that the statement will apply when we move to 0.9


3. Should the statement:
 +4.  Perform a complete build (in a Python virtual environment):
be:
Perform complete build in python 2 and python 3 environments (ex. use Python virtual environments)

i.e. insure it runs in both environments locally. Probably not major issue since this is done by travis also, but the more you catch locally, the less pr redoes.

4. There is a statement about other tests like the run tests.  We should probably formalize this. somewhere (at least before we get to 0.9.0).  Since I will be running these (now that I can run at least some tests on broad set of servers) it will become more important) to be sure that the run test maintainer (me) does this and somehow submits results.   This last one is more a thought than final conclusion.

andy-maier:
Responses on your comments:

* On 1: Agree. Will update the file.
* On 2: This is the DEVELOP.md file for 0.8.x. We have a different one for 0.9.x, because we switched to Sphinx and that makes many things different.
* On 3: It is true that `make test` runs only in one Python environment. However, step 8 uses `tox` which runs `make test` in all Python environments configured for tox. So the `make test` is meant to be your quick-turnaround test that is used to fix issues. Will update the file with a note pointing that out.
* On your last comment 1: Agree. We should post test results to the release PR. I volunteer for the local CI environment, and need to work on it a bit to make the result collection easier.
If you could volunteer for the real WBEM server, that would be good. Maybe you can also automate that so that all it takes is running a single command, which produces a binary output. Will update the file.

andy-maier:
While re-reading it, I found some more things that I thought needed improvement:
- Introduced a $MNU shell variable so that the commands can now copy&pasted.
- Added `git pull` after checkout to make sure local is up to date.
- Other minor improvements.

andy-maier:
I think the file requires another complete review. Sorry for that.

andy-maier:
After reviewing commit 900e657, +1 from me.

KSchopmeyer:
I agree with your comments above.

I volunteer to figure out how to commit results of running against servers to the release results.  I will create an issue for that.

Read through again. Lots of changes and I assume that there may be more revisions as we go through the process (but maybe not for v 0.8.x) since hopefully there will be only 1 or two more 0.8 releases before 0.9.0

+1

PR #244, opened 2016-04-19
Mof roundtrip test test added to mof_compile

KSchopmeyer:
Adds new test for mof roundtrip.  At this point, this test fails in the recompile; we get a number of compile errors during the recompile and I have not been able to isolate them.  We do have issues in the mof output apparently

When we can get through the recompile we can recombine this with the more schema test so that we only have to compile the schema one time, reducing the test time significantly.  I am keeping it separate to be sure that the FullSchema test actually runs.

We have two choices
1. add this test to master now and accept that we get failures
2. Try to get the mof output completely fixed.

NOTE: I thought that this would be easy but the output is very large and the placing of errors does not yet appear logical.

NOTE: I tried to figure out how to run the compile test one time and save the resulting output but short of a set of globals or putting the variables in the MOFTest class have not figured out yet.

### DIFF
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -1704,6 +1704,7 @@ class MOFWBEMConnection(BaseRepositoryConnection):
         self.qualifiers = {}
         self.instances = {}
         self.classes = {}
+        self.compile_ordered_classnames = []
         if conn is None:
             # This instance variable is used only to make get/set
             # of 'default_namespace' behave as it should, in the case
@@ -1837,6 +1838,9 @@ class MOFWBEMConnection(BaseRepositoryConnection):
                     raise
 
         try:
+            self.compile_ordered_classnames.append(cc.classname)
+
+            # The following generates an exception for each new ns
             self.classes[self.default_namespace][cc.classname] = cc
         except KeyError:
             self.classes[self.default_namespace] = \
@@ -2052,6 +2056,8 @@ class MOFCompiler(object):
         if ns not in self.parser.classnames:
             self.parser.classnames[ns] = []
         try:
+            # Call the parser.  To generate detailed output of states
+            # add debug=1 to following line.
             rv = self.parser.parse(mof, lexer=lexer)
             self.parser.file = oldfile
             self.parser.mof = oldmof
--- a/testsuite/test_mof_compiler.py
+++ b/testsuite/test_mof_compiler.py
@@ -1,6 +1,10 @@
 #!/usr/bin/env python
 #
 
+""" Test the mof compiler against both locally defined mof and
+    a version of the DMTF released Schema.
+"""
+
 from __future__ import print_function, absolute_import
 
 import sys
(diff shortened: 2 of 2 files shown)


andy-maier:
We should not merge it if it fails, because that will hide other failures.
Let's fix it on this branch until it works.

KSchopmeyer:
Agreed.   It appears that the problem is that my original code used a dummy compiler that compiled to lists; hus the list order was the same as the compile order so compile was from superclasses to subclasses.  The to memory options on the compiler compiles to a dictionary so all order is lost. Thus the tree order of the mof hierarchy must be recreated by building a tree ordered by superclass/subclass.  Three choices tonight: 1) build code to create hiearchy 2) go back to the dummycompiler that builds lists, 3) figure out how to do dictionary that keeps compile order.

KSchopmeyer:
Did figure out some of the issues, etc. with getting classes and qualifiers from the compiler (use the handle) and the lastest checkin just compares that with using getClass to get the classes.

andy-maier:
One way I have used in the past to get to a defined order using an unordered dict, was to maintain a list with the keys in parallel, that defines the order.

andy-maier:
Another option would be to use `collection.OrderedDict` but that was introduced in Python only in 2.7, so we have to figure out what we do for 2.6.

andy-maier:
The Travis run for the current code succeeds for Python 3.x, but fails for Python 2.x, with:

In `TestFullSchemaRoundTrip.test_all()` in `test_mof_compiler.py` line 157:

            for qd in sorted(qual_decls):
    >           print('{}'.format(qd.tomof()), file=mof_out_hndl)
    E           ValueError: zero length field name in format

Why is the approach with `format()` needed at all, can it not be this?

                print(qd.tomof(), file=mof_out_hndl)

KSchopmeyer:
My thought this morning was a parallel list of classnames that is the compile order. I like that better than trying to do an ordered dictionary or trying to reconstruct order from the superclass/class hiearchy (which I wrote late last night as a test. It will be useful later for a display tool).

The format, etc. is just because in earlier code I always used format and so adopted that as my standard and I did not bother to look up how you do file in other formats.  One of the continuing issues with python is that there are "so many options" to some of these interfaces, that I tend to stick with what I already know.

Short call on thursday morning (you pick time and we use cmpi again) but I think that the list of classnames in compile order seems simplest and will try it.  However, does that mean that we need a function to get classes in compile order or is that a do it yourself

andy-maier:
I found out what the "ValueError: zero length field name in format" means:
In Python 2.6, the `"{}"` was not yet supported;that was introduced only in Python 2.7.
For Python 2.6 support, one needs to add the field names or numbers, i.e. `"{0}"`.

andy-maier:
No discussion needed, it just needs to be worked.

KSchopmeyer:
This pr not ready for commit.  It finds at least 3 errors in tomof of which only one is fixed right now, pr #260 
The second may be an issue with methods that was masked by the first.
The third is that right now the compiler errors out with a lexer error in the middle of the very last class, in the middle of a string.  Want to fix so everything else works and then fight this one.

This upload tests for qualifier equality but only displays differences in classes to help dig out the problems.

KSchopmeyer:
Whoops.  Finger problems and did not test after rebase.

andy-maier:
Is it ready to be merged?

KSchopmeyer:
On 04/27/2016 02:25 AM, Andreas Maier wrote:
>
> Is it ready to be merged?
>
No. Tested and maybe hit another problem last night.  Also have to clean 
out some diagnostics.
>
> —
> You are receiving this because you were assigned.
> <https://github.com/pywbem/pywbem/pull/244#issuecomment-214992831>
>


-- 
================
Karl Schopmeyer
tel: 972-814-5581

KSchopmeyer:
+1 Passes all tests but one where we have issue over number of properties that could well be peg provider.

Andy: Review.

Fixed issue causing recompile fail and now completely recompiles the schema

andy-maier:
+1

The current code in `test_mof_compiler.py` accesses the new `compile_ordered_classnames` instance variable of `MofWBEMConnection` directly. The code even before this PR accessed other instance variables. I don't think they were meant to be public.

However, let's not charge this PR with cleaning this up. I have created issue #369 to follow up on this.

andy-maier:
We agreed to move this one forward.

PR #246, opened 2016-04-21
Updated DEVELOP.md to integrate changes from recent 0.8 update.

andy-maier:
Ready for merge. Please review.

Basically the v0.8.0 version,
* plus steps for creating new dist subdirectory for new M.N release
* minus the manual handling of doc updates because that is automated now with Sphinx and RTD.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -12,7 +12,8 @@ Git workflow
   - Based upon an earlier topic branch, in case of dependencies
   - It is valid to rebase topic branches and force-push them.
 * We use pull requests to review the branches.
-  - Make sure the correct long-lived branch is used as a merge target!
+  - Use the correct long-lived branch (e.g. `master` or `stable_0.8`) as a
+    merge target!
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
Read through and wonder about one thing:   in the following at about line 270

 +6.  Increase the package versions in the following files by changing the
+    old final version `M.N.U-1` to the new development version `M.N.U.dev0`:

Is it true that there is only one file to change??? and that is pywbem/NEWS???

I want to go through an simulate the steps for myself but other than above I put 

+1

on this one

andy-maier:
On your comment about step 6: The version change is actually described to be in `setup.py`, `pywbem/__init__.py`, and `pywbem/NEWS`, not just in `pywbem/NEWS`.

However, that is still incorrect, because these are the files for 0.8.x. Forgot to change them to the 0.9.0 files. Will fix that.

andy-maier:
+1 from me as well. Karl's issue has been fixed.
Merging.

PR #248, opened 2016-04-21
 Described a deprecation policy.

andy-maier:
Ready for merge.
Please review.

Details from the commit log:
* Added section 'Deprecation policy' that describes the rules.
* Changed makefile to enable `DeprecationWarning` messages during test.
* Added entry `DeprecationWarning` to section "Special types", for easier reference by description of deprecated functionality.

PR #250, opened 2016-04-21
Fixes #196: Deprecated imethodcall() and methodcall().

### LINKED ISSUE #196: Should methodcall() and imethodcall() be made private
In an earlier discussion, we said that they should not be part of the documented external API.
For compatibiity (in case someone has built operations on top of them), it would be appropriate to still have them public with their unchanged names, but just not part of the documented external API.

Here is why it is difficult to do both of that:
- The `__all__` approach does not work for them, because they are not module-global symbols, but members of a class.
- The `.. autoclass::` statement has the ability to specify those class members that are supposed to be shown. However, specifying the other names there causes the special members (e.g. `__str__()`) to no longer be shown. This seems to be a bug in Sphinx or its Autodoc extension.

In the current code, they show up in the documentation, and their description states that they are not part of the external API.

We could go beyond that by making them private, i.e. by adding an underscore. However, that changes the class interface incompatibly.

The discussion point is whether we are serious enough with the intention to remove them from the documented external API, that we would buy into the incompatibility of making them private. We did that for a number of other module-global functions, so it comes down to whether we think someone might have actually used them.

andy-maier:
Follows the deprecation policy added in PR #248.

Please review.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -17,6 +17,10 @@ pywbem v0.9.0.dev0
 
 Released: Not yet
 
+Deprecations
+^^^^^^^^^^^^
+
+
 Clean Code
 ^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


andy-maier:
+1

KSchopmeyer:
Isn't the deprecation policy definition also in pr 248.  Other than that question

+1

PR #252, opened 2016-04-22
Fixes #197: Fixed issues with namespace and path in WBEMConnection methods

### LINKED ISSUE #197: Issues around namespace and path in operation methods in WBEMConnection
These are those issues from #184 that need some discussion first:

* In `ModifyInstance()`, the instance to be modified is explained just based upon its namespace. That could be straightened out to be stated based upon the entire path. However, the namespace is called out in order to be clear that the connection default namespace does not play any role here.
* In association ops, default namespace of the connection is used if the input instance path does not specify one. This is inconsistent with ModifyInstance. Come to one consistent approach.
* In `CreateInstance()`, the creation class name is stated to be taken from the `classname` variable of the provided new instance object, while the namespace is taken from the `path` variable of that object. Why the inconsistency?
* In operations that take an instance path or class path as input, the host component of the path is not passed on to the server, but that is not specifically described anywhere. Should it?
* In operations that return instance paths: What should we say about the presence of namespace and host in the returned instance paths?
* In operations that return instances: What should we say about the presence of namespace and host in the `path` component of the returned instances?

andy-maier:
Ready to be merged. Please review!

Details:

- Added an optional `namespace` parameter to the `WBEMConnection.CreateInstance()` method, for consistency with other methods, and to have an explicit alternative to the namespace in the path component of  the `NewInstance` parameter.
- The `ClassName` parameter of several operation methods can be specified as both a string and a `CIMClassName` object. In the latter case, a namespace in that object was ignored so far. Now, it is honored. This affects the following `WBEMConnection` methods: `EnumerateInstanceNames`, `EnumerateInstances`, `EnumerateClassNames`, `EnumerateClasses`, `GetClass`, `DeleteClass`.
- In the `WBEMConnection.ModifyInstance()` method, the class names in the instance and path component of the `ModifiedInstance` parameter are required, but that was neither described nor checked. It is now described and checked.
- In the `WBEMConnection.ModifyInstance()` method, a host that was specified in the path component of the `ModifiedInstance` parameter incorrectly caused an INSTANCEPATH element to be created in the CIM-XML. This bug was fixed, and a host is now ignored.
- Improved diagnostic messages in test_client() assertions.

**Discussion point:** This PR implements the agreement described in issue #197, with the following exception: For `CreateInstance()`, the namespace when taken from the path of the `NewInstance` is not considered deprecated as we had agreed. The argument for not deprecating it is that it is very similar to how we allow namespaces on `ClassName` arguments. The argument for deprecating it is that having a path on a NewInstance is inconsistent with the idea that this operation gets an instance with no path and the server determines the path.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -90,6 +90,18 @@ Enhancements
 * Created tomof() for CIMProperty making common functionality available
   to both class and instance tomof() pr # 151
 
+* Added an optional `namespace` parameter to the
+  `WBEMConnection.CreateInstance()` method, for consistency with other methods,
+  and to have an explicit alternative to the namespace in the path component of
+  the `NewInstance` parameter.
+
+* The `ClassName` parameter of several operation methods can be specified
+  as both a string and a `CIMClassName` object. In the latter case, a namespace
+  in that object was ignored so far. Now, it is honored. This affects the
+  following `WBEMConnection` methods: `EnumerateInstanceNames`,
+  `EnumerateInstances`, `EnumerateClassNames`, `EnumerateClasses`, `GetClass`,
+  `DeleteClass`.
+
 Bug fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
We agreed to the item defined in the discussion above. That means that this one is ready for final review.

andy-maier:
+1

KSchopmeyer:
+1

PR #253, opened 2016-04-22
Fixes #188: Added section to DEVELOP.md about use of Python namespaces in the project

### LINKED ISSUE #188: Strategy for Python namespaces
At the point where we add more stuff, like higher level functions (e.g. get_interop_namespace, get_all_namespaces, get_central_instance, get_profiles), or the indication receiver, we will have them in separate Python namespaces.

We should figure out whether we just let that happen as it happens, vs. having some sort of strategy around that. For example, by using `pywbem` as the overall namespace, and `pywbem.client`, `pywbem.listener`, `pywbem.profile`, etc. as the namespaces that actually provide the function.

That would be an incompatible change for today's pywbem, but maybe it is worth it. Maybe the overall namespace should be a new name, then we can keep the old one for compatibility.

andy-maier:
Please review, this is just an initial stab but reflects the discussion on issue #188.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -333,3 +333,59 @@ version (e.g. `0.9.4`) whose development is started.
 14. On GitHub, list all open issues that still have a milestone of less than
     `M.N.U` set, and update them to target milestone `M.N.U`.
 
+Use of Python namespaces
+========================
+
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

KSchopmeyer:
+1

Issue #231, opened 2016-04-16
Reorg code so that CIMClass tomof and CIMInstance tomof use common mof output generator


KSchopmeyer:
These two objects generate similar mof so that there is no reason not to use a common code set for both.  tomof() for CIMProperty was added to the code in pr 151 and this can be modified slightly to output the property mof for classes also.

andy-maier:
Fixed by PR #260.

Issue #35, opened 2016-02-09
Mof compiler output overruns line length


KSchopmeyer:
The mof compiler consistently overruns the 80 col limit with description outputs because it does not
account for the qualifier name on the first line. For example:

Where the first line characters each " are beyond the 80 char limit. 
      [Description ("A user-friendly name for the object. This property allows each "
       "instance to define a user-friendly name in addition to its key "
       "properties, identity data, and description information. \nNote that "

This is a problem of the fold processor since it folds the string without knowing about the prepended Description, etc.

andy-maier:
Fixed by PR #260.

PR #260, opened 2016-04-24
Move CIMProperty.tomof processing for classes to CIMProperty class. #231 #223 #35

KSchopmeyer:
Move class  property tomof processing from CIMClass to CIMProperty. #231
Add processing of values to CIMClass.property processing.  This was completely ignored which resulted in about 100 issues in trying to recompile the schema mof.  This code already existed in 
CIMProperty.tomof() - Issue #223 

Add code to keep qualifier.tomof lines under 80 char long. Corrected error use of :term in doc issue -#35

Review but I need to update to add change log update for this change.
NOTE: We will not really know how this works until I finish commit the round-trip code and we can test
against the whole dmtf schema.


KSchopmeyer:
+1 Please review.

andy-maier:
-1

I found only some nits:
* The `bool` type can be specified directly as a parameter type and does not need `:class:'py:bool'`. This applies to `CIMProperty._array_val2mof()` and `CIMProperty.tomof()`.
* The new `isInstance` parameter of `CIMProperty.tomof()` does not follow the convention we use othwerise. Suggest to name it `is_instance`.

KSchopmeyer:
I agree with second but could you clarify first comment.

andy-maier:
On my first comment:

The confusing part may have been that I used straight quotes, which I did only because I don't know of a markup syntax in these comments here that allows to show a backquote in monospaced font (However, I just found out that in an indented example section, that is possible).

Anyway, here is an explanation:

In order to reference a built-in type from anywhere in a docstring, the explicit syntax you used is necessary. 

However, in the type portion of a parameter description, the short form using just the type names is sufficient (the type definitions are then search automatically). The explicit form works as well, though.

I am suggesting to use the short form because it is simpler and because it is already used in many places.

Example parameter docstring using the short form:

    folded (bool):
      Some description

The same parameter docstring using the explicit form:

    folded (:class:`py:bool`):
      Some description

Both generate the same link.

KSchopmeyer:
got it.  The sphinx rule set is complex

andy-maier:
+1

PR #265, opened 2016-04-26
Fixes #263: tomof() output of CIMClass / CIMMethod / CIMParameter drops array indicator

### LINKED ISSUE #263: CIMParameter mof round trip drops array indicator
Original MOF
   uint32 RemoveResourcesFromResourcePool(
         [IN, Description ( 
             "Array of CIM_LogicalDevice instances to remove from the Pool."
              )]
      CIM_LogicalDevice REF HostResources[], 
         [IN, Description ( 
             "The pool to remove the resources from." )]
Recompiled mof after compile of the above and output with tomof()
    uint32 RemoveResourcesFromResourcePool(
          [Description ("Array of CIM_LogicalDevice instances to remove from the Pool."),
          IN (True)]
        CIM_LogicalDevice REF HostResources,
          [Description ("The pool to remove the resources from."),
          IN (True)]

KSchopmeyer:
…omof() output drops array indicator

The mof should output an array indicator "[]" when CIMParameter is_array
is set and with the array size if array_size is not None.  The code
correctly issues the one if array size is set but ignores the array
indicator output if only is_array is set.

The CIMMethod.tomof() was outputting an extra \n after last method.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -117,6 +117,8 @@ Bug fixes
 
 * Modify qualifier.tomof processing to stay within 80 column limits. Issue #35
 
+* fix bug in method mof output where array flag "[]" left off array parameters.
+
 pywbem v0.8.2
 -------------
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
I had an error in this one and dropped a \n under some cases. That does not explain why one run passed however.  Will try again in morning

KSchopmeyer:
+1 finally.

andy-maier:
+1

PR #267, opened 2016-04-26
Fixes #259: Added section 'Standards conformance' to documentation

### LINKED ISSUE #259: We should document relation to specs
We should have a spec compiiance document that documents

1. Which specifications and which versions of the dmtf specs we support
2. Any compilance issues and functionality we specifically do not support in the dmtf specifications

andy-maier:
Ready to be merged.
Please review.

Note: This change already includes what the additions from the ``listener`` branch conform to, with a TODO to be finalized / adjusted / removed once that branch gets merged.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -62,6 +62,48 @@ try it out and `report any issues <https://github.com/pywbem/pywbem/issues>`_.
 
 .. _`Deprecation policy`:
 
+Standards conformance
+---------------------
+
+The ``pywbem`` package conforms to the following CIM and WBEM standards,
+in the version specified when following the links to the standards:
+
+* The protocol supported by the PyWBEM client and listener is CIM-XML; it
+  conforms to :term:`DSP0200` and :term:`DSP0201`.
+
+  Limitations:
+
+  - Pulled enumeration operations are not currently supported.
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

KSchopmeyer:
-1. I think we are going to have to get down to spec version numbers. 

Should we also mention DSP0004 for the model since we do conform to that?

andy-maier:
* On spec version numbers:
  They are in the references section, and the standards conformance section says that they apply.

* On DSP0004 and model:
  It would be the metamodel. I was thinking about where that would practiually apply, and found it applies to the CIM objects, where the current text states conformance to DSP0004.
  But maybe we should extend that to also mention the "metamodel supported by PyWBEM" or something like that. I'll do that.

andy-maier:
Commit 90c46e4 addresses the second comment.

Please let me know whether we need to do something about the first comment.

KSchopmeyer:
Lets just state the specs with the versions as defined in the conformance section. We don't want to duplicate but we do want people to understand that the spec version is important.

KSchopmeyer:
+1

andy-maier:
Karl reviewed the latest commit in the call today, and agrees to go forward with this one.

PR #272, opened 2016-04-28
Fixes #270, fixes #275: Add unit test cases for cim_types, fixed str() for CIMDateTime interval

### LINKED ISSUE #270: Testcases for cim_types.py are entirely missing
Add them.

### LINKED ISSUE #275: Incorrect str() value for CIMDateTime interval in Python 3
In Python 3, the str() function on a CIMDateTime interval value has incorrect minutes and seconds fields

```
>>> i = CIMDateTime('12345678224455.654321:000')
>>> str(i)
12345678220000.654321:000
```

Analysis shows that the following code in class `CIMDateTime` is incorrect:

```
def __str__(self):
    if self.is_interval:
        hour = self.timedelta.seconds / 3600
        minute = (self.timedelta.seconds - hour * 3600) / 60
        second = self.timedelta.seconds - hour * 3600 - minute * 60
        return '%08d%02d%02d%02d.%06d:000' % \
                (self.timedelta.days, hour, minute, second,
                 self.timedelta.microseconds)
    else:
        . . .
```

It uses the `/` operator, which performs an integer division in Python 2 (yielding the desired results) but performs a _true division_ resulting in the exact floating point result in Python 3.

**Solution:** Use _floor division_ `//`. It has been introduced in Python 2.2 via [PEP 238](http://docs.python.org/release/2.2.3/whatsnew/node7.html).

andy-maier:
Not yet ready to be merged.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -150,6 +150,10 @@ Bug fixes
   an INSTANCEPATH element to be created in the CIM-XML. This bug was fixed,
   and a host is now ignored.
 
+* Fixed a bug where the CIM datetime string returned by the `str()` function
+  on `CIMDateTime` interval objects contained incorrect values for the minutes
+  and seconds fields on Python 3 (issue #275).
+
 
 pywbem v0.8.2
 -------------
--- a/pywbem/cim_types.py
+++ b/pywbem/cim_types.py
@@ -276,6 +276,8 @@ class CIMDateTime(CIMType, _CIMComparisonMixin):
     interval.
     """
 
+    cimtype = 'datetime'
+
     def __init__(self, dtarg):
         """
         Parameters:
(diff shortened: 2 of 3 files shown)


andy-maier:
This PR is ready to be merged, please review.

It is the first use of the py.test capability to run parametrized tests.

It still has a few TODOs for CIMDateTIme methods that are not yet tested, but I suggest we integrate this and add those tests later.

Besides adding tests for cim_types, this PR also fixes issue #275.

andy-maier:
+1

andy-maier:
Rebased to have the fix first, and then the addition of the unit tests. No change in code since I gave the +1.

KSchopmeyer:
+1

Issue #274, opened 2016-04-28
SSL connection issue with python 2.7


KSchopmeyer:
Now that I have a server consistently working with other clients (cimcli) I am getting errors on connection with pywbem 8.2 and 8.3 and the 8.3 with my working patches.

It is generating the response:

    raise ConnectionError("Socket error: %s" % exc)
ConnectionError: Socket error: certificate verify failed

when connection is made either with test code or with wbemcli using the same credentials that work for cimcli.

KSchopmeyer:
We are getting cert files from the os so that cim_http ends up with /etc/ssl/certs as a proposed cert file and that is an indication that the certs must be checked.  We need to discuss why this code is getting the file from the system.  Not sure of the reason for the logic there yet. (get
When I force the certs file and ca_certs to none internal to the function, it works. on pyython27

The good news is that python3 also works.  Now, I have to clean up the code mess since it is full of diagnostics.

KSchopmeyer:
I retake the previous statement about python3.  There is still a problem in that it reports previously connected.

KSchopmeyer:
I think this means that you cannot use wbemcli with ssh hosts unless the cert is in ssl/certs.

I added no_verification to the connect as a test and that worked.  Question is what parameters should we really have on wbemcli for security

KSchopmeyer:
The code makes the assumption that the certs in ssl/certs are valid options. Thus, the user not supplying a cert does not mean that verification is bypassed because the default certs are included so any certs = None is bypassed. Only the use of no_verification disables this and at least wbemcli does not provide either a cmd line option or default for this.  In open pegasus, if the user did not provide infor on a cert file that meant there were no certs.  

**Discussion:** What should the behavior be in cim_http.py. Should any certs from the system be included by default? If that is the case, then no_verification becomes more important in that it is the way a user defines ssl usage but does not care about verifying the server cert.

KSchopmeyer:
This is a problem across both python 2 and 3

KSchopmeyer:
The issue is that cim_http.py searches the system for default certificates in a number of files in /etc.
See the function def get_default_ca_certs().   This means that if there are any certificates in these directories, the user MUST set no_verification=True  to avoid having these certificates tested against the cert from the server.

andy-maier:
Karl,
I found the commit that introduced the search in the default locations:

[...]

Issue #216, opened 2016-04-13
mof_compiler; need to standardize the cmd line options to match our std


KSchopmeyer:
Document what we want, and whether it should be in the url, etc.

KSchopmeyer:
In discussion on 13 April telecon we concluded that the standard within pywbem should be to try to use the uri for everything and avoid extra parameters as much as possible.  Therefore, we need to provide alternatives for those script calls that use things like --no-ssl so that the url itself can be used.

KSchopmeyer:
Right now, I believe that this is wbemcli at least that uses the --no-ssl.

Somewhere we probably need to start our own style guide to capture style decisions like this so that they are general knowledge for the future.

andy-maier:
**Proposal:**
For specifying WBEM server & connection related things, I suggest to use this:

    -s, --server server
            URL of the WBEM server.
            - Scheme is optional and defaults to HTTP.
            - Host is required.
            - Port is optional and defaults to the standard ports, depending on HTTP/HTTPS.
    -n, --namespace namespace
            Namespace in the WBEM server to work against
            Default: root/cimv2
    -u, --username username
            Username for authenticating with the WBEM server
            Default: No username
    -p, --password password
            Password for authenticating with the WBEM server
            Default: Prompt for password if a username was specified
    -c, --cert certfile
            File path of certificate file to be used with HTTPS.
    -k, --key keyfile
            File path of key file to be used with HTTPS.
    -t, --timeout timeout
            Timeout in seconds
            Default: TBD

**Discussion:** Should we change wbemcli and mof_compiler to support only this proposal?

andy-maier:
Some data points:
* The current [PyWBEM utilities](https://pywbem.readthedocs.org/en/latest/utilities.html)
* [OpenPegasus cimcli man page](http://cvs.opengroup.org/cgi-bin/viewcvs.cgi/pegasus/src/Clients/cimcli/doc/Attic/cimcli.html?hidecvsroot=1&search=None&hideattic=1&sortby=file&logsort=date&rev=1.1&diff_format=h)
* [SFCB wbemcli](manpages.ubuntu.com/manpages/precise/man1/wbemcli.1.html)

andy-maier:
On compatibility:

* mof_compiler could find out that it is being invoked with the old options if:
  - the parameter after `-u`has URI format, or
  - the old option `--url` is specified, or
  - the required `-s` or `--server` option is missing
[...]

PR #279, opened 2016-05-03
Fixes #273, #274: Add multiple parameters to wbemcli hopefully consistent with our prop…

### LINKED ISSUE #273: wbemcli clarify help
the port help is as follows
usage: wbemcli [options] server

Provide an interactive shell for issuing operations against a WBEM server.

Positional arguments:
  server                Host name or IP address of the WBEM server.

Server related options:
  Specify the WBEM server and namespace

  --port port           Port where the WBEM server listens

However, it does not clarify if this is overridded by the choice of --no-ssl or non-existence of that parameter

KSchopmeyer:
…osed standards in issue # 216

Please review.

This is a demo of what we propose for standard cli arguments in issue #216 but  has the following deviations from our proposed standards in issue #216

1. server is positional parameter not the -s or --server.  See the issue for discussion but this seems logical since it is required unless we define a default.
2. keyfile and certfile do not have the short arg definitions, just the long arguments
3. This does not add config file which is discussed in issue 273. Created a new issue to do that since it is significant work and not sure we need it for 0.9.0

This work was done to be able to test ssl conveniently so it adds all of the ssl based arguments including:
1. certfile, keyfile arguments for client cert and keyfile
2. no-verification boolean argument to bypass client verification of server cert.
3. ca_cert argument for local store for certs to match with cert from server.
4. allows http:// and https:// on server name. 
5. allows:port on server name.
6. Extends descriptions to help users understand use of various security arguments.  This could probably be extended even more.
7. marks --ports and --no-ssl asdeprecated.
8. Breaks args down into 3 groups, positional, server (port, namespace, -nossl, timeout), and security (certs, etc.)

I have tested the basic argument  changes and am in process of testing the various cert arguments but that is going to take time.

This corrects #274 by making a specific no_verification parameter so that ssl test can be run against a server that returns certificates without testing those certificates.

Corrects #274 by clarifying port optionality


andy-maier:
Some comments:

* On description of `server`:
  Its description is not really sufficient (misses port, IPv4 vs IPv6, IPv6 zone id, etc.). See the description of the `url` argument in https://pywbem.readthedocs.io/en/latest/client.html#pywbem.WBEMConnection for what I think we need here.

* On description of `--port` option:
  - It is not clear whether a port in server or a port specified using this option has precedence. If we clarify that, we no longer need to say what it is deprecated in favour of.
  - The defaults are not 5988/5989, but whatever is specified or defaulted in the `server` argument.
  - The syntax of a port specified in the `server` argument should be described there, not here.
  - Change: "Default: Standard WBEM defined ports( non-ssl=5988 or ssl=5989).  This option is deprecated in favor of adding port to server as a suffix preceded by ":""
  - To: "Overrides the port specified or defaulted in the server argument."

* On description of `--no-ssl` option:
  - It is not clear whether a protocol specified in `server` or the use of HTTP specified using this option has precedence. If we clarify that, we no longer need to say what it is deprecated in favour of.
  - Change: "This option is deprecated in favor of using http or https on server argument."
  - To: "Overrides the protocol specified or defaulted in the server argument."

* On `-no` option:
  - Short options must be one character long. I suggest to drop the short option here.

* On `--ca_certs` option:
  - Rename its argument `CaCerts` to `cacerts` for consistency with other option arguments.
  - Rename the option to `--cacerts`, for simplicity.
  - The mentioning of `None` in the description of the default makes it more complicated than needed. Change text from: "Default: None. However, if None, pywbem will search for possible certificates in default system directories. Therefore, --no-verification must be supplied to bypass client verification of server sent certificaties." to: "Default: Use certificates in the following default system directories:" followed by a list of them.

* On `--no_verification` option:
  - Option name: GNU naming conventions for options suggest to use hyphen between `no` and the rest of it: `--no-verification`. See https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table
  Also, because there are client and server certificates and they are quite different things, I'd prefer to have some similarity in the names of this option and `--cacerts`. So I'd prefer: `--no-caverify`. I do understand that this way, it gets inconsistent with the argument names of `WBEMConnection`, so that may be worth a discussion.
  - Description: The current text is not 100% clear that no verification takes place, if the option is specified. Change text from: "Do not verify wbem server returned certificate against certs in ca_certs or system defined certs." to: "Do not verify a certificate returned by the WBEM server.".

* TBD: Does the use of certfile make the use of keyfile superfluos? We should clarify that.

andy-maier:
Also, I don't think this PR fixes issue #274.

KSchopmeyer:
Updated mostly per comments above:

* On description of server:

**Major extension to server description and more examples. This one became a book**

 *On description of --port option:

**Cleaned up per comments. See comment under --no-ssl**

 *On description of --no-ssl option:

Cleaned up per comments.  Note that on this and port, the server definition supersedes this and it would not bother me to simply remove this and even the port option.  Remember that this is not
really a scripting tool but completely manual

DISCUSSION: Why not just drop the old forms.

 *On -no option:
        Changes -t -x. No reason.  By the way, despite linux warnings, double character shorts are used.

**DISCUSSION:** That is a very popular options since much testing, even with ssl will be with no verification so I prefer some type of short form.

 *On --ca_certs option:

** Fixed**

*On --no_verification option:
        Option name: GNU naming conventions for options suggest to use hyphen between no and the rest of it: --no-verification. See https://www.gnu.org/prep/standards/html_node/Option-Table.html#Option-Table 

**DONE**

Also, because there are client and server certificates and they are quite different things, I'd prefer to have some similarity in the names of this option and --cacerts. So I'd prefer: --no-caverify. I do understand that this way, it gets inconsistent with the argument names of WBEMConnection, so that may be worth a discussion.

**DISCUSSION**  **I did not change this to no-caverify*

        Description: The current text is not 100% clear that no verification takes place, if the option is specified. Change text from: "Do not verify wbem server returned certificate against certs in ca_certs or system defined certs." to: "Do not verify a certificate returned by the WBEM server.".

** WHOOPS missed this one. Just an oversight but will have to do another commit, etc.***

    TBD: Does the use of certfile make the use of keyfile superfluous? We should clarify that.

** Not necessarily.  The key may be part of the cert or a separate file. Added comment to effect that the keyfile can be part of the certfile but if we are to explain keys in detail it is a complete extra document**


    TBD: Does the use of keyfile make the use of username/password superfluous? We should clarify that.

**That is a good question. In theory yes but I trust nothing with this stuff.**
**Added comment to effect that the keyfile can be part of the certfile**

andy-maier:
Comments on latest commit 79a5d0b:

* On help text for `server` argument:
  - Missing space in "serverin"
  - The `scheme` component does not contain `://` according to the syntax notation. however, its description offers `"https://"` and `"http://"`. Change to `"https"` and  `"https"`.
  - On "If there is no scheme prefix the `-no-ssl` option determines the connection type.": Not true, if the `-no-ssl` option is specified, that selects HTTPS regardless of whether the scheme prefix was specified. Also, it does not say what the default protocol is if neither `-no-ssl` is specified nor a scheme prefix is present. Suggestion to address both: Change to: "" .... more to come ...

KSchopmeyer:
1. We agreed to just dliminate the no-ssl option.
2. We agreed that the default when no scheme prefix is provided is https
3. Modify no-verification to no-verify-cert.
4. make short option nvc
5. Question. Can you have multiple short options. KS look.
6. Add directories searched to help text.

KSchopmeyer:
+1 Ready for hopefully final review.  I think I got everything on the list above including:
1. remove no-ssl and port options
2. https is default when no scheme
3. Add list of default ca certs to help
4. Fix minor errors
5. Added one TODO in that we cannot format the epilog.  Tried the R| option with smartformatter and that does not work on epilog so we live with what we have.
6. Note that this create 3 globals in cim_http (default ports, and default_ca_list)
7. change no_verification to no_verify_cert and -nvc

andy-maier:
Comments on the level as per commit 7c21efb. All comments apply to the help text:

1.  In the description of the server argument, the syntax notation has a leading hyphen. I suggest to remove that.

2. Terminology in description of scheme in server argument: It uses "connection type" to denote the values "http" and "https". I'm not necessarily opposed to that term, but I'd like us to think about alternatives ("protocol"?). DSP0200 does not use such a term. The Wikipadia articles on HTTPS, HTTP and URL also do not use such a term.

3. Again in the description of scheme: It talks about "ssl connection" and "non-ssl connection". Because SSL is deprecated in favor of TLS, we should not use "SSL" or "ssl" anymore, even when it is still widely used as a common term for TLS and SSL.

   Proposal: Change: `defines the connection type; "https" for an ssl connection or "http"" for a non-ssl connection.` to: `defines the connection type; "https" for an HTTPS connection or "http" for an HTTP connection.`.

4. Again in the description of scheme: The text: `The default if there is no scheme prefix is "https.` misses a closing double quote after https.

5. In description of port within server argument: The sentences: `if provided defines the WBEM server port to be used. If the port suffix is not provided the default ports are:` has too much boilerplate around being specified or defaulted. We do that shorter in other options. Suggestion: 
`defines the WBEM server port to be used. Defaults:`

6. In description of port within server argument: The sentence: `The port suffix superceeds the --port argument` no longer applies, because we removed the `--port` option.

7. In group "Server related options:": Its description text: `Specify the WBEM server connection type, port, and namespace` is no longer correct because we removed the options for connection type and port. Also, because only namespace is left now, we could also mention the timeout in the description.

8. Timeout option: The "None" in the text: `Default: No timeout(None)` is not really understandable at the level of command line options. Remove `(None)`.

9. In description of `--cacerts` option: There is a long line. Start a new line before: `However, if None, pywbem will search for`. Also, `for` is used two times in that sentence.

10. In description of `--certfile` option: On text: `If this option exists`. Suggest to change this to: `If specified`.

11. In description of `--certfile` option: On text: `Default: None.`. "None" is not really understandable at the level of command line options. Suggest to say: `Default: No client certificate file, using simple authentication.`. I found in the Wikipedia article about HTTPS that "simple authentication" is the term to be used if it is not mutual auth.

12. In description of `--keyfile` option: On text: `Default: None.`. Same comment as in 11.

13. This is now on the behavior, not on the help text: If the client certificate is sent to the server, is that done encrypted? If not, it is not a good idea to have the private key being part of the certificate file.

KSchopmeyer:
+1 I think I got the help cleaned up.  I am happy and ready for commit.

Note that I added comments to your list as I fixed things so you should be able to see details there.

I actually picked up an error and shortened up several if then else tests.   Filed new issue to add at least minimal test for wbemcli if possible.

KSchopmeyer:
Found error in doc and also squashed last couple of changes.  Note: I do not have any questions for discussion so it is your review next.

+1

andy-maier:
I have a few more comments:

1. On the following text in description of scheme in server argument:

        -scheme: Defines the url scheme;
          - "https" for ssl protocol
          - "http"" for non-ssl protocol.

   - If we want to use the term "protocol", we should use it also in the first sentence, above.
   - We agreed to no longer use "ssl" or "non-ssl" but "HTTPS" and "HTTP" instead.
   - duplicated closing double quote on `https""`.

   Suggestion: change these lines to:

       -scheme: Defines the protocol to use:
         - "https" for HTTPS
         - "http" for HTTP

2. On description of port component in server argument:
   - extra space in "us ed": `Defines the WBEM server port to be us ed`
   - Duplicated colon in: `Defaults::`.

KSchopmeyer:
+1
Found one more error, we were never setting the timeout.  The good news is we are now supplying the timeout.  The bad news is issue #287 when we use this argument in WBEMConnection

andy-maier:
The following issues from my list further up are still open:

* From 2: On description of port component in server argument:
   - extra space in "us ed": `Defines the WBEM server port to be us ed`

* From 4:  On description of `--cacerts` option:
   - On: "The --no-verify-cert option must be set to bypass client verification of the WBEM server certificaties.": I suggest to remove this sentence. That is already stated in the description of the `--no-verify-cert` option. If you like to keep the sentence, I'll not be particularly opposed, but let me know.

* Also from 4:  On description of `--cacerts` option:
   - On: "The system directories searched for certificates are:" and "Default: Search for matching certificates in the following system directories:": We don't need both sentences. Suggest to remove the first one.

andy-maier:
This PR addresses issue #216 for the `wbemcli` command, but the issue also applies to the mof_compiler command, so I removed #216 from the "Fixes:" list in the title of this PR.

KSchopmeyer:
Fix issue raised by issue #287 and also sets limits on size of timeout (0 - 300).  By the way, I have no idea what 0 would cause but probably nothing good :-)

Fix error when -t not specified. testing for integer sets variable to integer but out of range.

+1  Note that we have no autotests for this code

PR #290, opened 2016-05-09
Fixes #288: IndexError when handling exception from send()); Added debug prints for those ignored exceptions.

### LINKED ISSUE #288: Error in exception when calling with SSL and exception occurs
Python 2.7

The exception after first send data
With python 2.7 (m2crypto) and no_verification=False: 

When you do an https connect with a script you the the exception below:

              except Exception as exc: # socket.error as exc:
                    # TODO AM: Verify these errno numbers on Windows vs. Linux.
                    print('exception exc %s' % repr(exc))

fails:

      File "./dotestpeg.py", line 11, in <module>
        CLS = CONN.GetClass('CIM_ManagedElement')
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 2513, in GetClass
        **extra)
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 670, in _imethodcall
        timeout=self.timeout)
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_http.py", line 611, in wbem_request
        if exc.args[0] != 104 and exc.args[0] != 32:
    IndexError: tuple index out of range

**Edit:** Andy edited for better formatting.

andy-maier:
This PR makes some decisions on the discussion points stated in issue #288, and should be seen as a proposal on how to address them.

Otherwise, ready to be merged. Please review.

Details from the commit log:

- In `wbem_request()`, changed catching of `Exception` to `SocketErrors` to avoid secondary `IndexError` on access of `exc.args[0]`.
- Added debug prints if ECONNRESET or EPIPE are encountered. The debug prints are disabled by default and can be enabled by setting `conn.debug=True`. In support of that, passed 'conn.debug` to `wbem_request()`.
- In our copy of httplib `send()` in `HTTPBaseConnection`, merged the latest code (which supports data objects that have a read() method - which probably is not used in pywbem).
- Added Todo entry for turning the quick fix of having `_ensure_bytes()` in `send()`, into a better approach.

### DIFF
--- a/pywbem/NEWS
+++ b/pywbem/NEWS
@@ -1,7 +1,28 @@
 
-pywbem-0.8.4.dev0  2016-mm-dd
+pywbem-0.8.4.dev0  2016-05-09
 
   BUG FIXES:
+
+    * Fixed an IndexError in cim_http.wbem_request() that occurred during
+      handling of another exception.
+
+  ENHANCEMENTS:
+
+    * Fixed description in INSTALL.md to correctly describe how to establish
+      OS-level prerequisites.
+
+    * Improved Swig installation code by reinstalling Swig if it was installed
+      but still cannot be found in PATH (e.g. if the installation was tampered
+      with).
+
+    * Removed dependency on git (this was a leftover from when M2Crypto needed
+      to be obtained from its development repo).
+
+    * Added debug prints for two probably legitimate situations where socket
+      errors are ignored when the server closes or resets the connection.
+      These debug prints can be enabled via the `debug` instance variable
+      of the WBEMConnection object; they are targeted at development for
+      investigating these situations.
 
 
 pywbem-0.8.3  2016-04-15
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -34,6 +34,7 @@ from __future__ import print_function
 import re
 import os
 import sys
+import errno
 import socket
 import getpass
 from stat import S_ISSOCK
@@ -290,7 +291,6 @@ def wbem_request(url, data, creds, headers=[], debug=0, x509=None,
 
       debug : ``bool``
         Boolean indicating whether to create debug information.
-        Not currently used.
 
       x509
         Used for HTTPS with certificates.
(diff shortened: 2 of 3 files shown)


KSchopmeyer:
In order to really test this I put it into the new socket socket_err_check_hostname code for 0.9.0.

The good news is that the initial patch actually eliminates the original issue.  The bad news is that now that we can go further, we are going to have to improve the whole test environment to be able to test with both valid and invalid certs and with certs that have correct and bad hostname.  But that is beyond the scope of this patch.

The following are more detailed notes. Tested with the updated wbemcli using the options to select multiple security options.

Note for the future. We should probably try to start patches like this with the real head of tree.

With python 2.7 testing with ssl and pegasus with verification, it clearly avoids the original issue of index error.  +++.  With the -nvc option it works but the old code did also because there is no verification.

The good news is that with verification on python 2.7.9 it does work and gets through the exception.

The bad news is that when you set the --cacert (see below to show it actually works on python 3) it returns the error :

  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/wbemcli", line 564, in EnumerateClasses
    IncludeClassOrigin=ico)
  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/pywbem/cim_operations.py", line 2293, in EnumerateClassNames
    **extra)
  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/pywbem/cim_operations.py", line 670, in _imethodcall
    timeout=self.timeout)
  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/pywbem/cim_http.py", line 657, in wbem_request
    client.endheaders()
  File "/usr/lib/python2.7/httplib.py", line 975, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python2.7/httplib.py", line 835, in _send_output
    self.send(msg)
  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/pywbem/cim_http.py", line 354, in send
    self.connect()
  File "/home/kschopmeyer/pywbem/git2hubpywbem/pywbem/pywbem/cim_http.py", line 461, in connect
    if not check(self.sock.get_peer_cert(), self.host):
  File "/home/kschopmeyer/.virtualenvs/testssl/local/lib/python2.7/site-packages/M2Crypto-0.24.0-py2.7-linux-x86_64.egg/M2Crypto/SSL/Checker.py", line 138, in __call__
    fieldName='commonName')
WrongHost: Peer certificate commonName does not match host, expected sheldon, got PEGASUS TEST CERTIFICATE-DO NOT USE

Looks like this is an issue within openpegasus fixed install with the certificate we create in that we create and use a single cert with this blah name "PEGASUS TEST ..."  for a raw install. Within the test suite we provide for creating a cert specifically for a test run but then go back to the base cert.

On python 3,  testing against pegasus with the selfsigned pegasus cert(and this is really scary) it actually works and returns data.  

Also when I try with an invalid cert (nothing in --cacert) it returns "certificate verify failed".  What bothers me is that, whatever the reason, not including creds does not matter.  Do not know if this is pywbem or openpegasus. It may be the mode set in open pegasus.

What bothers me at this point is why, given pegasus 3, we do not get caught in the invalid cert issue like we do in python2.  Would appear that either I do not have the code correct or it is not really doing the match_hostname completely with python 3.4.

andy-maier:
Karl,
for this to work, we need to rollback the change from PR #179 into stable_0.8. Let's defer this one until we are ready to do that.

andy-maier:
In PR #179, Karl suggested to merge this PR and its 0.9 companion PR #293, and then to continue with #179 on top.

PR #293, opened 2016-05-09
IndexError when handling exception from send(); Added debug prints for those ignored exceptions.

andy-maier:
This PR merges the changes from PR #290 into the master branch.

Review should be done on PR #290. This PR can merge automatically once PR #290 is approved.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -125,6 +125,22 @@ Enhancements
   the server argument to use http:// or https:// prefix and suffix with
   :<port number> and drops the old arguments of --port and --no-ssl
 
+* Fixed description in INSTALL.md to correctly describe how to establish
+  OS-level prerequisites.
+
+* Improved Swig installation code by reinstalling Swig if it was installed
+  but still cannot be found in PATH (e.g. if the installation was tampered
+  with).
+
+* Removed dependency on git (this was a leftover from when M2Crypto needed
+  to be obtained from its development repo).
+
+* Added debug prints for two probably legitimate situations where socket
+  errors are ignored when the server closes or resets the connection.
+  These debug prints can be enabled via the `debug` instance variable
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
OK. for 2.7 now have pegasus fixed so the default cert now has correct hostname and for python 2.7, it appears to work without the no_verification, i.e it passes match_hostname.  Also the failed cert exception works. Nonexistant host works (Name or service not known). Invalid port works.

Since on the master the whole ssl python3 path is broken until we agree on pr #179 cannot test that as a standalone but only with pr #179

KSchopmeyer:
With python 3 and pr #179.

1. bad hostname generates good exception but different than 2.7 "No address associated with hostname" ranther than name or service unknown.  That difference is common in implementations.

2. Bad port.  Good exception

3. Good connection, good. gets data

4. use only system certs (no --cacerts) Good SSL error, "certificate verify failed)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_http.py", line 527, in connect
    "SSL error %s: %s" % (arg.__class__, arg))
pywbem.exceptions.ConnectionError: SSL error <class 'ssl.SSLError'>: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:600)

5. However, the auth names/pw do not cause failure when bad names.  Same for python 2 and 3.

+1 but remember that this was tested with pr 179 merged into the test.

andy-maier:
Rebased on master to resolve merge conflicts.

andy-maier:
On your comment #5: Does that mean that with certificates, username/password are ignored and can be ommitted?

andy-maier:
In PR #179, Karl suggested to merge this PR and its 0.8 companion PR #290, and then to continue with #179 on top.

PR #299, opened 2016-05-11
Fixes a couple of issues. 1) the wrap function needs the hostname, 2)…

KSchopmeyer:
… message misspelling, 3) insure check_hostname state

I ran through the tests again and hit a problem.  I cannot explain why the issue occurred because we passed the manual tests before this with no problem but I had to add server_hostname to the ssl wrap function to avoid an error.  Note that it is clear from the ssl python code that you must do this so I do not know how it worked before.

Somehow I messed up since when I went back to manually test, the python 3 tests with cacert failed (passed with python2)

### DIFF
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -502,6 +502,7 @@ def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
                         ctx.load_verify_locations(cafile=self.ca_certs)
                     ctx.check_hostname = True
                 else:
+                    ctx.check_hostname = False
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1  I did not add anything to changes.rst since this code should have been in pr179 so that comment applies.

It is clear that I missed something since in going through the ssl module passing the host address in the ssl wrap function is required and I did not have it.  Simply cannot explain how it tested but it is clear that the manual tests were not sufficient.  Writing an automated test tonight.

andy-maier:
+1

PR #303, opened 2016-05-11
Fixed FileNotFoundError not subscriptable problem in wbemcli for Python 3

andy-maier:
Ready to be merged - please review. This solves the issue for the master branch.

Details, from the commit log:

- This fixes issue #302 for the wbemcli command, by accessing the error code no longer by array subscription, but by accessing the errno attribute.
- Also, established compatibility between the different exceptions raised before and starting with Python 3.3 (IOError vs FileNotFoundError).

PR #304, opened 2016-05-11
Fixes #302: Fixed FileNotFoundError not subscriptable problem in wbemcli for Python 3

### LINKED ISSUE #302: FileNotFound exception on py3 when using wbemcli the first time
The handling of the situation where the history file of the interactive shell is not found (e.g. upon initial use) works for Python 2:

        try:
            _readline.read_history_file(histfile)
        except IOError as arg:
            if arg[0] != _errno.ENOENT:
                raise

However, in Python 3.3, the exception raised for that situation was changed from IOError to FileNotFoundError (subclassing OSError), and neither FileNotFoundError nor OSError are subscriptable anymore. So the handling code fails with an according exception:

    TypeError: 'FileNotFoundError' object is not subscriptable

Analysis shows that (1) both the IOError used before Python 3.3, and the OSError from Python 3.3 on, have an errno attribute. (2) the new FileNotFoundError also has that errno attribute.
So the code can be changed from access by subscript to accessing the errno attribute, and then the only difference that remains is the exception type (IOError vs OSError/FileNotFoundError ).

This applies to wbemcli and examples/listen.py (in the listener branch).

Credits: This problem was found by @skolsuper, and mentioned in issue #251.

andy-maier:
Ready to be merged - please review. This fixes the issue for the stable_0.8 branch. Otherwise, same as PR #303.

Details, from the commit log:

- This fixes issue #302 for the wbemcli command, by accessing the error code no longer by array subscription, but by accessing the errno attribute.
- Also, established compatibility between the different exceptions raised before and starting with Python 3.3 (IOError vs FileNotFoundError).

### DIFF
--- a/pywbem/NEWS
+++ b/pywbem/NEWS
@@ -6,6 +6,11 @@ pywbem-0.8.4.dev0  2016-05-09
     * Fixed an IndexError in cim_http.wbem_request() that occurred during
       handling of another exception.
 
+    * Fixed problem that wbemcli in Python 3 when used without existing history
+      file would fail with "TypeError: 'FileNotFoundError' object is not
+      subscriptable" (issue #302).
+
+
   ENHANCEMENTS:
 
     * Fixed description in INSTALL.md to correctly describe how to establish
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #315, opened 2016-05-12
Added escaping of single quotes to mofstr().

andy-maier:
Ready to be merged. Also updated the testcases.

Please review.


KSchopmeyer:
+1
Note that I have not tested through the recompile step but it is exactly what I had for the code.  I just wanted to separate the two solutions.

andy-maier:
We agreed on the call today that we do not roll that back to 0.8.

andy-maier:
+1

PR #317, opened 2016-05-13
Fixes #193: Remove toinstance flavor

### LINKED ISSUE #193: pr 139 was incomplete. We are adding toinstance to compiled mof if tosubclass set
1. We treat toinstance as illegal in the mof input.   While its use is specialized, it is not illegal
2. The compiler automatically sets toinstance flavor whenever it encounters tosubclass flavor and that does not make sense.

KSchopmeyer:
Fixed this issue and added tests to confirm behavior.  The one discussion item is that in addition to removing the function that forced the toinstance flavor if the tosubclass flavor was set, we also allow the toinstance to be set specifically if it is set in the mof.  

**DISCUSSION**: Should we allow the mof writer to ever input with the toinstance flavor set either on qualifier declaration, or class.  To me it would probably never be used but seems logical.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -200,6 +200,10 @@ Bug fixes
   
 * Fixed issue with tomof() output where datetime values were not quoted.
   (issue #289)
+  
+* Eliminate automatic setting of toinstance flavor in mof_compiler when
+  tosubclass is set.  Also enabled use of toinstance flavor if defined
+  in a class or qualifier declaration. (issue 193)
 
 * Fixed problem in class-level associator operations that namespace was classname
   when classname was passed as a string (issue #322).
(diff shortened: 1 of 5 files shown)


KSchopmeyer:
This code changes allows MOF to specifically set toinstance. It just seemed logical. Although there is probably no real reason for this in the future.

KSchopmeyer:
+1 except for discussion item

andy-maier:
+1, based on code rewiew.
Feel free to merge and delete this branch.

andy-maier:
On tge discussion item: If we support toinstance, we should fully support it, i.e. the MOF parser supports it, the internal representation does, and any MOF output as well.

andy-maier:
Please rebase these changes on master.

andy-maier:
I think there was nothing left to be done after the rebase. merging.

PR #318, opened 2016-05-16
Fixes #297: Add deprecated warning message if validate_callback parameter on WBEM…

### LINKED ISSUE #297: Set warnings for the ssl verify callback that is a wbemconnection parameter in python 2.7/m2crypto
In the future pywbem  we expect to move to at least ssl and or requests package there is apparently no callback mechanism within these ssl mechanisms so that this parameter will be removed at some point in the future

KSchopmeyer:
…Connection is used.

Added the warning if the parameter exists on the WBEMConnection call.  Also added documentation on the deprecated state and fact that not used with python 3

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -227,7 +227,7 @@ Bug fixes
   
 * Fixed issue with tomof() output where datetime values were not quoted.
   (issue #289)
-  
+
 * Eliminate automatic setting of toinstance flavor in mof_compiler when
   tosubclass is set.  Also enabled use of toinstance flavor if defined
   in a class or qualifier declaration. (issue 193)
@@ -245,6 +245,9 @@ Bug fixes
 * Extend PropertyList argument in request operations to be either list
   or tuple. (issue #347)
 
+*  Added deprecated warning to WBEMConnection verify_callback parameter
+   since it is not used with python ssl module and will probably be 
+   removed completely in the future. (issue # 297)
 
 pywbem v0.8.2
 -------------
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -44,6 +44,7 @@ import platform
 import base64
 import threading
 from datetime import datetime
+import warnings
 
 import six
 from six.moves import http_client as httplib
@@ -303,7 +304,11 @@ def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
         :meth:`WBEMConnection.__init__`.
 
       verify_callback:
-        Used for HTTPS with certificates.
+        Used for HTTPS with certificates but only for python 2. Ignored with
+        python 3 since the  python 3 ssl implementation does not implement
+        any callback mechanism so setting this variable gains the
+        user nothing.
+
         For details, see the ``verify_callback`` parameter of
         :meth:`WBEMConnection.__init__`.
 
@@ -394,6 +399,10 @@ def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
                                              timeout=timeout)
             self.ca_certs = ca_certs
             self.verify_callback = verify_callback
+            # issue 297: Verify_callback is  not used in py 3
+            if verify_callback is not None and six.PY3:
+                warnings.warn("verify_callback parameter ignored",
+                              UserWarning)
 
         def connect(self):
             # pylint: disable=too-many-branches
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -547,6 +547,9 @@ class WBEMConnection(object):
             does not need to be used. See :ref:`Verification of the X.509 server
             certificate` for details.
 
+            Warning: This parameter is not used when the python environment
+            is Python 3 because the ssl module does not support it.
+
             If `None`, no such callback function will be registered.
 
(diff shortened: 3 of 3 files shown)


andy-maier:
-1
The docstring of the parameter should state that it is deprecated.

KSchopmeyer:
+1 Fixed Andy's comments

andy-maier:
-1 sorry karl, but you once said we should not make statements about future releases. How about stating what the replacement is?

KSchopmeyer:
Whoops. My bad

KSchopmeyer:
Fixed by removing the reference to removing anything in the future.

andy-maier:
I did not mean the change log, but the docstring of the verify_callback arg

KSchopmeyer:
Ahha.  but I think I disagree in this case.  All we are noting is that it is deprecated because it may be removed as an option in a future version.   The only other thing I can say is that deprecated because not used across all version of Python.  In fact, the goal is really to eliminate differences and to do that we have to eliminate that callback or go to mycrypto for everything.  Note that I am not even sure why it is there since the general requirements of server verification (verify server cert against cacerts and verify hostname) are apparently covered in the standard code so other than the user wants to do something beyond this level of verification, the callback is not required.  And since the ssl module provides the same level of security it would mean that we offer the user a different level of security with python 2.7 than with 3.x.  Note that I did not promise anything specific just that this might be dropped as a part of the user api.   

I will go either way but just though I would add to the thought process.

KSchopmeyer:
How about this:
      verify_callback:
        Used for HTTPS with certificates but only for python 2. Ignored with
        python 3 since the  python 3 ssl implementation does not implement
        any callback mechanism so the callback would never be calles.
        Deprecated: May be removed in a future version of PyWBEM because it
        is not common across all versions of Python used by PyWBEM.

andy-maier:
I checked the 0.7 version of pywbem, and the verify_callback arg was there, i.e. before M2Crypto got used. At that point, the standard Python SSL module was used, wgich apparently supported verification by callback at the time. If py3 removed that support, it might be helpful to understand why, and these would be the reasons to state in our deprecation. I simply like to describe that from a functional perspective and not so much from a formal one. I think your latest suggestion goes in the right direction. Unless we understand better what the,reasons were for the removal of the callback in py3, my take is to say that additional verification provides no additional value.

Also, the current text (even your latest suggestion) does not explain what the standard verification is, to which the callback adds. As I understand it, the standard verification is to validate the server certificate against its chain of trust, and to validate the subject of the server cert against the connection hostname. We should state that.

KSchopmeyer:
I need to review becuase from what I see of the documentation, the ssl module does not have a callback mechanism.   I am going to move this one down in my priority right now because my major goal is to get the pull ops done this week.

KSchopmeyer:
Turns out that is the ssl package used for 0.7.0 was pyopenssl, which imports as openssl.  It is specifically a wrapper layer around openssl, originally created around Python 2.2 time because there was not much available then.

Apparently Redhat moved to M2Crypto during the time they were developing for OpenLMI as part of the commits that lead to the 0.8.0 release. I do not know why they made the move but would guess that for them, 1 single linux platform so may prerequisites were easy to handle, 2. Easier to use the PyOpenssl because at that time PyOpenssl had lots of c code. 3) maybe some security advantages although not really sure about that since the list of functionionality of each of the packages is a moving target.

Finally when we moved to release 0.8.0 and decided that we needed to support Python 3 we used the Ross Peoples code which used the SSL module (an integral module of Python releases today).  Voila, 2 separate ssl implementaitons

Sadly each one has some slight variations including:
1. Level of security
2. Existence of a callback function for verification of server cert.
3. exact interfaces
3. Python versions supported.

Note that m2Crypto and pyOpenssl are both more than just SSL connection support, they are cyptrographic packages so some of the arguments are about the general crypto features.

However, that begs the issue of what we do now with:

    a. The fact that the verify_callback is used by M2Crypto but not used by the SSL module.  Note that the SSL module does have a callback _SSLContext.set_servername_callback()_ but is is not the same functionality as the callback in OpenSSL itself and sort of passed through to M2Crypto and PyOpenSSL python interfaces (our verify_callback parameter).

b. It appears that the level of functionality within the SSL modules varies by Python version although there appears to be some version back to Python 2.6 and Python 2.7.9 appears to have received changes to extend the functionality.  Thus, using SSL means that the actual level of security for the client will be python version dependent.

So it appears that the issue of the callback is still open, Further, depending on what we do with the next version of PyWBEM (i.e. which http and ssl package we use) we may or may not have a callback function as part of the definition.

So we can either
a. Be sure that there is a note in the documentation for 0.9.0
b. or actually use the deprecated warning.

I suggest that we back off the deprecation and just be sure we have info in the documentation that the callback is not called if Python 3 is being used.

We can clarify other issues with the next version where we should probably pick a single package for all python versions.

KSchopmeyer:
We agreed that for now we want the warning to say that the variable is not used and only be activated for python3.  Remove all reference to this being deprecated.

andy-maier:
Sorry - another rebase is needed.

andy-maier:
+1 from me, pending the rebase.

KSchopmeyer:
+1 rebased.

PR #319, opened 2016-05-17
Expand run cim operations

KSchopmeyer:
Adds and expands tests in this test file. 

1. Expanded enumeratInstances to many tests
2. Added tests to a number of other operations.
3. Added TODOs where we probably want to expand next
3. Added tests for the server itself (interop namespace, etc.). Note that this does NOT use the new
Server class.  We need to add those tests also but wanted to close this work out
4. Add test subclass for openpegasus and some tests that are specifically OpenPegasus including
a class that return all property types.
5. Expanded the cli parameters for more clarification and to add some of the https parameters.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -142,6 +142,12 @@ Enhancements
   These debug prints can be enabled via the `debug` instance variable
   of the WBEMConnection object; they are targeted at development for
   investigating these situations.
+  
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

KSchopmeyer:
+1

Issue #294, opened 2016-05-10
Support logging to the system log


andy-maier:
Add infrastructure to log to the system log.

Also, change the debug prints added in PR #290 / #293 into system log calls.

KSchopmeyer:
Move to 0.10.0

andy-maier:
Log these specific situations:
- Unknown IPARAMVALUE elements for output parameters of operations (e.g. pull)
- Namespace not set in instance path, when returning instances with path (e.g. ei)

andy-maier:
We agreed not to implement creating logs for such cases, but to issue Python warnings (e.g. see UserWarning in tupleparse)

TODO: Verify what we should do in the cases listed above.

andy-maier:
The situations mentioned above are handled as follows:

* Unknown IPARAMVALUE elements for output parameters of operations (e.g. pull):
  Actually this is about PARAMVALUE. PR #1322 introduces checks for that and raises a ParseError in all cases of unexpected elements.

* Namespace not set in instance path, when returning instances with path (e.g. ei):
  In all cases where returned objects do not have a namespace by design of the CIM-XML protocol, pywbem adds the effective target namespace. In all cases where returned objects must have a namespace by design of the CIM-XML protocol, a missing namespace causes a ParseError to be raised. In all cases where returned objects have an optional namespace, the namespace will be present if returned by the server (ExecQuery is the only such case).

As a result, no further changes are needed. Closing this issue.

PR #320, opened 2016-05-21
Fixes #9, Fixes #329: implement DMTF Pull Operations with tests

### LINKED ISSUE #9: Pywbem does not implement DMTF pull operations
The **pull operations** defined in DMTF DSP0200 v 1.4 (See http://www.dmtf.org/standards/wbem) have not been implemented in pywbem.  These
operations are now being requested by users and are a required part of DSP0200 so the client should be extended to implement them consistent with the definitions of the DMTF specification.

### LINKED ISSUE #329: cim_operations EnumerateInstances
Code optionally gets namespace from ClassName parameter but documentation is unclear about this.

          ClassName (:term:`string` or :class:`~pywbem.CIMClassName`):
            Name of the class to be enumerated, in any lexical case.
            If specified as a :class:`~pywbem.CIMClassName` object, its host
            component will be ignored.

KSchopmeyer:
This commit includes OpenEnumerateInstancesWithPath,
PullInstancesWithPath, CloseEnumeration, OpenEnumerateInstancePaths,
and PullInstancePaths. The documentation is defined for all but
OpenEnumerateInstancePaths and PullInstancePaths.

The mock test is broken but there is a first set of tests in
run_cimoperations.py that executes a number of variations of
OpenEnumerateInstancePaths with pull and close against OpenPegasus.

NOT READY FOR MERGE.

Here is first cut of part ready for review and comment.  In particular, this currently fails mock and only about hallf of operaitons there.  Since the documentaiton for each operation is enormous would like to get the first one clean (OpenEnumerateInstances) before doing the other opens.

Also some code is hacky and we should discuss the enum_context thing class.

Anyway, it actually works now.

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -29,46 +29,74 @@ a WBEM server.
 All WBEM operations defined in :term:`DSP0200` can be issued across this connection.
 Each method of this class corresponds directly to a WBEM operation.
 
-======================================================  ==============================================================
-WBEMConnection method                                   Purpose
-======================================================  ==============================================================
-:meth:`~pywbem.WBEMConnection.EnumerateInstanceNames`   Enumerate the instance paths of instances of a class
-                                                        (including instances of its subclasses).
-:meth:`~pywbem.WBEMConnection.EnumerateInstances`       Enumerate the instances of a class (including instances of its
-                                                        subclasses)
-:meth:`~pywbem.WBEMConnection.GetInstance`              Retrieve an instance
-:meth:`~pywbem.WBEMConnection.ModifyInstance`           Modify the property values of an instance
-:meth:`~pywbem.WBEMConnection.CreateInstance`           Create an instance
-:meth:`~pywbem.WBEMConnection.DeleteInstance`           Delete an instance
-------------------------------------------------------  --------------------------------------------------------------
-:meth:`~pywbem.WBEMConnection.AssociatorNames`          Retrieve the instance paths of the instances (or classes)
-                                                        associated to a source instance (or source class)
-:meth:`~pywbem.WBEMConnection.Associators`              Retrieve the instances (or classes) associated to a source
-                                                        instance (or source class)
-:meth:`~pywbem.WBEMConnection.ReferenceNames`           Retrieve the instance paths of the association instances (or
-                                                        association classes) that reference a source instance (or
-                                                        source class)
-:meth:`~pywbem.WBEMConnection.References`               Retrieve the association instances (or association classes)
-                                                        that reference a source instance (or source class)
-------------------------------------------------------  --------------------------------------------------------------
-:meth:`~pywbem.WBEMConnection.InvokeMethod`             Invoke a method on a target instance or on a target class
-------------------------------------------------------  --------------------------------------------------------------
-:meth:`~pywbem.WBEMConnection.ExecQuery`                Execute a query in a namespace
-------------------------------------------------------  --------------------------------------------------------------
-:meth:`~pywbem.WBEMConnection.EnumerateClassNames`      Enumerate the names of subclasses of a class, or of the
-                                                        top-level classes in a namespace
-:meth:`~pywbem.WBEMConnection.EnumerateClasses`         Enumerate the subclasses of a class, or the top-level classes
-                                                        in a namespace
-:meth:`~pywbem.WBEMConnection.GetClass`                 Retrieve a class
-:meth:`~pywbem.WBEMConnection.ModifyClass`              Modify a class
-:meth:`~pywbem.WBEMConnection.CreateClass`              Create a class
-:meth:`~pywbem.WBEMConnection.DeleteClass`              Delete a class
-------------------------------------------------------  --------------------------------------------------------------
-:meth:`~pywbem.WBEMConnection.EnumerateQualifiers`      Enumerate qualifier declarations
-:meth:`~pywbem.WBEMConnection.GetQualifier`             Retrieve a qualifier declaration
-:meth:`~pywbem.WBEMConnection.SetQualifier`             Create or modify a qualifier declaration
-:meth:`~pywbem.WBEMConnection.DeleteQualifier`          Delete a qualifier declaration
-======================================================  ==============================================================
+==========================================================  ==============================================================
+WBEMConnection method                                       Purpose
+==========================================================  ==============================================================
+:meth:`~pywbem.WBEMConnection.EnumerateInstanceNames`       Enumerate the instance paths of instances of a class
+                                                            (including instances of its subclasses).
+:meth:`~pywbem.WBEMConnection.EnumerateInstances`           Enumerate the instances of a class (including instances of its
+                                                            subclasses)
+:meth:`~pywbem.WBEMConnection.GetInstance`                  Retrieve an instance
+:meth:`~pywbem.WBEMConnection.ModifyInstance`               Modify the property values of an instance
+:meth:`~pywbem.WBEMConnection.CreateInstance`               Create an instance
+:meth:`~pywbem.WBEMConnection.DeleteInstance`               Delete an instance
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.AssociatorNames`              Retrieve the instance paths of the instances (or classes)
+                                                            associated to a source instance (or source class)
+:meth:`~pywbem.WBEMConnection.Associators`                  Retrieve the instances (or classes) associated to a source
+                                                            instance (or source class)
+:meth:`~pywbem.WBEMConnection.ReferenceNames`               Retrieve the instance paths of the association instances (or
+                                                            association classes) that reference a source instance (or
+                                                            source class)
+:meth:`~pywbem.WBEMConnection.References`                   Retrieve the association instances (or association classes)
+                                                            that reference a source instance (or source class)
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.InvokeMethod`                 Invoke a method on a target instance or on a target class
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.ExecQuery`                    Execute a query in a namespace
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.OpenEnumerateInstances`       Open enumeration session to retrieve instances of
+                                                            of a class (including instances of its subclass)
+:meth:`~pywbem.WBEMConnection.OpenAssociatorInstances`      Open enumeration session to retrieve the instances
+                                                            associated to a source instance
+:meth:`~pywbem.WBEMConnection.OpenReferenceInstances`       Open enumeration session to retrieve the instances
+                                                            that reference a source instance
+:meth:`~pywbem.WBEMConnection.PullInstancesWithPath`        Continue enumeration session opened with
+                                                            OpenEnumerateInstances, OpenAssociatorInstances, or
+                                                            OpenReferenceinstances
+:meth:`~pywbem.WBEMConnection.OpenEnumerateInstancePaths`   Open enumeration session to retrieve instances of a class
+                                                            (including instances of its subclass)
+:meth:`~pywbem.WBEMConnection.OpenAssociatorInstancePaths`  Open enumeration session to retrieve the instances
+                                                            associated to a source instance
+:meth:`~pywbem.WBEMConnection.OpenReferenceInstancePaths`   Open enumeration session to retrieve the instances that
+                                                            reference a source instance
+:meth:`~pywbem.WBEMConnection.PullInstancePaths`            Continue enumeration session opened with
+                                                            OpenEnumerateInstancePaths, OpenAssociatorInstancePaths,
+                                                            or OpenReferenceInstancePaths
+:meth:`~pywbem.WBEMConnection.OpenQueryInstances`           Open query request to retrieve instances defined by
+                                                            the query parameter in a namespace
+:meth:`~pywbem.WBEMConnection.PullInstances`                Continue enumeration of enumeration session opened
+                                                            with OpenExecQuery
+:meth:`~pywbem.WBEMConnection.CloseEnumeration`             Close an enumeration session in process.
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.EnumerateClassNames`          Enumerate the names of subclasses of a class, or of the
+                                                            top-level classes in a namespace
+:meth:`~pywbem.WBEMConnection.EnumerateClasses`             Enumerate the subclasses of a class, or the top-level classes
+                                                            in a namespace
+:meth:`~pywbem.WBEMConnection.GetClass`                     Retrieve a class
+:meth:`~pywbem.WBEMConnection.ModifyClass`                  Modify a class
+:meth:`~pywbem.WBEMConnection.CreateClass`                  Create a class
+:meth:`~pywbem.WBEMConnection.DeleteClass`                  Delete a class
+----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.EnumerateQualifiers`          Enumerate qualifier declarations
+:meth:`~pywbem.WBEMConnection.GetQualifier`                 Retrieve a qualifier declaration
+:meth:`~pywbem.WBEMConnection.SetQualifier`                 Create or modify a qualifier declaration
+:meth:`~pywbem.WBEMConnection.DeleteQualifier`              Delete a qualifier declaration
+==========================================================  ==============================================================
+
+NOTE: The method EnumerationCount is to be deprecated from the DMTF specs
+and has not been implemented by any servers so was not implemented
+in pywbem
 """
 # pylint: enable=line-too-long
 # Note: When used before module docstrings, Pylint scopes the disable statement
@@ -83,9 +111,9 @@ from datetime import datetime, timedelta
 from xml.dom import minidom
 from xml.parsers.expat import ExpatError
 import warnings
+from collections import namedtuple
 
 import six
-
 from . import cim_xml
 from .cim_constants import DEFAULT_NAMESPACE
 from .cim_types import CIMType, CIMDateTime, atomic_to_cim_xml
@@ -97,11 +125,30 @@ from .tupleparse import parse_cim
 from .tupletree import dom_to_tupletree
 from .exceptions import Error, ParseError, AuthError, ConnectionError, \
                         TimeoutError, CIMError
+from .cim_constants import *  # pylint: disable=wildcard-import
 
 __all__ = ['WBEMConnection', 'PegasusUDSConnection', 'SFCBUDSConnection',
            'OpenWBEMUDSConnection']
 
+# Global named tuples. Used by the pull operation responses to return
+# (entities, end_of_sequence, and enumeration_context) to the caller.
 
+# openenumerateInstances, OpenAssociators, etc and PullInstanceWithPath
+# responses
+#pylint: disable=invalid-name
+pull_path_result_tuple = namedtuple("pull_path_result_tuple",
+                                    ["paths", "eos", "context"])
+
+# OpenEnumerateInstancePaths, etc.  and PullInstancePath responses
+pull_inst_result_tuple = namedtuple("pull_inst_result_tuple",
+                                    ["instances", "eos", "context"])
+
+# openqueryInstances and PullInstance responses
+pull_query_result_tuple = namedtuple("pull_query_result_tuple",
+                                     ["instances", "eos", "context",
+                                      "QueryResultClass"])
+
+# unicode test to set illegal xml char constant
 if len(u'\U00010122') == 2:
     # This is a "narrow" Unicode build of Python (the normal case).
     _ILLEGAL_XML_CHARS_RE = re.compile(
@@ -341,9 +388,9 @@ class WBEMConnection(object):
 
       - :exc:`~pywbem.AuthError` - Authentication failed with the WBEM server.
 
-      - :exc:`~pywbem.ParseError` - The response from the WBEM server cannot be
-        parsed (for example, invalid characters or UTF-8 sequences, ill-formed
-        XML, or invalid CIM-XML).
+      - :exc:`~pywbem.ParseError` - The response from the WBEM server cannot
+        be parsed (for example, invalid characters or UTF-8 sequences,
+        ill-formed XML, or invalid CIM-XML).
 
       - :exc:`~pywbem.CIMError` - The WBEM server returned an error response
         with a CIM status code.
@@ -368,8 +415,8 @@ class WBEMConnection(object):
       ... : All parameters of the :class:`~pywbem.WBEMConnection` constructor
         are set as public instance variables with the same name.
 
-      debug (:class:`py:bool`): A boolean indicating whether logging of the
-        last request and last reply is enabled.
+      debug (:class:`py:bool`): A boolean indicating whether logging of
+        the last request and last reply is enabled.
 
         The initial value of this instance variable is `False`.
         Debug logging can be enabled for future operations by setting this
@@ -629,7 +676,8 @@ class WBEMConnection(object):
                 self.default_namespace, self.x509, self.verify_callback,
                 self.ca_certs, self.no_verification, self.timeout)
 
-    def imethodcall(self, methodname, namespace, **params):
+    def imethodcall(self, methodname, namespace, response_params_rqd=None, \
+                    **params):
         """
         This is a low-level method that is used by the operation-specific
         methods of this class
@@ -645,9 +693,12 @@ class WBEMConnection(object):
         warnings.warn(
             "Calling imethodcall() directly is deprecated",
             DeprecationWarning)
-        return self._imethodcall(methodname, namespace, **params)
+        return self._imethodcall(methodname, namespace,
+                                 response_params_rqd=response_params_rqd,
+                                 **params)
 
-    def _imethodcall(self, methodname, namespace, **params):
+    def _imethodcall(self, methodname, namespace, response_params_rqd=None, \
+                     **params):
         """
         Perform an intrinsic CIM-XML operation.
         """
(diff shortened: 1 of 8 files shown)


pywbem/cim_operations.py
```diff
@@ -68,6 +68,30 @@
 :meth:`~pywbem.WBEMConnection.GetQualifier`             Retrieve a qualifier declaration
 :meth:`~pywbem.WBEMConnection.SetQualifier`             Create or modify a qualifier declaration
 :meth:`~pywbem.WBEMConnection.DeleteQualifier`          Delete a qualifier declaration
+------------------------------------------------------- ---------------------------------------------------------
```
andy-maier: Suggest to move the entire group before class ops.

testsuite/test_client/openenuminstances.yaml
```diff
@@ -0,0 +1,161 @@
...
+                                            <VALUE>Alice Town</VALUE>
+                                        </PROPERTY>
+                                    </INSTANCE>
+                                </VALUE.INSTANCEWITHPATH>
+                            </IRETURNVALUE>
+                            <PARAMVALUE NAME="EndOfSequence">
+                                <VALUE>TRUE</VALUE>
+                            </PARAMVALUE>
+                            <PARAMVALUE NAME="EnumerationContext">
+                                <VALUE></VALUE>
```
andy-maier: DSP0200 1.4 states that the EnumerationContext may be NULL if EndOfSequence is true. To follow that, the VALUE element needs to be removed.

pywbem/tupleparse.py
```diff
@@ -1615,7 +1639,8 @@ def parse_ireturnvalue(tup_tree):
                                 VALUE.OBJECTWITHLOCALPATH* | VALUE.OBJECT* |
                                 OBJECTPATH* | QUALIFIER.DECLARATION* |
                                 VALUE.ARRAY? | VALUE.REFERENCE? | CLASS* |
-                                INSTANCE* | VALUE.NAMEDINSTANCE*)>
+                                INSTANCE* | VALUE.NAMEDINSTANCE* |
+                                VALUE.INSTANCEWITHPATH)>
```
andy-maier: I was wondering why not zero to many of VALUE.INSTANCEWITHPATH, and it turns out that DSP0201 allows only one, while the Open and Pull ops defined in DSP0200 need zero to many. Please double check, but I believe we have an issue for DSP0201 here.

pywbem/cim_operations.py
```diff
@@ -755,18 +817,32 @@ def _imethodcall(self, methodname, namespace, **params):
...
-                raise CIMError(code, tup_tree[1]['DESCRIPTION'])
-            raise CIMError(code, 'Error code %s' % tup_tree[1]['CODE'])
-
-        if tup_tree[0] != 'IRETURNVALUE':
-            raise ParseError('Expecting IRETURNVALUE element, got %s' \
-                             % tup_tree[0])
-
-        return tup_tree
+        if len(tup_tree) == 0:
+            return None
```
andy-maier: I don't think that a zero length list or tuple is possible?

pywbem/cim_operations.py
```diff
@@ -755,18 +817,32 @@ def _imethodcall(self, methodname, namespace, **params):
...
+        # ERROR | ...
+
+        if tup_tree[0][0] == 'ERROR':
```
andy-maier: The non-pulled ops should have the name of the ERROR element still at tup_tree[0], and so should the pulled ops. Only when it is successful, do thexpulled ops have a list of IRETURNVALUE and PARAMVALUEs.

pywbem/cim_operations.py
```diff
@@ -755,18 +817,32 @@ def _imethodcall(self, methodname, namespace, **params):
...
+            err = tup_tree[0]
+            code = int(err[1]['CODE'])
+            if 'DESCRIPTION' in err[1]:
+                raise CIMError(code, err[1]['DESCRIPTION'])
+            raise CIMError(code, 'Error code %s' % err[1]['CODE'])
+        if pull_op is None:
+            #expect either ERROR | IRETURNVALUE*
+            err = tup_tree[0]
+            if err[0] != 'IRETURNVALUE':
+                raise ParseError('Expecting IRETURNVALUE element, got %s' \
```
andy-maier: Because ERROR has been checked before, it cannot be there anymore.

pywbem/cim_operations.py
```diff
@@ -755,18 +817,32 @@ def _imethodcall(self, methodname, namespace, **params):
...
+                                 % err[0])
+            return tup_tree
+
+        # At this point should have optional RETURNVALUE and at maybe one
+        # paramvalue element representing the pull return parameters
+        # of end_of_sequence/enumeration_context
+        # (IRETURNVALUE*, PARAMVALUE?)
```
andy-maier: It is exactly one IRETURNVALUE and zero to more PARAMVALUE elements.

pywbem/cim_operations.py
```diff
@@ -1124,7 +1200,7 @@ def EnumerateInstanceNames(self, ClassName, namespace=None, **extra):
 
         instancenames = []
         if result is not None:
-            instancenames = result[2]
+            instancenames = result[0][2]
```
andy-maier: Why does this change?

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+            An empty iterable indicates to include no properties.
+
+            If `None`, all properties are included.
+
+          FilterQueryLanguage (:term:`string`):
+            Optional: a String defining the name of the query language
+            used for the Query argument. This argument must exist if the
+            filterQuery argument is provided.
+
+          QueryLanguage (:term:`string`):
```
andy-maier: Should be called FilterQuery.

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+            to the WBEM server.
+          #TODO define typing for this correctly
+          OperationTimeout (term: `Uint32`):
+            Optional: If provided it defines a timeout value to be
+            sent to the WBEM Server
+            * If `Uint32` value, this value is sent to the server as the
+              proposed timeout for the operation sequence. If the client
+              waits longer than this timeout in seconds, the server may
+              terminate the operation
+            * if `None`
```
andy-maier: add: "this parameter is not sent..."

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+          ContinueOnError (:class:`py:bool`):
+            Indicates to the server to continue sending responses
+            after errorresponse has been sent.
+            * if `True` the server may continue sending responses after
+              sending an error response.
+            * If `False` the server must not continue sending responses after
+              and error response is sent and must close the enumeration
+              sequence after sending any error response.
+            * if `None` this parameter is not sent to the WBEM server and the
+              WBEM server is to close the enumeration sequence upon error.
```
andy-maier: Instead, say: "this parameter is not passed to the WBEM server, and  causes the server-implemented default to be used. :term:`DSP0200`  defines that the server-implemented default is ......"

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+
+          MaxObjectCount (term: `Uint32`)
+            Defines a value that sets the maximum number of instances the
+            WBEM server can return for this request. any Uint32 number including
+            zero is valid.
+            * If non-zero `Uint32` the WBEM server must return no more than the
+              number of instances defined by this value.
+            * If zero the WBEM server must return zero instances.  This may
+              be used by a client to restart the timeout.
+            * If `None`  a value of zero is sent to the WBEM server.
```
andy-maier: That is not good, becauae it prevents seeing what the server implemented default is. None should always cause the parameter not to be sent. Instead, say: "this parameter is not passed to the WBEM server, and  causes the server-implemented default to be used. :term:`DSP0200`  defines that the server-implemented default is ......"

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+
+            A tuple containing:
+            * A list of :class:`~pywbem.CIMInstance` objects
+              that are representations of the enumerated instances.
+            * instance of enumeration_context which contains
+              information about the state returned by the successfully
+              completed operation. This object may be queried with
+              the more() method to determine if this response
+              terminates the enumeration sequence.
+
```
andy-maier: I don' t like the merging of the output params into EnumContext. Can we not have the same output params asin DSYP0200, returned as a tuple? For consistency with the spec. that's how we do it for the other intrjnsic ops as well.

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+            LocalOnly=LocalOnly,
+            DeepInheritance=DeepInheritance,
+            IncludeQualifiers=IncludeQualifiers,
+            IncludeClassOrigin=IncludeClassOrigin,
+            PropertyList=PropertyList,
+            FilterQueryLanguage=FilterQueryLanguage,
+            FilterQuery=FilterQuery,
+            ContinueOnError=ContinueOnError,
+            MaxObjectCount=MaxObjectCount,
+            pull_op=True,
```
andy-maier: I don't like the pull_op arg and suggest we get rid of it. The method needs to deal with all kinds of return  values already.

pywbem/cim_operations.py
```diff
@@ -114,6 +139,42 @@
 _ILL_FORMED_UTF8_RE = re.compile(
     b'(\xED[\xA0-\xBF][\x80-\xBF])')    # U+D800...U+DFFF
 
+class EnumContext(object):
```
andy-maier: I believe that this class is not needed at all. Its namespace attr is not needed. Its max_obj_count prevents specifying a different obj count on pull (depending on pull interface). maybe you want zero on open and some non-zero value on pull. the end of seq can as well be a boolean out parm. that leaves the op level enum context which is a string. thexout params can be represented as returned tuples in python, and the current approach needs to do that anyway. overall, i think not enough advantages to have the class.

pywbem/cim_operations.py
```diff
@@ -1255,14 +1331,404 @@ class are to be included in the returned instances, as follows:
...
+            including zero is valid. This is used only to change the value
+            from a previous request of the sequence.  If the MaxObjectCount
+            is to be the same as the previous request it is already defined
+            bu the enum_sequence_state variable.
+            * If non-zero `Uint32` the WBEM server must return no more than the
+              number of instances defined by this value.
+            * If zero the WBEM server must return zero instances.  This may
+              be used by a client to restart the timeout.
+            * If `None` the method uses the max_object_count contained in
+              the enum_sequence_state object.
```
andy-maier: Here we see one of the disadvantages of having a default max countvin the context: None is now used to indicate the enum context default and can no longer bexused to indicate using the server defined default. one important use case for pywbem is server testing.

KSchopmeyer:
Update with new proposal for the interface.  This does not change the internal issues between tupleparse and  the response processing but it passes the limited tests in run_cimoperations.py. Mock not fixed at this point so will fail tests.

andy-maier:
I like the change to return a tuple, including the idea to hide the namespace in an inner tuple.

We could use namedtuple for the outer tuple.

pywbem/cim_operations.py
```diff
@@ -1255,14 +1298,386 @@ class are to be included in the returned instances, as follows:
...
+              number of instances defined by this value.
+            * If zero the WBEM server must return zero instances.  This may
+              be used by a client to restart the timeout.
+
+          :Returns:
+
+            A tuple containing:
+            * A list of :class:`~pywbem.CIMInstance` objects
+              that are representations of the enumerated instances.
+              end_of-sequence (:class:`py:bool`). If `True` this enumeration
```
andy-maier: Arg name: Use underscore instead of hyphen.
Missing asterisk before arg name.

KSchopmeyer:
On 05/23/2016 08:43 AM, Andreas Maier wrote:
>
> I like the change to return a tuple, including the idea to hide the 
> namespace in an inner tuple.
>
> We could use namedtuple for the outer tuple.
>

My only concern is to be sure the namedtuple applies all the way back to 
2.6.  Will check
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/pywbem/pywbem/pull/320#issuecomment-221000656>
>


-- 
================
Karl Schopmeyer
tel: 972-814-5581

andy-maier:
I just checked, and namedtuple was added in py 2.6:
https://docs.python.org/2/library/collections.html#collections.namedtuple

pywbem/cim_operations.py
```diff
@@ -54,6 +54,30 @@
...
+:meth:`~pywbem.WBEMConnection.OpenReferenceInstancesPaths`  Open an enumeration session to retrieve the instances
+                                                        that reference a source instance
+:meth:`~pywbem.WBEMConnection.PullInstance Paths`       Continue enumeration of an enumeration session opened
+                                                        with OpenEnumerateInstancePaths, OpenAssociatorInstancePaths,
+                                                        or OpenReferenceInstancePaths
+:meth:`~pywbem.WBEMConnection.OpenExecQuery`            Open query request to retrieve instances defined by
+                                                        the query parameter in a namespace
+:meth:`~pywbem.WBEMConnection.PullInstances`            Continue enumeration of an enumeration session opened
+                                                        with OpenExecQuery
+:meth:`~pywbem.WBEMConnection.CloseEnumeration`         Close an enumeration session in process.
```
andy-maier: I suggest adding a note at the bottom pointing out that EnumerationCount is not supported.

pywbem/cim_operations.py
```diff
@@ -1255,14 +1298,386 @@ class are to be included in the returned instances, as follows:
...
+            included in the returned instances, in any lexical case.
+
+            An empty iterable indicates to include no properties.
+
+            If `None`, all properties are included.
+
+          FilterQueryLanguage (:term:`string`):
+            Optional: a String defining the name of the query language
+            used for the Query argument. This argument must exist if the
+            filterQuery argument is provided.
```
andy-maier: This is a named arg with a defaullt value of `None`. So it always exists from the perspective of the called method, and it does not matter whether it was provided by the caller (it could have been provided as `None`).

Therefore, let's use `None` instead of "exists" or "provided", and let's omit "Optional:".

pywbem/cim_operations.py
```diff
@@ -1255,14 +1298,386 @@ class are to be included in the returned instances, as follows:
...
+
+          FilterQuery (:term:`string`):
+            Optional: a String defining the query that is to be sent
+            to the WBEM server.
+          #TODO define typing for this correctly
+          OperationTimeout (term: `Uint32`):
```
andy-maier: The type should be `:class:'~pywbem.Uint32'` where the single quotes should be back quotes.

andy-maier:
Karl, please rebase this PR on branch fix_331_builddoc-failure-on-py26 (or review that branch quickly so I can merge it, and then rebase this one on master).

andy-maier:
Review comments on documentation of OpenEnumerateInstances() as per commit e38fc65. Line numbers refer to file `pywbem/cim_operations.py`.

1. Line 1402 (ClassName arg): I posted a change proposal in issue #329, because this applies to all four EnumerateInstance operations.

2. Line 1477 (FilterQueryLanguage arg): Change "the Query argument" to "the FilterQuery argument".

3. Line 1481 (FilterQuery arg): Change "to the WBEM server in using the query language" to "to the WBEM server using the query language".

4. Line 1484 (OperationTimeout): Change:

        OperationTimeout (:class: `~pywbem.Uint32`):

   To:

        OperationTimeout (:class:`~pywbem.Uint32`):

   (i.e. remove the blank after `:class:`)

5. Line 1485 (OperationTimeout): Change: "Determines the minimum time in seconds" to "Minimum time in seconds".

6. Lines 1490-1498 (OperationTimeout): Change this paragraph, in order to remove redundancies:

        If provided it defines a timeout value to be sent to the WBEM
        Server. If not provided, this parameter is not passed to the
        WBEM server and the server uses its own default:
        * If non-zero `Uint32` value, this value is sent to the server
          as the proposed timeout for the enumeration session.
        * If value is zero, the server is expected to never timeout
          or to reject the open request if this value is not allowed.
        * If `None` this parameter is not passed to the WBEM server
          and the WBEM server uses its own default.

   To:

        * If not `None`, this parameter is sent to the WBEM server as the
          proposed timeout for the enumeration session. A value of 0
          indicates that the server is expected to never time out. The server
          may reject the proposed value, causing a
          :class:`~pywbem.CIMError` to be raised with status code
          :attr:`~pywbem.CIM_ERR_INVALID_OPERATION_TIMEOUT`.
        * If `None`, this parameter is not passed to the WBEM server, and
          causes the server-implemented default timeout to be used.

7. Lines 1501-1510 (ContinueOnError): Change this paragraph (which has some issues of ambiguity and misses specifics on the not supported error):

        Indicates to the server to continue sending responses
        after errorresponse has been sent.
        * if `True` the server may continue sending responses after
          sending an error response.
        * If `False` the server must not continue sending responses after
          and error response is sent and must close the enumeration
          session after sending any error response.
        * if `None` this parameter is not sent to the WBEM Server and
          the WBEM server uses its own default.
        NOTE: Not all servers allow this parameter.

   To:

        Indicates to the WBEM server to continue sending responses
        after an error response has been sent.
        * If `True`, the server is to continue sending responses after
          sending an error response. Not all servers support continuation
          on error; a server that does not support it must send an error
          response if `True` was specified, causing
          :class:`~pywbem.CIMError` to be raised with status code
          :attr:`~pywbem.CIM_ERR_CONTINUATION_ON_ERROR_NOT_SUPPORTED`.
        * If `False`, the server is requested to close the enumeration after
          sending an error response.
        * If `None`, this parameter is not passed to the WBEM server, and
          causes the server-implemented default behaviour to be used.
          :term:`DSP0200` defines that the server-implemented default is
          `False`.

8. Line 1512 (MaxObjectCount): Change:

       MaxObjectCount (term: `Uint32`):

   To:

        MaxObjectCount (:class:`~pywbem.Uint32`):

9. Lines 1513-1524 (MaxObjectCount): Change this paragraph (which has some issues):

        Optional Parameter (default None). Defines a value that sets
        the maximum number of instances the WBEM server shall return
        for this request. Any Uint32 number including zero is valid.
        If None, nothing is sent to the WBEM server. The DMTF
        Specification states that in this case, the server returns
        zero instances.
        * If non-zero `Uint32` the WBEM server must return no more than the
          number of instances defined by this value.
        * If zero the WBEM server must return zero instances.  This may
          be used by a client to restart the timeout.

To:

        Maximum number of instances the WBEM server may return
        for this request.

        * If positive, the WBEM server is to return no more than the
          specified number of instances.
        * If zero, the WBEM server is to return no instances. This may
          be used by a client to leave the handling of any returned instances
          to a loop of Pull operations.
        * If `None`, this parameter is not passed to the WBEM server, and
          causes the server-implemented default behaviour to be used.
          :term:`DSP0200` defines that the server-implemented default is
          to return zero instances.

10. Line 1528 (Returns): Change:

        A namedtuple containing:

   To:

        A :class:`py:namedtuple` containing:

11. Line 1528 (Returns): In the list of namedtuple members, i suggest to use the `name (type):` syntax. Also, the description of enumeration_context needs some cleanup w.r.t. the different cases, and misses a hint what it is needed for.

   Change:

        * instances A list of :class:`~pywbem.CIMInstance` objects
          that are representations of the enumerated instances.
        * end_of-sequence (:class:`py:bool`). If `True` this enumeration
          session is complete. The WBEM server has sent all of the
          instances possible and has closed the enumeration session.
        * enumeration_context (tuple of (:term:`string`) and `namespace`.
          `None` if the end-of-sequence == False. Otherwise this is
          the context for the next operation (PullInstancesWith Path
          or CloseEnumeration) of this enumeration session.
          This should be considered as an opaque value.

   To:

        * instances (list of :class:`~pywbem.CIMInstance`):
          Representations of the initial set of enumerated instances.
        * end_of-sequence (:class:`py:bool`):
          Indicates whether the enumeration session is exhausted
          after returning the initial set of enumerated instances.
          - If `True`, the enumeration session is exhausted, and the
            server has closed the enumeration session.
          - If `False`, the enumeration session is not exhausted.
        * enumeration_context (tuple of (ecs, namespace)):
           Enumeration context that identifies the opened enumeration
           session, and that the client needs to provide for subsequent
           operations on this enumeration session. The tuple items are:
           - ecs (:term:`string`):
             Enumeration context string returned by the server if
             the session is not exhausted, or `None` otherwise. This string
             is opaque for the client.
           - namespace (:term:`string`):
             Name of the CIM namespace that was used for this operation.
           NOTE: The reason for having this inner tuple is to hide the need
           for a CIM namespace. CIM operations always need a target
           namespace, but it never makes sense to specify a different one
           in subsequent operations.

andy-maier:
You can now rebase this branch on master.

KSchopmeyer:
Updated to complete documentation and validate builddoc output.  The problem with the operation table was an issue in my formatting. Now we can work on mock.
This is still WIP until I finish some more pegasus tests (assoc, ref, query) and at least one mock test setup. 
However, it appears that the code is running except for the openqueryinstances (George and his @#$%^ extra ideas like the temp return class.)
Discussion; The spec for OpenQueryInstances states that it has a FilterQuery and language.  That is messed up in that it is really query, not filterQuery I think

andy-maier:
Comments on the updated documentation as of commit 0ecdc24:

1. lines 1422 to 1424 (OperationTimeout in OpenEnumerateInstancePaths): This paragraph is redundant with the list of values below it and should be removed.

   This comment applies similarly to the other open and pull operations, as applicable.

2. lines 1490 to 1492 (context in OpenEnumerateInstancePaths): Removing the text "Enumeration context" left some gramatically strange text, but removing it seems a good change to me, so let's fix the rest.

   Change:

        that identifies the opened enumeration session, and that the
        client needs to provide for subsequent operations on this
        enumeration session. The tuple items are:

   To:

        Identifies the opened enumeration session. The client needs
        to provide this value for subsequent operations on this
        enumeration session. The tuple items are:

   This comment applies similarly to the other open and pull operations, as applicable.

3. lines 1489 and 1494 (SvrCtxt in OpenEnumerateInstancePaths): Using mixed case on `SvrCtxt` looks a bit inconsistent to me. How about using `server_context` instead?

   This comment applies similarly to the other open and pull operations, as applicable.

4. line 1478 (paths in OpenEnumerateInstancePaths): The type says `CIMInstancePath`, but that does not exist and what you want instead is `CIMInstanceName`.

   This comment applies similarly to the other open and pull operations, as applicable.

5. line 2293 (instances in OpenAssociatorPaths): This is `OpenAssociatorPaths`, but its return description says `instances`. Change to `paths`.

6. line 1788 (ObjectName in OpenReferencePaths): The argument name used in the function header is `InstanceName`, but in the description it is called `ObjectName`. Change to `InstanceName` in description.

7. comment 6 also applies to `OpenAssociatorPaths`.

8. line 2168 (in OpenAssociatorPaths): This line refers to OpenAssociatorInstances from DSP0200, but this is not correct and should be changed to refer to OpenAssociatorPaths.

9. comment 8 similarly applies to OpenReferencePaths (uses ReferenceNames on line 1780).

10. comment 8 similarly applies to OpenAssociatorInstances (uses OpenAssociatorInstancess with one s too much on line 2376).

11. lines 1386 to 1388 (in OpenEnumerateInstancePaths): Change the subsequent operations to be links to the corresponding methods.

   This comment also applies to the other open operations (sometimes this sentence still needs to be added).

12. lines 2176 to 2178 (warning about eos in OpenAssociatorPaths): I would not start the sentence with the condition that eos needs to be False, but just mention the function without the condition:

        Use :meth:`PullInstancesPaths` to retrieve the next set of
        instance paths in this enumeration session.

   This comment also applies to the other open operations (sometimes this sentence still needs to be added).

13. lines 2190 to 2182 (in OpenAssociatorPaths): This warning can be misunderstood, because there are multiple kinds of filtering. The warning does not apply to the AssocClass, ResultClass, Role, and ResultRole filters; it only applies to the FilterQuery filter. In addition, I'm not sure it makes sense to talk about the reasons why a WBEM server might not support it. In addition, we should say what happens if a WBEM server does not support it. Finally, given all that, I think the warning would be in a better place in the description of the FilterQuery argument.

   This comment also applies to the other open operations (sometimes this sentence still needs to be added).

14. lines 2781 and 2765 (enumeration_context in PullInstancesWithPath): Given that we renamed this to `context` in the tuple returned by the open operations, I think we should rename this argument as well, to `context`. Plus, I would rephrase the two sentences. Most importantly however, this is not the string returned by the server, but the inner tuple of (SvrCtxt, namespace)!

Change to:

        * `context` (tuple of (server_context, namespace)):
          Identifies the enumeraton session, including its current enumeration state.
          This must be the value of the `context` item in the :class:`py:namedtuple`
          returned from the previous open or pull operation for this enumeration
          session.
          The tuple items are:

          - SvrCtxt (:term:`string`):
            Enumeration context string returned by the server. This string is opaque
            for the client.
          - namespace (:term:`string`):
            Name of the CIM namespace to be used for this operation.

   This comment also applies to the other pull operations and to close.

15. lines 3049 to 3050 (in CloseEnumeration): I would remove this sentence about passing the namespace. If we update the context parameter as proposed in comment 14, that should be sufficient.

andy-maier:
On the FilterQuery and FilterQueryLanguage arguments of OpenQueryInstances: Because we generically transform the method arguments into IPARAMVALUE elements, I don't think we should fix that here.

On the returning of the class declaration by OpenQueryInstances: I understand it is a bit inconvenient, but a client does not have any other way of finding out about the structure of the returned instances, if the query specifies an artificial result "table".

andy-maier:
Onle one more comment:

1. In `CloseEnumeration()`, the `enumeration_context` argument should be renamed to `context`, for consistency with the other open and pull operations.

andy-maier:
+1 on the documentation.

andy-maier:
Please rebase.

KSchopmeyer:
Agreed with rebase requirement.  Having a real issue here that I need another eye on in that rebase tries to discard my code so I have done something nasty in some past push/rebase/etc. that is causing real confusion for git.  This is what happened the other day and it happened again last night. Just drops most of my changes.

andy-maier:
See my mail on cleaning it up.

KSchopmeyer:
+1

KSchopmeyer:
We discussed two issues here based on the latest push;

1. We ignore unknown iparamvalue names in the response. Should we error, iginore, log?
2. We bypassed the test of namespace on responses.  Should we leave it as is, check and log issue, or fix the responses? Remember that while theoriginal enum did add them, this is a high cost function for thousands of responses and it is clearly a server error.

andy-maier:
TODO Andy: Final review of the code.

andy-maier:
On Karl's items:
1. handling of unknown iparamvalues in reponse: ignore and log. Added the logging part to issue #294. That leaves the ignore part to be implemented.
2. i'd prefer if we could set the namespace in the path if has not been set by the server, and also log that fact. Added the logging part to issue #294. That leaves the settung it if not set part to be implemented.

andy-maier:
Additional comments, focusing on the code:

1. type name of nameetuple used for results: e.g. line 1587 of `cim_operations.py`:

        rslttup = namedtuple("result", ["paths", "eos", "context"])

   `rslttup` is a type and "result" is its name (i.e. `rslttup.__name__ = "result"`). From a type `T`, one expects `T.__name__ == "T"`.

   Also, the type `rslttup` is a bit too abbreviated, in my mind.

   How about:

        pull_result_tuple = namedtuple("pull_result_tuple", ["paths", "eos", "context"])

   Appears a number of times.

andy-maier:
We agreed in the call to:

* Move the two different result tuple types to become global variables in the `cim_operations` module (they stay within that module, i.e. are not added to `__all__` and do not get otherwise added to the `pywbem` namespace. The types will have a name that is their type name (i.e. `T.__name__ == "T"`). For example:

        pull_path_tuple = namedtuple("pull_path_tuple", ["paths", "eos", "context"])

* In each function that returns these named tuples, create the tuple explicitly instead of hiding it in a function, e.g.:

        return pull_path_tuple(self._get_rslt_params(result, namespace))

KSchopmeyer:
+1 Changes made from comments plus some other issues I found.

PR #328, opened 2016-05-28
FIxes #322: Fixed problem in class-level associator operations that namespace was classname

### LINKED ISSUE #322: Issue with associator and associatornames operations
Testing with run_cimoperations.py. Problem is that the functions inside AssociatorNames, Associators, References, ... Are not correctly sorting out classname, namespace, etc. per the documentation for the class version of this function. From the doc:

For class level usage: The class path of the source class, as a `string` or `CIMClassName` object.
If specified as a string, the string is interpreted as a class name in the default namespace of the connection, and can have any lexical case.
If specified as a `CIMClassName` object that does not specify a namespace, the default namespace of the connection is used.

It is getting the classname as the namespacename.

I have not yet explored any futher.

Call:

        classes = self.cimcall(self.conn.Associators, TEST_CLASS)

Error:

    ERROR: test_all (__main__.AssociatorNames)
      File "./run_cim_operations.py", line 687, in test_all
        names = self.cimcall(self.conn.AssociatorNames, TEST_CLASS)
      File "./run_cim_operations.py", line 78, in cimcall
        result = fn(*pargs, **kwargs)
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 2092, in AssociatorNames
        **extra)
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 795, in _imethodcall
        raise CIMError(code, err[1]['DESCRIPTION'])
    CIMError: 3: CIM_ERR_INVALID_NAMESPACE: CIM_ComputerSystem

print inside AssociatorNames:

        namespace = self._iparam_namespace_from(ObjectName)
        objectname = self._iparam_objectname(ObjectName)
        
        print('namespace=%s, objectname=%s, ObjectName=%s' % (namespace, objectname, ObjectName))

    namespace=CIM_ComputerSystem, objectname=CIM_ComputerSystem, ObjectName=CIM_ComputerSystem

**Edit:** Andy updated for better readability.

andy-maier:
Ready to be merged.
Please review.

I tested only with `make test`, plus the Travis CI test runs fine, so please run your real test, Karl.

Details from the commit log:

* Issue #322: When the ObjectName argument was passed as a class name string,
  the target namespace was incorrectly determined to be the class name,
  instead of using the default namespace of the connection (as documented).
  This was fixed by splitting the namespace extraction function into one for
  namespaces and one for object names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,16 @@ install:
   - pip list
   - make develop
   - pip list
+  - python -c "import setuptools,os; print('Version of GitPython package:'); sp=os.path.dirname(os.path.dirname(setuptools.__file__)); print(open(sp+'/git/__init__.py','r').read())"
+
+# TODO 2016-05 AM: Remove displaying of GitPython version, once GitPython 2.0.4 is released.
+
+# commands to run builds & tests
+script:
   - make check
   - make build
-  - make builddoc
-
-# commands to run tests
-script:
+  - if [[ $(python -c "import sys; print('%s.%s'%sys.version_info[0:2])") != 2.6 ]]; then make builddoc; fi
   - make test
+
+# TODO 2016-05 AM: Remove disabling of "make builddoc" for Python 2.6, once GitPython 2.0.4 is released.
+
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -201,6 +201,9 @@ Bug fixes
 * Fixed issue with tomof() output where datetime values were not quoted.
   (issue #289)
 
+* Fixed problem in class-level associator operations that namespace was classname
+  when classname was passed as a string (issue #322).
+
 
 pywbem v0.8.2
 -------------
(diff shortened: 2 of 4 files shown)


andy-maier:
Travis fails for Python 2.6 in make builddoc because it cannot import OrderedDict. The original message said the PyPi ordereddict package needs to be installed, but this did not do the trick.
Need to investigate further.

KSchopmeyer:
The good news is that it passes the currently defined run_cimoperations.py test. with the fix both for python 2.7 and python 3.  Since this test specifically includes a test what was causing the original problem, looks like you fixed that one.

andy-maier:
I have created PR #332 which fixes the Travis CI issue.
Rebasing this PR here on top of PR #332.

andy-maier:
+1

KSchopmeyer:
+1

I think I am beginning to lose context here with changes on top of changes but will sort it out.

PR #337, opened 2016-06-01
issue # 324 Adds an example of EnumerateInstances with ability for command line

KSchopmeyer:
Adds an example of EnumerateInstances with ability for command line to define host, namespace, classname, user, password and display of exceptions. Note that this example displays details of CIMError.

Tried to keep this simple by not using argparse.  That creates a messy cmd line but wanted to keep the example simple and still allow it to be used against both a predefined server/class and also allow the user to work against arbitary server/class.


andy-maier:
+1

PR #338, opened 2016-06-01
Fixes #277: Added description of authentication types in WBEM client API

### LINKED ISSUE #277: Supported authentication mechanisms are not described
The PyWBEM Client does not document which authentication mechsnisms it supports.

andy-maier:
This PR updates the description of the `WBEMConnection` constructor somewhat, and mainly adds a new section "Security Considerations" in the WBEM Client Library API.

This PR is ready to be merged. Please review (This is best done via `make builddoc` and reading the HTML output).

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -149,6 +149,8 @@ Enhancements
   subclass to specifically test against OpenPegasus if OpenPegasus is
   detected as the server.
 
+* Added description of supported authentication types in WBEM client API.
+
 Bug fixes
 ^^^^^^^^^
 
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -211,9 +211,6 @@ Authentication is the act of establishing the identity of a user on the
 client side to the server, and possibly also of establishing the identity of a
 server to the client.
 
-Because authentication in HTTP/HTTPS and therefore in CIM-XML is a complex
-topic, it is briefly described.
-
 There are two levels of authentication in CIM-XML:
 
 * TLS/SSL level authentication (only when HTTPS is used):
@@ -223,7 +220,7 @@ There are two levels of authentication in CIM-XML:
   flow.
 
   In almost all cases (unless an anonymous cipher is used), this involves
-  an X.509 certificate that is presented by the server (therefore called
+  an :term:`X.509` certificate that is presented by the server (therefore called
   *server certificate*) and that allows the client to establish the identity
   of the server.
 
@@ -244,7 +241,7 @@ There are two levels of authentication in CIM-XML:
 * HTTP level authentication:
 
   This kind of authentication is used in HTTP/HTTPS requests and responses (in
-  case of HTTPS, that is after the TLS/SSL handshake protocol has completed).
+  case of HTTPS, after the TLS/SSL handshake protocol has completed).
 
   In case of *Basic Authentication* and *Digest Authentication* (see
   :term:`RFC2617`), it involves passing credentials (userid and password) via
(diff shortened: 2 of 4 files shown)


KSchopmeyer:
**1. Line 320** 
 +* The `ca_certs` parameter may specify X.509 CA certificates that are used
+  to validate the X.509 server certificate returned by the WBEM server.


**COMMENT:** Add "or valid CA certificates for this server are located in the client SSL certificate library since PyWBEM defaults to a system level directory to search for certificates if none are specified
in the `ca_certs` parameter.

Same issue on line 348, the default is actually not no certificates but to point to a system level
directory as a possible use of certificates to verify server if ca_certs is None

**ANDY:** remember that we added to the help display to show these directories for wbemcli.

**2 line 403**
 +
+* When the WBEM server returns HTTP status 404 "Forbidden". It typically
+  does that if it does not allow the client to issue CIM-XML requests.

**COMMENT:** Not sure about this one and PyWBEM I don't see that we actually test for 404 anywhere at this point.

**3. line 214:**
+Because authentication in HTTP/HTTPS and therefore in CIM-XML is a complex
+topic, it is briefly described.

**PROPOSAL:** I would just remove the above sentence.

**line 246:**
+  This kind of authentication is used in HTTP/HTTPS requests and responses (in
+  case of HTTPS, that is after the TLS/SSL handshake protocol has completed).

**PROPOSAL:** change "that is after" to "after"

**4. Line 348:**
+It is important to understand which side actually makes decisions about the
+authentication type: The client only decides whether HTTP or HTTPS is used. The
+server decides everything else: Which HTTP authentication type is used
+(None, Basic, Digest), and whether an X.509 client certificate is requested
+from the client and if so, whether it tolerates a client not providing one.

**COMMENT:** While technically accurate in that this is defining possible authentication types, it ignores
the idea that the client has the option of not verifying the server certificate which is part of the
decision process.

Generally this does a very good job of documenting the overall security concepts.

+1 except for the above comments

andy-maier:
Responsesd to Karl's comments, above:

1. DONE, using different words:

        * The `ca_certs` parameter may specify the location of X.509 CA certificates
          that are used to validate the X.509 server certificate returned by the WBEM
          server.
          If not specified, the PyWBEM client assumes default locations for these
          certificates.

   I don't think we need to be more specific about the default locations, because this is an overview section.

2. **Discussion point:** DSP0200 defines the use of 403 "Forbidden" with:

   "The WBEM server or WBEM listener does not allow the client to issue CIM-XML message requests. The WBEM server or WBEM listener may alternatively respond with a "404 Not Found" if it does not wish to reveal this information to the client."

    It is true that the code has no provisions to handle 403 in any special way. It causes a `ConnectionError` to be raised, like any other non-200 and non-401 status will do.

    Need to discuss whether something needs to be done in the code to handle 403.

3. DONE.

4. DONE.

5. DONE (Added)

andy-maier:
COmments are addressed via commit 72db4f5.

andy-maier:
**MORE DISCUSSION POINTS:**

* The code contains some handling of a 401 status, but I cannot make sense of it. For example, on line 679 of cim_http.py, it raises AuthError if not local (local = file based connection). I.e. it does not process a challenge at all. That is before it attempts to process a challenge presented by OpenWBEM, i.e. it does not even get there. Bottom line, that means my statement in the description that 401 challenges are handled, is not correct.

* On lines 752 and 754, `ConnectionError` is raised. That is at the end of the 401 status processing. So my statement in the new description that 401 leads to `AuthError`, is not entirely correct.

* The `elif` on line 725 seems to contain OpenPegasus specific code. Is the condition selective enough that it is not used e.g. for SFCB or other servers?

I'm not sure I even want to describe the current (weird) situation precisely. We probably should fix the code as well, so that it shows a consistent behavior in 401 status processing, and then have the description in sync with the code.

Let's discuss what a consistent behavior for 401 processing would be.

One thing to note is that pywbem always has the `Authenticate` header in the request, and it supports only one authentication scheme (Basic). So when the server responds with a challenge (401 status with a `WWW-Authenticate` header), the only thing that could be done by the client is to retry the same request with the same `Authenticate` header. The question is, are there servers that respond with a challenge on the first such request and accept the second, exactly same request?

The other question I have is how a file based (local) connection works. I'm not sure that is needed. A local server can be reached using HTTP/HTTPS on localhost.

What would really help here is to have a set of servers set up, where the auth related parameters can be varied.

andy-maier:
We agreed in the call today to merge this and to address the coding issues in separate issues.

PR #340, opened 2016-06-02
Fix a number of pylint errors. 

KSchopmeyer:
Fixed some lint errors in various modules we are not touching right now.

 There was one code change. change variable that was unknown. keyword was unknown variable in list of tests against word. changed it to word

That was the only code change.  Just trying to cut the pylint down a bit to see my new ones.


andy-maier:
+1

PR #348, opened 2016-06-06
issue# 347 Allow tuple as input for PropertyList parameter in cim_operations

KSchopmeyer:
Extends test on input to allow both tuple and list. Extends tests in
run_cimoperations.py for more complete testing of propertylist and
some other tests.
Adds specific test for tuple input for propertylist.

Adds comment to changes.rst

This specifically allows the tuple as input.  Couple of minor issues here:

1. Do we really want to allow the tuple?
2. the tuple with single entry is a mess (requires a comma after the entry
3. If you do a tuple instead of a list with a single entry, it actually goes through but generates xml error because the tuple is not being handled as a tuple in cim_obj.py
4. Documentation defines list in some places and iterable in others.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -221,6 +221,9 @@ Bug fixes
   as list of classes for class level return but it actually list of
   tuples of classname,class. (issue #339)
 
+* Extend PropertyList argument in request operations to be either list
+  or tuple. (issue #347)
+
 
 pywbem v0.8.2
 -------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -151,6 +151,10 @@ Enhancements
 
 * Added description of supported authentication types in WBEM client API.
 
+* Allow tuple as input for PropertyList parameter of cim_operations.
+  Documentation indicated that iterable was allowed but was limited
+  to list. (issue #347)
+
 Bug fixes
 ^^^^^^^^^
 
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -3858,7 +3858,7 @@ def tocimxml(value):
 
     # List of values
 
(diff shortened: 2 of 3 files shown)


andy-maier:
+1 on the changes to add support for tuples.
However, let's have the discussion.
Another discussion point is whether we can actually support arbitrary iterables (iterables are broader than list+tuple)

KSchopmeyer:
I should never do a change like this and submit it in the same evening.   Agree with broader discussion and not sure that the list is sufficient but the code issue iteslf is that something like (propname) gets through and causes bad xml.  In this case, In this case, a) we agree on what iterables are allowed, b) we enforce that where the property is input, not way down in the toxml code where apparently it was interpreting my (propname) as a Value, not value.array.

we sit on this one until tuesday morning.

andy-maier:
Here is a Stackoverflow answer that lists all four possible ways to implement iterables:
http://stackoverflow.com/a/7542261

The trouble is, given these multiple ways, there does not seem to be an easy way to test whether an object is an iterable, other than looping over it.

andy-maier:
This article: http://nvie.com/posts/iterators-vs-generators/ describes the difference between iterable and iterator. Going by its definition "An iterable is any object [...] that can return an iterator", the test for an iterable would be to invoke `iter()` on it successfully. This is apparently what the for loop also does.

KSchopmeyer:
We agreed that:

1. We do not want to do checks that the server can do, i.e. single value for propertylist.
2. We should support both tuple and list for propertylist.
3. We want to limit this property
4. We could extend the test to do the iter test rather than is instance since string has already been tested.

KSchopmeyer:
+1

andy-maier:
+1

PR #350, opened 2016-06-06
fixes issue#339, Documentation of class level return for associators,…

KSchopmeyer:
… references

Updates documentation to reflect that class level return is actually a
list of tuples of classname, class and not just list of classes.

I slightly reformatted the associatornames and referencenames also.

Added doc to changes.rst


andy-maier:
+1

PR #356, opened 2016-06-08
Fixes #242: Using twine for upload to Pypi and other improvements in build

### LINKED ISSUE #242: Use twine for uploading to PyPI
It seems that the missing information about dependencies on PyPI is caused by uploading via `setup.py upload`. The [twine](https://pypi.python.org/pypi/twine) package on PyPI does it correctly, plus it uses secure comunication with PyPI, which `setup.py upload` does not when running on Python 2.6.

We should start using `twine` for the upload. Yet another development dependency, but it seems to be worth it.

See issue #113 for details.

andy-maier:

Note: The upload now requires to build on Linux (for the source archive and binary wheel archive) and to build on WIndows (for the WIndows installable). Upload works on both.

Details:

- Changed to use twine for upload to Pypi.
- Added wheel format to distribution archives (built by build target) and to Pypi upload.
- Added Windows installable .exe (built by buildwin target) to Pypi upload.
- Added check for build target to run on Linux or OSX (because sdist creates zip file on Windows).
- Added check for buildwin target to run on Windows.
- Changed package classifiers to indicate status: 5 - stable.
- Changed package classifiers to add Python 2 and Python 3 language versions.
- Replaced use of shell command in makefile with other means, because the shell command on Windows runs with the user's home directory as current directory.


andy-maier:
+1

KSchopmeyer:
+1

PR #357, opened 2016-06-09
Fixes #323: Added links to WBEM operations section to WBEMConnection and to all of its operation methods

### LINKED ISSUE #323: I think we lost the simple list of all operations in the move to new doc
In the old doc there was a list of the operations which represented a sort of toc to get to the operations easily.  In the new doc, they are all detailed in a long html page but not easy to get to each one.

I don't have a solutions but with 20+ operations, we need to do something like have an extra layer in the
left side menu, etc.

andy-maier:


andy-maier:
I assume your edit means +1. Merging.

PR #360, opened 2016-06-09
WIP fixes #215 enable the timeout for m2crypto

### LINKED ISSUE #215: under https m2crypto, the client timeout code is disabled in cim_http
With pr#376 for issue #363 we are now capable of demonstrating the exact behavior of timeout over a wide range of server delayed responses and timeouts.  It appears that the modifications to exceptions have fixed the issues we found here earlier.  The new openpegasus tests seem to demonstrate that time timeout works now for http and https over a wide range of timeouts and server delays so I propose closing this as fixed

Note that even the issue of very uncertain timeout time no longer exists since the new test against open pegasus demos that the timeout now appear to occur at about the correct time.

The one remaining test might be to confirm this with wbemcli which is where the original issues were found.

KSchopmeyer:
Teste manually with this simple patch and a hack to pegasus that disables responses so server accepts requests but does not respond.

Captured details in issue #215 but the general conclusion is:

1. enabling does not kill the server. It passes python 2 and 3 tests on my system.

2. It does cause the client to timeout with m2crypto.  However, the timeout times can run from correct to 300% of correct. (i.e. a 4 sec timeout causes timeout in from 4 to 12 seconds with most in 4 sec.)

3. We have another issue in that in when https verification is set we get a verification error exception and not a timeout exception.  See the details recorded in the issue

I had to do this manually since there is no automatic way to set the pegasus server to delay responses right now, it was a code hack.

I want to set up a mock to do this but have not completed that and not sure it is always the same thing since not all of the timers are in http itself.

### DIFF
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -450,14 +450,12 @@ def wbem_request(url, data, creds, headers=None, debug=False, x509=None,
                     # Therefore, we set the timeout at the level of the outer
                     # M2Crypto socket object.
                     # pylint: disable=using-constant-test
-                    if False:
-                        # TODO 2/16 AM: Currently disabled, figure out how to
-                        #               reenable.
-                        if self.timeout is not None:
-                            self.sock.set_socket_read_timeout(
-                                SSL.timeout(self.timeout))
-                            self.sock.set_socket_write_timeout(
-                                SSL.timeout(self.timeout))
+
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

andy-maier:
This one was not yet ready to be merged, but it also did not hurt. Karl will open another PR for the rest of it.

PR #365, opened 2016-06-10
fixes #364 Extend test_client.py to handle timeout exception and add test.

### LINKED ISSUE #364: test_client.py does not handle timeouts exception
Need to add callback for timeout exception since that is another special path for responses.

KSchopmeyer:
Extends test_client.py to add new callback for socket.timeout and
creates a first use of this function in a test that executes the timeout
exception return.

NOTE: Test marked HTTP because there is an issue with timeout and
HTTPS. See issue # 363

### DIFF
--- a/testsuite/test_client.py
+++ b/testsuite/test_client.py
@@ -185,7 +185,7 @@ class Callback(object):
       * `headers`: list of strings with HTTP headers of request
 
     Return value:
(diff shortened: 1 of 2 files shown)


andy-maier:
+1.
Is this ready to be merged?

KSchopmeyer:
yes, +1.

Issue #351, opened 2016-06-07
Generate test_client yaml from within the WBEMConnection operation methods


andy-maier:
In order to produce more test_client yaml testcases, it would be beneficial to generate those, instead of making them up manually. This can be achieved by having code in the WBEMConnection class in `imethodcall()` and `methodcall()` at the end when the result is available, that emits the yaml testcase for this operation. At this point, everything is available: Python input and output, and CIM-XML request and response (via debug). This could be controlled by another WBEMConnection instance attribute.

andy-maier:
Would like to wait with this until PR #320 has been merged.

andy-maier:
PR #320 has been merged; started working on this one.

andy-maier:
PR #366 is an initial stab at an operation recorder, but its TestClientRecorder only displays the operation, and needs to be implemented for writing a test_client testcase.

andy-maier:
PR #384 is the next step. Its TestClientRecorder supports operation args and results, but still has some limitations (see the PR for details).

andy-maier:
PR #384 has been updated and is now the complete implementation.

PR #366, opened 2016-06-10
Issue #351: Initial stab at operation recording support.

andy-maier:
**Ready to be merged** even though this is only a first stage in addressing issue #351.

Please review and test.

The purpose of merging this stage now is that the changes in cim_operations are large but still simple, and we should not mix that with small but difficult changes. Also, merging this stage allows testing with it, and that gives higher confidence.

The implementation provides a working operation recording, but the class that will produce the `test_client` yaml testcases at this point only outputs a debug print.

Details from the commit log:

- Added `_recorder` module with abstract base class `BaseOperationRecorder` and testcase generator class `TestClientRecorder` and a few helper classes.
- Added `operation_recorder` instance variable to the `WBEMConnection` class that allows to control the operation recording.
- Added support for recording the HTTP request and response to an operation recorder object, in the `wbem_request()` function. Reorganized the production of the HTTP header fields in that function.
- Added support for recording the `WBEMConnection` method input and result to an operation recorder in each of its operation methods.
- Improved handling of the result in `GetQualifier()`.
- Fixed incorrect parameter names in the description of qualifier methods (`Qualifier` -> `QualifierName`, `Qualfier` -> `QualifierDeclaration`). This affected only the documentation.
- Editorial changes in the docstrings of the `WBEMConnection` methods.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -148,7 +148,7 @@ Enhancements
   These debug prints can be enabled via the `debug` instance variable
   of the WBEMConnection object; they are targeted at development for
   investigating these situations.
-  
+
 * Extended run_cim_operations.py which is a live test against a server.
   It has only been tested against OpenPegasus but was extended to cover
   more details on more of the operation types and to create a test
@@ -177,9 +177,14 @@ Enhancements
     - CloseEnumeration
   PyWBEM does NOT implement the EnumerationCount operation as it
   is both deprecated and unusable. (Issue #9)
-  
+
 * Added a tutorial section to the generated documentation, using
   Jupyter Notebooks for each tutorial page. (issue #324)
+
+* Added the concept of operation recording on WBEM connections, that supports
+  user-written operation recorders e.g. for tracing purposes. Added an
+  operation recorder that generates test cases for the `test_client`
+  unit test module. (issue #351)
 
 
 Bug fixes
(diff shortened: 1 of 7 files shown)


KSchopmeyer:
Two comments.

1. I note that you had to put the recorder around every single cim_operation.  I think I see why, since before these methods go to imthodcall, we process the input to create separate namespace and name parameters (and maybe other inconsistencies) but it is a shame to have create a lot of extra code just for the recorder which only we will use.  Wonder if we could not have done that by generalizing imethodcall a bit (with some extra parameters) to handle the separation of arguments within imethod call since that is where it all comes back together again.

2. Same thing applies to having to add one more try block around each of the operations just to record.  since it will probably be largely us who does the recording.

3. Sorry about the : errors on the Returns and Parameters.

4. MaxobjectCount is recorded in CloseEnumeration but there is no MaxObjectCount in the call.

andy-maier:
On 1. and 2.: I thought about that but if we move the defaulting into imethodcall it becomes a quite unmanageable IMO. If we pass the original args in addition but leave the defaulting outside, that would have been possible to have the recorder code inside of imethodcall. However, the reason against that was the exception handling. I needed a single place where any exception can be catched, so that would have required the extra try block anyway. And try blocks are supposedly cheap in Python.

Bottom line: I could have moved the recorder code into imethodcall with an extra exception around its current code, and with some extra args that represent the original args, but I did not like that because of the additional args that would have shown up twice (original namespace and defaulted) and make it harder to understand.

On 4. An error.

KSchopmeyer:
I guess that the only other way to do it would have been to create another layer between imethodcall and this and that is bad also.

Again I am happy to have this committed piece by piece so if there is no way to simplify the code, lets go with it since my primary concert was simply the added complexity duplicated over a lot of functions.

+1

andy-maier:
Commit 98596c3 fixed Karl's comment 4.

andy-maier:
+1. Merging.

PR #367, opened 2016-06-11
An example of indication management that runs against openPegasus

KSchopmeyer:
This is a complete of an in process listener with a client that
generates a number of indications for that listener and then
closes the server and finally exits.

Please review code.  Also see below for issue (pegasus or this
dropping indications)

Note that this does process indications but there is a problem
here between pegasus and this example.  We never receive the
number of indications expected.  At this point we can review and
even commit but there is an issue somewhere between 
pegasus and the listener in losing indications. 

Andy please review the code to be sure it matches your needs and
requirements for good code and the listener requirements.
Thus, for example, I put a listener.log in the process_indications
function  and used logger throughout.  Here is an example display

(pywbem27)kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem/examples$ ./pegasusindicationtest.py  http://sheldon ks ks 10
2016-06-11 13:10:30,747 Client       INFO     WBEMListener started http port 50000 
++++++++++Received CIM indication #1: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,800 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #1: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
++++++++++Received CIM indication #2: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,804 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #2: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,805 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,807 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
++++++++++Received CIM indication #3: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,810 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #3: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
++++++++++Received CIM indication #4: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,814 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #4: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,817 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,820 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
++++++++++Received CIM indication #5: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,821 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #5: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
++++++++++Received CIM indication #6: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
++++++++++Received CIM indication #7: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,825 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #6: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
2016-06-11 13:10:30,828 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,830 pywbem.listener.139637088698320 INFO     ++++++++++Received CIM indication #7: host=127.0.0.1
CIMInstance(classname=u'Test_IndicationProviderClass', path=None, ...)
HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,831 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:30,832 pywbem.listener.139637088698320 INFO     HTTP/1.1 POST from 127.0.0.1: HTTP status 200
2016-06-11 13:10:35,334 Client       ERROR    Insufficient indications received exp=10 recvd=7
(pywbem27)kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem/examples$

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -186,6 +186,9 @@ Enhancements
   operation recorder that generates test cases for the `test_client`
   unit test module. (issue #351)
 
+* Added example of inprocess indication subscription and handling in the
+  examples directory.
+
 
 Bug fixes
 ^^^^^^^^^
--- a/examples/pegasusindicationtest.py
+++ b/examples/pegasusindicationtest.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+
+"""
+Example of handling subscriptions and indications from a particular
+provider in openpegasus. This example depends on a test class and
+method in OpenPegasus.  It creates a server and a listener, starts
+the listener and then requests the indications from the server.
+It waits for a defined time for all indications to be received and
+then terminates either normally if all are received or in error if
+not all were received.
+
+"""
+from __future__ import print_function, absolute_import
+import sys
+import time
+import logging
+from socket import getfqdn
+from pywbem import WBEMConnection, WBEMServer, WBEMListener, CIMClassName, \
+                   Error, Uint32
+
+# definition of the filter.  This is openpegasus specific and uses
+# a class that generates one indication for each call of a method
+TEST_CLASS = 'Test_IndicationProviderClass'
+TEST_CLASS_NAMESPACE = 'test/TestProvider'
+TEST_QUERY = 'SELECT * from %s' % TEST_CLASS
+
+# global count of indications recived by the local indication processor
+RECEIVED_INDICATION_COUNT = 0
+LISTENER = None
+
+
+def process_indication(indication, host):
+    """This function gets called when an indication is received.
+       Depends on logger inside listener for output
+    """
+
+    #pylint: disable=global-variable-not-assigned
+    global RECEIVED_INDICATION_COUNT, LISTENER
+    # increment count. This is thread safe because of GIL
+    RECEIVED_INDICATION_COUNT += 1
+
+    LISTENER.logger.info("++++++++++Received CIM indication #%s: host=%s\n%s",
+                         RECEIVED_INDICATION_COUNT, host, indication)
+
+
+def run_test(url, user, password, requested_indication_count, verbose):
+    """
+        Runs a test that:
+        1. Creates a server
+        2. Creates a dynamic listener and starts ti
+        3. Creates a filter and subscription
+        4. Calls the server to execute a method that creates an indication
+        5. waits for indications to be received.
+        6. Removes the filter and subscription, stop the listener, and stop
+           the server.
+
+    """
+
+    logging.basicConfig(stream=sys.stderr, level=logging.INFO,
+                        format= \
+                        '%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
+
+    conn = WBEMConnection(url, (user, password), no_verification=True)
+    server = WBEMServer(conn)
+
+    # Set arbitrary port for the http listener
+    http_listener_port = 50000
+
+    # Create the listener and listener call back and start the listener
+
+    global LISTENER
+    LISTENER = WBEMListener(getfqdn(), http_port=http_listener_port,
+                            https_port=None)
+
+    # connect the listener callback, add to server, and start listener
+    LISTENER.add_callback(process_indication)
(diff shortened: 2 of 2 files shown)


andy-maier:
Comments on the code:

1. The docstring of run_test() mentions the creation of the listener three times.
2. There is  comment somewhere that the filter gets removed implicitly with server removal, but it is removed explicitly.
3. I would stop the listener before removing the server etc.
4. The waiting time does not seem to be sufficient.
5. The logging happens twice, you don't need a handler because the root logger already writes to stderr.

Otherwise, great!

andy-maier:
One more:

* I don't understand why the code in main() chechs for `len(argv) < 5`, if there are only 4 args.

andy-maier:
Forget my last comment ... max index vs length confusion....

KSchopmeyer:
+1

Found the problem that was causing compile error and no everything integrated.

andy-maier:
Remaining issues:

* On  "6. Remove server which should remove the filter." in `run_test()`:
  There filter is removed explicitly, so that docstring should be adjusted to the code.

* Please double check whether all indications can now be received during the waiting time.

KSchopmeyer:
+1 for the example but we have a real problem in that not all indications are received.

I asked for 50 indications with print for each loop. Recieved 8 (all in first second) but then it just sat waiting for about 13 seconds until it terminated telling me that 8 of 50 had been received.

Lets separate the commit for this code from the issue of missing indications.   I think that there is a timing issue somewhere but have not gone through the pegasus trace to assure that the required number has been created so far.  In an earlier version I used a pegasus function that sends one indication per method call and when I set a delay between method calls, it increased the percentage received by process_indication but even then it was not 100%.

I will run a test after we finish this pr and actually count in the pegasus trace so we can be sure it is not pegasus.  However, these providers are used in a number of pegasus internal tests and the tests pass.

The whole thing should probably be wrapped in a with or some similar cover to assure that we always remove the filter, etc. at the end of the program but want to get some example in so we can test more.

andy-maier:
Agree on separating the issue of missing indications.
+1 otherwise.

andy-maier:
Created issue #375 to follow up on the missing indications.

PR #372, opened 2016-06-15
fixes #241 Add the individual pull operations to wbemcli

### LINKED ISSUE #241: Document strategy for pywbem command line tools in contributors guide
Karl pointed out in PR #236 that we need to come up with a strategy for where to have the sript-ish code when we create more pywbem command line tools.

Here is a stab at that, for discussion:

- Anything that could be useful as a module goes into a module (not in the script).
- Command line argument processing, plus all of the rest related to the script (i.e. that is not useful as a module) goes into an extension-less script file, whose hash-bang states it is Python (using `/usr/bin/env python`, to pick up any virtual python).
- There is a .bat file for Windows that invokes python with the suffix-less script of the same name.

That strategy provides for:

- commands with a clear command name without being tied to a suffix revealing the implementation language
- same command name whether on Linux or Windows
- as a result, the same single documentation applies to both
- nicely supported by Python setup scripts (both the suffix-less file and the Windows .bat file get installed into the script directory and thus are available in the PATH).

I do realize that these benefits can also be achieved by what we had before this PR, but it allowed running the module as a script to have a command, and now users are forced to use the suffix-less command. Maybe that is the real reason.

KSchopmeyer:
Adds each operation with its parameters and detailed help to wbemcli.
This includes all of the pull operations.

Added note to changes.rst.

Note that not all have been tested however since that is a manual
operation.

I am not happy with the prefixes. First there are so many of them I cannot remember them and second, it really does not work to manually type these things in one at a time; the sequences timeout before you get the second part typed in.  But in any case, this is the logical extension to wbemcli for each pull operation.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -189,6 +189,8 @@ Enhancements
 * Added example of inprocess indication subscription and handling in the
   examples directory.
 
+* Extended wbemcli for all pull operations. (issue #341)
+
 
 Bug fixes
 ^^^^^^^^^
--- a/wbemcli
+++ b/wbemcli
@@ -22,7 +22,9 @@
 
 """
 wbemcli - Provide an interactive shell for issuing operations against a WBEM
-server.
+server. The server connection (WBEMConnection) is executed when during
+startup and any python statements including the pywbem wbem requests defined
+in cim_operaitons.py are executed from the python command line.
 
 Invoke with `-h` or `--help` for a help message.
 """
@@ -327,6 +329,7 @@ def AssociatorNames(op, ac=None, rc=None, r=None, rr=None):
                                 ResultRole=rr)
 
 
+# pylint: disable=too-many-arguments,redefined-outer-name
 def Associators(op, ac=None, rc=None, r=None, rr=None, iq=None, ico=None,
                 pl=None):
     """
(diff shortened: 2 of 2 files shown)


andy-maier:
Just a few comments on the abbreviations:

1. The abbreviation for `OpenEnumerateInstancePaths` is `oep` and thus dies not follow the rule of using the upper case chars of the full name, which would result in `oeip`.

   Same commemt applies to `OpenAssociatorInstancePaths` and `OpenRefetenceInstancePaths`.

2. Same applies to the `FilterQuery` and `ContinueOnError` args of the pull ops.

KSchopmeyer:
Went ahead and made the changes to follow the standard. Still not tested all of the operations. Bet we have never tested all of the operations in wbemcli.

KSchopmeyer:
+1  except only minimal testing.

andy-maier:
My take is: Having a simple rule to translate the long names into the short names is beneficial.

I have two more:

* `cor` for `ContinueOnError` does not follow the simple rule. Does `coe` stand for anything bad?

* The short name for the `OperationTimeout` arg is `ot` in the function but `it` in the docstring.

KSchopmeyer:
Don't ask me why I chose cor that night.  And I will review operation timeout since really should be interoperation timeout.

andy-maier:
+1

PR #380, opened 2016-06-20
fixes #346 Add tests to run_cimoperations.py for server class

### LINKED ISSUE #346: Add server tests to run_cimoperations.py
Extend this test so that it executes the new server tests.

I set as release mandatory since I think we should have some confirmation of this code set before releasing other than the example.

KSchopmeyer:
Add tests of all of the functions of the server class
to run_cim_operations including comparison of results
against results from other function to access data
and/or expected results.

Add note to changes.rst

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -191,6 +191,9 @@ Enhancements
 
 * Extended wbemcli for all pull operations. (issue #341)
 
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

KSchopmeyer:
+1

PR #382, opened 2016-06-21
Fixes #377: Incorrect lexical case of Interop namespace name with OpenPegasus

### LINKED ISSUE #377: case issue in PyWBEM server namespace access
The functions that  return the interop namespaces seem to have an issue with casing and the interop namespace:

Note below that the explore.py example displays the interop namespace with an lowercase o and the same namespace in the namespace list with an uppercase O.

Interop namespace:
  root/PG_Interop
All namespaces:
  root/PG_InterOp
  root/SampleProvider
  root/benchmark
  test/CimsubTestNS2
  test/CimsubTestNS3
  test/CimsubTestNS0
  test/CimsubTestNS1
  root/PG_Internal
  test/WsmTest
  test/TestIndSrcNS1
  test/EmbeddedInstance/Static
  test/TestINdSrcNS2
  test/TestProvider
  test/EmbeddedInstance/Dynamic
  root/cimv2
  root
  test/cimv2
  test/static

andy-maier:
Ready to be merged.
Please review.

Details: The root cause for showing `root/PG_Interop` as the Interop nanmespace for OpenPegasus was that that was the attempted namespace name, and the current code used the attempted name if successful.
This has been changed to inspect the returned CIM_Namespace instance paths and to use their `Name` keybinding, if possible. Otherwise, the attempted name is still used.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -273,6 +273,10 @@ Bug fixes
    module).  The minimum level protocol set by the client is TLSV1 now whereas
    in previous versions of pywbem it was SSLV23. (issue # 295)
 
+* WBEMServer.interop_ns now contains the returned interop namespace name,
+  if possible, instead of the attempted one.
+
+
 pywbem v0.8.2
 -------------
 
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

KSchopmeyer:
+1

PR #386, opened 2016-06-23
Fixes #381: Added support for a config variable to control enforcing the value range of CIM integer types

### LINKED ISSUE #381: Introduce a strict mode
Issue #374 shows that there is usage out in the field where the value range used on a CIM numeric datatype exceeds the allowed value range for that datatype. This may happen more often, so users should have a way to disable the additional checking for CIM integer values that is being introduced in pywbem 0.9.0.

Also, need to investigate what other checks have been introduced that may cause similar issues.

It coul dbe a module-global boolean constant that is used to switch to non-strict, with strict being the default so users learn what's used incorrectly.

andy-maier:
Ready to be merged.
Please review.

This PR introduces a `pywbem.config` module for setting config variables.

It defines a first config variable `ENFORCE_INTEGER_RANGE` that controls whether the value range of CIM integer types is enforced or not.

The config variables are documented in a new chapter "Configuration".

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -125,7 +125,9 @@ Enhancements
   input parameters that are supported by `int()`.
 
 * Added the concept of a valid value range for the CIM integer data types, that
-  is enforced at construction time.
+  is enforced at construction time. For compatibility, this strict checking can
+  be turned off via a config variable:
+  `pywbem.config.ENFORCE_INTEGER_RANGE = False`.
 
 * Extended wbemcli arguments to include all possible arguments that would
   be logical for a ssl or non-ssl client. This included arguments for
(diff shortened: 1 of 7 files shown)


andy-maier:
+1

KSchopmeyer:
+1  I only have one question. Is there going to be one for real32 also at some point?

KSchopmeyer:
One other possible thought.  Would it be more logical to have a general strict and just document what it meant rather than a number of separate variables the define strict vs. non-strict??? That is just a question now.

andy-maier:
We discussed this in the call today and Karl approves.
We are going for granular switches if we need more of them.
There is no range checking for real right now.

PR #388, opened 2016-06-24
Fixes #379: Add test to execute listener code

### LINKED ISSUE #379: run_cimoperations.py. Add tests for listener
Need some tests for listener.  Logical place is run_cimoperations.py for a live test.  Added this one subsequent to finishing first cut at adding the server tests to run_cim_operation.py.

KSchopmeyer:
Ready for review and merge.

This is a single test in in run_cimoperations.py that creates and deletes server, listener, subscription, filter.  It does not process indications nor test for errors.  It does verify that the host has the defined filter and subscription and that it deletes them.  We need to add more tests but wanted to get the basic structure running.  Note that this test runs only against pegasus because it depends on a particular class/provider (pegasus validates subscriptions against providers)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -202,6 +202,10 @@ Enhancements
   message to be displayed. Note, this is an incompatible change in the
   command line options. (issue #216)
 
+* Extended run_cimoperation.py to include basic listener test to create and
+  delete indication subscriptions in the server.  This test does not process
+  indicaitons from the server. (issue#379)
+
 
 Bug fixes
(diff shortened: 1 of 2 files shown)


andy-maier:
We already use the server URL as an identification in the listener methods.

What we could do as an improvement, is to change the documentation to talk about server ID instead of server URL.

Plus, of course to fix the 60000 char issue ...

andy-maier:
+1.
Please rebase.

KSchopmeyer:
Rebased

PR #390, opened 2016-06-25
Fixes #378: Multiple profile example and add select_profiles function

### LINKED ISSUE #378: examples/explore.py issue with multiple profiles
For the indication and server profiles, the code tests org and name but not version.  This means that
the result could be multiple profiles returned but only one is considered in the code following.

Shouldn't this also test against version to be complete?

KSchopmeyer:
Review requested but DO NOT commit since have couple more cleanups if you agree with function.

I used this to add another function to server.  Please review to see if you approve.  This gets
a set of profiles with a single request which seemed like a logical simplification.

I used it in the explore example.

It is not ready to commit now.  1 Have the old function I wrote in the example code and have not
done test for it.  But if you agree the next steps are simple and will complete it.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -206,6 +206,11 @@ Enhancements
   delete indication subscriptions in the server.  This test does not process
   indicaitons from the server. (issue#379)
 
+* Extend the server class with a function (filterProfiles) to select a
+  subset of profiles from the server based on organization, profile name,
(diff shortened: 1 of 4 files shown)


andy-maier:
+1 based on rewiewing.

andy-maier:
Is this ready to be merged?

KSchopmeyer:
Yes.  Somehow I got a conflict in it for a while but that is cleared.

+1

PR #392, opened 2016-06-27
Fixes #389: Changing 'array' to 'list' in cim_http.py

### LINKED ISSUE #389: Pylint finds isinstance with the target type of array
E:422,64: Undefined variable 'array' (undefined-variable) in cim_http.py

andy-maier:
Ready to be merged.
Please review.

The reason the change loig is not updated with a fix entry, is that this error never showed up to the user, because in our case, the data object never has a ` read()` method.


KSchopmeyer:
+1

PR #398, opened 2016-06-28
Fixes #354: Added info about paths in returned objects, and other docu fixes

### LINKED ISSUE #354: Document what will be in path component of all enumerate responses
There is no documentation today on the expectations of what should be in the path or path component of classes on enum/assoc/reference responses.

andy-maier:

This one ended up to cover more than originally was intended.

Details, from the commit log:

There are these code (=non-docu) changes:

- Changed name of namedtuple item 'QueryResultClass' in result of `OpenQueryInstances()` to 'query_class'.
- Made `DEFAULT_CA_CERT_PATHS` constant in `cim_http` module public and documented it. It was referenced in the description of `WBEMConnection()`.

The rest are docu-only changes:

- Added missing information about paths in returned objects to all operations (it was missing in most of them). This was the actual reason for this change.
- Added missing information about paths in input objects in all operations (it was missing in only a few).
- Fixed incorrect description of `host` attribute of `CIMInstanceName` and `CIMClassName` which stated to be the WBEM server URL, to be just `host[:port]` instead.
- Fixed the format of return value descriptions to avoid "Result Type:" section in generated documentation. This sometimes could be achieved by certain words in the first sentence, but in two cases that was not sufficient and I discovered that inserting ": " before the first line did the trick.
- Fixed the format of exception descriptions to use "Raises:" instead of "Exceptions:" because the latter was not recognized by Sphinx as a valid alias to "Raises".
- Fixed documentation for `Pull*()` methods to return *next* set of items and not *initial* set of items.
- Made sure all pulled association methods state that they cannot be used to retrieve classes and made these statements consistent.
- Added statements in descriptions of CIM objects as to which attributes are guaranteed to never be `None`.
- Made statements about `extra` keyword argument consistent in all ops. w.r.t. wording and position.
- Swapped real descriptions and references to details between instance variables and ctor args, in all CIM object classes.
- Made terminology consistent between 'argument' vs. 'parameter'; used 'attribute' in place of 'instance variable' or 'component'.
- Fixed incorrect use of 'named argument' to become 'positional argument'.

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -456,3 +456,10 @@ The operation methods of :class:`~pywbem.WBEMConnection` raise
   - authentication failed
   - user is not authorized to access resource
 
+.. _`Default CA certificate paths`:
+
+Default CA certificate paths
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autodata:: pywbem.cim_http.DEFAULT_CA_CERT_PATHS
+
--- a/docs/mof_compiler.help.txt
+++ b/docs/mof_compiler.help.txt
@@ -1,40 +1,88 @@
 usage: mof_compiler [options] moffile ...
 
-Compile MOF files, and update the repository of a WBEM server with the result.
+Compile MOF files, and update a namespace in a WBEM server with the result.
 
 Positional arguments:
   moffile               Path name of the MOF file to be compiled.
                         Can be specified multiple times.
 
 Server related options:
-  Specify the WBEM server and namespace
+  Specify the WBEM server and namespace the MOF compiler works against, for
+  looking up existing elements, and for applying the MOF compilation results
+  to.
 
-  -u url, --url url     URL of the WBEM server.
-                        Default: /var/run/tog-pegasus/cimxml.socket
+  -s url, --server url  Host name or URL of the WBEM server (required),
+                        in this format:
+                            [scheme://]host[:port]
+                        - scheme: Defines the protocol to use:
+                            - "https" for HTTPS protocol
+                            - "http" for HTTP protocol
+                          Default: "https".
+                        - host: Defines host name as follows:
+                             - short or fully qualified DNS hostname
+                             - literal IPV4 address(dotted)
+                             - literal IPV6 address (RFC 3986) with zone
+                               identifier extensions(RFC 6874)
+                               supporting "-" or %25 for the delimiter
+                        - port: Defines the WBEM server port to be used.
+                          Defaults:
+                             - 5988, when using HTTP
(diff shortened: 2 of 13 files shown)


andy-maier:
+1

KSchopmeyer:
taking a bit of time go through this one..

**cim_operations.py**

1. line 2010 of cim_operation.py.  Shouldn't that be :class:`~pywbem.CIMInstanceName` and samefor others like  line 2221, 2471, 2704, 3630. However, I note that this actually works with just the class name in the output documentation so guess you can ignore this comment. guess I don't understand all of the options of RestructureTxt and Sphinx

2. I think that the change to you made to guery_class should probably be query_result_class just to keep the names consistent with the spec parameters

3. Line 4722.  The text ("without parenthesis or any parameter signature)" leaves me confused (not that this was not a change you made but I am not sure what it means. You have any idea what it means?

4. In some cases we use the following terminology  (see line  4722, MethodName)         If specified as a string, the string is interpreted as a class name in the default namespace of the connection, and can have any  lexical case.

It would be simpler to say case independent  name or  name (case independent) than " can have any lexical case.

That is a nitpick but just makes thing clearer.  That one is just a suggestion since there are lots of that usage so probably not worth the effort.

5. Line 4729. "For instance level usage"  probably should be for instance-level usage. Bet there are more of these also.

6. Another nitpick.  You changed "containing" to "that contains" in several places.  Either is correct but containing is just a bit shorter.

7. Noted that in places we use "**instances** (list of :class:`~pywbem.CIMInstance`): (line 2839). It is logical that this is a python list  but should be be setting some attribute on list to clarify that??? Not sure, just reading through and wondered.  Adding examples will help this a lot but that is not for today.  Actually in cim_obj.py there are cases where list in list of is tagged. (line 1224)


**cim_obj.py**

8. cim_obj.py  - Line 1204 states that the properties will never be None.  I assume that means that you can have an empty dictionary???. Same for qualifiers.  Think we are trying to say that it will always be a dictionary but that the dictionary may be empty. 

9. See cim_objpy line 1224.  That  attribute is a very confusing thing to be attached to the instance itself. I am not sure exactly what the user is supposed to do with it.  Filtered by who, when, etc.   We can make this a separate bug but should clarify at some point. Not part of this change but I had the questions when I read through to document.

andy-maier:
Commit dc20050 addresses all issues in Karl's comment that are marked with DONE. The remaining ones need discussion.

andy-maier:
We agreed in the call yesterday that the remaining work is to deprecate property_list on CIMnstance.

KSchopmeyer:
+1

PR #404, opened 2016-07-01
Fixes #402: Modify compiler to separate class properties from instance properties

### LINKED ISSUE #402: Creating instance in compiler sets values into class
Example: testmofs/TestTypes.mof.  Currently instances disabled but in file. If you try to compile with
the instances enabled it sets the instance values into the class.

Probably line 1379 of mof compiler which uses the class properties as the instance properties
    inst = CIMInstance(cname, properties=cc.properties,
                       qualifiers=quals, path=path)

KSchopmeyer:
Issues agreed on I hope and fixed. **Ready For Review** with update 2 July morning.

When instances were created by the compiler, it got information from the
corresponding class but put the instance property values into the class
itself since the compiler was not copying the properties from the class to the instance.
Separated these so the instance properties go into the instance. This means:

1. We are not trashing the class when we do an instance compile.
2. Each instance receives its own properties if compiling more than one instance. Before the property values for the last instance compiled went into all instances and the class.
3. Originally it build instances with all properties from the class and included whatever value was in the class itself. Now only properties defined in the mof instance definition are put into the new instance.
4. Originally all the qualifiers from the class properties were seen in the instance properties.  Now we do not copy qualifiers so the instance created has only qualifiers that it may set.

Note: Removed the instances from TestTypes.mof to keep it a pure class test since that already
exists as a unit test and  set them into new class TestInstance.mof.

Added test mof TestInstance.mof for create instances.  This is based on the class used in the
TestType.mof file using a pragma to copy the class into a new TestInstance.mof file.   

The embedded instance properties in the instances are commented out because there is still an issue with those properties (see issue #147) (testing for 147 is where I found this issue this morning)

**DISCUSSION:** There are two issues for discussion:

1. Should we be moving the class-level property qualifiers to the instance properties.  Proposal NO.
**CONCLUSION:**  2 jul We agreed that we should not be putting class qualifiers into Instances.

2. The last lines of the instance mof builder in the compiler tries to build a path using the qualifiers
in the instance mof;  no key qualifier, therefore, no  property values assigned to path.  The question is whether we should even be trying to create the path part of the  instance mof-compile.
**CONCLUSION:** 2 jul We agreed that we should not try to build path

Since this was completely broken before this change. We blew up the class mof in trying to create the instance mof,  I would guess we are free to do what is right. There is no guarantee when you build instance mof that there are all the properties to build the path so I think that building the
path is illogical and we should just remove that code and document that.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -297,6 +297,16 @@ Bug fixes
 * WBEMServer.interop_ns now contains the returned interop namespace name,
   if possible, instead of the attempted one.
 
+* Fix issue where mof_compiler was setting values for compile of instances into
+  the class object and also setting the values for the last compiled instance
+  in a compile unit into all other compiled instances for the same class. Since
+  the concept of compiling a path into compiled instances is flawed (there is
+  no requirement to include all properties into a instance to compile that code
+  was removed so that the path is NOT build into a compiled instance. Finally
+  the qualifiers from the class were also included in compiled instances which
+  was incorrect and an accident of the code.  They are no longer included into
+  the compiled instances.)  (issue # 402)
+
 
 pywbem v0.8.2
 -------------
--- a/docs/mof_compiler.help.txt
+++ b/docs/mof_compiler.help.txt
@@ -1,40 +1,88 @@
 usage: mof_compiler [options] moffile ...
 
-Compile MOF files, and update the repository of a WBEM server with the result.
+Compile MOF files, and update a namespace in a WBEM server with the result.
 
 Positional arguments:
   moffile               Path name of the MOF file to be compiled.
                         Can be specified multiple times.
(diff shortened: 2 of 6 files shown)


andy-maier:
Wow..... 
I agree with the proposal for the first remaining issue and for the second issue, I think the path should not be built for an instance in MOF. It simply does not have an address.

KSchopmeyer:
I think we agree.  Will make one more change to remove the code that trys to build the path in the mof_compiler and resubmit for final review.

Instances will be compiled without the class qualifiers for properties and without attempting to build
a path element for the instance

andy-maier:
+1

andy-maier:
This PR fixes both issues #402 and #406.

Issue #420, opened 2016-07-05
More clearly id server subscription objects


KSchopmeyer:
When pywbem creates objects for the server for subscriptions, destinations, and filters it uses the naming convention cimfilter<binary_time> or cimlistener<binary_time>

This causes a couple of issues:
1. There is a small possibility of duplication for indications created, probably not in the same listener but for different listeners.  Probably simpler to use uuid rather that time.time.

2. There is no way for this listener to attach something that could it could use to identify itself in the future to clean up subscriptions stuck in the server.

We propose the following changes:

1. Change to use uuid in place of time.time to generate the final part.
2. change the fixed test from cimfilter and cimlistener to pywbemfilter and pywbem listener
3. Allow a tag to be attached for a) each listener creation and b) each filter creation as an optional parameter of the create_filter and
2. Allow the listener to attach a prefix for the listener itself to each indication that would be part of the listener init and to each filter created.  These would be optional parameters where the default is None

KSchopmeyer:
See pr #422

PR #422, opened 2016-07-07
fixes issue #420 Add an optional string id to listener and filter

KSchopmeyer:
Ready for review:

Andreas: you may elect to put this on off until you finish restructuring listener

Adds an optional user defined string id to the listener
definition an another one to the filter definition.

The listener id is prepended to the name of all destination and
listener objects if it exists.  The filter id is prepended to the
name property of all filters after the listener id.

These are an informal mechanism for the user to identify instances
of filters and destination ojects that it has registered with the
WBEM server.

Nothing is prepended if these attributes are None

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -93,13 +93,13 @@ any indications it receives, in MOF format.
 """
 
 import re
-import time
 from socket import getfqdn
 import logging
 import ssl
 from xml.dom import minidom
 from xml.parsers.expat import ExpatError
 import threading
+import uuid
 import six
 from six.moves import BaseHTTPServer
 from six.moves import socketserver
@@ -110,7 +110,7 @@ from ._server import WBEMServer
 from ._version import __version__
 from .cim_obj import CIMInstance, CIMInstanceName, _ensure_unicode
 from .cim_operations import check_utf8_xml_chars
-from .cim_constants import CIM_ERR_FAILED, CIM_ERR_NOT_SUPPORTED, \
+from .cim_constants import CIM_ERR_NOT_SUPPORTED, \
                            CIM_ERR_INVALID_PARAMETER, _statuscode2name
 from .tupleparse import parse_cim
 from .tupletree import dom_to_tupletree
@@ -573,6 +573,7 @@ class ListenerRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
                  logging.INFO)
 
     def _get_log_prefix(self):
+        """Return the prefix components for a log entry"""
         return '%s %s from %s' % \
                (self.request_version, self.command, self.client_address[0])
 
@@ -584,7 +585,7 @@ class ListenerRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
             (__version__, self.server_version, self.sys_version)
 
 
-
+#pylint: disable=too-many-instance-attributes
 class WBEMListener(object):
     """
     A WBEM listener.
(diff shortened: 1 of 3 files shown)


pywbem/_listener.py
```diff
@@ -632,6 +632,11 @@ def __init__(self, host, http_port=DEFAULT_LISTENER_PORT_HTTP,
 
             `None` means not to use a private key file. Setting up a port
             for HTTPS requires specifying a private key file.
+
+          listener_id (:term:`string`):
+            If not `None` a string that prefixes the name property of filters
+            and handler instances to help the listener identify its filters and
```
andy-maier: On the term "handler": We use the term "listener destination" elsewhere, and I suggest to stay consistent. If you think it is needed, we could introduce the term "handler" somewhere, e.g. in `WBEMListener.add_server()`.

Related to consistency: I realized that in `add_server()`, the description uses the term "indication listener", which is also inconsistent and should be changed to "listener destination".

pywbem/_listener.py
```diff
@@ -632,6 +632,11 @@ def __init__(self, host, http_port=DEFAULT_LISTENER_PORT_HTTP,
...
+            handler instances in a WBEMServer.
```
andy-maier: On term "WBEMServer": Here we talk about the actual WBEM server, not about the class `WBEMServer`, so I suggest to change this to "WBEM server.".

pywbem/_listener.py
```diff
@@ -893,6 +905,9 @@ def add_server(self, server):
         else:
             raise ValueError("Invalid scheme in server URL: %s" % \
                              server.conn.url)
+                             
+        # set the prefix into the server object
+        server.listener_id = self._listener_id
```
andy-maier: I don't like that at all; it creates a new instance variable from outside of the class, that has no real semantic meaning for that class, and that uses the object just as a store to pass it on to when it is needed.

Alternative proposal, removing the need for adding the `listener_id` instance variable to `WBEMServer`:
* For `_create_destination()`, add a `listener_id` parameter, and when calling it in `add_server()`, pass `self.__listener_id` as its value.
* For `_create_filter()`, add a `listener_id` parameter, and when calling it in `add_dynamic_filter()`, pass `self.__listener_id` as its value.

pywbem/_listener.py
```diff
@@ -1250,13 +1265,15 @@ def _create_destination(server, dest_url):
...
+    dest_inst['Name'] = '%spywbemlistener%s' % (server.listener_id,
+                                                uuid.uuid4())
     dest_inst['Destination'] = dest_url
 
     dest_path = server.conn.CreateInstance(dest_inst)
     return dest_path
 
-def _create_filter(server, source_namespace, query, query_language):
+def _create_filter(server, source_namespace, query, query_language, \
+                   indication_id_prefix=None):
```
andy-maier: The `indication_id_prefix` parameter is not used in `add_dynamic_filter()` which is the only place that calls this internal function. As a result, the user so far has no way to define it.

If it is really necessary to have this ID in addition to the listener ID, it should be externalized in `add_dynamic_filter()`. If not, it should be removed.

andy-maier:
-1
See comments.

KSchopmeyer:
Incorporated comments above and extended the pegasus tests to test most of the id alternatives.

KSchopmeyer:
+1

andy-maier:
+1
Except, you could add a test for the name strings.
Please rebase.

KSchopmeyer:
Added more tests so that now the IDs formatted names are completely tested.  Also cleaned up other pegasus based listener tests and one issue in the listener itself.  This passes the pywbem tests and also the pegasus tests.

+1

Issue #169, opened 2016-03-31
MOFParseError does not have attributes for line/column in error


andy-maier:
See PR # 423
The `MOFParseError` exception class is a subclass of `ValueError` and does not provide any additional attributes.
- It would be good if it indicated the line and column numbers of the problem it communicates. And the file name.
- Also, it is somehow strange to see it derived from `ValueError`. Ideas for different classes, or reminders on why it needs to be that way, are welcome.

KSchopmeyer:
Note, however, that when it displays an error it puts the line with the error and then another line with a ^ at the position of the error.  However, It would be better to actually do some format that editors can handle to get the user back to the error itself.

andy-maier:
That's fine for a user of the command line script. But if we also want to support programmatic callers of the API well, then the MOFParseError exception is probably a better interface for them compared to having their own logger and parsing what's printed on it (including anything that is not a syntax error at all).

KSchopmeyer:
Interesting. Note that tools like pylint do both. Which was one of the things that was causing issues with my pylint.awk program but which geany handles just fine.

KSchopmeyer:
Propose we follow the pylint model.

andy-maier:
I'm not arguing against logging the syntax errors, I just think we need the exception based interface in addition. After all, the exception is thrown, it has just no meaningful information inside.

andy-maier:
We agreed to have the information in the exception object.
In addition, the pylint comment by Karl means that the log messages should be in pylint standard format which can be understood by some editors (such as geany)

KSchopmeyer:
Andy. Maybe I am confused but MOFParseError subclasses from exceptions.py.Error which is the base for pywbem and subclasses from Exception.  MOFParseError does not subclass from Value.
In fact, MOFParseError takes line and file information directly from ply yacc in the p_error(p) function when called from ply.  It gets file name, line number, line position,  and context(i.e. that line) from the parser in p_error and stuffs them into MOFParseError instance.

One issue is that it does not get this correctly when called from elsewhere (i.e from a mof_production)  because for whatever the reason, when you use the p parameter that is part of the mof_compiler definition of the parser, not all of the same functionality is visible.  Means digging further into the parser to find out why.  Once we are getting the same information universally and using the same exception for all mof_errors, we can generalize the result.  See the issue #204 where we proposed a second exception class.

andy-maier:
[...]

Issue #204, opened 2016-04-11
MOF parser: Conflicting qualifier flavors go undetected


andy-maier:
If the qualifier declaration in MOF specifies conflicting flavors (e.g. both EnableOverride and DisableOverride), the last one encountered currently wins in the pywbem MOF compiler (see `_build_flavors()` in `mof_compiler.py`).

DSP0004 defines the following rules:
- Flavor (ToSubclass) and flavor (Restricted) shall not be specified both on the same qualifier type. If none
  of these two flavors is specified on a qualifier type, flavor (ToSubclass) shall be the implied default.
- If flavor (Restricted) is specified, override permission is meaningless. Thus, flavor (EnableOverride) and flavor (DisableOverride) should not be specified and are meaningless if specified.
- Flavor (EnableOverride) and flavor (DisableOverride) shall not be specified both on the same qualifier
  type. If none of these two flavors is specified on a qualifier type, flavor (EnableOverride) shall be the
  implied default.

**Proposed solution:**
1. Detect conflicting flavors (e.g. in `_build_flavors()` in `mof_compiler.py`) and raise MOFParseError in case of conflicts. Conflicting flavors are ToSubclass & Restricted, and EnableOverride & DisableOverride.
2. The rule not to specify EnableOverride or DisableOverride in case of Restricted, is only a recommendation. No exception should be raised, if violated. Instead, the notion of "meaningless" should be implemented somehow, e.g. by setting the `overridable` attribute to `None` if `tosubclass` ends up being `False`. This question is basically in the same category as the application of default values (see issue #203) and should be handled consistently with that.

andy-maier:
**Discussion:** Discuss proposed solution.

KSchopmeyer:
The problem presented by the code is more complex than the discussion above in that _build_flavors is used at two points in the compiler
1. Building Flavors when processing a CIMQualifierDeclaration
In this case the conclusions are correct in that we want to assure that conflicting flavors are rejected.
2. Processing a qualifier usually in a class where a CIMQualifier declaration already exists
In this case we not only want to assure that conflicting flavors are rejected but since there is already an existing qualifierDeclaration we need to determine if we want to really process the qualifier in terms of that declaration. Thus, if Override is not allowed in the decl do we want to allow in in the class definition or let this be passed on to the repository? Note that we process only for the existing class and do not take into account the superclass.

KSchopmeyer:
[...]

PR #423, opened 2016-07-12
fixes issues #169 and #204 Clean up MofParseError Class and test conflicting flavors

KSchopmeyer:
**Andy:** Ready for Review. See one question below.

Note: resubmitted with fix for issue#204 added since that is a test for new MOFParseError exception.

Note: For Tuesday and Wednesday morning travis repeatedly failed.  However, it passed completely as of about 13:00 US CDT Wednesday.

Adds attributes at the class level for the MOFParseError class and
cleans up the error processing. This identifies the attributes,
centralizes
how they are acquired (several are acquired from the parser token
returned).  It prints out the same error message as before these
changes.

Adds a simple test for conflicting flavors (tosubclass and restricted or enableoverride and disable override). Also adds tests for case insensitivity in flavors and invalid flavor names.

Note that this does NOT test for conflicts between flavors in a qualifier and its corresponding qualifier declaraction (ex. qualifier declaration has disableoveride and qualifier has enable override)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -238,6 +238,15 @@ Enhancements
   CIM_IndicationSubscription objects from WBEM Server. (issue #421).
   )
 
+* Clarify MofParseError by defining attributes as part of the class init and
+  moving some code from productions to the class itself (issue #169). This
+  make MofParseError exception more suitable for use from the productions
+  themselves. The original definition was really only for use as a call from
+  ply. Add tests for invalid qualifier flavors to unit tests and add test in
+  mof_compiler.py for conflicting flavors ex. tosubclass and restricted in
+  the same definition. This test uses the new MofParseError (issue #204)
+
+
 Bug fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


KSchopmeyer:
+1

andy-maier:
Please rebase.

andy-maier:
On the question about using `pe.msg`: I think this is all correct code; every `MOFParseError` object should have an attribute named `msg` now.

What exactly fails, is it the access to the attribute, or the assertion on its value?

KSchopmeyer:
Updated for review

+1

KSchopmeyer:
Rebased.

andy-maier:
Karl,
I can reproduce the PyLint issue with this branch.

I believe this is a PyLint bug and have appended a simplified test case to the existing [PyLint issue 811](https://github.com/PyCQA/pylint/issues/811). For now, I suggest to live with it.

It works to disable the warning on the line where it is raised, but I'm not sure it is worth the effort.

andy-maier:
+1

Issue #421, opened 2016-07-06
Add methods to get all subscriptions and destinations objects from server.


KSchopmeyer:
To help the user sort out what is actually in the server add new functions to get all subscriptions and destination objects from a server

KSchopmeyer:
See pr #431

andy-maier:
Closing it, as it is fixed by #431.

PR #431, opened 2016-07-30
fixes issue #421 Add functions to get listener destination and subscriptions from Server

KSchopmeyer:
Ready For Review:

Adds two new functions to Listener.  These simply get the instances from
the corresponding CIM Class in the server so that a client can get
the server side of all of the various subscription classes.

COMMENT: I thought about adding a filter mechanism to this so you could filter by regex against particular properties (name, etc.) but concluded that if we want to filter lets make it a separate function.

Also, I am asking whether we should be getting the paths for these or the instances themselves but for the moment simply got the paths.  

DISCUSSION OF FUTURE REQUIREMENTS: should we add function or option to get instances rather than paths???

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -234,6 +234,10 @@ Enhancements
   `pywbem.ConnectionError` in one case), and any other bad HTTP responses
   are now raised as a new exception `pywbem.HTTPError`.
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #438, opened 2016-08-10
Fixed '301 Redirect' error when 'make develop' accesses pypi.

andy-maier:
The `make develop` target invokes `python setup.py develop` which accesses Pypi in order to verify availability of dependent packages. This currently fails because the new Pypi warehoue at pypi.io moved to pypi.org and the old site returns a HTTP 301 "Redirect" status which the xmlrpc client used to access it does not handle.

The fix is to simply use the new site directly.

This fixes the Travis CI errors we currently have.


andy-maier:
+1

KSchopmeyer:
+1

PR #439, opened 2016-08-10
Fixes #433: Bad exception. Fix couple of pylints in example.

### LINKED ISSUE #433: Issue with getinterop_ns in _server on some servers
On some servers in test environment, the call to server.interop_ns returns:

WBEM server URL:  https://10.1.134.98
KeyError=%s "Key 'interop' not found"

No Further info at this point.

This only happens with certain servers in the smi lab and apparently only with python2. However, I am still gathering information.

Turns out it is difficult to actually get a detailed trace from the operations in the client.  See next issue

KSchopmeyer:
Ready for Review:
This fixes the known issue (incorrect exception catch.

It fixes a couple of pylint issues in one of the examples.

Note that the fact that the server in error is doing other strange things is outside scope of this fix.

### DIFF
--- a/examples/explore.py
+++ b/examples/explore.py
@@ -40,14 +40,14 @@ def explore_server(server_url, username, password):
     for inst in server.profiles:
         print_profile_info(org_vm, inst)
 
-    indication_profiles = server.get_selected_profiles( 'DMTF', 'Indications')
-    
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #441, opened 2016-08-10
Fixes #153: Completed test_client test cases for EnumerateInstances and EnumerateInstanceNames

### LINKED ISSUE #153: mock test for enumerate instances incomplete.
A lot of these test groups (ex. getInstance) are incomplete but thought I would document this one for now.  We can never get to 100% but we do need to add many more tests to each group.  I will try to create the working list for enumerate instances in this issue.

TODO indicate which of these already done.
The test for EnumerateInstance should include:

1. Standard good delivery with no, single, multiple objects
2. Return instances that have
      no properties
      all properties
3. Test with property list
4. test with all request parameters set both ways.
5. Test of the following errors
    TODO

andy-maier:
Ready for review and merge.

The tests cover all successful cases, including special handling of input parameters (namespace and classname) and output parameters (namespace set in returned path), and the documented server-originated error cases.

Any other error cases are not covered; they should be covered in separate test cases, because they are not so much specific to an operation.

### DIFF
--- a/testsuite/test_client/enumerateinstancenames.yaml
+++ b/testsuite/test_client/enumerateinstancenames.yaml
@@ -1,0 +1,949 @@
+-
+    name: EnumerateInstanceNames1
+    description: EnumerateInstanceNames succeeds returning 3 instance paths; test that returned instance paths get namespace in set.
+    pywbem_request:
+        url: http://acme.com:80
+        creds:
+            - username
+            - password
+        namespace: root/cimv2
+        timeout: 10
+        debug: false
+        operation:
+            pywbem_method: EnumerateInstanceNames
+            ClassName: PyWBEM_Person
+    pywbem_response:
+        result:
+            -
+                pywbem_object: CIMInstanceName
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

KSchopmeyer:
+1

Good step up.
I would not that this still does not test all options, specifically some of the request attributes such as localonly, etc.

andy-maier:
Karl: 

The variations on the input parms I think are not needed, because the code just passes them through (e.g. for LocalOnly).

What is needed is that the specified operation input value and its representation in the CIM-XML is verified. That is done in the test cases for all cases of True, False, None, and omitted. I admit that not all values are tested with all parameters, but for the parameters that just pass through, each of the values above is tested for some parameter.

KSchopmeyer:
+1 . Lets just note the issues and get this code committed.  Actually we probably need to do all of the parameters simply to assure that they get passed through correctly. 

We will always find tets don done

PR #448, opened 2016-08-14
Fixes #414, fixes #454: Separate subscriptions/listeners  into separate modules

### LINKED ISSUE #414: Separate _listener so we have separate components for listener and subscription management
There are real cases where a listener and the subscription creation tool are separate.  The subscription mechanism creates subscriptions but does NOT remain alive.  The listener is a separate entity with a separate life cycle.

Currently the pywbem listener strategy is that subscription lifecycle is within the lifecycle of the listener, created when the listener starts and removed when the listener terminates. 

While a user can always fall back to using the subscription create instance, etc. directly the question is whether our listener strategy should allow for async lifecycle of subscriptions independent of the lifecycle of the listener that creates/deletes the indications.  We have demonstrated that you can create an independent listener with the examples/listen.py code so that it is clear that we can actually make such a listener.

### LINKED ISSUE #454: Python 2.6 does not include NullHandler defined in listener and subscription manager
We include the NullHandler in _listener at least but it is not implemented in Python 2.6.  This apparently
can be fixed with:

if sys.version_info < (2, 7):
    class NullHandler(logging.Handler):
    """Implement logging NullHandler for python 2.6"""
        def emit(self, record):
            pass

or:
try:  # Python 2.7+
    from logging import NullHandler
except ImportError:
    class NullHandler(logging.Handler):
        def emit(self, record):
            pass

We only discovered this when we created a first test for indications that runs in travis.

KSchopmeyer:
Please Review
**COMMENT 18 aug: KS** WIP now because run_cimoperations.py needs update to match last changes. Only changes for subscription and listener code is a) commments, b) run_cimoperaitons test fail.

Breaks listener and subscription manager apart. Adds at least some support for static filters, subscriptions.

**Discussion:** Known issues:

1. I used two different patterns for the add functions and remove functions. In the add_filters i added an optional attribute to define if the filter is dynamic or static.   In the remove indications and filters I modified so we have two different functions remove_static_filter, and remove_dynamic_filter. That creates inconsistency.  **DISCUSION:** What is the most logical pattern for creates and removals now that we really will support static and dynamic filters. Thus, for example
```
     sub_mgr.add_filter(...dynamic_filter=false)
     sub_mgr.remove_static_filter(...)
```
2. When creating a static filter, it always creates a new destination. That is probably OK for the moment but not efficient if we are going to do may filters.

3. Right now I have the indication_id attribute in both the listener and subscription manager.  I need to sort out where it really should be.  Is this an id for the listener or for the subscription manager. 
**DISCUSSION:** Still thinking that one out and not sure if this is really a listener thing or subscription_manager. Note that we really only have one listener for subscription manager at this point.

4. The use of static is different here than in dsp docs.  Our definition is a filter, subscripiton that is not controlled iwhtin the life cycle of a subscription manager instance.  The subscripition manager may create and delete static subscriptions but there may also be real static subscriptions in an implementation which cannot ever be deleted.
**DISCUSSION:** Should we be using a different name

5. the code does not exercise any control over static subscriptions mixed with dynamic filters so far.  Should really add a test so that a subsubscription must be static if the filter is not in the dynamic_filter list, etc.

6. We are not marking static vs dynamic in the name component, etc. for filters and subscriptions. The difference is that dynamics are in the lists maintained in the subscription_manager. But there is no function such is is_dynamic_filter.  That is just a comment now.

Passes tests with the examples code and have not yet run the run_cimoperation.py test.  I have not created tests for manual filters, subscriptions.

This is still a work in progress for today but functionally complete.  I need to run tests and we may want to refactor a bit.  Note that I am inconsistent between create/remove, etc. on whether the static vs dynamic is a attribute of the call or separate function.


andy-maier:
Comments on the code:

* Line 364 in `_listener.py`: I did not test it, but I think this code cannot run because I don't think the self object has a `server` attribute. It should have a `listener` attribute, though.

* Line 699 in `_listener.py`:  The following four instance variables can be deleted:
    ```
    self._servers = {}  # WBEMServer objects for the WBEM servers
    self._subscription_paths = {}  # CIMInstanceName of subscriptions
    self._dynamic_filter_paths = {}  # CIMInstanceName of dynamic filters
    self._destination_path = {}  #
    ```

* In `_listener.py`, the following module level constants can be deleted:
    ```
    DEFAULT_QUERY_LANGUAGE
    SUBSCRIPTION_CLASSNAME
    DESTINATION_CLASSNAME
    FILTER_CLASSNAME
    SYSTEM_CREATION_CLASSNAME
    ```

* On the `listener_id`: In the `WBEMListener` class, it is not used anymore, and I think that is correct. So it can be removed from that class entirely.

* In `_subscription_manager.py`, the following module level constants can be removed:
   ```
DEFAULT_LISTENER_PORT_HTTP = 5988
DEFAULT_LISTENER_PORT_HTTPS = 5989
IMPLEMENTED_CIM_VERSION = '2.0'
IMPLEMENTED_DTD_VERSION = '2.4'
IMPLEMENTED_PROTOCOL_VERSION = '1.4'
   ```

andy-maier:
On your comments:

On 1: We should use the definition of static filters from DSP1054. In that definition, a client cannot create a static filter. That is the definition of a static filter: They exist in the server and cannot be managed by clients. Therefore, a parameter `dynamic_filter` on the filter creation method should not be there.

On 2nd 1: I don't see in the code where the `add_dynamic_filter()` would create a listener destination instance in the server. That instance is created by `add_server()` instead, and only once (if you add the server just once), for use by all subscriptions to all filters.

On 2: Is this about `listener_id` instead? If so, I think it should be on the `WBEMSubscriptionManager`, because it does not seem to be used by the `WBEMListener`.

On 3: The only deviation I see is that the current code thinks it can create static filters. Which by definition is not possible. Any filter the client creates will always be a dynamic filter. Given that, is there still an issue with different meaning of our code vs. DSP?

On 4: I believe this one goes away when we stick with the standard notion of dynamic vs static.

On 5: Not sure I understand. Is this still relevant if we stick with the standard notion of dynamic vs static?

KSchopmeyer:
There is another catagory, those that have beem created created by another manager.

Add method to close all servers.

Lets call this owned.

use listener url, not listener object

Change listener_id to subscription_manager_id

Allow multiple listeners.

move to add.listener as function where we have attributes of listener url and list of servers.

KSchopmeyer:
We had a discussion on the work and proposed the following changes:

1. there is no listener_id, only a subscription_manager_id
2. There should be only one function for add and remove for filter and subscription.  Should not distinguish between dynamic, etc.

3. We call filters and subscriptions over which the server instance within the subscription manager has lifecylce automatic control, owned filters and subscriptions. These are the ones that are in the owned_filters and owned_subscriptions list.

4

KSchopmeyer:
we proposed changing so that you add listeners to a server as a separate function and add lintener filter to add.subscriptions

add.server()
add_listener_dest
..

Also go back to using the complete listener url as a single entity throughout.

Finally, change add and remove subscription to and and remove subscriptions

KSchopmeyer:
We agreed to use the definition/implementation of url in WBEMConnection as the basis for the listener url in subscription manager.  Found several issues iwth that including:
1. While the uri documentation in wBEMConnection is good, the actual url processing is a function buried in cim_http and simply breadks out to host, port, ssl_flag.
2. That function breaks up the url but does nto really validate it.  Thus, for example, it tests for https but accepts anything for the schema if not https.
3. That function assumes that port is an option and substitutes the server ports if the prot is not provided.  That whole question of default for listener ports is an open issue.
Since the correct url is even more important for this than for operations since a bad url would only be found when the server tried to deliver indications making it difficult to diagnose we really do need a good validator for the url for indication destinations.

andy-maier:
For default ports of the listener, we should double check as to whether the default ports defined in DSP0200 apply to listeners as well. If so, they should be defaulted to, unless specified in the URL.

For URL validation, that is a general issue and not pywbem specific. We coukd have the submgr send,an invalid indication delivery request to the listener. That would verify the presence of the listener and the authentication (SSL/TLS handshake including possibly certificates). However, a server could theoretically see the listener under a different IP address, and in case of IPv6 zone IDs it would very likely fail.

Anyway, it might be a reasonable enough check to do. We could offer a validate() method on the submgr that takes the listener URL and does that.

KSchopmeyer:
updated for agreed approach to add_server, add_listener.  Simple test works and will clean up run pegasus tests thursday morning.  This does not validate listener urls.  Sadly I don't think the sendinvalid to listener really works out.  Don't have a good answer so best is just to assure as much as we can like a good url.  Will add url validate thursday.

KSchopmeyer:
**Discussion:** I left the logger in the subscription manager but since that is really a client tool I wonder if we should remove it.

andy-maier:
On fixing NullHandler: For py>=2.7, do `import NullHandler from logging` and when using it, use it as `NullHandler` not as `logging.NullHandler`.

On using logging in submgr: I also think it can be removed there, particularly because nothing is being logged right now.

andy-maier:
Comments on  `_subscription_manager.py`, mostly on descriptions:

1. Description of `subscription_manager_id` in `__init__()`:

    * On "to help the subscription manager identify its filters and": This suggests that the code uses the ID for idemtification purposes, but that is not the case. This is purely for the user. Suggest to change to something like: "to help the user identify its filters and". 
    * Along with the above, I suggest to state that there is no hard requirement that the ID be unique.
    * I don't think we need to show the detail of how the name property of a filter is made up.

2. On appending a colon to `subscription_manager_id` in `__init__()`: That leads to the fact that `__repr__()` would print it (when added, see comment 4) with the colon, which is a bit unfamiliar. It also causes the lower level functions `_create_destination()` and `_create_filter()` to have a  `subscription_manager_id` parameter that has the colon at the end, even though that is not documented. In fact, `_create_filter()` even documents that it must not contain a colon. Admittedly these are just internal functions, but I still think we should clean that up, by appending the colon only at the time the Name property value is put together in these lower level functions. That would also make it clearer should we ever use the ID for anything else but the Name property.

3. Logger: I think we can remove it from here (not used, anyway).

4. In `__repr__()`, the `_subscription_manager_id` could be added.

5. In `add_listener_destinations()`, I'd prefer if it returned after processing the single URLs, and the else path was outdented and `else:` removed. Also, you could have a variable in the former else path: `listener_url = listener_urls` along with the comment that we process a single listener URL, and use that, to make the code clearer.

6. On terminology for listener URLs in `add_listener_destinations()`:

   * The second paragraph says "for each listener_url in listener_urls". listener_url is a local variable from the code and should not be used in the description. Use e.g. "for each specified listener URL".
   * On "Each listener_url": Change to "Each listener URL".
   * Generally, for "url", use upper case "URL".

7. On format for listener URLs as described in `add_listener_destinations()`: The documented syntax rule leaves the scheme optional. I think that is how it should be. However, we then need to describe which scheme is the default (I suggest HTTP, for consistency with `WBEMConnection`, see `cim_http.parse_url()` secure by default), and the statement "This url must include scheme, address, and port components." further up is not correct, because both scheme and port are optional. Because we explain the format with a detailed syntax rule, I think that sentence further up can be removed.

8. In `add_server()`: the comment "#Create a new dictionary for this server" is confusing. It really creates new dictionary entries for this server.

9. In `remove_server()`, first paragraph of description: Change "from the listener" to "from the subscription manager", and change "that were created by this listener" to "that were created by this subscription manager". Lateron, same issue for comment "# Remove server from this listener".

10. In `add_filter()`, the second paragraph of the description still talks about the notion of removing non-dynamic filters. How about this replacement paragraph:

        Dynamic indication filters are those that are created by clients.
        Indication filters that pre-exist in the WBEM server are termed *static
        indication filters* and cannot be removed by clients. See
        :term:`DSP1054` for details about indication filters.

   plus, to motivate the second paragraph, change first sentence of description from "Add an indication filter" to "Add a dynamic indication filter".

10. In `add_filter()`, the description of the `owned` parameter should be moved one parameter down, to after `query_language`.

11. There are a few occurrences of spaces in Sphinx references that should be removed:

    ```
    pywbem/_subscription_manager.py:420:          owned (:class: `pybool`) Defines whether a filter whose lifecycle
    pywbem/_subscription_manager.py:434:        filter_id(:term: `string`)
    pywbem/_subscription_manager.py:823:      subscription_manager_id (:term: `string`))
    ```

    Also, the type of parameter `owned` should be (insert a colon):
    ```
    :class:`py:bool`
    ```

12. In `remove_filter()`, the second paragraph of the description is confusing because it seems to describe internals, and uses inconsistent terms to talk about the WBEM server in the two cases it has. I think we can remove that entire paragraph; we don't need to explain that we do a correct job in updating our control structures.

13. In `remove_filter()`, the logic in the code is partly confusing and partly incorrect.

    The bug is that in the else clause, there is no `path` variable defined, it should be `filter_path`.

    The confusing part is in the then clause:

    * First, the `server_id` is always in `self._owned_filter_paths`. If the server ID specified in the `server_id` parameter is not a valid server, the use of `_get_server()` already raises an exception.
    * Then, I would simply have only one unconditional instance deletion (first, in case it fails) and then the conditional cleanup of the owned paths.
    * Finally, I was wondering whether it could be possible that we have the same filter path more than once in `_owned_filter_paths`. If the WBEM server behaves correctly, it would return a different filter path if a create was done twice with the same input data. But who knows whether it behaves correctly, and then we may get back the same filter path for the new (or same) instance, and will store it in `_owned_filter_paths` because we don't check that. I suggest that during cleanup, we take care of that case, i.e. remove the `break` from the loop to continue removing that path. I'm not sure it is helpful to check for duplicate paths on creation, because what do you do when detecting it.

    How about:

    ```
        server = self._get_server(server_id)
        server.conn.DeleteInstance(path)
        paths = self._owned_filter_paths[server_id]
        for i, path in enumerate(paths):
            if path == filter_path:
                del paths[i]
                # we continue to the end to find possible duplicates
    ```

andy-maier:
One more comment on `_listener.py`:

* We should document where the `server` attribute comes from, on line 346:

```
            self.server.listener.deliver_indication(indication_inst,
                                                    self.client_address[0])
```

KSchopmeyer:
Is the filter path owned  when you create the subscription then the owned matches.

Could we do all this checking in the create subscription.
if not owned subscription, if we have filter in owned, reject.
if destination is owned, reject.

Add an optional explicit destination instance path to subscription so user has to create the dest if the subscription is not Owned.

KSchopmeyer:
New comments:,.

1. We have two functions independent of the subscriptionmanager class.  One of them is _create_destination which was originally iintended for internal use only.  With the manual subscription manipulation, this should probably be moved into the class.

KSchopmeyer:
Think I covered the comments, changes, etc.  The one thing not complete is all of the run_cimoperations.py tests. For some reason Pegasus jumped ship and I will waste an hour rebuilding it now. I wanted to get this to you for review and will fix any remaining tests tonight.

andy-maier:
I reviewed `_subscription_manager.py` and have the following remaining comments:

1. In `__repr__()`, the attribute `subscription_manager_id` in the returned string should have a leading underscore, consistent with the other attributes.

2. In `add_listener_destinations()`, the return value in case of passing a list of URLs is `None`. I think that should be changed to be a list of instance paths of the created instances. The docstring should be updated accordingly.

3. Item 16 from the previous set of comments is still open: `add_subscriptions()` creates the listener destination instances as well. I don't think it should do that. We discussed that briefly, and I think we landed on agreeing that it should not.

   Related to that: The logic we have now in `add_subscriptions()` is confusing to me:

   * I don't understand why we need both `listener_urls` and `destination_instance_path` as input parameters. I think just having `destination_paths` that can be a single instance path or a list of instance paths or `None` (defaulting to all owned listener destinations for that server), should be sufficient, if we agree that this method relies on previously created or pre-existing listener destination instances.

   * I don't think the listener URL item in the tuples in attribute `_server_destinations` would be used anymore if we change to just instance paths as input to `add_subscriptions()`. If so, it can be removed and the dict value becomes a list of instance paths. Also, the `_server_destinations` attribute contains only owned destinations, so a name change to `_owned_destination_paths` would be good, for consistency with the other attributes.

   * I don't understand the logic of the current code in `add_subscriptions()`. I think the logic should be something like:

        - If `destination_paths` is `None`, set it to the list of all owned destination paths (from the `_owned_destination_paths` attribute). Else if it is not a list, make it a list. The resulting list of destination paths is what the following logic uses.
        - If the subscription is not owned, enforce that all destinations (from the resulting list of destination paths) and the filter are not owned. Note that the current check is the other way around and is only for filters.
        - For each destination path in  the resulting list of destination paths, create the subscription instance (from that destination path and the specified filter path), and if owned, remember its instance path in the `_owned_subscription_paths` attribute.

4. Item 17 from the previous set of comments is still open: `_create_destination()` uses the specified listener URL unchanged and unvalidated. That means we leave it to the server to apply defaults.

   I think we should apply the defaults on the client side, and in the discussion we agreed on that. Also, we should validate the URL.

   Both of that can be done by invoking `parse_url()` and reconstructing the URL from the returned components:
   ```
listener_url = "{}://{}:{}".format(*parse_url(listener_url))
   ```

   I think that should be done in `_create_destination()`. The URL validation in `add_listener_destinations()` I think should be removed.

5. The `Name` property constructed in `_create_destination()` has issues:
 
   * The description shows the `subscription_manager_id` twice.
   * The code does not create the `subscription_manager_id` component in the `Name` value conditionally as suggested by the syntax definition. Also, it creates the string `None` is it is `None`. Both can be handled by special-casing the value `None` in the code.

6. The `Name` property constructed in `_create_filter()` has issues:
 
   * The special-casing of `subscription_manager_id is None` should be the other way around (i.e. if not None).
   * A nit: The code used to special-case `filter_id` being None, and `subscription_manager_id` being `None` is needlessly different and could be made consistent.
   * I'm not sure we need the type check on `filter_id` being a string. We don't have it for IDs in the other methods either.

KSchopmeyer:
Think I made the fixes.  However, I have only tested with cim_operations.py and make all.  Need to run the examples larer tonight.

I hope I caught everything.

andy-maier:
I commented on the last commit cdf508f in the previous comment, marked with **COMMENT(AM)**.

KSchopmeyer:
At least part of the lost indications problem is that += is NOT thread safe.  Think I cleaned up most comments and submitted a push request.  Note that what happens now is that it goes out to several thousand indications and then stalls.  No futher info so far.

andy-maier:
Remaining comments, as of commit 431be3e:

1. The use of `parse_url()` in `_create_destination()` defaults the port and scheme. I think we said we do not want to default them at this point.

2. My little confusion about `dest_paths.extend(dest_path)` in `add_listener_destinations()`, as discussed in the call. Address by renaming `dest_path`.

3. The double occurrence of `<subscription_manager_id>` and `<fsubscription_manager_id>` in the syntax string shown in the docstring(!) of `_create_destination()`.

4. In `add_subscriptions()`, can we swap the order of the `owned` and `destination_paths` parameters ? (both in the function header and in the docstring). it is somehow more logical to me to have the input paths before a modifier like owned.

5. In `add_subscriptions()` on line 617+618, the `if not owned:` and `if path in ...` statements could be combined into a single `if` stmt by `and`-ing the conditions. This would be leaner and consistent with how it is done for filters a few lines up.

andy-maier:
Remaining open issues as of commit 9dc6ee9, consolidating Karl's list and Andy's list:

1. Needs more tests in cim_operations.

2. Needs a multi server test but cannot do without multiple servers since it tries to send subscriptions to each server.

3. Needs multi-listener test and we can do that.

4. In cim_http.py line 277, in `parse_url()`, there is:

    ```
    if not allow_defaults and _scheme != 'http':
    ```

The comparison of `_scheme` in this if-stmt is needless, because `_scheme` is always `None` in this else-path. That comparison can be removed.

KSchopmeyer:
Done as of Tuesday evening:
1. Add tests for non-owned indications and misex owned and non-owned indications

2. Added a remove_destinations function.  This will only remove non-owned destination instances.  That function was missing.

Remaining tasks:

1. Final review
2. Add test of multiple listeners 
3. Add test to prove non-owned indications remain when subscription manager stopped. Same as tests we added today except stop the subscription manager and restart it and then confirm the non-owned subscriptions are still in place.

pywbem/_subscription_manager.py
```diff
@@ -15,11 +15,29 @@
...
+This `WBEM API`_ provides functionality for managing subscriptions for
+indications from one or more WBEM servers and one or more WBEM listeners.
+
+This class provides api for creating and removing indications for multiple
+WBEM servers and multiple indication listeners and for getting information
+about existing indication subscriptions.
+
+This class allows 2 types of indication subscriptions:
+
+    - Oned subscriptions and filters - These are subscriptions and filters
```
andy-maier: Typo: "Oned" -> "Owned".

pywbem/_subscription_manager.py
```diff
@@ -15,11 +15,29 @@
...
+
+    - client defined filters, destinations, and subscriptions that may be
+      created and deleted by the methods in this class but which are not
+      removed from the WBEM Server when the servers are removed.  It is
+      the user responsibility to eventually remove these subscriptions,
+      filters, and destination instances from the WBEM server.
 .. note::
 
-   At this point, the WBEM subscription managerAPI is experimental.
+   At this point, the WBEM subscription managerAPI is **experimental**.
```
andy-maier: Typo: Add blank in "managerAPI".

pywbem/_subscription_manager.py
```diff
@@ -172,101 +187,141 @@ def __repr__(self):
...
+            If no port is specified in the URL, the default ports are:
+
+            * If HTTPS is used, port 5989.
+            * If HTTP is used, port 5988.
+
+            See `WBEMConnection` for examples of valid urls.
+
+          owned (:class:`py:bool`)
+            Defines whether a destination whose lifecycle is owned by this
+            server or not is to be created. If True, a filter is
```
andy-maier: Change "a filter is created" to "a listener destination is created". One more occurrence of this.

pywbem/_subscription_manager.py
```diff
@@ -172,101 +187,141 @@ def __repr__(self):
...
+            created that is defined by the life cycle of this server instance
+            within this subscription manager.  If False a  filter is created
+            that it may outlive the lifecycle of the server instance
+            and must be removed manually by the user (for example using
+            the `remove_filter` function.
+
+        Returns:
+
+            list of :class:`~pywbem.CIMInstanceName`: The indication
```
andy-maier: Change "indication destinations" to "listener destination".

pywbem/_subscription_manager.py
```diff
@@ -172,101 +187,141 @@ def __repr__(self):
...
+            destination instance paths for the defined listener_urls.
+            `Owned` destinations are recorded in the subscription manager for
+            the target 1server_id` so is not needed by the user. However
```
andy-maier: * I don't think we need to mention that it is recorded by server.
* Incorrect markup "1server_id`".
* "so is not needed by the user" is incomplete.

pywbem/_subscription_manager.py
```diff
@@ -172,101 +187,141 @@ def __repr__(self):
...
+            if a destinations are not owned, it is the users responsibility
```
andy-maier: grammar: "a destinations are not owned"

andy-maier:
Remaining comments, as of commit 1f0ca83:

1. See the comments on diffs since that commit.

2. In `_listener.py`, the module description says: "See _subscription_manager.py and the examples directory for an example of a subscription_manager and listener defined in the same executable."

   I don't see an example of a subscription manager and listener defined in the same executable, nor does the `_subscription_manager.py` module contain that. Unless there is such an example, I suggest to remove the quoted sentence.

3. In `_listener.py`, the example in the module description misses the "::", which is why it formats as text.

4. In the description of the `WBEMListener` class, the following sentence is no longer true and can be removed: "The listener also supports the management of subscriptions ...". 

5. In the description of the `WBEMSubscriptionManager` class, the following sentence can be removed: "This `WBEM API`_  ...", because the sentence following it basically says the same.

6. In the description of the `WBEMSubscriptionManager` class, the following sentence: "This class provides api for creating and removing indications for multiple WBEM servers and multiple indication listeners and for getting information about existing indication subscriptions." has some issues:

   a. grammar: change "provides api for" to "provides for".
   b. change "creating and removing indications" to "creating and removing subscriptions".

7. In the description of the `WBEMSubscriptionManager` class, the following wording of the second list item: "client defined ..." is not distinguishing it clearly enough from the first item, which is also about client-defined objects. How about: "Unowned ...".

8. In the description of the `WBEMSubscriptionManager` class, the first item "Oned subscriptions and filters ..." misses to mention listener destinations, which can also be owned.

8. In the description of the `WBEMSubscriptionManager` class, the first item references the add_listener_url function with improper syntax. Use:

        (see :meth:`~pywbem.add_listener_url`)

9. Typo and grammar errors in the example section of the class description of `WBEMSubscriptionManager`: Change "listeener to subscribes" to "listener to subscribe".

10. In the description of the `subscription_manager_id` parameter of `WBEMSubscriptionManager`, the syntax description must be more indented (probably by 8 positions).

pywbem/_subscription_manager.py
```diff
@@ -278,46 +333,45 @@ def add_server(self, server, listener_urls=None):
...
 
     def remove_server(self, server_id):
         """
-        Remove a WBEM server from the listener and unregister the listener
-        from the server by deleting all indication subscriptions, dynamic
-        indication filters and listener destinations in the server that were
-        created by this listener.
+        Remove a WBEM server from the subscription manager and unregister the
+        listeners from the server by deleting all owned indication
+        subscriptions, owned indication filters and ownedlistener destinations
```
andy-maier: Add blank in "ownedlistener"

pywbem/_subscription_manager.py
```diff
@@ -385,21 +432,22 @@ def add_filter(self, server_id, source_namespace, query,
...
 
-        filter_id(:term: `string`)
+          owned (:class:`py:bool`)
+            Defines whether a filter whose lifecycle is owned by this server
+            or not is to be created. If True, a filter is
+            created that is defined by the life cycle of this server instance
+            within this subscription manager.  If False a  filter is created
+            that it may outlive the lifecycle of the server instance
+            and must be removed manually by the user (for example using
+            the `remove_filter` function.
```
andy-maier: Incorrect markup for reference to `remove_filter()`. Change to:

    :meth:`~pywbem.WBEMSubscriptionManager.remove_filter`

KSchopmeyer:
Added tests.  Note that the multiple listener test proved to be a problem so I propose we make that a separate issue.  Fixed doc comments from Andy and some others I found.  reviewed documentation and code. Ran tests.

I think this is clean enough to commit

+1

**NOTE:** The run_cimoperations.py indication tests are proving to be very fragile because 

a. we do not completely clean up the server between tests and any test error stops that particular test without doing the stop_listener() (leaves indications in the server and the port open), Thus, I ignore the problem of not delivering the correct number since it does happen at random and messes up the whole test as well as forcing those subscriptions to be cleaned out manually.

b. The are indications lost sometimes.  Therefore, I had to bypass the indication count test for now. We just output a message if the counts do not match.

I must rewrite this set of tests at some point so we clean up the server between tests and force the listener stop (ex. try: finally).  However, we appear to now have all of the functions covered and are removing all subscriptions, etc.

I did accidentally find a way to deliver lots of indications when I accidentally left the old subscriptions in the server.  At one point we were delivering 400 indications when we asked for 20 because of existing subscriptions.  strange in that they were delivered VERY rapidly.

KSchopmeyer:
+1

andy-maier:
+1

PR #450, opened 2016-08-16
Fixes #203: Document limitations of flavors

### LINKED ISSUE #203: MOF parsing: Qualifier declaration does not apply flavor defaults
In the standard qualifiers.mof (CIM Schema 2.45), the EmbeddedInstance qualifier is one example of a qualifier that is defined without explicit flavors, relying on the defaults:

    Qualifier EmbeddedInstance : string = null, 
        Scope(property, method, parameter);

When the MOF compiler parses this qualifier declaration, the resulting CIMQualifierDeclaration object contains `None` for the flavors that have not been specified. While that is formally correct according to how the interface is defined (`None`means "information not available"), it is not very practical.

**Proposed solution:** Set the flavors to the DSP0004/DSP0201 defined defaults, if not specified in the qualifier declaration MOF (e.g. in `_build_flavors()` in `mof_compiler.py`).

**DIscussion:** Proposed solution needs discussion.

Considerations:

* Should the defaults be applied at this level, vs. the object just shows exactly what was specified, leaving application of defaults to some higher level?

KSchopmeyer:
Ready for review and merge.
Initial commit had an accidentally folded comment causing build failures.  Update committed.

Documents the issues with flavors and the defaults required. Also adds
todo to consider removal of toinstance in future

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -640,7 +640,7 @@ def mofstr(strvalue, indent=MOF_INDENT, maxline=MAX_MOF_LINE):
     This function handles MOF escaping and breaking the string into multiple
     lines according to the `maxline` and `indent` parameters.
 
-    DSP0004 defines that the character repertoire for MOF string constants
+    `DSP0004` defines that the character repertoire for MOF string constants
     is the entire repertoire for the CIM string datatype. That is, the entire
     Unicode character repertoire except for U+0000.
 
@@ -660,7 +660,7 @@ def mofstr(strvalue, indent=MOF_INDENT, maxline=MAX_MOF_LINE):
     U+0001..U+001F, it uses generic MOF escape sequences (e.g. U+0001 becomes
     "\x0001")
 
-    The following table shows the MOF escape sequences defined in DSP0004:
+    The following table shows the MOF escape sequences defined in `DSP0004`:
 
     ============  =============================================================
     MOF escape    Character
@@ -3277,6 +3277,35 @@ class CIMQualifier(_CIMComparisonMixin):
     specifies information such as a documentation string or whether a property
     is a key.
 
+    :class:`CIMQualifier` objects can be used to represent the qualifier values
+    that are specified on a CIM element (e.g. on a CIM class). In that case, the
+    :attr:`propagated` property is always `False`, and the effective values of
+    applicable but unspecified qualifiers need to be determined by users, by
+    considering the default value of the corresponding qualifier type, the
+    propagation and override flavors of the qualifier, and the qualifier values
+    that have been specified in the class ancestry of the CIM element in
+    question.
+
+    :class:`CIMQualifier` objects can also be used to represent the effective
+    values of all applicable qualifiers on a CIM element, including those that
+    have not been specified, e.g. in the MOF declaration of the CIM element.
+    In this case, the :class:`CIMQualifier` objects for qualifier values that
+    are specified in MOF represent the specified values, and their
+    :attr:`propagated` property is `False`. The :class:`CIMQualifier` objects
+    for qualifier values that are not specified in MOF represent the effective
+    values, and their :attr:`propagated` property is `True`.
+
+    Whether a set of :class:`CIMQualifier` objects on a CIM object represents
+    just the specified qualifiers or all applicable qualifiers needs to be known
+    from the context.
+
+    :class:`CIMQualifier` has properties that represent qualifier flavors
+    (:attr:`tosubclass`, :attr:`toinstance`, :attr:`overridable`, and
+    :attr:`translatable`). If any of these flavor properties is not `None`, the
+    qualifier value represented by the :class:`CIMQualifier` object implicitly
+    defines a qualifier type. Implicitly defined qualifier types have been
+    deprecated in :term:`DSP0004`.
+
     Attributes:
 
       name (:term:`unicode string`):
@@ -3305,12 +3334,21 @@ class CIMQualifier(_CIMComparisonMixin):
         `None` means that this information is not available.
 
       tosubclass (:class:`py:bool`):
-        Indicates whether the qualifier value propagates to subclasses.
+        If not `None`, defines that there is an implicit qualifier type for
+        this qualifier.
+
+        If `True`, specifies the ToSubclass flavor (the qualifier value
+        propagates to subclasses); if `False` specifies the Restricted flavor
+        (the qualifier values does not propagate to subclasses).
 
         `None` means that this information is not available.
 
       toinstance (:class:`py:bool`):
-        Indicates whether the qualifier value propagates to instances.
+        If not None, defines that there is an implicit qualifier type for
+        this qualifier.
+
+        If `True` specifies the ToInstance flavor(the qualifier value
+        propagates to instances.
 
         `None` means that this information is not available.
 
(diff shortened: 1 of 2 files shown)


pywbem/cim_obj.py
```diff
@@ -3659,28 +3675,32 @@ class CIMQualifierDeclaration(_CIMComparisonMixin):
         `None` means that this information is not available.
 
       toinstance (:class:`py:bool`):
-        If not `None`, specifies whether the qualifier value propagates
-        to instances.
+        If not `None`, specifies the flavor that defines whether the qualifier
+        value propagates to instances.
```
andy-maier: I don't like the new wording: "specifies the flavor that defines whether" is an indirection that does not help me to understand it.

We may need some change to better express that it only represents what the qualifier declaration specified. However, that would be a change on the sentence describing None, e.g.:

`None` means that the qualifier declaration does not specify this information.

And that change would need to be done on all flavor parameters.

andy-maier:
-1, see comments in the code.

KSchopmeyer:
Added more documentation on flavors.

pywbem/cim_obj.py
```diff
@@ -3275,6 +3275,19 @@ class CIMQualifier(_CIMComparisonMixin):
...
 
+    The flavors defined in this class may reflect only flavor declaractions and
+    not the effective flavors (default value for flavors that are defined as
+    `None`) of qualifiers in a class hiearchy.
+
+    Use of qualifier declarations works for passing qualifiers to servers via a
+    protocol because the protocols allow missing flavors and are expected to
+    set the defaults on reception.
+
+    If a CIMQualifier is to be used in definition of a class hiearchy such as
```
andy-maier: "CIMQualifier" -> "CIMQualifier object"

pywbem/cim_obj.py
```diff
@@ -3653,33 +3701,43 @@ class CIMQualifierDeclaration(_CIMComparisonMixin):
         This variable will never be `None`.
 
       tosubclass (:class:`py:bool`):
-        If not `None`, specifies whether the qualifier value propagates
-        to subclasses.
+        If not `None`, specifies the tosubclass flavor. This flavor defines
```
andy-maier: Use mixed case for any flavor names, e.g. "ToSubclass" in this case.

This comment applies to all usages of flavor names in docstrings in this module.

pywbem/cim_obj.py
```diff
@@ -3653,33 +3701,43 @@ class CIMQualifierDeclaration(_CIMComparisonMixin):
...
+        whether the qualifier value propagates to subclasses. In MOF this
+        flavor is controlled by the `tosubclass` and `restricted` flavor
+        keywords.
```
andy-maier: This attribute specifies not only the ToSubclass flavor, but in addition the Restricted flavor (by means of a value of `False`). I think we should describe that with explicit text for the values `True`, `False`, and `None`.

pywbem/cim_obj.py
```diff
@@ -3653,33 +3701,43 @@ class CIMQualifierDeclaration(_CIMComparisonMixin):
...
         Note that :term:`DSP0200` has deprecated the presence of qualifier
         values on CIM instances.
 
       overridable (:class:`py:bool`):
-        If not `None`, specifies whether the qualifier value is overridable
-        in subclasses.
+        If not `None`, specifies the  overridable flavor. This flavor defines
+        whether the qualifier value is overridable in subclasses. Internally
+        the flavor is named overridable; setting of this flavor in MOF is defined by
+        the `enableoverride` and `disableoverride` keywords.
```
andy-maier: "Overrideable" is not a flavor as defined in DSP0004. We settled on using the MOF keywords also for the flavors in the metamodel, i.e. this attribute specifies the "EnableOverride" and "DisableOverride" flavors. I think we should describe that with explicit text for the values `True`, `False`, and `None`.

pywbem/cim_obj.py
andy-maier: We should say that for non-None values, this attribute specifies the flavor of an implicitly defined qualifier type, and that implicitly defined qualifier types are deprecated in DSP0004. This comment applies to all flavor attributes of this class.

pywbem/cim_obj.py
```diff
@@ -3723,28 +3781,30 @@ def __init__(self, name, type, value=None, is_array=False,
             attribute.
 
           overridable (:class:`py:bool`):
-            If not `None`, specifies whether the qualifier value is overridable
-            in subclasses.
+            If not `None`, defines the flavor that defines whether the
+            qualifier value is overridable in subclasses.
```
andy-maier: For all flavor parameters, see my comments above, on the corresponding attributes.

andy-maier:
see my comments in the changes.

andy-maier:
Remaining comments, as of commit 0c837f6:

1. In the last paragraph of the description of `CIMQualifier`, I think it would be helpful to clarify that the implicitly defined qualifier type is conceptual and is not materialized, by adding:

        The implicitly defined qualifier type is conceptual and is not materialized as a
        :class:`~pywbem.CIMQualifierType` object.

2. In the description of the `toinstance` variable of `CIMQualifier`, there is a statement missing about the meaning of `False`. I am aware that the difficulty to express this is the fact that there is no flavor such as "not-to-instance".

   How about adding:

        If `False`, specifies that qualifier values do not propagate to instances. There is no
        flavor corresponding to `toinstance=False`.

3. The descriptions of the last four instance variables of `CIMQualifier` say: "If not None, defines that there is an implicit qualifier type for this qualifier." I think it would be good to also state what the flavor of that qualifier type is.

   How about, instead of the quoted sentence:

        If not `None`, causes an implicit qualifier type to be defined for this qualifier that
        has the specified flavor.

4. In the description of the `toinstance` variable of `CIMQualifierDeclaration`, there is a statement missing about the meaning of `False`. How about adding:

        If `False`, specifies that qualifier values do not propagate to instances. There is no
        flavor corresponding to `toinstance=False`.

5. In the description of the `translatable` variable of `CIMQualifierDeclaration`, there is a statement missing about the meaning of `False`. How about using the correspondng text from `CIMQualifier`:

        If `False`, specifies that the qualfier is not translatable. There is no flavor corresponding
        to `translatable=False`.

6. Typo "tdefault" in the description of `CIMQualifierDeclaration`:

        This differs from the DMTF specification DSP0004 where tdefault values are defined as follows:

7. The list of default values in the description of `CIMQualifierDeclaration`:

      * EnableOverride flavor, overridable attribute = True
      * ToSubClass flavor, tosubclass attribute = True
      * Translatable flavor,translatable attribute = False
      * ToInstance flavor, toinstance flavor = False. Not defined in DSP0004 and as
        deprecated in the DMTF protocol specification DSP0200

   has some issues:

   a. The flavors are not true or false, the qualifier type has the flavor or does not have the flavor.
   b. In ToInstance, the second "flavor" should be "attribute".

   How about:

      * Has the EnableOverride flavor; `overridable=True`
      * Has the ToSubClass flavor; `tosubclass=True`
      * Does not have the Translatable flavor; `translatable=False`
      * Does not have the ToInstance flavor; `toinstance=False`. Not defined in DSP0004 and as
        deprecated in the DMTF protocol specification DSP0200

KSchopmeyer:
+1

andy-maier:
+1

PR #451, opened 2016-08-17
Add WBEMSubscriptionManager to listener.rst to include it in the docs

andy-maier:
This small doc fix adds on top of #448.

The change is only in commit 484de31.

Ready for review and merge.


andy-maier:
+1

KSchopmeyer:
+1

PR #467, opened 2016-09-02
Cleaned up change log.

andy-maier:
Some items have been merged (e.g. items for indication support).
Cleaned up formatting.
Moved a few items between Enhancements and Bug fixes.

PR #468, opened 2016-09-02
Removed `value` from str() and repr() of CIMParameter

andy-maier:
We deprecated the use of `CIMParameter.value`, and so we should not show it in `__str__()` and `__repr__()`. Besides that, accessing it in these functions would cause a `DeprecationWarning` to be issued.

PR #467 already removed the item in the change log that was stating that `value` was added to `CIMParameter.__str__()`.


KSchopmeyer:
+1

andy-maier:
+1.
Rebased to make it independent of the change log update.

PR #469, opened 2016-09-02
Improvements in documentation of WBEM listener API.

andy-maier:
Ready for review.

Besides the renaming of a local variable, this contains only documentation changes in the "WBEM listener API" chapter. It is best to review that when looking at the HTML output, instead of looking at the diff.
  
Details (of all commits of this PR):

* commit c42ce7c:

  Reordered methods in WBEMSubscriptionManager to be more logical.

* commit 191d21d

  Further updates to subscription manager documentation.

  Details:

    - Clarified when instances to be removed may/must be owned/not-owned.
    - better described the two types of subscriptions (owned / not-owned), and how tha trelates to dynamic/static filters.
    - Various wording changes for consistency and clarity.

* commit cab82fe

  Improvements in documentation of WBEM listener API.
    
  Details:

    - Added _subscription_manager.py to doc dependent files in makefile.
    - Moved general description of WBEM listener API from WBEMListener class into the WBEM listener API section (i.e. one level up).
    - Improved wording of several class and method descriptions in WBEMSubscriptionManager.
    - Fixed links and typos.


KSchopmeyer:
1. There are some trailing white spaces that git caught in pylint.

2. I am uncomfortable with the WBEMListener API for the title to section 5.  It should probably be something like WBEM Indication Support or WBEM indication/subscription API.  It is definitly much more than the listener.

3. Why to the examples (section 5.1.1) go first in the documentation? Not really important be we are getting the examples before we get the API definitions.

4. This is a general comment for the future.  Should we be showing the __repr__ as part of the interface API?  Lets not change anything now however.

5. in start() the ref to the stop() method does not generate a link. line 770

6. If we are describing APIs, should not the deliver_indication() be considered an internal method?

7. My BooBoo.  doc for function callback_interface, I left the Raises: as a TODO.  Not sure if this raises any exceptions.

8. Line 56   Confusing text. The following is hard to read:

```
This restriction is enforced by the :class:`~pywbem.WBEMSubscriptionManager`
class, and it is motivated by the fact that an indication subscription
depends on the indication filter and on the listener destination it relates.
The automatic removal of an owned filter or an owned listener destination
would therefore not be possible if a subscription relating them was not-owned.
```
 I am trying to work out a better description and will provide it in the morning.

9. line 248  lateron  either make two words or just leave it out since the work prequisite is sufficient.

10. line 383.  I missed removing this when we dropped the optional port definition.  That means that the port component of line 367 needs to be required also.

11. line 521 We are using the word dynamic before we define it.  The definition is in line 525 but should probably be moved to general description part of file.

12. Line 599 "Of course, the user may still delete these filter instances explicitly (for example by using remove_filter())."

13. Line 672 "The indication filter must be a dynamic filter (static filters
        cannot be removed by clients) but may be owned or not-owned."
  
    Again issue of dynamic. Since we use it in several places lets put sentence in beginning defining it.
    Actually the definition is on line 525, after the first use.

I want to read through once more in the morning since my thinking mechanism degrades by the end of the day


    Remove the leading phrase "Of course", suberfulous

andy-maier:
Karl, I have addressed your comments in commit eb8431e. It also contains a couple code changes, see the details of the commit message. Please review.

There are some remaining **DISCUSSION POINTS**.

andy-maier:
Done for now, waiting for review.

KSchopmeyer:
Here is my proposal for item 8:

Owned and not-owned subscriptions, filters, and listener destinations can be
arbitrarily mixed, with one exception:

a not-owned subscription cannot be created including either an owned filter
or owned destination because the automatic removal of the owned filter or
destination by the subscription manager would break the subscription without
removing it.

This restriction is enforced by the :class:`~pywbem.WBEMSubscriptionManager`
class.

===========
DISCUSSION: Note that we allow explicit removal of owned flters and owned subscriptions, but not of owned destinations. Therefore, we only have a remove_destinations() that rejects removal of owned destinations. The reason stated in its description was copied from another place in the previous version, and I don't think it is good enough of a reason.

'''
        The listener destination instances must be not-owned. Owned listener
        destination instances are not allowed to be removed using this method,
        because the subscription manager will automatically remove them when
        the WBEM server is deregistered from it.
'''

Since owned is just an artifact of the existence of the destination in the owned_destination_paths list, if for some reason the sub mgr dies, they will be considered not-owned in the next restart of the sub mgr.

andy-maier:
On not allowing removal of owned destinations: I understand what you say, but I don't see how that is a reason not to allow removal of owned destinations.

andy-maier:
On your proposal for item 8: I don't think this reasoning reflects what would happen. The automatic removal of the owned filter or destination by the subscription manager would not break the subscription, because the WBEM server would (hopefully) reject the DeleteInstance on either of them as long as there is still a subscription referencing them. So I changed the reasoning as follows:

```
* A not-owned subscription cannot be created with an owned filter and/or an
  owned listener destination because that would prevent the automatic life
  cycle management of the owned filter or listener destination by the
  subscription manager. This restriction is enforced by the
  :class:`~pywbem.WBEMSubscriptionManager` class.
```

andy-maier:
I'll be gone for the day...

andy-maier:
I think item 8 is done, but item 12 has still a **discussion point**.

KSchopmeyer:
In thinking through your #8, I agree that we should allow removing them.  I did bring up one more point however,  should we do a check to see if the destination is linked to any subscription before removing it and should that apply to filters also.  That would prevent breaking subscriptions.   At this point I even if we agree to the additional test, I would propose it for 0.10.0, assuming that 0.10.0 is not months away.  The down side of that check would be that each removal would require reading all subscriptions and we have no idea how many there are so it would be high overhead.

andy-maier:
On tge check before removal: Do you think it is common that a server would delete an instance referenced by an association, leaving that association with a broken reference? IMO, that is not allowed as per DSP0004. It may still be common. If so, a check may make sense, but tgat would be the case for both owned and not-owned.

andy-maier:
Commit 284c2e2 allows the explicit removal of owned listener destinations.

That leaves only the discussion point about checking for broken references in subscriptions.

KSchopmeyer:
Yes. +1

PR #473, opened 2016-09-03
Fixes #472: Documentation and help editorial cleanup for run_operationtimeouts.py

### LINKED ISSUE #472: test program run_operationtimeouts.py
The help is not well formatted and I had commented out the forced password code.

This only impacts the execution of this test and it is a test that can only be run on the new unreleased open pegasus because it adds a special method.

KSchopmeyer:
+1

This is editorial cleanup to a test program that only runs against OpenPegasus and only I run.  There should be little risk to doing this as part of 0.9.0.  I listed the issue as 0.9.0 optional but might as well get this in.  Fixed the help where some EOLs were missing, a variable name, and added back the else component of getting name password on cmd line.

### DIFF
--- a/testsuite/run_operationtimeouts.py
+++ b/testsuite/run_operationtimeouts.py
@@ -21,6 +21,7 @@ import sys
 import unittest
 import warnings
 import datetime
+from getpass import getpass
 
 import six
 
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

PR #475, opened 2016-09-04
Removed Windows installable.

andy-maier:
Note that support for Windows is still claimed. I added a disclaimer that testing happens mostly on Linux. This now also covers the previous disclaimer that we did not try out OS-X, so I removed that as well.

Please review.


andy-maier:
I tested on Windows 7 (with CygWin) and "make test" on Python 2.7 passed.

andy-maier:
I tested building the dist archives on Windows 7 (with CygWin, and in the CygWin bash shell), and it worked fine. I compared the content of the source and wheel archives, and they were correct, compared to Linux. Differences were only version numbers of supporting packages, and order of dict entries.

As a result, I removed the restriction to not build the dist archives on Windows.

KSchopmeyer:
+1

PR #476, opened 2016-09-04
[WIP] Release v0.9.0

andy-maier:
This is the decision to release pywbem v0.9.0.

The actual release date should match what is stated in `docs/changes.rst`.

Please vote +1 only if you are sure. All tests should be done with all other PRs targeting 0.9.0 merged, so we test the latest level.

TODOs:

* Karl: Run tests against a real WBEM server
* The actual release (continue in DEVELOP.md at step 16) 

DONE:

* Discussion point about having the dist archives in the repo. We agreed to no longer have them, and this PR has been updated to remove them.
* Discussion point about explicit removal of owned destinations (and possibly abut removal of owned filters and subscriptions, too). See PR #469, we agreed to allow the removal, and issue #483 has been created for the check for referencing subscriptions.
* Merge PR #469.

Note: Andy's tests in multi-Linux CI will need to be skipped this time because home environment is not yet operational.


KSchopmeyer:
I have  run the cim_operations with a repeat command several hundred times today.  That is why issue #479.  At this point I am getting indication count errors and a couple of time an error in getting the profiles.  Also ran run_cimtimetouts.py several times OK

finally once, I got an error in one of the pull tests (the one where we get the objects one by one. It stated that there was a discrepancy of 2 paths in the count of about 600 objects received.  All other times it returned a good response.

You get to decide if you want to commit the code from issue #479 (pr is 0..9.0 optional) but I am not sure how much more repeat testing I can do without extending tests more.

I think that within the limits of the testing and known issues we are ready for release.  I  am prepared to release at this point.

Suggest that we make the final decision on telecon tuesday morning.

andy-maier:
We missed to state two known issues in the change log, that were stated for v0.8.1:

```
Known Issues
^^^^^^^^^^^^

* Installing PyWBEM on Python 2.6 has a conflict with the `pbr` package
  from PyPI, resulting in a TypeError: "dist must be a Distribution
  instance". This issue is specific to Python 2.6 and does not occur in
  any of the other supported Python versions (2.7, 3.4, 3.5). This issue
  can be mitigated by uninstalling the `pbr` package, or if that is not
  possible, by migrating to Python 2.7. See issue #26 on GitHub.

* MOF using names that are reserved keywords will fail to compile in the
  MOF compiler. For example, a CIM property named 'indication'.
  See issue #62 on GitHub.
```

andy-maier:
Discussion point:

Should we have the distribution archives in the Git repo?

I tend to think we should not have them. For example, the Pypi upload creates the distribution archives at the time of upload. We need to manually make sure that we have freshly built the distribution archives in the repo. This is a process complexity we could get rid of.

KSchopmeyer:
My tendency is to say we should not have distribution archives in Git repo but that is probably because I just never used them.  Any work or extra sync we do not have to do just makes life simpler and it is already very complex as it is.

andy-maier:
I like that. Then let's remove them starting with 0.9.0.

andy-maier:
In commit 8a5ec1d, removed the dist archives from the repo, plus all other changes needed for that. See the commit message for details.

andy-maier:
We agreed we are ready for release. We'll merge this PR and then do a final test.

KSchopmeyer:
+1

PR #480, opened 2016-09-05
Fixes #479: Clean up run_cimoperations.py

### LINKED ISSUE #479: Clean up run_CimOperations.py
Found some issues in this test program.  Would like to clean up but since I am the only one using it, I would think that this is 0.9.0 optional.

The issues include:
1. Poor names on some variables.
2. Incomplete tests for equality of instances between pull nad not-pull operations.  Actually only the
   counts were being compared, not the instances returned..
3. Couple of code errors in the pull tests leading to incomplete pull tests.  Partly this is because the pull tests are not really determininstic since the server can return a variable number of responses to any single request.
   Specifically this fixes the issue where we were getting different counts of responses when pull sequence compared to an enumerateInstances return.

KSchopmeyer:
Found some code errors in the pull tests and also cleaned up to code and
added another test for instance equality.

Fixed those issues.  Note that since I am the only user and we are down to the end, I made this one 0.9.0 optional

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -210,11 +210,12 @@ class ClientTest(unittest.TestCase):
     def inst_in_list(self, inst, list_):
         """ Determine if an instance is in a list of instances. Return
             True if the instance is in the list. Otherwise return False.
+            Tests on path only.
+
         """
         for i in list_:
             if i.path == inst.path:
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

andy-maier:
I restarted the two Travis runs that failed (they failed with a connection error to Pypi, which is quite common these days).

PR #482, opened 2016-09-05
Fixes #481: Changes to examples for indications

### LINKED ISSUE #481: Issues in the  indications examples code
1. the listen.py has some minor issues including case of the variable LISTENER, not thread save, print of indication is disabled and documentation says it is enabled.  Also added a status function that shows count and more to the startup help. Finally if --help entered on cmd line, it returns usage.

2. the send_indications.py function does not handle https indications. It is a parallel to the send_indication.sh but we never enabled the https component.

KSchopmeyer:
1. Correct some minor issues in listen.py including a)lowercase the
object name since it is easier to type, b) add status function c) enable
output of indication with logging function. d) add thread protection to
the processing function

2. Extend the send_indications.py example to allow https indications.
However, this is currently set to force verify=false so that the client
does not verify the server cert because the requests code does not like
our current cert (it is unhappy about a name element). Will have to recreate the
tmp.pem to pass requests ssl

Again marked this as 0.9.0 optional.  Certainly not required since these are examples and
do not show up in the pip.

### DIFF
--- a/examples/listen.py
+++ b/examples/listen.py
@@ -1,11 +1,16 @@
 #!/usr/bin/env python
 
+''' Test listener. When started this code creates a WBEMListener on the
+    address/ports defined by the input parameters.  This listener
+    displays any indications received as log entries.
+'''
+
 import sys as _sys
 import os as _os
 import code as _code
-import six as _six
 import errno as _errno
 import logging as _logging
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

PR #491, opened 2016-09-06
Fixes #470: Make usage of Jupyter tutorials more obvious

### LINKED ISSUE #470: Improve usage/documentation of ipython notebook
Karl noted that he found use of the ipython notebook difficult. We agreed that we should work to make it clearer how to incorporate notebooks, etc. so that using the notebooks becomes simple. Otherwise only those who already use notebooks will use them.

We noted that we should be able to use notebooks for executable examples, demos, some presentations, etc but we had to be able to easily integrate pieces into notebooks and create both local and remote notebooks to make this work.

andy-maier:
Please review.

Best way to review this is building the documentation using `make builddoc` and looking at the Tutorials page of the documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -472,7 +472,7 @@ intersphinx_cache_limit = 5
 # results in the link caption "this issue".
 
 extlinks = {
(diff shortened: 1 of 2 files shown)


andy-maier:
That's strange. Let's find out more about this before we merge this one.

PR #499, opened 2016-09-09
Fixes #466: Implements the enumerate generator methods

### LINKED ISSUE #466: Add Iter... methods to WBEMConnection to hide open... vs enumerate operations
We will want to extend the enumerates so that the user does not have to make decisions about whether to use pull or not.

If we do it at the WBEMConnection level, the first enumerate request could test for existence of pull operations and set a flag that would last for the life of the WBEMConnection.  I think that the it is logical to do this at the WBEM.

This would also require that the user be able to set some configuration information:
1. interoperation timeout
2. Max objects on each pull

These could be added to the WBEM connection optional variables with a default.

new new operations would probably be based on a generator so that the user would call the operation and get a generator returned.

KSchopmeyer:
**STATUS 11 Dec:** for review: All functions implemented along with a _use... parameter for each.  Removed the method to set use_pull_operations.  Some initial code for these operations is now in run_cim operations but have more work there.  I have not yet sorted out how to test this through mock yet. First goal is complete test with run_cimoperations.py I have at least a day of adding tests just in run_cim_operations.py

Added almost all basic tests (except invoke method) to run_cimoperations.py so we are exercising at least the main paths. Working now to try to get them into mock but still trying to sort out how to handle the generate response(probably use it to drive a for loop)

**Discussion:** The IterQueryInstances works but want to discuss the fact that what I did is a hack. Using generators has a number of limitations, specifically it is not easy to create results that return the generator type and some other entity.  I created a class that is returned and has two properties, the query_result_class (on None) and the generator object.

However, making the chain so that I can feed this return class from the pull responses proved difficutltso for now I get all and then return them IterQueryInstancesReturn. The alternative is another class wrapping the whole implementation of IterQueryInstances so that we could wrap multiple levels of generators and I ran out of time.

**DISCUSSION:** restructure with iter function in separate class: See last comment for my conclusion after implementing. **CONCLUSION:** Agreed not to do this

There is also a sample program in examples right now.  The example does execute but is incomplete.

Known Questions:
2. One of the inefficiencies of this is that for every connection we go through the pull dance first try pull and then if pull fails, non-pull. For scripts, where there is just one connection, this is not a real problem. However,  it does assume that the user makes a WBEMConnection with a WEBServer and keeps using it to be efficient. No solution. This is just a note for now.

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -55,6 +55,35 @@ WBEMConnection method                                       Purpose
 ----------------------------------------------------------  --------------------------------------------------------------
 :meth:`~pywbem.WBEMConnection.ExecQuery`                    Execute a query in a namespace
 ----------------------------------------------------------  --------------------------------------------------------------
+:meth:`~pywbem.WBEMConnection.IterEnumerateInstances`       Iterator API that uses either OpenEnumerateInstances and
+                                                            PullInstancesWithPath or EnumerateInstances depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterEnumerateInstancePaths`   Iterator API that uses either OpenEnumerateInstances and
+                                                            PullInstancesWithPath or EnumerateInstances depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterReferenceInstances`       Iterator API that uses either OpenReferenceInstances and
+                                                            PullInstancesWithPath or References depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterReferenceInstancePaths`   Iterator API that uses either OpenReferenceInstances and
+                                                            PullInstancesWithPath or References depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterAssociatorInstances`      Iterator API that uses either OpenAssociatorInstances and
+                                                            PullInstancesWithPath or Associators depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterAssociatorInstancePaths`  Iterator API that uses either OpenAssociatorInstances and
+                                                            PullInstancesWithPath or Associators depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+:meth:`~pywbem.WBEMConnection.IterQueryInstances`           Iterator API that uses either OpenQueryInstances and
+                                                            PullInstances or ExecQuery depending on
+                                                            the attributes and existence of pull operations in the
+                                                            server.
+----------------------------------------------------------  --------------------------------------------------------------
 :meth:`~pywbem.WBEMConnection.OpenEnumerateInstances`       Open enumeration session to retrieve instances of
                                                             of a class (including instances of its subclass)
 :meth:`~pywbem.WBEMConnection.OpenAssociatorInstances`      Open enumeration session to retrieve the instances
@@ -116,13 +145,16 @@ from collections import namedtuple
 
 import six
 from . import cim_xml
-from .cim_constants import DEFAULT_NAMESPACE, CIM_ERR_INVALID_PARAMETER
+from .config import DEFAULT_ITER_MAXOBJECTCOUNT
+from .cim_constants import DEFAULT_NAMESPACE, CIM_ERR_INVALID_PARAMETER, \
+    CIM_ERR_NOT_SUPPORTED
 from .cim_types import CIMType, CIMDateTime, atomic_to_cim_xml
 from .cim_obj import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
     NocaseDict, _ensure_unicode, tocimxml, tocimobj
 from .cim_http import get_object_header, wbem_request
 from .tupleparse import parse_cim
 from .tupletree import xml_to_tupletree_sax
+from .cim_http import parse_url
 from .exceptions import Error, ParseError, AuthError, ConnectionError, \
     TimeoutError, CIMError
 
@@ -368,7 +400,27 @@ def check_utf8_xml_chars(utf8_xml, meaning):
     return utf8_xml
 
 
-class WBEMConnection(object):
+def _validateIterCommonParams(MaxObjectCount, OperationTimeout):
+    """
+    Validate common parameters for an iter... operation.
+
+    MaxObjectCount must be a positive non-zero integer or None.
+
+    OperationTimeout must be positive integer or zero
+
+    Raises:
+      ValueError if these parameters are invalid
+    """
+    if MaxObjectCount is None or MaxObjectCount <= 0:
+        raise ValueError('MaxObjectCount must be > 0 but is %s' %
+                         MaxObjectCount)
+
+    if OperationTimeout is not None and OperationTimeout < 0:
+        raise ValueError('OperationTimeout must be >= 0 but is %s' %
+                         OperationTimeout)
+
+
+class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
     """
     A client's connection to a WBEM server. This is the main class of the
     WBEM client library API.
@@ -482,11 +534,26 @@ class WBEMConnection(object):
         :class:`~pywbem.BaseOperationRecorder`, each operation
         that is executed on this connection will be recorded by
         invoking its :meth:`~pywbem.BaseOperationRecorder.record` method.
+
+      use_pull_operations (:class:`py:bool`):
+        Indicates whether the client will attempt the use of pull operations in
+        any `Iter...()` methods. However, each operation has its own internal
+        equivalent flag based on this flag to allow for implementations that
+        implement only some of the pull operations. The method
+        `set_use_pull_operations` may be used to reset all of these flags to
+        a defined value.
+
+        This variable is initially set from the same-named input argument.
+        If it is `Nnone` is modified when `Iter...()` methods are invoked,
+        dependent on the support for pull operations in the WBEM server.
+        After the first call , the varaible will be either `True` or `False`
+        and indicating whether pull operations will be used in this and
+        subsequent `Iter...()` methods.
     """
 
     def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
                  x509=None, verify_callback=None, ca_certs=None,
-                 no_verification=False, timeout=None):
+                 no_verification=False, timeout=None, use_pull_operations=None):
         """
         Parameters:
 
@@ -672,6 +739,26 @@ class WBEMConnection(object):
             Note that not all situations can be handled within this timeout, so
             for some issues, operations may take longer before raising an
             exception.
+
+          use_pull_operations (:class:`py:bool`):
+            Controls the use of pull operations in any `Iter...()` methods, by
+            intializing a corresponding instance variable that is used during
+            any `Iter...()` methods.
+            The default value (`None`) will work on any WBEM server whether or
+            not it supports pull operations.
+
+            `None` means that the `Iter...()` methods will attempt a pull
+            operation first, and if the WBEM server does not support it, will
+            use a traditional operation. The corresponding instance variable
+            will be adjusted to indicate what the WBEM server supports.
+
+            `True` means that the `Iter...()` methods will only use pull
+            operations. If the WBEM server does not support pull operations, a
+            :exc:`~pywbem.CIMError` with status code `CIM_ERR_NOT_SUPPORTED`
+            will be raised.
+
+            `False` means that the `Iter...()` methods will only use traditional
+            operations.
         """
 
         self.url = url
@@ -689,6 +776,16 @@ class WBEMConnection(object):
         self.last_request = None
         self.last_reply = None
         self.operation_recorder = None
+        self.use_pull_operations = use_pull_operations
+        # set the flags for each individual operation to the initial
+        # value defined by use_pull_operations
+        self._use_enum_inst_pull_operations = use_pull_operations
+        self._use_enum_path_pull_operations = use_pull_operations
+        self._use_ref_inst_pull_operations = use_pull_operations
+        self._use_ref_path_pull_operations = use_pull_operations
+        self._use_assoc_inst_pull_operations = use_pull_operations
+        self._use_assoc_path_pull_operations = use_pull_operations
+        self._use_query_pull_operations = use_pull_operations
 
     def __str__(self):
         """
@@ -720,10 +817,12 @@ class WBEMConnection(object):
             creds_repr = repr(self.creds)
         return "%s(url=%r, creds=%s, " \
                "default_namespace=%r, x509=%r, verify_callback=%r, " \
-               "ca_certs=%r, no_verification=%r, timeout=%r)" % \
+               "ca_certs=%r, no_verification=%r, timeout=%r, " \
+               "use_pull_operations=%r)" % \
                (self.__class__.__name__, self.url, creds_repr,
                 self.default_namespace, self.x509, self.verify_callback,
-                self.ca_certs, self.no_verification, self.timeout)
+                self.ca_certs, self.no_verification, self.timeout,
+                self.use_pull_operations)
 
     def imethodcall(self, methodname, namespace, response_params_rqd=None,
                     **params):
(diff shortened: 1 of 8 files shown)


pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+
+            Exceptions described in :class:`~pywbem.WBEMConnection`.
+
+        Returns:
+
+          A Python iterator object. Instances can be retrieved
+          by iterating through the iterator.
+
+        Example:
+            enum_gen = conn.EnumerateInstancesWithPathGen('CIM_Blah')
```
andy-maier: - The function is now called differently.
- The return value is an iterator not a generator and the variable storing it should not end with `_gen`. Maybe just  `instances`?

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                if inst.get('thisPropertyName') == 0
+                    enum_get.close()
+                    break;
+                else:
+                    print('instance %s' % inst.tomof())
+
+        """
+
+        if MaxObjectCount is not None and MaxObjectCount <= 0:
+            raise ValueError('MaxObjectCount=%s. Invalid" le 0' % \
```
andy-maier: There's probably something wrong in the format string.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                        pull_result = self.PullInstancesWithPath(
+                            pull_result.context, MaxObjectCount=MaxObjectCount)
+
+                        while inst in pull_result.instances:
+                            yield inst
+                    return
+
+                # if Not supported, set flag and try alternative operation.
+                except CIMError as ce:
+                    if ce.args[0] != CIM_ERR_NOT_SUPPORTED:
```
andy-maier: Use `ce.status_code` instead.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                        self.has_pull_operations = False
+                    else:
+                        raise
```
andy-maier: Here is a logic error. The description says we want to try pull enumerations if `has_pull_operations` is None, and set it to True if that succeeds. That second part does not happen in the current code, it can be done in an else path of the try/except.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+
```
andy-maier: How about inserting an `assert self.has_pull_operations == False` to document this?

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+            # Alternate request if Pull not implemented. This does not allow
+            # the FilterQuery
+            if FilterQuery is not None or FilterQueryLanguage is not None:
+                raise MaxObjectCount(
```
andy-maier: - MaxObjectCount is not an exception.
- In addition, we do not have a CIM status 'CIM_ERR_FILTERED_ENUMERATION_NOT_SUPPORTED'.

The discussion point is whether we want to raise a `CIMError` (which is documented to be an error returned from a server) vs. some other exception Python typically uses for incorrect input parameters, e.g. `ValueError`.

I think I developed an opinion on this: As a user of this function I'd expect a `CIMError(CIM_ERR_NOT_SUPPORTED)` to indicate that the CIMOM or provider did not implement enumeration of instances in any way, so I would find it confusing to raise that when the expectation of the user about pull support was wrong. I'd argue that the user should look at `conn.has_pull_operations` before it attempts to use pull-specific parameters. If they missed that, they deserve that these parameters are considered incorrect input parameters, i.e. I think we should raise a `ValueError` in this case.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                # if Not supported, set flag and try alternative operation.
+                except CIMError as ce:
+                    if ce.args[0] != CIM_ERR_NOT_SUPPORTED:
+                        self.has_pull_operations = False
+                    else:
+                        raise
+
+            # Alternate request if Pull not implemented. This does not allow
+            # the FilterQuery
+            if FilterQuery is not None or FilterQueryLanguage is not None:
```
andy-maier: I think `OperationTimeout` should also be rejected.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+
+                    for inst in pull_result.instances:
+                        yield inst
+
+                    # loop to pull while more exist.
+                    while not pull_result.eos:
+                        pull_result = self.PullInstancesWithPath(
+                            pull_result.context, MaxObjectCount=MaxObjectCount)
+
+                        while inst in pull_result.instances:
```
andy-maier: `while` -> `for`

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                ClassName, namespace=namespace, LocalOnly=LocalOnly,
+                DeepInheritance=DeepInheritance,
+                IncludeQualifiers=IncludeQualifiers,
+                IncludeClassOrigin=IncludeClassOrigin,
+                PropertyList=PropertyList, **extra)
+                
+            for inst in enum_rslt:
+                yield inst
+
+        # used to clean up if close() executed
```
andy-maier: Don't understand this comment. The code in `finally` gets executed in all cases where one of the operations fails or exhausts the result set. This code is not triggered by the user calling `close()` on the returned iterator.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+        finally:
+            if pull_result is not None:
+                if not pull_result.eos:
+                    self.CloseEnumeration(pull_result.eos)
```
andy-maier: `pull_result.eos` -> `pull_result.context`

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
                 self.operation_recorder.record_staged()
             return result_tuple
 
+    # TODO overriding the internal choice of pull vs enum. Should we be able
+    #      force one or the other for a connection?
+    # TODO is this a good name for the generator function
+    def EnumerateInstancesGen(self, ClassName, namespace=None,
```
andy-maier: Some material on discussing naming conventions for generator functions:
- https://bytes.com/topic/python/answers/506531-generator-naming-convention
- https://stackoverflow.com/questions/9039265/name-of-a-function-returning-a-generator

I think given that we have different flavors of methods, we cannot just use no prefix or no suffix. It seems to me that using an `iter` prefix would be most Pythonese. Maybe in mixed case to be consistent with our naming style, i.e. `Iter`. Basically, use the name of the open operation and replace `Open` with `Iter`.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1721,269 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
                 self.operation_recorder.record_staged()
             return result_tuple
 
+    # TODO overriding the internal choice of pull vs enum. Should we be able
+    #      force one or the other for a connection?
```
andy-maier: On overriding: No. The user can already control the behavior of the generator method using the `has_pulled_operations` capability of the connection. If the user wants a different behavior on that same connection than the behavior selected with `has_pulled_operations`, they can use the specific pulled or non-pulled operation methods directly. First, there is no performance disadvantage of doing that compared to using the generator method, because the generator method still materializes the result set of each pull or (in case of non-pulled) of the entire operation, before it yields its instances. Admittedly, a little more coding is needed for using the pull operations directly. Second, cases where it makes sense to use the opposite of what was selected when creating the connection should be rare.

pywbem/cim_operations.py
```diff
@@ -453,11 +453,18 @@ class WBEMConnection(object):
...
+        It can be set by the attribute or if the default of None is set, the
+        first request that uses the pull operations sets the value to True.
+        If set to False, the generator functions will default to using the
+        non-pull equivalent methods
     """
 
     def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
                  x509=None, verify_callback=None, ca_certs=None,
-                 no_verification=False, timeout=None):
+                 no_verification=False, timeout=None, has_pull_operations=None):
```
andy-maier: This parameter does not indicate what the server has, but what this connection wants to use, and we even adjust that as we go, to what the server has.

So I suggest renaming this parameter and the corresponding instance variable to `use_pull_operations`.

andy-maier:
See my comments in the code.

pywbem/cim_operations.py
KSchopmeyer: Extra quote

pywbem/cim_operations.py
KSchopmeyer: Actually wanted to use the pull_result and use None as a flag.  I did not set it back to None when done however

pywbem/cim_operations.py
KSchopmeyer: Used insts_iterator

pywbem/cim_operations.py
KSchopmeyer: Will do

KSchopmeyer: Done

pywbem/cim_operations.py
KSchopmeyer: Since these do not cause any issues or really change the performance if enumerates are issue I propose that we not flag them.  The goal is to provide as compatible an interface as possible so that the maximum number of parameters do not have to be set dependent on the  type of call, not to detail out each difference.
The timeout simply does not matter if EnumerateInstances is used. It only applies to the pull.

The continue on error does have a user impact. 

I propose test the continueOnError but not the timeout.

pywbem/cim_operations.py
KSchopmeyer: Done

pywbem/cim_operations.py
KSchopmeyer: Right now I kept them both but want to look at this again

andy-maier:
PR #546 addresses the issue the one Travis CI run encountered by chance. I restarted the run.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+
+          A Python iterator object. Instances can be retrieved
+          by iterating through the iterator.
+
+        Example:
+            insts_iterator = conn.IterEnumerateInstances('CIM_Blah')
+            for inst in insts_iterator:
+                # close if a particular property value found
+                if inst.get('thisPropertyName') == 0
+                    enum_get.close()
```
andy-maier: Change old variable name `enum_get` to `insts_iterator`.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                    break;
```
andy-maier: Semicolon not needed after `break`.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                else:
+                    print('instance %s' % inst.tomof())
+
+        """
+
+
+        if MaxObjectCount is not None and MaxObjectCount == 0:
+            raise ValueError('MaxObjectCount=%s. Invalid leq 0' % \
+                             MaxObjectCount)
```
andy-maier: The message still does not reflect the comparison. In addition, we should not use abbreviations such as `leq`.

pywbem/cim_operations.py
andy-maier: Done - Was changed as proposed.

pywbem/cim_operations.py
```diff
@@ -453,11 +459,23 @@ class WBEMConnection(object):
...
+        Attribute that defines whether the WBEM server defined by this
+        connection has or is to use the pull operations
+
+        if `None` no pull operation has been attempted on this connection
+        If `True` the user has specified pull operations only or the client
+        has received a successful response to a Iter... method.
+        If `False` the  the user set use_pull_operations=False in the WBEM
+        connection creation or the server has responded with not supported
+        indicating that the pull operations are not supported on the server.
+
```
andy-maier: I can meanwhile understand why you had named the instance variable `has_pull_operations` at some point (if i understood you correctly in the call): After the first call to an `Iter...()` method it is set to reflect the capability of the server, but only if it was not initially passed as `False`. I think naming the instance variable `use_pull_operations` like it is currently is better, because it does not truly reflect the server capability, but what is used in the `Iter...()` methods. No change needed, I just wanted to provide my thoughts on this.

Comments on the description:

1. On "... the WBEM server ... has or is to use ...": It is the client that uses them, not the server.

2. The description of the three values mostly repeats the algorithm that is implemented. I think it should be written by focusing on the use of pull.

3. As an instance variable, I prefer wording that explains what the variable indicates instead of what it controls or defines.

How about this description:

```
        Indicates whether the client will attempt the use of pull operations in any `Iter...()` methods.

        This variable is initially set from the same-named input argument, and is
        adjusted when `Iter...()` methods are invoked, dependent on the support
        for pull operations in the WBEM server.
        After adjusting the variable, it is either `True` or `False` and indicates whether
        pull operations will be used in any `Iter...()` methods.
```

pywbem/cim_operations.py
```diff
@@ -643,6 +661,19 @@ def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
...
+            or only the non-pull operations for the Iter... methods. Normally
+            this should NOT be used but it may prove useful in some test or
+            error cases.
+            If `None` ignored and the server sets the attribute based on the
+            response to the first pull operation.
+            If `True` forces the client to use only the pull operations for
+            the Iter... methods and return any response  to a pull operation
+            including `CIM_ERR_NOT_SUPPORTED` as an error on this connection.
+            If `False` forces the client to use only the non-pull operations
+            (ex. EnumerateInstances) for the Iter... methods.
```
andy-maier: Comments on the description of the `use_pull_operation`  ctor argument:

1. The "force" only works in one direction, at least in the current implementation: `False` forces non-use even if pull is supported by the server, but `True` specifies that an attempt to use pull is made, which may fall back to using non-pull.

2. On "Normally this should not be used...": It is not clear what the "this" refers to.

3. For `None`, I would like to see a more explicit statement that pull is attempted.

4. For `True`, it does not force the use of only pull. That may have been an a prior implementation, but the current code attempts pull and if it fails with not supported, goes on to attempt non-pull. From a perspective of the (unadjusted) input argument, `None` specifies exactly the same as `True`. Which is fine IMO, just the description needs to be fixed.

5. On "If `False` forces ...": I don't like this super-short form. Better use "`False` forces ...".

How about this description:

```
            Controls the use of pull operations in any `Iter...()` methods. The default
            value will work fine on any WBEM server whether or not it supports pull
            operations.
            `None` and `True` mean that pull operations will be attempted, and if
            the WBEM server does not support pull operations, the corresponding
            instance variable is set to `False`, causing the traditional operations
            to be used in the future.
            `False` means that only the traditional operations will be used,
            regardless of whether or not the WBEM server also supports
            pull operations.
```

pywbem/cim_operations.py
andy-maier: Done, ValueError is now raised.

pywbem/cim_operations.py
andy-maier: Done: Karl added checking for ContinueOnError, as proposed.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+                ClassName, namespace=namespace, LocalOnly=LocalOnly,
+                DeepInheritance=DeepInheritance,
+                IncludeQualifiers=IncludeQualifiers,
+                IncludeClassOrigin=IncludeClassOrigin,
+                PropertyList=PropertyList, **extra)
+
+            for inst in enum_rslt:
+                yield inst
+
+        # Clean up if close() executed and  pull operation is active
```
andy-maier: I finally understand the situation this comment describes: The user is in the loop iterating over the items yielded by this method (i.e. iterates over the returned iterator), and for some reason calls close() on the returned iterator before the iterator is exhausted. Using the wording "and pull operation is active" is a bit confusing to me, because it is not the pull operation against the server that is active, but the iterator is not yet exhausted.

Can we change this comment to:

```
# Clean up if the caller closes the iterator before exhausting it
```

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+        finally:
+            if pull_result is not None and not pull_result.eos:
+                self.CloseEnumeration(pull_result.eos)
```
andy-maier: Error: `CloseEnumeration()` must be called with the context.

To fix this, change line 2034 to:

```
self.CloseEnumeration(pull_result.context)
```

pywbem/cim_operations.py
```diff
@@ -1944,6 +2269,9 @@ class are to be included in the returned instances, as follows:
                 MaxObjectCount=MaxObjectCount,
                 **extra)
 
+        if MaxObjectCount is not None and MaxObjectCount < 0:
+            raise ValueError('MaxObjectCount=%s. Invalid lt 0' % MaxObjectCount)
```
andy-maier: Suggest to change the message to be a human readable statement, e.g.:

```
"MaxObjectCount must be >= 0, but is %s" % MaxObjectCount
```

pywbem/cim_operations.py
andy-maier: The comment applies to line 1984. I raised this issue again on the latest version.

pywbem/cim_operations.py
andy-maier: Done, Karl changed it to `Iter...()`

pywbem/cim_operations.py
andy-maier: Done - comment was removed, and no override control is provided.

pywbem/cim_operations.py
andy-maier: Done - renamed to `use_pull_operations`.

andy-maier:
I have some more comments on the last commit, see above or the "Files changed" tab.

pywbem/cim_operations.py
KSchopmeyer: Done

pywbem/cim_operations.py
KSchopmeyer: Done

pywbem/cim_operations.py
KSchopmeyer: It was originally just internal, there has ... When we extended it to cover both what the client sees and what the user wants, we sort of mixed things up.

Done

pywbem/cim_operations.py
KSchopmeyer: Done. Added comment explaining why only on pull below the finally also

pywbem/cim_operations.py
KSchopmeyer: Good catch.

pywbem/cim_operations.py
KSchopmeyer: Done.  should be > 0, not >= 0

KSchopmeyer: There were multiples and fixed them all.  Note that since we have to add this test to all open/pulls I will make it a common function but right now concentrating on the Iter...

pywbem/cim_operations.py
andy-maier: Done by Karl.

pywbem/cim_operations.py
```diff
@@ -1709,6 +1741,299 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+            execution of this generator.
+
+            * If positive, the WBEM server is to return no more than the
+              specified number of instances.
+            * Zero is not allowed in that it would mean that zero instances
+              are to be returned for every request issued.
+            * If `None`, this parameter is not passed to the WBEM server, and
+              causes the server-implemented default behaviour to be used.
+              :term:`DSP0200` defines that the server-implemented default is
+              to return zero instances.
```
andy-maier: This one is still open, we discussed in the call today that we would address it as proposed.

pywbem/cim_operations.py
```diff
@@ -453,11 +459,33 @@ class WBEMConnection(object):
...
+        Indicates whether the client will attempt the use of pull operations in
+        any `Iter...()` methods.
+
+        This variable is initially set from the same-named input argument, and
+        is modified when `Iter...()` methods are invoked, dependent on the
+        support for pull operations in the WBEM server.
+        After modifying the variable, it is either `True` or `False` and
+        indicates whether pull operations will be used in any subsequent
+        `Iter...()` methods.
+
```
andy-maier: The new text for the `use_pull_operations` instance variable (above this comment) makes the part of the old text that is till below this comment (lines 474 to 482) superfluous. They should be removed.

andy-maier:
I had three more comments, see the "Files changed" tab.

pywbem/cim_operations.py
```diff
@@ -643,6 +671,24 @@ def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
...
+            not it supports pull operations so that first the pull operation is
+            attempted and if the WBEM server does not support pull operations,
+            the corresponding instance variable is set to `False` the
+            traditional (non-pull) operation will be executed for the current
+            operation and all subsequent  Iter...()operations for this
+            WBEM Connection.
+
+            `True` means that pull operations will be attempted; all errors
+            including `CIM_ERR_NOT_SUPPORTED` are returned to the caller. The
+            traditional operations will not be attempted.
```
KSchopmeyer: No.  The way the code works is that if True is set on the attribute, a failure of the open or pull goes directly to raise in all cases.  The assumption we are making is that if true is set either by the attribute or initial success, all operations MUST be pull.   Unless I misread my code that is what it does.  The only time it first tries pull and then falls back to enum is if use... is None and the error is CIM_ERR_NOT_SUPPORTED. 

                except CIMError as ce:
                    if self.use_pull_operations is None and \
                       ce.status_code == CIM_ERR_NOT_SUPPORTED:
                        self.use_pull_operations = False
                    else:
                        raise

Any other exception gets raised directly.

pywbem/cim_operations.py
KSchopmeyer: I agree.  Removed

pywbem/cim_operations.py
andy-maier: Karl, you did not miss anything. My reading of the code was wrong. So in that case, I have a few other comments on the description, which I will make in the file again.

pywbem/cim_operations.py
```diff
@@ -453,11 +460,22 @@ class WBEMConnection(object):
...
+      use_pull_operations (:class:`py:bool`):
+        Indicates whether the client will attempt the use of pull operations in
+        any `Iter...()` methods.
+
+        This variable is initially set from the same-named input argument, and
+        is modified when `Iter...()` methods are invoked, dependent on the
+        support for pull operations in the WBEM server.
+        After modifying the variable, it is either `True` or `False` and
+        indicates whether pull operations will be used in any subsequent
+        `Iter...()` methods.
```
andy-maier: In order to reflect the fact that only for an initial value of `None`, the instance variable will be adjusted, I suggest to replace the paragraph above with this:

```
        This variable is initially set from the same-named input argument. If
        it is `None`, it is adjusted when `Iter...()` methods are invoked, dependent
        on the support for pull operations in the WBEM server. After the first call
        to an `Iter...()` method, the variable will be either `True` or `False`.
```

pywbem/cim_operations.py
```diff
@@ -643,6 +661,24 @@ def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
...
+
+          use_pull_operations (:class:`py:bool`):
+            Controls the use of pull operations in any `Iter...()` methods.
+            The default value (`None`) will work on any WBEM server whether or
+            not it supports pull operations; first the pull operation is
+            attempted and if the WBEM server does not support pull operations,
+            the corresponding instance variable is set to `False` and the
+            traditional (non-pull) operation will be executed for the current
+            operation and all subsequent Iter...()operations for this
+            WBEM Connection.
```
andy-maier: The paragraph above is not explicit about the fact that the instance variable is set to True is the server supports pull operations. I suggest to replace it with this text:

```
            Controls the use of pull operations in any `Iter...()` methods.
            The default value (`None`) will work on any WBEM server whether or
            not it supports pull operations.

            `None` means that first the pull operation is attempted. If that
            succeeds, corresponding instance variable is set to `True`. If the
            WBEM server does not support pull operations, the corresponding
            instance variable is set to `False` and the traditional (non-pull)
            operation will be executed.
```

pywbem/cim_operations.py
```diff
@@ -643,6 +661,24 @@ def __init__(self, url, creds=None, default_namespace=DEFAULT_NAMESPACE,
...
+
+            `True` means that pull operations will be attempted; all errors
+            including `CIM_ERR_NOT_SUPPORTED` are returned to the caller. The
+            traditional operations will not be attempted.
+
+            `False` means that only the traditional operations will be used,
+            regardless of whether or not the WBEM server also supports
+            pull operations.
```
andy-maier: The description is not explicit about the fact that the instance variable is set to True is the server supports pull operations, and we could do a few editorial improvements as well.

I suggest to replace the description with this one:

```
            Controls the use of pull operations in any `Iter...()` methods, by
            intializing a corresponding instance variable that is used during
            any `Iter...()` methods.
            The default value (`None`) will work on any WBEM server whether or
            not it supports pull operations.

            `None` means that the `Iter...()` methods will attempt a pull operation
            first, and if the WBEM server does not support it, will use a traditional
            operation. The corresponding instance variable will be adjusted to
            indicate what the WBEM server supports.

            `True` means that the `Iter...()` methods will only use pull operations.
            If the WBEM server does not support pull operations, a
            :exc:`~pywbem.CIMError` with status code `CIM_ERR_NOT_SUPPORTED`
            will be raised.

            `False` means that the `Iter...()` methods will only use traditional operations.
```

pywbem/cim_operations.py
```diff
@@ -1709,6 +1745,311 @@ def OpenEnumerateInstancePaths(self, ClassName, namespace=None,
...
+
+        This method performs either the `OpenEnumerationInstances` and
+        `PullInstancesWithPath` operations (pull requests) or the
+        EnumerateInstances operation if the WBEM server does not support the
+        DMTF pull operations. It is an alternate to using `EnumerateInstances`
+        or a `OpenEnumerateInstances`/`PullInstancesWithPath` sequence.
+
+        The `use_pull_operations` attribute of the WBEMConnection constructor
+        may be used to control whether pull or traditional requests are
+        used.
```
andy-maier: The two paragraphs above could be improved to talk about the default behavior and that the input argument can be used to override the default. Also, I would use "argument" and not "attribute".
Finally, links could be used where possible.

To address all of that, I suggest to replace the two paragraphs above with these:

```
        This method performs either the
        :meth:`~pywbem.WBEMConnection.OpenEnumerationInstances` and
        :meth:`~pywbem.WBEMConnection.PullInstancesWithPath`
        operations (pull operations) or the
        :meth:`~pywbem.WBEMConnection.EnumerateInstances` operation
        (traditional operation) if the WBEM server does not support the
        pull operations. It is an alternative to using these operations that
        automatically uses the pull operations if supported.

        The `use_pull_operations` argument of the
        :class:`~pywbem.WBEMConnection` constructor can be used to
        override that default behavior to either insist on pull operations or
        to avoid their use.
```

andy-maier:
Some more comments in the "Files changed" section.

Plus this one:

* By having only one variable for the use of pull operations, we assume that a WBEM server has either implemented all pull operations or none of them. An implementation of only OpenEnumerateInstances() but not of "OpenAssociators()" and "OpenReferences()" will cause failures, because we are setting the instance variable to True if the first pull operation in any Iter...() method succeeds.

  This problem becomes more likely considering that CIM_ERR_NOT_SUPPORTED is probably also returned if the provider does not support the attempted pull operation. That could even happen when the CIMOM transforms the pull client operation into a non-pull provider operation, and the provider simply does not support it (e.g. an association provider supports Associators but not References).

**Discussion point:** Does the client need to accomodate for the case that a WBEM server supports only a subset of the pull operations?

KSchopmeyer:
One more interesting comment. We have a public attribute use_pull_operations.  The user can reset that at any time to start over.  In fact I am using it in the tests to first force pull and then non-pull rather than just setting it when the connection is made. That is just a side effect but it is the api right now

KSchopmeyer:
We propose to just extend the documentation so that users understand the potential issue.

andy-maier:
To get the Travis issues fixed, please rebase this on branch `issue550-pin-gitpython-208`.

andy-maier:
I have merged PR #551 and deleted branch `issue550-pin-gitpython-208`, so you now need to rebase on `master`.

KSchopmeyer:
Discussion:  The individual flags.

andy-maier:
Result of the discussion about the individual flags:
- code sets the individual flags from the connection level flag. The individual flags get updated based upon what the server supports. The individual flags are not fed back to update the connection level flag.
- there is a new function set_enumerate...() that allows changing the connection level flag and essentially starts over again with the automatic updating of the individual flags.

andy-maier:
Remaining Todos:
* Review whether we want the set_enumerate....() function
* Review entire code
Once both of that has happened:
* Implement the rest of the functions.

KSchopmeyer:
Updated to rebase to master that has all flake8 fixes and rebase into single set of changes.

andy-maier:
On the need for the `set_use_pull_operations()` method: The CIMOM's support for pull operations does not change for a given connection. However, provider support may vary across the different providers.

If a CIMOM compensates for that, i.e. if it services a client's pull operation by invoking the classic provider operation in case a provider does not support pull, then we don't need the method. Unfortunately, the need for such a compensating behavior is not specified anywhere.

Still, I have a hard time imagining a client program using the `set_use_pull_operations()` method to adjust to pull vs. classic at a provider granularity, and possibly different for each server. Having said that, the alternative to using the method is to have one connection with, and one without pull enabled, and to use them dependent on the provider implementation. That is not any less ugly, but also not significantly more ugly, IMO. The ugliness is not so much the second connection, but the need to understand which provider on which type of server environment implemented pull.

In other words, the `set_use_pull_operations()` method does not reduce the ugliness in any significant way.

I believe the way out is the deprecation of the classic operations in DSP0200. That will cause CIMOM implementers to support the compensating behavior described above.

Overall, I think we should not have the `set_use_pull_operations()` method.

andy-maier:
Otherwise, +1 on the implemented operations.

KSchopmeyer:
We agreed to add the missing path information back in so old and new operations match

Also proposed as separate issue to fix the original operations based on a config variable to that paths include namespace and host.

KSchopmeyer:
We think this should be a higher level class, ie. IterWBEMConnection.  We will do that before commit.

We also concluded that the recorder is at the lowest level.

pywbem/cim_operations.py
```diff
@@ -482,11 +534,26 @@ class WBEMConnection(object):
         :class:`~pywbem.BaseOperationRecorder`, each operation
         that is executed on this connection will be recorded by
         invoking its :meth:`~pywbem.BaseOperationRecorder.record` method.
```
andy-maier: This is where we can add a statement that for the `Iter*()` methods, the recording will be done at the level of the lower level methods they invoke.

andy-maier:
I agree with the approach, and have made a comment where I suggested a good place for adding the statement about recording for the Iter functions.

andy-maier:
+1

PR #501, opened 2016-09-10
Fixes #500: Correct var used before setting issue

### LINKED ISSUE #500: subscription manager, variable used before setting
We are using the dest_path variable before setting it.

KSchopmeyer:
Corrected just plain code error.  Meant that remove subscription does not actually remove subscription from server.

Think I need to do more through indication test since the test should have caught this.  But will make that part of general cleanup

Note that this is actually a BUG and should go into version 0.9.1.  However, we probably have a few days since I would not expect much use of the new code immediatly.

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -506,11 +506,11 @@ class WBEMSubscriptionManager(object):
             return
 
         server = self._get_server(server_id)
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #504, opened 2016-09-13
Fixes #502: Extend examples for indications

### LINKED ISSUE #502: example/listener only displays overall count
Should be easy to expand listener.py to count by host.

KSchopmeyer:
Extend listen.py to count indications received by host rather than
overall count. Change name of logger from _logger to logger to
be easier to work with and consistent with python logger examples

Change send_indications.py to correct issue in argparser.

Clean up send_indicatons server cert verification

Add a new property to indication so in the future we can id source.

Clean up code around server cert verification. We had disabled it before
since the requests module has a different pattern than other ssl code in
that it combines the no-verification and server cert into a single
variable. If False, no verification of server cert. If a file, it is the
server cert file.

### DIFF
--- a/examples/listen.py
+++ b/examples/listen.py
@@ -9,7 +9,7 @@ import sys as _sys
 import os as _os
 import code as _code
 import errno as _errno
-import logging as _logging
+import logging
 import threading
 # Conditional support of readline module
 try:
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

PR #505, opened 2016-09-13
Fixes #485: run_cimoperations.py fail with instance count error

### LINKED ISSUE #485: Pull operation response count error
In run_cimoperations.py we run tests of the pull operations by running both the pull operation and the corresponding  non-pull and comparing results.   We occasionally hit one error, in the pull of paths with maxobjectcount = 1 where we are tryng to retrieve about 700 paths.  This does not happen all the time but it does happen.

```
IL: test_get_onebyone (__main__.PullEnumerateInstancePaths)
Enumerate instances with MaxObjectCount = 1)
  File "./run_cim_operations.py", line 1124, in test_get_onebyone
    self.assertEqual(len(paths_pulled), len(paths_enum))
AssertionError: 702 != 700
```

We had a couple other of this type of issue and found that they were actually errors in the test code, not PyWBEM.  In addition in all cases I have documented it is a difference of two paths.

As an example, I ran a repeat 100 times of this one test and the problem happened 2 times. In each case,  the pull showed 2 fewer instances than the enumerate response.

It now only happens with this boundary case of asking for instances one at a time so we released 0.9.0 with the issue open

KSchopmeyer:
Ready for review

Fixed the existing tests so that it does not fail.  The issue was not the code but the fact that we were enumerating CIM_ManagedElement and one of the subclasses is threads.  The count of threads varies continuously so that even a open/pull sequence immediatly followed by an enumerate often ended up being slightly different. Changed the test so that we do approximate counts and it does not fail if the counts are slightly different.

We will add one more test with another issue where we know the exact count expected and can test easily but that will be pegasus only since it depends on a specific provider.

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -50,6 +50,12 @@ TEST_CLASS_NAMESPACE = 'root/cimv2'
 TEST_CLASS_PROPERTY1 = 'Name'
 TEST_CLASS_PROPERTY2 = 'CreationClassName'
 
+# TOP level class that can be used for tests. Note that this class delivers
+# differing numbers of instances over time because it is reporting on the
+# running server.
+TOP_CLASS = 'CIM_ManagedElement'
+TOP_CLASS_NAMESPACE = 'root/cimv2'
+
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

Issue #506, opened 2016-09-16
pywbem 0.9.0 installation over pip fails


Napsty:
Tried to install the newest 0.9.0 from pip but it fails.
I tested on Linux Mint 17.3 (based on Ubuntu 14.04) and also on openSUSE 13.2. The result is the same; error message is 

```
 SWIG/_m2crypto_wrap.c:127:20: fatal error: Python.h: No such file or directory
```

Full output (this is from openSUSE):

```
linux-uevk:~ # pip install pywbem
Downloading/unpacking pywbem
  Downloading pywbem-0.9.0-py2.py3-none-any.whl (188kB): 188kB downloaded
Downloading/unpacking six (from pywbem)
  Downloading six-1.10.0-py2.py3-none-any.whl
Downloading/unpacking M2Crypto>=0.24 (from pywbem)
  Downloading M2Crypto-0.25.1.tar.gz (300kB): 300kB downloaded
  Running setup.py (path:/tmp/pip_build_root/M2Crypto/setup.py) egg_info for package M2Crypto

Downloading/unpacking PyYAML (from pywbem)
  Downloading PyYAML-3.12.tar.gz (253kB): 253kB downloaded
  Running setup.py (path:/tmp/pip_build_root/PyYAML/setup.py) egg_info for package PyYAML

Downloading/unpacking ply (from pywbem)
  Downloading ply-3.9.tar.gz (150kB): 150kB downloaded
  Running setup.py (path:/tmp/pip_build_root/ply/setup.py) egg_info for package ply

    warning: no previously-included files matching '*.pyc' found anywhere in distribution
Downloading/unpacking typing (from M2Crypto>=0.24->pywbem)
  Downloading typing-3.5.2.2.tar.gz (51kB): 51kB downloaded
  Running setup.py (path:/tmp/pip_build_root/typing/setup.py) egg_info for package typing

Installing collected packages: pywbem, six, M2Crypto, PyYAML, ply, typing
  Running setup.py install for M2Crypto
    building 'M2Crypto.__m2crypto' extension
[...]

PR #509, opened 2016-09-16
Docs: Added installation and development sections.

andy-maier:
Please review.

This is improves the installation documentation in response to issue #506.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -1,0 +1,210 @@
+
+.. _`Appendix`:
+
+Appendix
+========
+
+This section contains information that is referenced from other sections,
+and that does not really need to be read in sequence.
+
+
+.. _'Special type names`:
+
+Special type names
+------------------
+
+This documentation uses a few special terms to refer to Python types:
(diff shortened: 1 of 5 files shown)


KSchopmeyer:
Couple of comments:

1. Intro.rst - The initial statement about it being pure python is correct but it into the next  paragraph before there is info about the issues with the dependencies. One sort of gets the wrong impression from that first paragraph.

3. Is there not one more way to install, pip directly from the git repository?

And NO I was not about to try the new review process that theyu announced this moring

2. Under installation we are pushing the use of virtual environments (word preferably) while serious users that are building other packages on top of this would use the virtual environments, as we put out more script level, I am assuming that these people don't care about virtual environments.  We probably need to clarify that.

KSchopmeyer:
+1 with minor changes.

andy-maier:
Made the suggested minor change, nearly as proposed.

andy-maier:
Merging, as karl had already voted +1 with the minor change.

Issue #511, opened 2016-09-20
Compatibility across versions


tasleson:
What's the projects stance with regards to compatibility?  Is the path to version 1 paved with incompatible code?

I'm seeing the following inconsistencies with running 0.7.0 vs. 0.10.0:

`pywbem.cim_http.Error -> pywbem.exception.Error or pywbem.Error`
`pywbem.cim_http.AuthError -> pywbem.exception.AuthError or pywbem.AuthError`

0.7.0 does not offer pywbem.Error & pywbem.AuthError

In addition CIMException on 0.7.0 uses [] to retrieve error code [0] and error msg [1].   Version 0.10.0 appears to use CIMException.status_code & CIMException.status_description.

There are likely other differences too, that I haven't stumbled upon.

KSchopmeyer:
You are correct. There are some compatibility issues.  To be honest, we tried to keep compatibility but clearly that did not work out perfectly.  One thing we did not do was to keep a complete set of 7.0 tests just to insure comparability. We did significantly redo exceptions, partly to get them under control with the ssl but partly just for a cleaner interfaice but we ignored compatibility.  I am sure that there are some others but hopefully not many.  See the changes.rst for more details on our changes.

If you find others, please update us.

andy-maier:
I'd like to add to what Karl said, by pointing to the [Deprecation policy](https://pywbem.readthedocs.io/en/latest/intro.html#deprecation-policy) section of the pywbem documentation. Even below 1.0.0, we are trying to stick to these rules as much as possible (i.e. 0.7 -> 0.8 -> 0.9 -> 0.10.0 are all considered new minor releases which can add new deprecations but are compatible). The text actually says "as backwards compatible as possible", leaving some wiggle room there. i think we should not allow any incompatibility in a minor release when we reach 1.0.0.

On the path to 1.0.0, my take is we try hard to stay compatible. Note that we have issue #497 open in order to discuss what we would like to change before we release a 1.0.0. Note that this is a wishlist, and we may end up not doing some of these due to compatibility issues. You can help us by commenting on that issue in case you have issues with the proposed items.

Having said all that, I think we can and should resolve the incompatibilities you listed:

* pywbem.cim_http.Error -> pywbem.Error: We can make that class available in the pywbem.cim_obj namespace, in addition. We likely cannot track its use and issue DeprecationWarning warnings.
* pywbem.cim_http.AuthError -> pywbem.AuthError: ditto
[...]

PR #513, opened 2016-09-21
For 0.9: Fixed compatibility issue with Error; Clarified CIMError; Added exceptions unit test

andy-maier:
This PR rolls back the fix for issue #511 into the 0.9 fix stream.

Details:
- Added the `pywbem.Error` base exception class to the `pywbem.cim_http` namespace where it existed in 0.7.0, for compatibility reasons.
- In the description of `CIMError`, clarified that accessing the exception object itself by index or slice will be delegated to its `args` instance variable.
- In the description of `CIMError`, clarified that `args[1]` is the *input* status description, not the value of the `status_description` property. The difference is that the latter defaults to a standard text if it is
  not provided as input, but `args[1]` does not (for compatibility reasons).
- Added a unit testcase `test_exceptions.py` for the exceptions module.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -25,6 +25,15 @@ Bug fixes
 * Fixed the use of a variable before it was set in the `remove_destinations()`
   method of class `WBEMSubscriptionManager`.
 
+* Fixed a compatibility issue relative to pywbem 0.7.0, where the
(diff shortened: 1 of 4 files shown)


andy-maier:
+1

KSchopmeyer:
+1

PR #514, opened 2016-09-21
Fixes #511: Fixed compatibility issue with Error; Clarified CIMError; Added exceptions unit test

### LINKED ISSUE #511: Compatibility across versions

andy-maier:
Please review.

Details:
- Added the `pywbem.Error` base exception class to the `pywbem.cim_http` namespace where it existed in 0.7.0, for compatibility reasons.
- In the description of `CIMError`, clarified that accessing the exception object itself by index or slice will be delegated to its `args` instance variable.
- In the description of `CIMError`, clarified that `args[1]` is the *input* status description, not the value of the `status_description` property. The difference is that the latter defaults to a standard text if it is
  not provided as input, but `args[1]` does not (for compatibility reasons).
- Added a unit testcase `test_exceptions.py` for the exceptions module.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -45,6 +45,15 @@ Bug fixes
 * Fixed the use of a variable before it was set in the `remove_destinations()`
   method of class `WBEMSubscriptionManager`.
 
+* Fixed a compatibility issue relative to pywbem 0.7.0, where the
+  `pywbem.Error` class was no longer available in the `pywbem.cim_http`
+  namespace. It has been made available in that namespace again, for
+  compatibility reasons. Note that using sub-namespaces of the `pywbem`
+  namespace such as `pywbem.cim_http` has been deprecated in pywbem 0.8.0.
+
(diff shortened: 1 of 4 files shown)


andy-maier:
+1

tasleson:
+1

KSchopmeyer:
+1

PR #516, opened 2016-09-22
Fixes #512: Add flake8 as lint test to the make and setup.py.

### LINKED ISSUE #512: Add flake8 lint to makefile
We agreed in a telecon on 11 Sept that it would be worthwhile to add flake8 as an automated target in the makefile in addition to the existing pylint.

KSchopmeyer:
Adds flake8 as lint tool to setup.py and makefile. 
It is executed as a component of check:. However
whereas pylint is python 2 only, flake8 is both
python2 and python 3.

Includes a first rc file. The rc file was created primarily to reduce those minor warnings, etc.
that are found many times.  This reduces the size of the file from about
140k to 40 k for the moment. One change to the config file was to set the line length to 80 from 79 because apparently we have pylint set to 80 so this does not show extra warnings like that.
All of the excluded warnings are documented in the pyflake8.log with comments defining what
is being warned.

Note that this does NOT include any of the many flake8 extra plugins.

### DIFF
--- a/flake8.rc
+++ b/flake8.rc
@@ -1,0 +1,32 @@
+[flake8]
+max_line_length = 80
+
+exclude =
+    .git,
+    .tox,
+    __pycache__,
+    *.pyc,
+    attic,
+    packaging,
+    build_doc,
+    docs,
+    dist,
+    pywbem.egg-info,
+    
+ignore =
+    # Ignore extra blank lines
+    E302,
+    # the backslash is redundant between brackets
+    E502,
+    # E303 too many blank lines
+    E303,
+    # E265 block comment should start with #
+    E265,
+    # E128 continuation line under-indented for visual indent
+    E128,
+    # E261 at least two spaces before inline comment
+    E261,
+    # E203 whitespace before ':'
+    E203,
+    # E231 missing whitespace after ':'
+    E231
--- a/makefile
+++ b/makefile
@@ -102,7 +102,18 @@ pylint_py_files := \
     setup.py \
     os_setup.py \
     $(filter-out $(moftab_files), $(wildcard $(package_name)/*.py)) \
-    $(wildcard testsuite/test*.py) \
+    $(wildcard testsuite/*.py) \
+    testsuite/validate.py \
+
+# Flake8 config file
+flake8_rc_file := flake8.rc
+
+# Flake8 source files to check
+flake8_py_files := \
+    setup.py \
+    os_setup.py \
+    $(filter-out $(moftab_files), $(wildcard $(package_name)/*.py)) \
+    $(wildcard testsuite/*.py) \
     testsuite/validate.py \
 
 # Test log
@@ -141,7 +152,7 @@ help:
 	@echo '  develop    - Prepare the development environment by installing prerequisites'
 	@echo '  build      - Build the distribution files in: $(dist_dir)'
 	@echo '  builddoc   - Build documentation in: $(doc_build_dir)'
-	@echo '  check      - Run PyLint on sources and save results in: pylint.log'
+	@echo '  check      - Run PyLint&Flake8 on sources and save results in: pylint.log&flake8.log'
 	@echo '  test       - Run unit tests and save results in: $(test_log_file)'
 	@echo '  all        - Do all of the above'
 	@echo '  install    - Install distribution archive to active Python environment'
@@ -209,7 +220,7 @@ doccoverage:
 	@echo '$@ done.'
 
 .PHONY: check
-check: pylint.log
+check: pylint.log flake8.log
 	@echo '$@ done.'
 
 .PHONY: install
@@ -227,7 +238,7 @@ test: $(test_log_file)
 
 .PHONY: clobber
 clobber: clean
-	rm -f pylint.log epydoc.log test_*.log $(moftab_files) $(dist_files)
+	rm -f pylint.log  flake8.log epydoc.log test_*.log $(moftab_files) $(dist_files)
 	rm -Rf $(doc_build_dir) .tox
 	@echo 'Done: Removed everything to get to a fresh state.'
 	@echo '$@ done.'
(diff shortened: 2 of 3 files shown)


makefile
```diff
@@ -213,7 +224,11 @@ doccoverage:
 	@echo '$@ done.'
 
 .PHONY: check
-check: pylint.log
+check: pylint.log flake8.log
+	@echo '$@ done.'
+
+.PHONY: flake8
+flake8: flake8.log
 	@echo '$@ done.'
```
andy-maier: If you really thing we should have a `flake8` virtuial target in addition to the `check` target, then please add it to the `help` target.

makefile
```diff
@@ -291,6 +306,16 @@ else
...
 
+flake8.log: makefile $(flake8_rc_file) $(flake8_py_files)
+	rm -f flake8.log
+	-bash -c "set -o pipefail; PYTHONPATH=. flake8 --config=$(flake8_rc_file) $(flake8_py_files) 2>&1 |tee flake8.tmp.log"
+	mv -f flake8.tmp.log flake8.log
+	@echo 'Done: Created flake8 log file: $@'
+
+.PHONY: check
+check: flake8.log
+	@echo '$@ done.'
```
andy-maier: There is already a `check` target further up, with both log files.

docs/intro.rst
```diff
@@ -63,11 +63,11 @@ Installation
...
-However, some of the Python packages used by pywbem have dependencies on
-operating-system packages (e.g. M2Crypto, or some packages when developing on
-pywbem). As of pywbem v0.8, its setup script has support for installing
-prerequisite OS-level packages, and installing (or in some cases, building)
-Swig, if needed. Swig is needed during installation of the M2Crypto package.
+Some of the Python packages used by pywbem have dependencies on
+operating system packages (e.g. M2Crypto, or some packages when developing on
+pywbem). As of pywbem v0.8, the setup script has support for installing
+prerequisite operating system packages, and building them if needed (e.g. Swig
+which is needed during installation of the M2Crypto package).
```
andy-maier: This change should not be in this PR (I have that change already in PR #509, which is merged).

flake8.rc
```diff
@@ -0,0 +1,19 @@
+[flake8]
+max_line_length = 80
+ignore =
+    # Ignore extra blank lines
+    E302,
```
andy-maier: I suggest to enable E302. It does not hurt, once you stick to the blank line rules.

flake8.rc
```diff
@@ -0,0 +1,19 @@
...
+    # the backslash is redundant between brackets
+    E502,
```
andy-maier: Suggest to enable E502.

flake8.rc
```diff
@@ -0,0 +1,19 @@
...
+    # E303 too many blank lines
+    E303,
```
andy-maier: Suggest to enable E303.

flake8.rc
```diff
@@ -0,0 +1,19 @@
+[flake8]
+max_line_length = 80
+ignore =
```
andy-maier: I suggest to enable all these checks. If not now, then in a second PR that fixes them.
None of them hurts, and all of them lead to more consistent code, if only optically.

flake8.rc
```diff
@@ -0,0 +1,19 @@
...
+    # E265 block comment should start with #
+    E265,
+    # E128 continuation line under-indented for visual indent
+    E128,
+    # E261 at least two spaces before inline comment
+    E261,
+    # E203 whitespace before ':'
+    E203,
+    # E231 missing whitespace after ':'
+    E231
```
andy-maier: I think we want to exclude some files, by adding these lines:

```
exclude =
    .git,
    .tox,
    __pycache__,
    *.pyc,
    attic,
    packaging,
    build_doc,
    docs,
    dist,
    pywbem.egg-info,
```

makefile
```diff
@@ -109,6 +109,17 @@ pylint_py_files := \
...
 
+# flake8 config file
+flake8_rc_file := flake8.rc
+
+# PyLint source files to check
+flake8_py_files := \
+    setup.py \
+    os_setup.py \
+    $(filter-out $(moftab_files), $(wildcard $(package_name)/*.py)) \
+    $(wildcard testsuite/test*.py) \
```
andy-maier: We have files in `testsuite` that do not start with `test*` and that should be checked. Change `test*.py` to `*.py`.

andy-maier:
see my comments.

makefile
```diff
@@ -291,6 +306,16 @@ else
 	@echo 'Info: Pylint requires Python 2; skipping this step on Python $(python_major_version)'
 endif
 
+flake8.log: makefile $(flake8_rc_file) $(flake8_py_files)
+	rm -f flake8.log
+	-bash -c "set -o pipefail; PYTHONPATH=. flake8 --config=$(flake8_rc_file) $(flake8_py_files) 2>&1 |tee flake8.tmp.log"
```
andy-maier: Removing the minus from the bash command would mean we enable flake8 checking for real. I suggest we do that.

makefile
KSchopmeyer: Not really needed. I just put that in to test. Removed

docs/intro.rst
KSchopmeyer: I sure blew that one. Not sure how that got in. Will redo.

flake8.rc
KSchopmeyer: Trying for compatibility now and we start to remove as we fix things.

flake8.rc
KSchopmeyer: Done

makefile
KSchopmeyer: Ouch. That is a direct copy from pylint_py_files so should change that one also or simply change it to line_py_files for both

makefile
KSchopmeyer: I propose to make this last one a separate issue since it is currently finding a number of things. Lets get the function in then make changes to the checks, etc.  Note that today, we also see a lot of errors in flake8 that are disabled in pylint.  To disable these we will have to go back and do disable in the the code for lots of things.

andy-maier:
rebased it and squashed all commits into one.

makefile
andy-maier: Agree.

flake8.rc
andy-maier: Agreed

andy-maier:
+1 from me - merging.

PR #521, opened 2016-09-23
Added branch to readme button for coverage.

andy-maier:


andy-maier:
+1. Merging because it is really minor.

PR #525, opened 2016-09-25
Appveyor CI - Fixed installation of lxml install prereqs and enabled tests.

andy-maier:
After many trial&error attempts, this PR is now ready for merge. Please review.

This PR fixes the installation of the Python lxml package on Windows in the Appveyor CI environment by manually downloading the binary libraries needed for the lxml build that happens during its Python installation.

It also disables the test case with sending 1000 indications in `testsuite/test_indicationlistener.py`, because that takes about 30min. Not sure what causes this slowness. The Appveyor CI environment is generally very fast (generally faster than Travis CI, it seems).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,16 +2,22 @@
 environment:
   matrix:
     - PYTHON: C:\Python27
-      PYTHON_VERSION: 2.7.8
+      PYTHON_VERSION: 2.7.12
       PYTHON_ARCH: 32
 
 install:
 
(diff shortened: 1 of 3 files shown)


andy-maier:
+1.
Suggest to merge this quickly in order to give it more exposure. We can always quickly disable it again in case of problems.

KSchopmeyer:
+1. I filed an issue on the speed of the 100 indications test.

PR #527, opened 2016-09-25
fixes #523:  flake8 cleanup of issues and more ignores

### LINKED ISSUE #523: pyflake returns way to many errors
We need to get the pyflake list of reported issues much lower than it is before we can consider trying to enable it as a test fail script.  This issue cleans up a number of issues in the code, and adds some more the the ignore list.

Note: While we have tried not to change the code any more than necessary, we have make a few code changes to fix issues.

There are still more cleanup that we can do to reduce flake8 issue list but it is already lower than pylint (but does not catch the TODOs, etc. so the comparison is unfair.

NOTE: Also added statistics to the flake8 log output.

KSchopmeyer:
Make a number of minor changes including both editorial changes to eliminate flake8 issues and some actual code changes.

Ready For Review:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,6 +31,9 @@ Known Issues
 Enhancements
 ^^^^^^^^^^^^
 
+* add flake8 as a lint tool.  It is executed with makefile check target.
+  see issues $512, cd
+
 * Improved usage information for Tutorial section, to make usage of Jupyter
   tutorials more obvious.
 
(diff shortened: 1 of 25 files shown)


KSchopmeyer:
-1. I have one more change to add

KSchopmeyer:
+1  Note that I have converted most of the from cim_constants import * to a list of variables.  This is because pflake8 reports these as questionable.   Also, in general, the pythonic way discourages the use of the * on import.

andy-maier:
+1

PR #529, opened 2016-09-26
Fixes #483: Added check for referencing subscriptions in remove filter + destination

### LINKED ISSUE #483: Check for referencing subscriptions before removing filters or destinations
The `remove_filter()` and `remove_destinations()` methods of `WBEMSubscriptionManager` currently first perform the DeleteInstance operation against the specified instance path, and then remove the instance path from the instance variable for owned filters or destinations.

There is not currently any check for subscriptions that reference the filter or destination to be removed.

If the WBEM server rejects the DeleteInstance for the filter or destination instance in case of referencing subscription associations, no check is needed on the client side: The DeleteInstance will raise a `CIMError` exception, which will percolate up. Because the DeleteInstance is the first thing that is done in the removal method (besides checking the WBEM server for being registered), this causes the removal method effectively to not change anything, and we are good.

However, if the WBEM server allows successful deletion of a filter or destination instance while referencing subscription instances are present, this would leave the referencing subscriptions with a broken reference. That is a state that should be avoided.

In order to check for that, the removal methods would have to find out whether there are referencing subscriptions in the WBEM server. This can be done with a ReferenceNames operation with the source instance being the instance to be deleted, and its result filter being the subscription class name. That operation returns any subscriptions referencing the instance to be removed. Given that there is anyway an interaction with the WBEM server for the deletion, this one additional operation is probably not a problem.

Just checking the instance variables for owned subscriptions referencing the owned filter or destination to be removed is not sufficient, because a user of another subscription manager that tries to reuse existing destinations or filters would find them based upon their instance in the WBEM server and would consider them to be not-owned.

So the main question is: Can we assume that WBEM servers are properly implemented and refuse deletion of a filter or destination instance when there are referencing subscription instances?

andy-maier:
Please review.

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -151,6 +151,8 @@ from ._server import WBEMServer
 from ._version import __version__
 from .cim_obj import CIMInstance, CIMInstanceName
 from .cim_http import parse_url
+from .cim_constants import CIM_ERR_NOT_FOUND
+from .exceptions import CIMError
 
 # CIM model classnames for subscription components
 SUBSCRIPTION_CLASSNAME = 'CIM_IndicationSubscription'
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1

PR #530, opened 2016-09-27
fixes #401: Modify run_cimoperations.py to rename yaml file before each test

### LINKED ISSUE #401: Should we set up separate file for run_cim_operatons for each test
When Andy committed the recorder changes part 2 a simple extension was added to run_cim_operations.py

Couple of questions:

1. Thought we were going to do a file per test??? since if anybody really uses this on the full test, The file will be enormous.
2. Should we be deleting existing file before running tests.
3. Should we find some naming combo that allows us to assure files are ignored by git and we do not accidentally get one in the repository (i.e. a defined prefix)???

KSchopmeyer:
Please Review

Issue #401 brought up the question of whether we should clean out the yaml file before each use (with some other issues).  However, the conclusion is that there is not an issue in the recorder itself since it starts with an open file.

Previously run_cimoperations.py  just appended to an existing file making the tests messy.  Note that the issue itself now requires no change to _recorder.py

I used this pr to fix the yaml file lifecycle in run_cimoperations.py  and as an easy way to close the issue.  I don't think there is any issue with the r_recorder itself but we were not cleaning out the old yaml file in cim_operations so added code to create a backup if the file exists so we start each run_cim_operations.py test with an empty yaml file.

Note that this does not create a separate file for each test but one file for the complete execution of the test program.

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -12,6 +12,7 @@ from __future__ import absolute_import
 
 # pylint: disable=missing-docstring,superfluous-parens,no-self-use
 import sys
+import os.path
 import threading
 from datetime import timedelta
 import unittest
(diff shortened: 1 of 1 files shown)


andy-maier:
+1

KSchopmeyer:
+1

PR #532, opened 2016-09-28
Added severity-sensitive return code checking for PyLint in makefile.

andy-maier:
Please review.

The change causes the PyLint make rule to fail for error messages and fatal messages (and invocation errors). other types of messages are ignored.

At this point, the failure is ignored (using a dash with the make command), so this change is just a preparation for when we want to activate it.


andy-maier:
+1

KSchopmeyer:
+1

Issue #540, opened 2016-10-04
Filter Name must be user-definable


andy-maier:
Some management profiles require certain values for the `Name` property of indication filters. 

However, the current implementation of the subscription manager sets the `Name` property to a certain format.

andy-maier:
This is what DSP1054 1.2.2 says about the `Name` property of `CIM_IndicationFilter`:

---------------------------

**7.3.11.3.2 Property: Name**

The value of the Name property shall be the name of the indication filter; it shall be formatted as defined by the following ABNF rule:

```
    OrgID ":" RegisteredName ":" UniqueID
```

`OrgID` shall identify the business entity owning the referencing profile. `OrgID` shall include a copyrighted, trademarked, or otherwise unique name that is owned by that business entity or that is a registered ID assigned to that business entity by a recognized global authority. In addition, to ensure uniqueness, `OrgID` shall not contain a colon (:). For referencing profiles owned by DMTF, `OrgID` shall match "DMTF".

`RegisteredName` shall be the registered name of the referencing profile, as defined by the value of the RegisteredName property in the RegisteredProfile instance representing the implemented version of that profile.

`UniqueID` shall uniquely identify the represented indication filter within the referencing profile.

**DEPRECATED (begin)**

For compatibility with version 1.0 of this profile, referencing profiles owned by business entities other than DMTF may in addition define values for the Name property that are formatted as defined by the following ABNF rule:

```
    OrgID ":" UniqueID
```

Where:
* `OrgID` is defined above in this subclause.
* `UniqueID` shall uniquely identify the instance within the business entity owning the referencing profile.

Version 1.1 of this profile has deprecated this additional format.

**DEPRECATED (end)**

---------------------------

andy-maier:
Karl to find out how SMI-S treats the `Name` property.

andy-maier:
It seems we do not have any place in a filter where we can safely store a pywbem-specific identification:
- The `Name` property may be determined by profiles
- The other key properties (`SystemName` and the two class names) also have a defined semantics
- None of the other properties covered in the indications profile can be used

In theory we could use one of the properties defines on the class but not used in the indications profile, e.g. `ElementName`. The question is whether the server implementations support that.

KSchopmeyer:
As of smis 1.7 SMI places not place any requirements on the name property above any beyond DSP1054. All of the specific information on the characteristics of the name property were removed.

[...]

PR #541, opened 2016-10-04
Fixes #461: Recovering owned destination, filter and subscriptions upon server add.

### LINKED ISSUE #461: Recover owned instances when restarting subscription manager
When a Python program that uses a `SubscriptionManager` instance dies, there may have been CIM instances created that were owned by the `SubscriptionManager` instance. Currently, these instances are neither cleaned up while dying, nor are they rediscovered when the subscription manager restarts.

It would be good to come up with an approach by which a subscription manager could rediscover existing CIM instances on a server that were owned, and again consider them owned instances.

The difficulty to resolve is how the new `SubscriptionManager` instance can safely be considered a restarted old one? Maybe via the `subscription_manager_id` value? Do we need to prevent that two `SubscriptionManager` instances have the same `subscription_manager_id` value and thus both assume ownership of existing previously owned instances?

andy-maier:
This PR does not include a testcase.

Please review.

Details:
- Inserted an indicator for ownership into the `Name` property of filters and destinations (`"owned"` vs. `"permanent"`).
- Made subscription manager ID mandatory (must not be None), because it is now used for recognizing owned instances.
- Made filter ID mandatory for the same reason.
- Added code in `add_server()` to determine owned destinations and filters by matching their `Name` property, and to determine owned subscriptions if they reference an owned filter or an owned destination.

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -143,6 +143,7 @@ Another more practical example is in the script
 project).
 """
 
+import re
 from socket import getfqdn
 import uuid
 import six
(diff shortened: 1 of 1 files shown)


andy-maier:
We need to discuss the approach, as well as issue #540.

KSchopmeyer:
We concluded that we may have one significant problem with this in that profiles may set requirements on the Name property. Andy opened issue

andy-maier:
Issue #540 describes that remaining problem, and we agreed that we would handle that in a separate PR.

So this PR is ready to be merged (whtout adressing issue #540). Please review and vote.

andy-maier:
+1

KSchopmeyer:
+1

PR #543, opened 2016-10-05
fixes #542 redundant backslash flake8

### LINKED ISSUE #542: remove more flake8 msgs.
This issue is specifically to remove the "redundant backslah' flake8 message.

KSchopmeyer:
Ready for review

Removes a large number of flake8 messages.  Also fixed a few formatting issues noted by pylint but this did not really change pylint message count much.  Note that the current number of messages from flake8 is about the same but we now enabled the E502 error (there were over 400 of them)

This is completely aesthetic but makes the code cleaner.

the flake 8 message log summary is now:

3     E131 continuation line unaligned for hanging indent
33    E501 line too long (115 > 80 characters)
1     F401 '._version.__version__' imported but unused
2     F821 undefined name 'ResourceWarning'
1     F841 local variable 'insts_enum' is assigned to but never used
2     W601 .has_key() is deprecated, use 'in'

A number of these I just do not know how to resolve because of the way flake8 works..  In particular the line_too long messages are primarily a) table in cim_operations which is completely within a comment b) some long lines in documentation due to the specification of attribute typing that must go on a single line.

Since the standard way to ignore issues in flake 8 is to put on the same line '  # noqa Exxx' this does not work within docstrings.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -44,10 +44,7 @@ ignore =
     # E402 module level import not at top of file
     # This occurs primarily because of conditonal imports including six
     # ex. if six...
-    E402,
-    # E502 the backslash is redundant between brackets. Remove when cleaner
-    # Right now about 500 of these messages
-    E502,  
+    E402, 
     # E265 block comment should start with #
     E562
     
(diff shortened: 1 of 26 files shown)


andy-maier:
We agreed in the call today that we would handle the remaining issues separately.

+1 for this PR, merging.

PR #544, opened 2016-10-05
Fixes #520: Resolved all flake8 issues and fixed a string formatting error in add_server()

### LINKED ISSUE #520: Reduce pylint and flake8 issues to where we can successfully test return code
The release is mandatory is for flake8 and any non-zero returns.

In pylint, the return code returned is as follows:

   Output status code:
      Pylint should leave with following status code:
     \* 0 if everything went fine
     \* 1 if some fatal message issued
     \* 2 if some error message issued
     \* 4 if some warning message issued
     \* 8 if some refactor message issued
     \* 16 if some convention message issued
     \* 32 on usage error
     status 1 to 16 will be bit-ORed so you can know which different
     categories has been issued by analysing pylint output status code

So far I have not identified what the return codes are for flake8 other than it appears to return 1 for any issue (informal documentation)

andy-maier:
WHOOPS. Sorry. Karl forgot to take off the review needed last week +1
Activated all checks in .flake8 and resolved all issues.
Found one error in string format arguments in `WBEMSubscriptionManager.add_server()`.
Removed the ignoring ot the flake8 check in the makefile.

Please review.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -17,34 +17,4 @@ exclude =
 #disable_noqa
     
 ignore =
-    # E126 continuation line over-indented for hanging indent
-    # Currently does not appear to match our style
-    E126,
-    # E127 continuation line over-indented for visual indent.
-    # Disabled because does not appear to match pylint rules
-    E127,
-    # E128 continuation line under-indented for visual indent
-    E128,
-    # E261 at least two spaces before inline comment
-    # Not sure it matches our style and generates many of these messages
-    E261,
-    # E265 block comment should start with '# '
-    # over 100 of these messages right now
-    E265,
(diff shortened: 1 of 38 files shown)


KSchopmeyer:
_subscription manager line 1135. I really wish that we could push to get rid of the line continuation  backslash in statements like this

couldn't we resolve even statements like:

                duration_sec = (float(duration.microseconds) / 1000000) + \
                    duration.seconds + (duration.days * 24 * 3600)

with paren around whole statement so it was interpreted as a single statement

I reviewed almost all of it, tested it in both pywbem 2 and 3 and ran run_cimoperations.py. No problems amd the above are just comments.

+1

PR #559, opened 2016-11-05
ERROR IN THIS PR Cancel: Fix formatting issue with argparse epilog

KSchopmeyer:
**I just realized that somehow I messed this one up and it changed 40 files..  Not sure why but Ignore it and will resubmit. Will remove pr once I understand what I did wrong.**

This fixes formatting of the argparse desc and epilog by adding a custom formatter that has multiple  formatters.  I am not sure how far I trust this fix in the future but found the proposal on stackoverflow and it is also recommended solution. in a python issue. That is documented in the code

In addition it:

- expanded the desc slightly.

- changes main and main call to get script name from args.

- fixes an issue with a missing parameter on one of the command definitions. This one popped up during testing with on pylint (this actually showed up after I had run a number of pylints that did not show the error) but it announced that the rc  attribute for ClassName did not exist. Fixed

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -17,34 +17,4 @@ exclude =
 #disable_noqa
     
 ignore =
-    # E126 continuation line over-indented for hanging indent
-    # Currently does not appear to match our style
-    E126,
-    # E127 continuation line over-indented for visual indent.
-    # Disabled because does not appear to match pylint rules
-    E127,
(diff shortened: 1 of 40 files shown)


KSchopmeyer:
I closed this one because it must have been a fingers error.

PR #561, opened 2016-11-05
fixes #558 Correct formating of argparse epilog

### LINKED ISSUE #558: wbemcli epilog not correctly formatted
Despite the fact that the epilog is correctly formated in the source, argparse removes all whitespace from the epilog and therefore the formatting when smart formatter is used.  Note that this is documented in the smart formatter.

KSchopmeyer:
Ready for review. Complete and two +1 votes.

This fixes formatting of the argparse desc and epilog by adding a custom
formatter that has multiple  formatters.  I am not sure how far I trust
this fix in the future but found the proposal on stackoverflow and it is
also recommended solution. in a python issue. That is documented in the
code

In addition it:

- expanded the desc slightly.

- changes main and main call to get script name from args.

- fixes an issue with a missing parameter on the openQueryInstances
operation definition in wbemcli. This issue popped up during testing
with on pylint (this actually showed up after I had run a number of
pylints that did not show the error) but it announced that the rc
attribute for ClassName did not exist.

There may be more effective ways to do this than force wbemcli to create a
custom formatter but at least this one works.

### DIFF
--- a/docs/wbemcli.help.txt
+++ b/docs/wbemcli.help.txt
@@ -2,6 +2,11 @@ usage: wbemcli [options] server
 
 Provide an interactive shell for issuing operations against a WBEM server.
 
+wbemcli executes the WBEMConnection as part of initialization so the user can
+input requests as soon as the interactive shell is started.
+
+Use h() in thenteractive shell for help for wbemcli methods and variables.
+
 Positional arguments:
   server                Host name or url of the WBEM server in this format:
                             [{scheme}://]{host}[:{port}]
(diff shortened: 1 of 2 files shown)


andy-maier:
If that formatter works, I'm fine with it.
+1.
Please resolve the merge conflicts.

KSchopmeyer:
Not really happy about it since it is really a mixin  of two formatters, ours and theirs but it was easier for now than trying to get their code into our smartformatter for desc, epilog, etc.

+1

andy-maier:
It now fails...

andy-maier:
Karl,
this is also caused by the new version of flake8. PR #574 fixes that, please rebase this PR to the latest master branch, and it should be resolved.

Issue #498, opened 2016-09-07
Memory inefficiency with DOM


willspearman:
## Problem:

The DOM is extremely inefficient for converting XML serially to objects. For large responses in memory constrained environments, this causes MemoryError exceptions.
## Setup:
- pywbem-0.9.0
- XML with 18k small objects (15MB)
- mock script
## Result:
- 1500 MiB peak memory usage
- 100 seconds runtime

Run the attached script like:

`python -m memory_profiler ./pywbem_memory_demo.py pywbem_memory_demo.xml`

[pywbem_memory_demo.zip](https://github.com/pywbem/pywbem/files/459650/pywbem_memory_demo.zip)

Example script execution:

```
(pywbem-released) [silo2000 pywbem_memory_demo (master)]$ time python -m memory_profiler ./pywbem_memory_demo.py pywbem_memory_demo.xml
Found 18000 results.
Filename: ./pywbem_memory_demo.py

Line #    Mem usage    Increment   Line Contents
================================================
    21   26.520 MiB    0.000 MiB   @profile
    22                             def main():
    23   26.520 MiB    0.000 MiB       conn = pywbem.WBEMConnection('http://10.10.10.10',
    24   26.520 MiB    0.000 MiB                                    ('None', 'None'), default_namespace='root/emc')
    25   26.520 MiB    0.000 MiB       conn.debug = False
    26
    27   26.520 MiB    0.000 MiB       instances = ['Clar_StorageVolume']
    28
    29   26.523 MiB    0.004 MiB       patch = mock.patch(pywbem_patch)
    30   26.527 MiB    0.004 MiB       pmock = patch.start()
    31
    32   26.527 MiB    0.000 MiB       with open(sys.argv[1], 'r') as fh:
    33   41.723 MiB   15.195 MiB           response = fh.read()
    34
    35   41.723 MiB    0.000 MiB       pmock.return_value = response
    36
    37   41.723 MiB    0.000 MiB       parent = pywbem.CIMInstanceName('Foo')
    38 1584.145 MiB 1542.422 MiB       for i in instances:
    39 1584.145 MiB    0.000 MiB           results = conn.Associators(parent, ResultClass=i)
    40 1584.145 MiB    0.000 MiB           print 'Found %s results.' % len(results)



real       1m41.468s
user       1m38.771s
sys        0m1.732s
(pywbem-released) [silo2000 pywbem_memory_demo (master)]$
```
## Suggestion:
- POC with SAX parsing shows:
[...]

PR #562, opened 2016-11-06
fix issue #498 Replace the DOM XML parser with sax parser 

KSchopmeyer:
Ready for review: DOM removed everywhere but the prettyprinter.  See below for more details. Passes tests. Karl voted positive on current push.

I took Will's fork, and worked it a bit to clear up any outstanding issues with pylint, and flake8.  Since I could not commit this back to Will's repository I created a new branch in our repository. Also added note to changes.rst

CONCLUSION: We really must format for the debug output to make it useful.  However, this means either running everything through the dom for debug or finding another prettyprinter for xml

DISCUSSION: prettyprinter, see my proposal below

This change significantly improves efficiency of response parsing, in particular memory usage during response parse. This completely removed the DOM parser.


willspearman:
Regarding the `re.match`, that line is there to make the current output white-space compatible with the old output so that the unit test check for no regressions works.

If we truly don’t need the old DOM parser anymore, then the unit test’s only purpose is obsolete. If you remove the DOM code, you could delete the unittest and delete the line patching up the whitespace.

As for the pretty printing, I don't think there is any cheap option for replacing this as that is a benefit of DOM.

Thanks so much for handling the merge and cleanup.

KSchopmeyer:
Karl:

1. Sort out Will's comment.

2. Propose prettyprint alternatives.

andy-maier:
I think the old minidom-based implementation should be removed completely. This translates to:
* In `_listener.py`, change `ListenerRequestHandler.parse_export_request()` to use the sax-based implementation.
* In `tupleparse.py`, change `parse_embeddedObject()` to use the sax-based implementation.
* In `tupletree.py`, remove `xml_to_tupletree()` and `dom_to_tupletree()`.
* Update `test_tupleparse.py` and `test_tupletree.py` accordingly.

andy-maier:
On pretty-printing request and response XML when debug is on: Can we try to do that based on sax?

KSchopmeyer:
Found the issue here.  Support for str type was added in python 3.5 but does not exist in python 3.4.  Thus the checks failures with python 3.4.

KSchopmeyer:
Note that the change to protect xml_to_tuple from str inputs (problem with python 3.4) is placed in several places, not just in the function itself to minimize how much it is used.  It is in the embedded_object handler and the test_tupleparser.py.  Thus, the main stream of responses assume that utf-8 is received, a logical solution.

Andreas: Feel free to comment on this one.

Second, we still use the dom for the prettyprint.  That will take more work;

**DISCUSSION:** Should we leave the dom implementation since there appears to be no readily available alternative so we would have to write our own.  Remember that this is used only when WBEMConnection debug is used.

KSchopmeyer:
The one failure wS a pip error. However, in retrospect I want to move the test for valid xml to xml_to_tuple. Simpler and clearer. Also not much real overhead.

KSchopmeyer:
+1. I am happy at this point. 

**DISCUSSION:**  Note that the one outstanding issue is the prettyprint for debug output. It still uses the minidom.  However, I propose we accept that for now and write separate issue to write a new prettyprinter based on the sax processor.

andy-maier:
Created issue #579 to capture the discussion item in the previous comment.

PR #566, opened 2016-11-08
WIP fixes #563 add wbemcli script capability.

### LINKED ISSUE #563: Expand WBEMCLI to execute a script as part of startup
There could be cases where users of wbemcli want to setup an environment beyond what we do as standard as part of the startup of wbemcli or execute specific python code before the interactive console is made available to the user.  As an example today, set up a server object but that is just one simplistic environment.

This could involve executing imports, creating new functions or classes, and possibly executing additional functions/methods before the user gets control in the wbemcli interactive console.

Python has several components that interact with the python environmnent including to include/execute, etc. code dynamically:

**exec statement** - supports dynamic execution of Python code. if the optional parts are omitted, the code is executed in the current scope. If only the first expression after in is specified, it should be a dictionary, which will be used for both the global and the local variables. If two expressions are given, they are used for the global and local variables, respectively.

**execfile statement** - This function is similar to the exec statement, but parses a file instead of a string. It is different from the import statement in that it does not use the module administration — it reads the file unconditionally and does not create a new module.

**import** - The import statement combines two operations; it searches for the named module, then it binds the results of that search to a name in the local scope.

either exec or execfile seem to do what I think I want.

I did the work and created a PR.  Have that pr up for discussion

KSchopmeyer:
Ready for final review: waiting on pr 576 to fix checks failure.

WIP but the current code works and there is a first example of a script. that does work.  The first script just sets up the wbemserver object including a valuemap so that the user has all that available.

I created some examples and put them into examples.  The script test-wbemcli-script.sh executes a simple example that does not require a running server (just displays the args for wbemcli).  Note that I made args into ARGS global to do this for this demo.

This could be expanded to actually do something useful and we could have any number of example scriptlets if we thought this was useful. Remember at this point this is prototype.

Further, since it will allow executing the script and quitting so that it never actually starts the interactive console is could be used to, for example, create a very simple script to show namespaces simply by having a scriptlet that did what the example does, then executed pp(SERVER.namespaces) and then quit.

If this is logical some of the questions include:

1. If we do scriptlets where should they go since I assume that they should be part of the release. Examples is just a temp holding place.
2. How much documentation does this need to be usable for pywbem users.

### DIFF
--- a/docs/wbemcli.help.txt
+++ b/docs/wbemcli.help.txt
@@ -68,6 +68,12 @@ Connection security related options:
                         then be part  of the certfile
 
 General options:
+  -s [scripts [scripts ...]], --scripts [scripts [scripts ...]]
+                        Execute the python code defined by the script before the
+                        user gets control. This argument may be repeated to load
+                        multiple scripts or multiple scripts may be listed for a
+                        single use of the option. Scripts are executed after the
+                        WBEMConnection call
   -v, --verbose         Print more messages while processing
   -h, --help            Show this help message and exit
 
--- a/examples/test_wbemcli_script.sh
+++ b/examples/test_wbemcli_script.sh
@@ -1,0 +1,10 @@
+#
+#  Demo of python scripts with wbemcli
+#
+#  This pair of scripts just starts wbemcli, displays the command line
+#  arguments args and exits wbemcli
+#  It works as a simple demo because it does not actually call the server
+#  with an operation so does not understand if there is actually no
+#  running server.
+#
+wbemcli http://localhost -s wbemcli_display_args.py wbemcli_quit.py
--- a/examples/wbemcli_display_args.py
+++ b/examples/wbemcli_display_args.py
@@ -1,0 +1,5 @@
+"""
+    Display the cli input arguments with their names
+"""
+attrs = vars(ARGS)
+print('\n'.join("%s: %s" % item for item in attrs.items()))
--- a/examples/wbemcli_quit.py
+++ b/examples/wbemcli_quit.py
@@ -1,0 +1,5 @@
+"""
+Pywbem scriptlet to quit wbemcli
+"""
+print('Quit Scriplet. Quiting wbemcli')
+quit()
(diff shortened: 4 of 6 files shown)


andy-maier:
Please resolve the merge conflict.

andy-maier:
Otherwise, +1.

KSchopmeyer:
I am ready to commit this but still have the open question about where we should put things like the example scripts since examples do not get distributed with the release.

It appears that the problem causing the fail is from flake8 and specifically

setup.py:537:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/cim_xml.py:90:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/cim_http.py:121:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/mof_compiler.py:269:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/mof_compiler.py:282:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/mof_compiler.py:289:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/mof_compiler.py:305:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/mof_compiler.py:1909:1: E305 expected 2 blank lines after class or function definition, found 1
pywbem/_recorder.py:79:1: E305 expected 2 blank lines after class or function definition, found 1
testsuite/test_cim_http.py:350:1: E305 expected 2 blank lines after class or function definition, found 1
testsuite/run_cim_operations.py:3659:1: E305 expected 2 blank lines after class or function definition, found 1
testsuite/run_cim_operations.py:3886:1: E305 expected 2 blank lines after class or function definition, found 1
testsuite/validate.py:86:1: E305 expected 2 blank lines after class or function definition, found 1
13    E305 expected 2 blank lines after class or function definition, found 1
make: *** [makefile:316: flake8.log] Error 1
ERROR: InvocationError: 'C:\\cygwin\\bin\\make.EXE check'
___________________________________ summary ___________________________________
ERROR:   pywin: commands failed

The only think I do not understand is why this was not caught on the pr where the changes were made.

andy-maier:
Karl,
this is caused by a new version of flake8. PR #574 fixes that and has been merged just now.

So please rebase again to the latest master branch, and it should be resolved.

KSchopmeyer:
+1

PR #549, opened 2016-10-20
Added description of prereq OS packages.

andy-maier:
PLease review. Ready for merge.

PR #573, opened 2016-11-13
Added description of prereq OS packages.

andy-maier:
This PR rolls back the change from PR #549 into pywbem v0.9.


andy-maier:
Because PR #549 has already been reviewed, I will merge this without further review.

PR #574, opened 2016-11-15
Fixes #560: Added scripts to flake8 and pylint checks; fixed flake8 erros in mof_compiler.

### LINKED ISSUE #560: flake8 does not run against scripts without a .py file suffix
I only noticed that it appears to do nothing when I call it from within geany so far.  But I noted some issues in wbemcli that flake8 should have caught and it did not catch them in make all.

Note in particular that the \ on help lines is still in the code and flake8 is not complaining.

Confirmed it does not run against wbemcli: Any files without extensions are simply ignored.  Even the command

    flake8 wbemcli 

which produces no results and if the -v option is set says it is checking 0 files.

copying wbemcli to wbemcli.py

produces lots of flake8 errors.

**DISCUSSION:**  To date the only proposed solution is to temporarilyu copy the scripts to a corresponding *.py file before executing flake8 in the makefile and then remove the temp copy.

Should we do this?

andy-maier:
Ready for merge.

### DIFF
--- a/makefile
+++ b/makefile
@@ -95,22 +95,17 @@ doc_dependent_files := \
 # PyLint config file
 pylint_rc_file := pylintrc
 
-# PyLint source files to check
-pylint_py_files := \
-    setup.py \
-    os_setup.py \
-    $(filter-out $(moftab_files), $(wildcard $(package_name)/*.py)) \
-    $(wildcard testsuite/*.py)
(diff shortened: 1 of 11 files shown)


andy-maier:
+1

KSchopmeyer:
+1

PR #581, opened 2016-11-21
Fixes #556: Added git as an OS-level dependency for development.

### LINKED ISSUE #556: CentOS 7 install of pywbem has prerequisites beyond defined list.
When installing pywbem into a newly created CentOS system we found that we had to install the following manually:

  - gcc
  - Python Development lib
  - git

Not sure if any of this should be part of the os_prerequisites for pywbem but thought it was worth the comment.

This is because the CentOS core is typically a minimal install and installs fewer components than many other distributions.

andy-maier:
Please review.


KSchopmeyer:
+1

PR #583, opened 2016-11-21
Fixes #407: Added context manager capability to WBEMSubscriptionManager

### LINKED ISSUE #407: Automatic cleanup of owned destinations, filters, subscriptions
Pywbem takes no part in any cleanup of subscriptions and filters that might be left in the server when the listener is stopped without removing these subscriptions/filters.

We do not have a way to identify them as created by our client, etc. so they could be eliminated subsequent at a later time.

At this point I just want to start the conversation since after a bunch of testing where we were terminating the test programs early we ended up with a bunch of outdated filters in the server all with the name cimfilter<uid>

andy-maier:
Also, corrected DSP1054 references in the docs of `WBEMSubscriptionManager`.

The only testcase is in `run_cim_operations.py`, and I did not test with it.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -81,6 +81,10 @@ Bug fixes
   version to <=2.0.8, due to its use of unittest.case which is not available
   on Python 2.6.
 
+* Fixed an `AttributeError` in the `remove_all_servers()` method of
+  `WBEMSubscriptionManager` and dictionary iteration errors in its
+  `remove_server()` method. PR #583.
+
 
 pywbem v0.9.0
 -------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -54,6 +54,10 @@ Enhancements
 * Added `git` as an OS-level dependency for development (it is used by GitPython
   when building the documentation).
 
+* Added enter and exit methods to `WBEMSubscriptionManager` to enable using it
+  as a context manager, whose exit method automatically cleans up by calling
+  `remove_all_servers()`.
+
 
 Bug fixes
 ^^^^^^^^^
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -143,7 +143,9 @@ as a context manager, in order to get automatic cleanup::
 
     with WBEMSubscriptionManager('fred') as subscription_manager:
         server1_id = subscription_manager.add_server(server1)
-    # exit method automatically calls remove_all_servers()
+        . . .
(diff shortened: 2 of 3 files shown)


andy-maier:
Karl, can you run the `run_cim_operations.py` test in your environment?

KSchopmeyer:
Comments:

1. Comment on the doc
    with WBEMSubscriptionManager('fred') as subscription_manager:
        server1_id = subscription_manager.add_server(server1)
    # exit method automatically calls remove_all_servers()

a. add a line with . . . for other work to be done

b. Add to comment on last line " which removes all owned filters, etc.

2. Question. Is there any reason to further docment what a context manager does either on line
181 (class description) or line 239 (the _enter_ function)?? I.e. are we going to get users who do
not clearly understand what a context manager is?

KSchopmeyer:
running run_cim_operation.py I get the following error:

ERROR: test_subscription_context_manager (__main__.PyWBEMListenerClass)
Test that the WBEMSUbscriptionmanager class can be used as a
  File "testsuite/run_cim_operations.py", line 3673, in test_subscription_context_manager
    sub_mgr.remove_all_servers()
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_subscription_manager.py", line 424, in remove_all_servers
    self.remove_server(server.url)
AttributeError: 'str' object has no attribute 'url'

NOTE: I only tested this shortly before the meeting and for some reason am getting issue with not enough indications received in some of the other tests but not  yet sure why.  However, those are count errors, this is a syntax/semantics issue.

andy-maier:
In addition, there is an error in `remove_server()` where it deletes the entries in the various dictionaries. Python does not allow deleting items in a dictionary while iterating over it. So the list of keys to iterate over needs to be instantiated into a separate list object that remains unchanged when deleting the items. The `list()` constructor is needed to make that work on Python 3, because `keys()` returns an iterator in Python 3. Fixed that as well.

andy-maier:
Karl,
this PR is now ready for a retest with `run_cim_operations.py`.

andy-maier:
+1

andy-maier:
Commit dff3989 is a fix that should be rolled back into the 0.9 stream.

andy-maier:
Rebased on master after merging PR #604, to fix the Sphinx build issue on Python 2.6.

andy-maier:
We agreed to merge this one, and to do the test as part of PR #607.

PR #584, opened 2016-11-21
Fixes #526: Moved compat_args.py from testsuite/ to try/

### LINKED ISSUE #526: testsuite/compat_args.py. Is there any reason to keep this in testsuite?
This is not really part of the test suite.  Should we move it to tools?

andy-maier:
As agreed for issue #526.
Please review


andy-maier:
+1

KSchopmeyer:
+1 My only confusion is the name of the directory try.  We agreed to that but in restrospect I am not sure what a directory named try really means.

andy-maier:
**Discussion point:** Directory name

andy-maier:
We used `try` only because that's my standard directory name for things to "try out". We discussed this again in the call today and because none of us was able to come up with a better name, we decided to go with `try`.

PR #586, opened 2016-11-23
Fixes #548: Fixes and improvements to global ops functions in wbemcli

### LINKED ISSUE #548: wbemcli does not accept optional attributes when abbreviation used
The following works

```
>>> x = oei('CIM_ManagedElement')
```

The following fails

```
>>> y = oei('CIM_ManagedElement', MaxObjectCount=10)
TypeError: OpenEnumerateInstances() got an unexpected keyword argument 'MaxObjectCount'
```

But the non-alias cmd works:

```
>>> y = CONN.OpenEnumerateInstances('CIM_ManagedElement', MaxObjectCount=10)
```

andy-maier:
This PR is now ready to be reviewed.

It contains many docstring changes, but also a few code fixes.

Details from the commit log:
- Removed long global function names (e.g. `EnumerateInstances`), and kept the short names (e.g. `ei`).
- Fixed missing `ns` parameter of `ci()`.
- Fixed incorrect invocation of underlying operation in several short functions:
  - `orip()` invoked `OpenReferenceInstances()`.
  - `oaip()` invoked `OpenAssociatorInstances()` and incorrectly had `iq`, `ico`, `pl` parameters.
- Fixed incorrect position of `rc` parameter in `oqi()` function.
- In `mc()`, removed returning the result of the underlying operation that doesn't return anything.
- Fixed errors in docstrings and improved the docstrings of the short functions.

### DIFF
--- a/pywbem/tupletree.py
+++ b/pywbem/tupletree.py
@@ -58,6 +58,7 @@ import xml.dom.minidom
 import xml.sax
 import re
 import sys
+import six
 
 __all__ = []
 
@@ -142,10 +143,11 @@ def xml_to_tupletree_sax(xml_string):
     """
     handler = CIMContentHandler()
 
-    # The following is required because python version 3.4 does not accept
-    # str for the sax processor xml input.
-    if sys.version_info[0:2] == (3, 4):
-        if isinstance(xml_string, str):
+    # The following is required because the SAX parser in these Python
+    # versions does not accept unicode strings, raising:
+    # SAXParseException: "<unknown>:1:1: not well-formed (invalid token)"
+    if sys.version_info[0:2] in ((2, 6), (3, 4)):
+        if isinstance(xml_string, six.text_type):
             xml_string = xml_string.encode("utf-8")
 
     xml.sax.parseString(xml_string, handler, None)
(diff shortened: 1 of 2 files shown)


andy-maier:
We got another Python 2.6 error. The changed packages since the last time it ran fine on Python 2.6, are:
- numpy 1.11.1 -> 1.11.2
- pytest 3.0.0 -> 3.0.4
- setuptools 12.0.5 -> 28.8.0
- wheel 0.29.0 -> 0.30.0a0

KSchopmeyer:
I agree except for the removal of the long form op names and that is primarily because the long form names make better documentation in the scripts.  I have now written several scripts already and I found that it was much easier to read with the long form names.

andy-maier:
There is something else: The docstrings of the global ops functions in wbemcli are not included in the generated documentation. They are available in the help system of wbemcli only.

Should we add them to the generated documentation?

On the long names: Should we add them as    aliases?

andy-maier:
We agreed to open an issue for adding the short global functions to the RTD documentation.

We agreed not to add the long global function names back in again, because we already have the choice between global short functions with short arguments, and the long named methods

PR #593, opened 2016-11-30
Fixed AttributeError and dict iteration errors in remove_all_servers and remove_server

andy-maier:
This rolls back a fix from PR #583 into v0.9.

Details:
- In remove_all_servers(), fixed an AttributeError (str has no 'url' attribute).
- In remove_server() and in remove_all_servers(), fixed dictionary iteration errors where dictionary items were deleted but the iteration was based upon an iterator over the dictionary. Fixed that by making the loop based upon a copied list of the dictionary keys.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -25,6 +25,10 @@ Enhancements
 Bug fixes
 ^^^^^^^^^
 
+* Fixed an `AttributeError` in the `remove_all_servers()` method of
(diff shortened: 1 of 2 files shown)


andy-maier:
Repushing this commit to trigger verification runs in order to find out whether `stable_0.9` is also affected from the Python 2.6 issue solved with PR #596.

KSchopmeyer:
+1

PR #596, opened 2016-12-01
Fixed SAX parse error with unicode strings on Python 2.6.

andy-maier:
This PR fixes the SAX parse error that showed up on Python 2.6 starting on 2016-12-01, by applying the same fix that was already implemented for Python 3.4: Encode the unicode string into a byte string.

So far, this fix is not needed on Python 2.7 and Python 3.5, for unknown reasons.

The symptom is that `test_tupleparse.py` fails in the test run on Python 2.6 with:

```
SAXParseException: "<unknown>:1:1: not well-formed (invalid token)"
```

If you encounter this, use this PR via one of these:
- rebase your PR on the branch of this PR (as long as it is not yet merged)
- rebase your PR on master (once this PR has been merged)
- cherry-pick the commit of this PR into your branch

### DIFF
--- a/pywbem/tupletree.py
+++ b/pywbem/tupletree.py
@@ -58,6 +58,7 @@ import xml.dom.minidom
 import xml.sax
 import re
 import sys
+import six
 
 __all__ = []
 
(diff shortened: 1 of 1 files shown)


andy-maier:
Apart from not understanding why this happened and why it does not happen on Python 2.7 or 3.5, I'll give this a +1.

KSchopmeyer:
+1

PR #607, opened 2016-12-09
Fixes #540, fixes #462: Several changes, fixes and improvements on WBEMSubscriptionManager

### LINKED ISSUE #462: Example with shutdown and eror handling of subscription manager
A need has been expressed to see such examples.

### LINKED ISSUE #540: Filter Name must be user-definable

andy-maier:
This PR changes a number of things in `WBEMSubscriptionManager`.

The main motivation was to make it easier for users to use it.

This PR should be reviewed in three ways:
- review the documentation (`make builddoc` -> review WBEMSubscription chapter), particularly `add_filter()` and the introduction text in chapter `WBEMSubscriptionManager` (i.e. the module docstring).
- review the tutorial (subscriptionmanager.ipynb)
- test the subscription manager code with `run_cim_operations.py`

Details on code changes:

- In order to support conditional creation of owned filters, destinations,
  and subscriptions, it became necessary to verify the properties of the
  instances, and not just their Name property or instance paths.
  This caused a number of changes, some of which are changes at the
  external interface of class `WBEMSubscriptionManager`:

  - Changed the internal instance variables that register owned instances, to
    hold the `CIMInstance` objects instead of the `CIMInstanceName` objects.
  - In all `_create_*()` functions, returned the instances instead of the
    instance paths, after retrieving the created instances using GetInstance in
    order to make sure we have the exact property values as created by the
    server (not just the ones we have passed in to CreateInstance).
  - In addition, these functions now use GetInstance to make sure the sub.mgr.
    has the correct properties in case the server implementation does not honor
    all properties provided in CreateInstance.
  - In addition, these functions now perform ModifyInstance for owned instances,
    when they find that an instance with the desired instance path exists.
  - In `add_*()` and `get_*()`, changed the return value to now return
    `CIMInstance` objects instead of the `CIMInstanceName` objects (or lists
    thereof).

- Added the possibility to directly specify a value for the `Name` property
  of indication filters.

- Added a `get_owned_destinations()` method that was missing so far, for
  consistency with similar methods for filters and subscriptions.

- Changed logic in `add_subscriptions()` to be recursive instead of iterative,
  for consistency with the other `add_*()` methods, particularly because that
  allows for lists witin lists (even though that is not documented).

- In `add_subscriptions()`, performed the check for not allowing permanent
  subscriptions on owned destinations for all destinations, before any
  creation is attempted. This makes the error behavior more transparent.

- In `remove_destinations()`, `remove_filter()`, and `remove_subscriptions()`,
  fixed a bug in the loop that iterates over the owned instances to delete
  the list items that match, that would have deleted the wrong item in case
  multiple items had matched.

- Removed default `None` from `subscription_manager_id` argument of
  `__init__()`, because it is stated and enforced to be not `None`.

Details on documentation related changes:

- Added a Jupyter notebook 'subscriptionmanager.ipynb' with a tutorial about
  the subscription manager.

- Improved the description of the ownership types in the module docstring.
  It now describes three types of ownership: owned, permanent, and static.

- Fixed and simplified the code example in the module docstring.

- Simplified the lengthy description of the `owned` argument in all methods
  that have it, based upon the better explanation in the module docstring.

- In the documentation of `get_owned_filters()` and
  `get_owned_subscriptions()`, removed the mentioning of raised exceptions,
  because these methods do not raise any exceptions.

Details on code quality changes:

- Removed pylint disable=line-too-long for some functions that meanwhile no
  longer needed it, and added it to some functions that missed it.

- Some clarifications in comments in the code.


andy-maier:
Rebased this PR on master after merging PR #583 (add context manager), so that `run_cim_operations.py` can now be run to test both of them.

KSchopmeyer:
1. See changes.rst. line 69. reated should be created. Still trying to figure out how to use the review stuff.

andy-maier:
+1

KSchopmeyer:
+  Owned CIM instances are created via the subscription manager and their life
+  cycle is bound to the life cycle of the registration of that WBEM server with
+  the subscription manager. 

The word registration is confusing.  I had to read it a couple of times to understand what it means. Sadly I don't have a better word right now, probably something like registration of that WBEM server
 via successful add.server(url).   Also on line 99 of _subscriptionmanager.py where there is an example of add, put in the comment register the wbemserver.

line 83 of subscription mgr
+The following example code demosntrates the use of a subscription manager
+to subscribe for a CIM alert indication on a WBEM server. The WBEM listener

spelling of 'demonstrates'

about line 130 of _subscriptionmanager.py
-``examples/pegIndicationTest.py`` (when you clone the GitHub pywbem/pywbem

The one difference is that the implication here is that the example demos the context manager and it doesn't

No big deal except I wonder if we should not be replacing all examples that create a submgr, do things and then release subscriptions, etc. and exit with the context manager to encourage use of the context manager. 

I have tried to go through the code, examples, etc. and to be honest, it is hard for me to detect issues in the code unless they are real design issues and I don't see any of those.  At the same time this needs more tests and writing the next set of tests probably comes after we commit this code.

This is good work and I think it is time to fix the minor issue, commit it and write tests.

I will vote ye
+The :ref:`Tutorial` section contains a tutorial about the subscription manager.

andy-maier:
Addressed your comments as described in your comment further up, and rebased the PR.

KSchopmeyer:
+1. Lets commit and then update any examples.  Much easier.

PR #610, opened 2016-12-10
Fixes #49: Added CIM Schema to repo to avoid downloading.

### LINKED ISSUE #49: CIM Schema gets downloaded on every Travis CI run
Which translates to twice per push, because for some reason, Travis runs two jobs for each push.

A possible solution would be to have the CIM Schema files in the PyWBEM git repo, probably best one by one (instead of the zip file).

andy-maier:


andy-maier:
+1

KSchopmeyer:
+1

PR #611, opened 2016-12-10
Fixed python devel package name for 3.4 and 3.5

andy-maier:
For RedHat 7, Python3.4 and 3.5 comes from two sources:
- package python34 from EPEL (or standard RHEL?)
- package34u from IUS community (supposedly better).

The problem is that for RedHat, the setup.py file specified a list of multiple choices (e.g. python34-devel, python34u-devel, python3-devel). However, the python34-devel works only with the python34 and python34u-devel only with python34u.

In order to be as standard as possible, this PR changes the setup.py file to pull in only python34-devel in case of RedHat/CentOS/Fedora, and assumes that python34 is installed.


andy-maier:
+1

KSchopmeyer:
+1

PR #612, opened 2016-12-11
Fixes #587: Added a check for unset URL target in recorder.

### LINKED ISSUE #587: Recorder issue with request exception
I believe this and #588 are duplicates.

In testing with run_cimoperations.py, I generated an openenumerateInstances exception by using a host name that was invalid.  This generated an exception (ConnectionError). However, we were using the recorder and the call to the recorder generated a second exception.
```
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 3915, in OpenEnumerateInstances
    **extra)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 866, in _imethodcall
    recorder=self.operation_recorder)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_http.py", line 684, in wbem_request
    raise ConnectionError('Error with file URL %s: %s' % (url, exc))
pywbem.exceptions.ConnectionError: Error with file URL localhost.py: [Errno 2] No such file or directory: 'localhost.  Note that we had not included the http:// so that the request was invalid.
```

It appears that the statement in _recorder.py at 410 & 411
```
            tc_http_request['url'] = TestClientRecorder.TESTCASE_URL + \
                http_request.target
```
cannot handle a NoneType in http_request.target. at this point.

During handling of the above exception, another exception occurred:

```
  File "testsuite/run_cim_operations.py", line 3191, in test_simple_iter_enum
    for inst in iterator:
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1859, in IterEnumerateInstances
    MaxObjectCount=MaxObjectCount, **extra)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 3923, in OpenEnumerateInstances
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 300, in record_staged
    self.record(pwargs, pwresult, httpreq, httpresp)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 411, in record
    http_request.target
TypeError: Can't convert 'NoneType' object to str implicitly
```

andy-maier:
Please review.
Also, it would be good to regenerate the connection error using run_cim_operations.py.

From the commit message:

Analysis of the issue:

After an exception (e.g. ConnectionError), the recorder still gets control
to do its thing. The `http_request.target` may be `None` in such a case.
However, the code assumed it could be concatenated with a string, which
fails in that case with:

```
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 411, in record
    http_request.target
TypeError: Can't convert 'NoneType' object to str implicitly
```

Solution:
- The `target` member will only be appended if not `None`.


andy-maier:
Per Karl's comment on PR #613, the solution for this issue seems to be verified (this commit is also part of PR #613).

Issue #588, opened 2016-11-30
yaml RepresenterError in TestClientRecorder for CIM objects


KSchopmeyer:
When executing one of the inovoke methods in run_cimoperations.py that returns an exception
we get the following issue from the recorder.
```
======================================================================
ERROR: test_all (__main__.InvokeMethod)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 4808, in InvokeMethod
    result = self._methodcall(MethodName, obj, Params, **params)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1112, in _methodcall
    raise CIMError(code, tt[0][1]['DESCRIPTION'])
pywbem.exceptions.CIMError: 16: CIM_ERR_METHOD_NOT_AVAILABLE: FooMethod

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "testsuite/run_cim_operations.py", line 1939, in test_all
    Ref=name)
  File "testsuite/run_cim_operations.py", line 126, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 4834, in InvokeMethod
    self.operation_recorder.record_staged()
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 300, in record_staged
    self.record(pwargs, pwresult, httpreq, httpresp)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 453, in record
    default_flow_style=False)
  File "/home/kschopmeyer/.virtualenvs/pywbem3/lib/python3.4/site-packages/yaml/__init__.py", line 216, in safe_dump
    return dump_all([data], stream, Dumper=SafeDumper, **kwds)
  File "/home/kschopmeyer/.virtualenvs/pywbem3/lib/python3.4/site-packages/yaml/__init__.py", line 188, in dump_all
    dumper.represent(data)
  File "/home/kschopmeyer/.virtualenvs/pywbem3/lib/python3.4/site-packages/yaml/representer.py", line 26, in represent
    node = self.represent_data(data)
[...]

PR #613, opened 2016-12-11
Improved diagnostics when PyYAML cannot represent an object.

andy-maier:
Karl tested. See below: I voted +1 on this based on the test.

This PR improves the diganostics for issue #588, but does not solve it.

Please reproduce the issue and report the results.

KS marked this update needed since I made comments

### DIFF
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -26,6 +26,7 @@ except ImportError:
     from ordereddict import OrderedDict
 from datetime import datetime, timedelta
 import yaml
+from yaml.representer import RepresenterError
 import six
 
 from .cim_obj import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
@@ -81,6 +82,18 @@ yaml.SafeDumper.add_representer(
     OrderedDict,
     lambda dumper, value:
     _represent_ordereddict(dumper, u'tag:yaml.org,2002:map', value))
+
+
+# Some monkey-patching for better diagnostics:
+def _represent_undefined(self, data):
+    raise RepresenterError("cannot represent an object: %s of type: %s; "
+                           "yaml_representers: %r, "
+                           "yaml_multi_representers: %r" %
+                           (data, type(data), self.yaml_representers.keys(),
+                            self.yaml_multi_representers.keys()))
+
+
+yaml.SafeDumper.represent_undefined = _represent_undefined
 
 
 class OpArgs(OpArgs_tuple):


KSchopmeyer:
This does correct the problem where we are dying if there is an exception generated once the connection is established with a yaml file.

However, if I try run_cimoperations.py with yaml file where the client catches the connect problem (ex.  just use blah as the name) I get the following error.

ERROR: test_nonexistent_namespace (__main__.EnumerateInstances)
Confirm correct error with nonexistent namespace.
  File "testsuite/run_cim_operations.py", line 733, in test_nonexistent_namespace
    namespace='root/blah')
  File "testsuite/run_cim_operations.py", line 140, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/pywbem-0.10.0.dev0-py2.7.egg/pywbem/cim_operations.py", line 1485, in EnumerateInstances
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/pywbem-0.10.0.dev0-py2.7.egg/pywbem/_recorder.py", line 334, in record_staged
    self.record(pwargs, pwresult, httpreq, httpresp)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/pywbem-0.10.0.dev0-py2.7.egg/pywbem/_recorder.py", line 458, in record
    data = http_request.payload.decode('utf-8')
AttributeError: 'NoneType' object has no attribute 'decode'

Finally, for whatever reason (since it always seemed to work before), if I put in a server url like http://blah, I now get a very long delay when using a url like http://blah.  But the good news is that the yaml output works.

KSchopmeyer:
I will go either way.  Commit this and write new issue on the problem where run_cimoperations.py --yamlfile bomb.yaml blah x%x causes an exception (i.e. exception in the client itself) and no file output and commit the current because it fixes the other issues.

or find remainder of issues.

andy-maier:
Karl, I have merged PR #612 which fixes a TypeError.
I was able to reproduce the new `AttributeError: 'NoneType' object has no attribute 'decode'` and will fix it.

andy-maier:
The latest commit fixes the `AttributeError`. Invoking:
```
testsuite/run_cim_operations.py --yamlfile bomb.yaml blah 
```
now results in:
```
ConnectionError: Error with file URL blah: [Errno 2] No such file or directory: 'blah'
```
which is expected.

Can you try to reproduce the yaml representation error with the CIMDateTime object? (I don't have a CIM server running yet in my new installation).

andy-maier:
Karl, I am reluctant to merge this without a test using `run_cim_operations.py` against a WBEM server, because the PR contains a monkey patch for PyYAML that improves the diagnostics.

The expected result would be that the issue reported in issue #588:
```
yaml.representer.RepresenterError: cannot represent an object: 00000060000000.000000:000
```
shows up again, just now with better diagnostics.

andy-maier:
With PR #621, this PR has been reduced to just the diagnostics improvements.
It still needs to be verified using `run_cim_operations.py`

KSchopmeyer:
The following was done by simply checking out your pr without starting from a fresh Centos7.

With the branch I first hit a problem with pip version, the version from the EPEL is VERY old (1.4.1)  but
a pip upgrade fixes that. Upgrades to 9.0.1.  Ran make all again and this time it went through the develop build, etc and finally died at the following:

Done: Created LEX/YACC table modules: pywbem/mofparsetab.py
rm -rf MANIFEST pywbem.egg-info .eggs
python setup.py sdist -d dist bdist_wheel -d dist --universal
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'bdist_wheel'
make: *** [dist/pywbem-0.10.0.dev0-py2.py3-none-any.whl] Error 1


Note: this means  that before you start you have to install pip which means getting the EPEL package since that is the source of all this.  So part of our setup definition for pywbem is to get pip and upgrade it.

The good news at this point is that it is going through the install process with the branch.

Here is what pip shows as installed.

andy-maier:
Please provide more details.
- Did this happen when you ran `make develop`?
- Were you using `sudo` or the root user?
- What is the output of `pip list` before you run this? (in the same environment you ran `make`)

KSchopmeyer:
Works.  I have another issue in that run_cim_operations.py goes through all of the tests (very rapidly) but it does generate yaml and not an exception.

andy-maier:
Given that we move issue #588 to after 0.10.0, we can as well move this PR here, to avoid the monkey patching it does in the released version.

KSchopmeyer:
I agree with your conclusion.

andy-maier:
Changed this to WIP and set discussion needed.

Discussion point: How do we go about this change. On the up side, it improves diagnostics. On the down side, it does this by monkey-patching a method in PyYAML.

andy-maier:
we agreed to merge it.

PR #614, opened 2016-12-11
Fixes #493: Added support to the operation recorder for an on/off switch.

### LINKED ISSUE #493: Add support for turning the operation recorder on/off
It is often necessary to turn the recorder on and off, if only a subset of the operations is to be recorded. It would be more convenient tohave a on/off switch on it, comared to having to deregister/register it.

andy-maier:
Please review.

Details:
- Added enable() and disable() methods to class BaseOperationRecorder, and changed the invocaiton of the record() method such that it only happens when the recorder is enabled.
- by default, the recorder is enabled.
- for completeness, added a property enabled that indicates whether the recorder is currently enabled.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -58,6 +58,8 @@ Enhancements
   as a context manager, whose exit method automatically cleans up by calling
   `remove_all_servers()`.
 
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
+1

PR #619, opened 2016-12-12
Fix # 616 Some issues with the latest subscription manager

KSchopmeyer:
Ready for review:

Fixed several issues in the new sub mgr and the tests for this in run_cimoperations.py

1. Correct code issues in subscription manager. See the issue for details of the fixes.
2. Correct some tests that were failing in run_cimoperations.py, mostly the move to use return instances rather than paths.
3. Minor extension to an example

Note that there is some crossover between the fixes to run_cimoperations.py and some other fixes to
cim_operations.py from the pr for issue #466 (implementing the iter functions). There may be a couple of conflicts which we will sort out when we try to merge.


andy-maier:
+1

PR #623, opened 2016-12-14
Two more working examples of scriptlets for wbemcli

KSchopmeyer:
Ready for review and commit:

I used these extensively in testing of the subscription manager changes.
Allowed me to see what was done and clean up the server easily.
At the same time they demo the weakness of the scriplets.

The scriptlets are:

1. View counts and paths for subscriptions, filters, destinations on a server
2. Remove all subscriptions, filters, destinations on a server.

Note: They do not use subscription manager today because I did them to help test the changes.
Probably should update to use subscription manager before committing.

Second they do not pass flake8 because of the ein, etc. functions from wbemcli.  Not importing anything to get around that.

### DIFF
--- a/examples/wbemcli_clean_subscriptions.py
+++ b/examples/wbemcli_clean_subscriptions.py
@@ -1,0 +1,54 @@
+"""
+Wbemcli scriptlet that removes all subscriptions, filters, and destination
+objects from the defined wbem connection
+"""
+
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Sort of rewrote them both since the code was a hack the first time. Andy. tks for review.

+1

andy-maier:
+1

Issue #643, opened 2016-12-21
test_cim_obj.py tests for CIMQualifierDeclaration are all empty


KSchopmeyer:
Since these are really tested through the mof_compiler tests which creates many qualifier decls as part of the compile of the cim model, these are not essential tests in this test file but we should complete this set of tests.

andy-maier:
Fixed by PR #647.

Issue #598, opened 2016-12-05
reference instances return REFERENCECLASS property in test_client


KSchopmeyer:
in the test case referenceinstance1 (part of pr589) we are getting the following error on the
response:

  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/pywbem-0.10.0.dev0-py2.7.egg/pywbem/cim_obj.py", line 498, in _intended_value
    % (name, actual, intended))
ValueError: Property u'Member' has a CIMInstanceName value with classname=u'PyWBEM_Person', but specifies reference_class=u'CIM_ManagedElement' (must be u'PyWBEM_Person')

Not yet sure what this means.  This works in run_cim_operations and I think that CIM_ManagedElement is the correct response so we may actually have an issue in cim_obj.py but not sure yet.

I wanted to gather as much data as possible now.
I think I have isolated the problems. See reasons in following comment block

andy-maier:
This is a check in `CIMProperty`, and we need to investigate as to whether this check is correct and the input data incorrect, vs. the check is overdoing it.

KSchopmeyer:
Agreed. I  moved this one back to mandatory since if it is an issue, it is a real issue.  Right now it is affecting our tests but want to clarify so we can fix tests. 

As an example:

OpenReferences yaml test on the PyWBEM_Person class.

In the test (recorder output) we set the reference_class to CIM_ManagedElement based on the response from OpenPegasus.

The mof for the reference class is:
```
class PyWBEM_MemberOfPersonCollection : CIM_MemberOfCollection {
};
```
and CIM_MemberOfCollection is:

```
class CIM_MemberOfCollection
{
    
        [Key, 
        Aggregate, 
        Description ( "The Collection that aggregates members." )]
    CIM_Collection REF Collection;
    
        [Key, 
        Description ( "The aggregated member of the Collection." )]
    CIM_ManagedElement REF Member;
};
```

The mof generated by the recorder has the property member as follows:

```
      </PROPERTY.REFERENCE>
      <PROPERTY.REFERENCE NAME="Member"  REFERENCECLASS="CIM_ManagedElement">
      <VALUE.REFERENCE>
```
So it would appear to me that the reference class is CIM_ManagedElement, not PyWBEM_Person.

From DSP2001

```
5.2.9 ReferenceClass 625
[...]

PR #647, opened 2016-12-26
fixes #620 and #643 Add testclient tests and add cim_obj tests

### LINKED ISSUE #620: Still missing tests in mock to cover all operations
At least the following tests need to be added to the yaml files to test all cim requests:

    finish invokemethod with input and output parameters test
    createclass
    modifyclass
    deleteclass
    setQualifier
    execquery
    openqueryinstances
    OpenQueryInstances
    PullInstances (part of the query instances sequence)

NOTE: Added opequeryInstances is part of the Iter pr, see pr for issue#466.  Sort of got there by accident while we were looking for another problem

KSchopmeyer:
**Status:** Ready for Review. 
This gets us up to 66% coverage so all this was about 1.4 % gain

Adds mock tests for CreateClass, DeleteClass, ModifyClass, SetQualifier,
DeleteQualifier, and ExecQuery.

Adds unittests to test_cim_obj.py for CIMQualifierDeclaration. This was part of getting
the mock tests running for setqualifier, etc.

Adds tests  to run_cim_operation.py for CreateClass, DeleteClass, ModifyClass, SetQualifier,
DeleteQualifier.

Corrects an error in CIM_operations where recorder was being called with
incorrect type.

At this point we have at least minimal mock tests for every operation  and most of them have both
success and fail tests.

There are several mock tests marked ignore_test for the following reasons(5 tests are marked ignore out of over 200 separate tests):
1. The reference_class issue (see issue #598) - two tests
2. The create_class, modify_class test for a class with both property and property.array elements has problems sorting the objects correctly so sometimes fails. No issue filed to date.

In addition the parameter component of invoke method is not exercised because of issues in test_client (issue not yet filed)

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -7880,7 +7880,7 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         if self.operation_recorder:
             self.operation_recorder.reset()
             self.operation_recorder.stage_pywbem_args(
-                method='SetQualifier',
+                method='DeleteQualifier',
                 QualifierName=QualifierName,
                 namespace=namespace,
                 **extra)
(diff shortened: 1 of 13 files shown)


andy-maier:
+1

PR #650, opened 2017-01-01
Fixes #649: coverage html report

### LINKED ISSUE #649: Add option to makefile to generate detailed coverage report
Add an optional report generation that shows line by line coverage of the files in the pywbem/pywbem
directory.  This will help analysis of coverage issues.

This is a feature of coverage and can be produced by adding the options for report output:
'''
--cov-report=annotate --cov-report=html
'''
to py.test

Make this optional since there is no reason to generate it for every test run, and particular the continuous integration tests under travis, etc. since it is just discarded.

An optional environment variable/command line variable can be used to control generation of the reports:

ex.
```
make test PYWBEM_COVERAGE_REPORT=1
```

KSchopmeyer:
Create optional detailed line by line report output showing code coverage.
    
 Adds an optional environment variable for the make file to control
output of a detailed line by line coverage report in html   If this
variable is set (export PYWBEM_COVERAGE_REPORT = true) html output is
generated in the directory coverage_html that shows the files in the
pywbem directory with coloring for which lines are covered and missed.
This also creates temporary files with the suffix ,cover in
pywbem/pywbem directory.  All these files are destroyed by clobber.
    
We made this an optional variable because it is only needed occasionally
when working on coverage issues. There is no need to generate these
reports for every make test.
    
To generate the reports:
      1) set the env var (export PYWBEM_COVERAGE_REPORT=true
      2) run make test
or
      1) run make test PYWBEM_COVERAGE_REPORT=true

Note that it is not necesary to use true, just some value since we only test for existence of the variable not for equality to true

### DIFF
--- a/coveragerc
+++ b/coveragerc
@@ -19,6 +19,7 @@ omit =
 # Controls whether lines without coverage are shown in the report.
 show_missing = True
 
+# defines location for coverage html output
 [html]
-directory = coverage_html_report
+directory = coverage_html
 
--- a/coveragerc
+++ b/coveragerc
@@ -19,3 +19,6 @@ omit =
 # Controls whether lines without coverage are shown in the report.
 show_missing = True
 
+[html]
+directory = coverage_html_report
+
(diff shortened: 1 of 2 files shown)


andy-maier:
+1

andy-maier:
+1

Issue #651, opened 2017-01-04
Add iter functions to wbemcli


KSchopmeyer:
Christmas makes me crazy.  I know I did this code but it is not in wbemcli.  We should have the short form commands for the iter... cimoperations.

KSchopmeyer:
Closed with pr # 655

PR #655, opened 2017-01-05
 fixes issue #651 Add iter functions to wbemcli

KSchopmeyer:
Status: ready for review and commit.  

This change does the following:

1.  adds the iter... functions to the list of shortcut aliases that are defined for wbemcli users.

2, Adds a test scriptlet in examples (wbemcli_testwbemcli_scriptlet.py that executes most of the shortcut aliases against a server.  This is not written as a unit test but simply generates asserts if the tests fail. The goal of this is not to run complete tests but to assure that the shortcuts and thier short-name attributes are correct and do execute.

3. fixes a couple of problems in the documentation for the new iter... operations in cim_operations.py.

This test scriptlet is only a temporary solution but this test runs a live test of a number of the functions against open pegasus including:

1. gc, ecn, ec, gq, eq

2. all the enumerates, associators, references and their path functions for all three variations  enumerate, open/pull, and iter... functins for enum, assoc, references. This pretty throughly tests the new iter functions through wbemcli against the pull and enumerate versions of the same functions.  Note that this also tests all of the various options for the refs, associators, etc. except for the filterquery, etc. attributes.

There are number of other functions it does not test including execquery, create/modify/delete instances and classes and the set qualifier operations.  Remember that we never tested any of this before.

However, I did not put this test into testsuite until we do some other work on wbemcli to avoid flake8 issues and pylint. At least we know that these functions including the new ones added do work.

Right now it passes the flake8 test but lots of comments from pylint about the functions, etc. from wbemcli.   Goal will be eventually to move this into testsuite and run as a normal part of the run... tests but that is another issue/pr.

### DIFF
--- a/coveragerc
+++ b/coveragerc
@@ -19,3 +19,7 @@ omit =
 # Controls whether lines without coverage are shown in the report.
 show_missing = True
 
+# defines location for coverage html output
+[html]
+directory = coverage_html
+
--- a/examples/wbemcli_count_instances.py
+++ b/examples/wbemcli_count_instances.py
@@ -1,4 +1,4 @@
-'''
+"""
     Pywbem wbemcli scriptlet that counts classes and instances in a WBEMServer.
 
     This class explored the complete server and retrieve all
@@ -7,7 +7,7 @@
 
     It includes an ignore list that ignores selected classes and another that
     will ignore selected namespaces.
-'''
+"""
 from collections import defaultdict
 import six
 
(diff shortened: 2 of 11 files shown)


andy-maier:
Karl, I rebased this va the "Update branch" button, because I tought it is ready for merge, and then I saw the "WIP" marker.
If you continue working on it, you need to pull this branch first.

KSchopmeyer:
+1

andy-maier:
+1

PR #657, opened 2017-01-16
fixes #656 pull maxobjectcount issue with pull... operations

### LINKED ISSUE #656: The cim_operations.py Pull... operations do not check for None for MaxObjectCount

    or

KSchopmeyer:
Ready for review:

Fix issue with MaxoObjectCount on PullInstances, PullInstancesWithPath and PullInstancePaths  CIM_Operations.py methods.  The MaxObjectCount was defined as a keyword  parameter where it should have been be positional.  This should NOT impact  clients unless they did not supply the parameter at all so that the result  was None which is illegal(Pull... operations MUST include MaxObjectCount).  Also extends these requests to test the Pull.. methods for valid  MaxObjectCount and context parameters.

This pr adds detailed checking for the input parameters for all of the Pull... operations to test:

Context for existence and minimum number of entires in tuple.

MaxObjectCount for existence, type (int), and value >= 0.

This pr adds new tests to mock tests and modifies run_cim_operations.py to use prefered form for the MaxObjectCount parameter.

Note that since we did not have a mock test for PullInstances, this adds that file but there are still couple of tests to add to that file.  Marked with a TODO and we will resolve in separate issue.

Possible question: I used ValueError as the exception for issues with both the MaxObjectCount and Context parameters for all of the Pull... operations.  However, the tests themselves are both for type (i.e non-integer, and value (negative) so the ValueError flag is not completely consistent.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -29,6 +29,14 @@ Enhancements
 Bug fixes
 ^^^^^^^^^
 
+* Fix issue with MaxoObjectCount on PullInstances and PullInstancePaths
+  CIM_Operations.py methods.  The MaxObjectCount was defined as a keyword
+  parameter where it should have been be positional.  This should NOT impact
+  clients unless they did not supply the parameter at all so that the result
+  was None which is illegal(Pull... operations MUST include MaxObjectCount).
+  Also extends these requests to test the Pull.. methods for valid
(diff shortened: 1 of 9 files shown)


KSchopmeyer:
This was a mistake.  The code was bad. Closing and canceling this pull Request

PR #658, opened 2017-01-16
Fixes 656: MaxObjectCount parameter of Pull operations inconsistent

### LINKED ISSUE #656: The cim_operations.py Pull... operations do not check for None for MaxObjectCount

KSchopmeyer:
This parameter was inconsistent between the different cim_operations.py pull...
operations methods (was keyword in 2 of the methods). The PR made it positional for all 3 operations.  Also added validation for the parameters for these operations (Context and MaxObjectCount).

Added tests to mock tests for the new method validation and modified
run_cim_operations.py to match the changes in cim_operations.py. Note
that these changes should not impact clients since the keyword parameter
is still accepted and the tests we added should be caught by the server
or pywbem code in any case.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -29,6 +29,15 @@ Enhancements
 Bug fixes
 ^^^^^^^^^
 
+* Fix issue with MaxObjectCount on PullInstances and PullInstancePaths
+  CIM_Operations.py methods.  The MaxObjectCount was defined as a keyword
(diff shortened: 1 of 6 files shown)


KSchopmeyer:
Note that this fix needs to be backed to 10 also

andy-maier:
+1

andy-maier:
I will leave issue #656 open when this PR gets merged, because that issue also needs to be fixed in 0.10.1.

PR #660, opened 2017-01-18
Fixes #648, fixes #601: WIP: Pull Operations yaml uses keyword 'pullresult' rather than 'result'

### LINKED ISSUE #601: recorder sets return type for openassociators to result, not p
I believe this is a duplicate of #648 in that they both cover the issues between pull results and results on the yaml.  

The recorder sets the return  type for open, and other pull operations to result where the required type is pullresult and shows the context items  wrong.

```
  pywbem_response:
    result:
    - - pywbem_object: CIMInstance
        classname: PyWBEM_PersonCollection
        properties:
          instanceid:
            pywbem_object: CIMProperty
            name: InstanceID
            value: PersonCollection
            type: string
            reference_class: null
            embedded_object: null
            is_array: false
            array_size: null
            class_origin: null
            propagated: null
            qualifiers: {}
          caption:
            pywbem_object: CIMProperty
            name: Caption
            value: null
            type: string
            reference_class: null
            embedded_object: null
            is_array: false
            array_size: null
            class_origin: null
            propagated: true
            qualifiers: {}
          elementname:
            pywbem_object: CIMProperty
            name: ElementName
            value: null
            type: string
            reference_class: null
            embedded_object: null
            is_array: false
            array_size: null
            class_origin: null
            propagated: true
            qualifiers: {}
          description:
            pywbem_object: CIMProperty
            name: Description
            value: null
            type: string
            reference_class: null
            embedded_object: null
            is_array: false
            array_size: null
            class_origin: null
            propagated: true
            qualifiers: {}
        path:
          pywbem_object: CIMInstanceName
          classname: PyWBEM_PersonCollection
          namespace: root/cimv2
          keybindings:
            instanceid: PersonCollection
    - false
    - - '500028'
      - root/cimv2
```

### LINKED ISSUE #648: Recorder issues with recording of open/pull requests.
We created another type for these requests, 'pullresult' rather than 'result' to get the context info into the response.

The recorder does not cover that difference nor does it report the context information correctly. It
reports it at the same level as the instances rather than within an envelop of its own.

See issue #601 for example of the problem outputting named tuples.  Since the recorder toyaml does not handle namedtuple nor does the python yaml outputter we have to add code to identify named tuples in the recorder and map them to something else.  This can probably be made to work by using th existing isinstance(xxx, tuple) and then a subtest to identify that the item is a named  tuple.  This secondary test will be a hack in that there is no thought-out way to id named tuples in python.  The best is probably to identify an attribute of the tuple (ex. (_asdict() or __dict__) and use this function to map the tuple to a dictionary.  This was one of those things that was clearly not though out in the design of nametuple in that an test for named tuple type is a hack and has variations on different versions of python (python 2.6 being the weakest). But, for example _asdict was obsoleted in 3.4 and then later.

KSchopmeyer:
Status: WIP. Code complete for change of yaml entry name but we have to add code to prcess named tuples ouput in the recorder.

Adds code to recorder to output yaml section definition of 'pullresult:' for pull operations that return tuples of info  vs. 'result:' for other operations

Also does small code cleanup in test_client.py so reduce code in hope that
we can reduce the result and pullresult to a single type in the future.

Adds options to identify pull operations in cim_operations call to recorder.

### DIFF
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -250,7 +250,7 @@ class BaseOperationRecorder(object):
         """Indicate whether the recorder is enabled."""
         return self._enabled
 
-    def reset(self):
+    def reset(self, pull_op=None):
         """Reset all the attributes in the class"""
         self._pywbem_method = None
         self._pywbem_args = None
@@ -270,6 +270,7 @@ class BaseOperationRecorder(object):
         self._http_response_reason = None
         self._http_response_headers = None
         self._http_response_payload = None
+        self._pull_op = pull_op
 
     def stage_pywbem_args(self, method, **kwargs):
         self._pywbem_method = method
@@ -418,7 +419,9 @@ class TestClientRecorder(BaseOperationRecorder):
 
         tc_pywbem_response = OrderedDict()
         if pywbem_result.ret is not None:
-            tc_pywbem_response['result'] = self.toyaml(pywbem_result.ret)
+                yaml_txt = 'pullresult' if self._pull_op else 'result'
+                tc_pywbem_response[yaml_txt] = self.toyaml(pywbem_result.ret)
+
         if pywbem_result.exc is not None:
             exc = pywbem_result.exc
             if isinstance(exc, CIMError):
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
We agreed to merge this change and Karl will write issue on the namedtuple output.

+1

andy-maier:
+1

PR #662, opened 2017-01-18
OBSOLETE. CloseMove fixes for pr656 back to 0.10.0

KSchopmeyer:
Here is hopefully the changes moved back to 0.10.0.  I am obviously missing something since this was not easy to do.  Suggestions welcome.  Also when I created the pr it pointed to the current head  as the merge target but I manually set it to stable-0.10.   

I think I have to learn more about this part of the process

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,6 +15,36 @@ contains the `master` branch up to this commit:
    :revisions: 1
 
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
This was obsolete pr that I marked obsolete

PR #665, opened 2017-01-21
fixes #656 Merge pull request #658 from pywbem/issue#656-pullops-maxobjcount

### LINKED ISSUE #656: The cim_operations.py Pull... operations do not check for None for MaxObjectCount

KSchopmeyer:
Fixes 656: MaxObjectCount parameter of Pull operations inconsistent
backported to 0.10.0 release. Backported with git cherry-pick

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,6 +15,37 @@ contains the `master` branch up to this commit:
    :revisions: 1
 
 
+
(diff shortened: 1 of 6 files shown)


KSchopmeyer:
Accidently defined this against the master, not the Stable_0.10 branch

PR #667, opened 2017-01-22
Fixes #646 Change recorder to fix issue of scope names case lose

### LINKED ISSUE #646: Mock test, setQualifier request fails scope name case
**STATUS 17 Jan 16:** This is not a pywbem request problem directly but an issue in the converson to yaml for the recorder. The scope name is converted to lower case. Requests are passed correctly to the server (scope name case preserved). However, when passed to the recorder and then converted to yaml, the scope name case is lost.   It appears to be an issue between the NocaseDict and the recorder. I created a test with both dict and NocaseDict and the dict entries were passed with case preserved but the NocaseDict were not.  See test in comment below.

DISCUSSION: Suggestions welcome.

In this test we send the request
```
  http_request:
    verb: POST
    url: http://acme.com:80/cimom
    headers:
      CIMOperation: MethodCall
      CIMMethod: SetQualifier
      CIMObject: root/cimv2
    data: '<?xml version="1.0" encoding="utf-8" ?>
      <SIMPLEREQ>
      <IMETHODCALL NAME="SetQualifier">
      <LOCALNAMESPACEPATH>
      <NAMESPACE NAME="root"/>
      <NAMESPACE NAME="cimv2"/>
      </LOCALNAMESPACEPATH>
      <IPARAMVALUE NAME="QualifierDeclaration">
      <QUALIFIER.DECLARATION ISARRAY="false" NAME="FooQualDecl" OVERRIDABLE="false"
      TOSUBCLASS="false" TYPE="string">
      <SCOPE CLASS="true"/>
      <VALUE>Some string</VALUE>
      </QUALIFIER.DECLARATION>
      </IPARAMVALUE>
      </IMETHODCALL>
      </SIMPLEREQ>
      </MESSAGE>
      </CIM>'
```
However, we get an error
```
Diff:
      <SIMPLEREQ>
        <IMETHODCALL NAME="SetQualifier">
          <LOCALNAMESPACEPATH>
            <NAMESPACE NAME="root"></NAMESPACE>
            <NAMESPACE NAME="cimv2"></NAMESPACE>
          </LOCALNAMESPACEPATH>
          <IPARAMVALUE NAME="QualifierDeclaration">
            <QUALIFIER.DECLARATION ISARRAY="false" NAME="FooQualDecl" OVERRIDABLE="false" TOSUBCLASS="false" TYPE="string">
              <SCOPE +class="true" -CLASS="true"></SCOPE>
              <VALUE>Some string</VALUE>
            </QUALIFIER.DECLARATION>
          </IPARAMVALUE>
        </IMETHODCALL>
      </SIMPLEREQ>
    </MESSAGE>
  </CIM>
```
Which indicates that the case of the scope names is important.  To pass the test, I changed the expected result to lower case class.  Test passes.  NOTE: Since scope names are supposed to be upper case, we should have changed the pywbem_request, not the http in the test to correct the issue.

KSchopmeyer:
Fixes the toyaml function to pass correct key case when converting from NoCaseDict to OrderedDictionary.

Adds tests to test_cimobj for this

We also cleaned all pylint messages out of _recorder.py. All of the changes were variable name changes or setting disable.

Fixes the data in testclient for setQualifier to correctly reflect what
the request and response should reflect (CLASS, not class).

Extends test in run_cim_operations.py

### DIFF
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -21,9 +21,9 @@ Operation recorder interface and implementations.
 from __future__ import print_function, absolute_import
 from collections import namedtuple
 try:
-    from collections import OrderedDict
+    from collections import OrderedDict  # pylint: disable=import-error
 except ImportError:
-    from ordereddict import OrderedDict
+    from ordereddict import OrderedDict  # pylint: disable=import-error
 from datetime import datetime, timedelta
 import yaml
 from yaml.representer import RepresenterError
@@ -43,7 +43,7 @@ if six.PY2:
 else:
     _Longint = int
 
-OpArgs_tuple = namedtuple("OpArgs_tuple", ["method", "args"])
+OpArgsTuple = namedtuple("OpArgsTuple", ["method", "args"])
 
 
 def _represent_ordereddict(dump, tag, mapping, flow_style=None):
@@ -86,6 +86,7 @@ yaml.SafeDumper.add_representer(
 
 # Some monkey-patching for better diagnostics:
 def _represent_undefined(self, data):
+    """Raises flag for objects that cannot be represented"""
     raise RepresenterError("cannot represent an object: %s of type: %s; "
                            "yaml_representers: %r, "
                            "yaml_multi_representers: %r" %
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #668, opened 2017-02-03
Fixes #661: Added Python 3.6 to Travis CI and 3.4-3.6 to Appveyor CI

### LINKED ISSUE #661: Add python 3.6 as a test target for continuous integration
Python 3.6 has been released.  We should be testing with it in the near future, at least in the travis continuous integration tests.

andy-maier:
Please review.

This one was larger than expected (with Python 3.6-specific errors on Linux and Windows, and with Windows-specific errors on Python 3.x that needed to be fixed).

Also, I had to document a limitation that Python 2.6 on Windows is not supported (because we use M2Crypto from the M2CryptoWin32/64 package, which does not support Python 2.6), and that 64-bit Python on Windows is not supported or at least was not tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - "2.7"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 # commands to install dependencies
 install:
@@ -30,4 +31,4 @@ script:
   - make test
 
 after_success:
-  - coveralls
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then coveralls; fi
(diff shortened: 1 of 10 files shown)


andy-maier:
On Travis with Python 3.6, the unit tests for `NocaseDict` fail in class `test_nocasedict.TestOrdering`: The assertion that `TypeError` is raised in Python 3 expected a particular string within the exception message. That string has changed in Python 3.6. The next update to this PR will fix that by tolerating both the old (3.4 - 3.5) and the new (3.6) message.

andy-maier:
+1

KSchopmeyer:
+1

One step closer to just killing off 2.6 support

Issue #671, opened 2017-02-24
Add cim_operation timing to pywbem


KSchopmeyer:
We propose to add a mechanism that will allow the execution time for each operation to be captured and made available to:

1. The user after the call
2. pywbem statistics keeper

This feature will be normally disabled but can be enabled by the user (ex. similar to the debug feature in WBEMConnection) either by a WBEMConnection constructor parameter or by setting an attribute in WBEMConnection.

Note that the execution time must capture the time both for normal completion and for exceptions.

DISCUSSION: What should we do about the iter functions that are actually a higher level and use the CIM_Operation request primitives?

andy-maier:
Addressed with PR #675.

PR #675, opened 2017-03-08
Fixes #761: Create classes for keeping statistics and add statistics

### LINKED ISSUE #761: duplicate reference target from cim_obj.py
Here is the documentation issue we were talking about a this morning:
```
/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_obj.py:docstring of pywbem.CIMInstanceName:: WARNING: more than one target found for cross-reference u'host': pywbem.WBEMListener.host, pywbem.WBEMConnection.host
/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_obj.py:docstring of pywbem.CIMClassName:: WARNING: more than one target found for cross-reference u'host': pywbem.WBEMListener.host, pywbem.WBEMConnection.host
```

KSchopmeyer:
Keeps time statistics on cim_operations.  This keeps statistics on time to
execute and request/response sizes for the cim operations.

Implements a Statistics class and connects cim_operations to that class to record execution time, request size, and reply size in that class for each operation type.  It saves count, avg time, request size, response size in and exception count in that class for each WBEM operation type by name.

The Statistics class provides for snapshotting and resetting the statistics in addition to the functions to record each new statistics with start_time() and stop_time()

NOTE: At this point the statistics are not really general.  The class OperationStatistics specializes the statistics for the operations and their is no provision for another class with its specialization at this point.

Finally, includes ability to disable statistics but I am not really using that capability. You enable statistics for each WBEMConnection.

In addition, the change adds three attributes to WBEMConnection.

last_req_len - Always
last_reply_len - Always
last_operation_time - Only when the enable-stats is set  

ANDY: I goofed and force updated over your changes before I realized you had made them.

NOTE: This does NOT capture info from the server statistics.

### DIFF
--- a/pywbem/_statistics.py
+++ b/pywbem/_statistics.py
@@ -16,26 +16,35 @@
 # Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 
 """
-The :class:`~pywbem.Statistics` class allows measuring the elapsed
-time and request/reply length of WBEM Operations and keeping statistics on the
-measured times
+Pywbem supports measuring the elapsed times of the WBEM operations that
+were performed in context of a connection, and maintaining a statistics
+over these times.
 
-It consists of two classes:
+This capability is disabled by default and can be enabled in either
+of these ways:
+
+* When creating a :class:`~pywbem.WBEMConnection` object, via its
+  ``enable_stats`` argument.
+
+* After the :class:`~pywbem.WBEMConnection` object has been created, by
+  modifying its :attr:`~pywbem.WBEMConnection.stats_enabled` instance
+  attribute.
+
+The :class:`~pywbem.Statistics` class maintains statistics over the measured
+elapsed times of the WBEM operations and is the interface for accessing the
+statistics. The statistics of a :class:`~pywbem.WBEMConnection` object are
+accessible via its :attr:`~pywbem.WBEMConnection.statistics` instance
+attribute.
 
 The :class:`~pywbem.OperationStatistic` class is a helper class that contains
-the actual measurement data for all invocations of a particular name. Its
-objects are under control of the :class:`~pywbem.Statistics` class.
-This class should be used independently.  Use the start_timer in the
-Statistics class to both start a timer and if it does not exist, create
-a new named statistic  category.
+the actual measurement data for one operation name.
+There will be one :class:`~pywbem.OperationStatistic` object each operation
+name (see the table of WBEMConnection methods in the :ref:`WBEM operations`
+section for the operation names).
+The :class:`~pywbem.OperationStatistic` objects are under control of the
+:class:`~pywbem.Statistics` class.
 
-The :class:`~pywbem.Statistics` class maintins time statistics over
-multiple separate named statistics gatherers (TimeStatistic class) and provides
-tools for reporting these statistics
-
-**Experimental:** The statistics functions are experimental for this
-release***.
-
+**Experimental:** The statistics support is experimental for this release.
 """
 
 from __future__ import absolute_import
@@ -43,38 +52,35 @@ from __future__ import absolute_import
 import time
 import copy
 
-__all__ = ['Statistics']
+__all__ = ['Statistics', 'OperationStatistic']
 
 
 class OperationStatistic(object):
     """
-    Elapsed time information for all invocations of a particular named
-    operation.
+    A statistics data keeper for the executions of all operations with the
+    same operation name.
 
-    This class maintains max, min, avg and count for a single named
-    statistic for time, request length and reply length.
+    This class maintains max, min, avg and count for elapsed time, request
+    size, and response size over the executed operations of a single
+    operation name.
 
-    Use :meth:`pywbem.Statistics.get_statistic` method  to
-    create objects of this class.
-
-    Ex:
+    Use the :meth:`pywbem.Statistics.start_timer` method to create objects
+    of this class::
 
         stats = container.start_timer('EnumerateInstances')
         ...
-        stats.stop_timer(req_len, reply_len, exception=True)
+        stats.stop_timer(request_len, reply_len, exc)
+
+    **Experimental:** This class is experimental for this release.
     """
-    display_col_hdr = \
-        'Count  Time    Time    Time  Req     Req   Req   Reply     Reply ' \
-        'Reply Exc Name\n' \
-        '       Avg     Min     Max   Avg     Min   Max   Avg       Min ' \
-        'Max   \n'
 
     def __init__(self, container, name):
         """
         Parameters:
 
-          container (Statistics):
-            The statistics container that holds this time statistics.
+          container (:class:`~pywbem.Statistics`):
+            The statistics container that holds this operation statistic
+            object.
 
           name (string):
             Name of the operation.
@@ -89,9 +95,9 @@ class OperationStatistic(object):
         self._time_max = float(0)
         self._start_time = None
 
-        self._req_len_sum = float(0)
-        self._req_len_min = float('inf')
-        self._req_len_max = float(0)
+        self._request_len_sum = float(0)
+        self._request_len_min = float('inf')
+        self._request_len_max = float(0)
 
         self._reply_len_sum = float(0)
         self._reply_len_min = float('inf')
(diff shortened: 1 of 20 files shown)


KSchopmeyer:
Tried the cleanup of except but that causes problems in some cases of exceptions:

>                                     raised_traceback_str))
E               AssertionError: Testcase OpenEnumerateInstancesF1: A CIMError exception was expected to be raised by PyWBEM operation OpenEnumerateInstances, but a different exception was actually raised:
E               Traceback (most recent call last):
E                 File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 4072, in OpenEnumerateInstances
E                   **extra)
E                 File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1045, in _imethodcall
E                   raise CIMError(code, err[1]['DESCRIPTION'])
E               pywbem.exceptions.CIMError: 2: CIM_ERR_ACCESS_DENIED: Access denied
E               
E               During handling of the above exception, another exception occurred:
E               
E               Traceback (most recent call last):
E                 File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/testsuite/test_client.py", line 613, in runtestcase
E                   result = op_call(**op_args)
E                 File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 4082, in OpenEnumerateInstances
E                   stats.stop_timer(exc)
E               UnboundLocalError: local variable 'exc' referenced before assignment

It appears that finally is picking up some cases in exceptions that are out of the scope of the exc variable that is set before the try block.

While this may point to a different problem with out exceptions, the way to solve it for now was to back of the change of function structure from using except... finally to the original structure that uses except... else

testsuite/test_statistics.py
```diff
@@ -0,0 +1,174 @@
...
+class StatisticsTests(unittest.TestCase):
+    """All tests for Statistics and TimeCapture."""
+
+    def test_enabling(self):
+        """Test enabling and disabling."""
+
+        statistics = Statistics()
+
+        self.assertFalse(statistics.enabled,
+                         "Error: initial state is disabled")
```
andy-maier: The message is printed if the assertion fails, so it should say what is not correct. In this case, what is not correct is that the initial state is *not* disabled.

KSchopmeyer:
Fixed Andy's comment. 

+1

KSchopmeyer:
+1.

Fixed an issue that had not shown up before (the scope of the variable x in Exception as x)   and added
to the statitistics a return from the stop so that the time of the last operation is also stored in WBEMConnection if a user was just access to that time and not all of the statistics.

Ran it through tests on pywbem2.7 and 3 and did run_cim_operation.py on both.  Note that I changed the cmd line parameters in run_cim_operations.py so the default is not to show stats and -stats turns on the stats.

Statistics (time in sec.):
Count   Avg     Min     Max     Exc Name
    1   0.005   0.005   0.005     0 OpenEnumerateInstances
    1   0.004   0.004   0.004     1 PullInstancesWithPath
    1   0.003   0.003   0.003     0 CloseEnumeration

The statics display looks like the above.

KSchopmeyer:
Mark this experimental and create new issue to add DMTF std statistics. We will try to commit this and then add the statistics from OpenPegasus.

We need to add the request and response size. This does come back from Pegasus but we really need to show it as produced by the client also.

pywbem/_statistics.py
```diff
@@ -0,0 +1,352 @@
+#!/usr/bin/env python
+#
+# (C) Copyright 2003-2007 Hewlett-Packard Development Company, L.P.
```
andy-maier: I think for new modules we add, they should not have these copyright statements, but those of the initial author (you)

pywbem/_statistics.py
```diff
@@ -0,0 +1,352 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+"""
+The :class:`~pywbem.Statistics` class allows measuring the elapsed
+time of instrumented coded and keeping statistics on the measured times
```
andy-maier: Grammar issues in this line.
Change line 22 to:

    time of WBEM operations and keeping statistics on the measured times.

andy-maier:
Karl, I had many small comments, so I decided to push a commit on top of yours, into this branch. Please pull this branch to get it.

Still to do for you (after pulling the branch):
- Rebase to master
- Mark as experimental
- Add request and response size to the statistics.

pywbem/_statistics.py
KSchopmeyer: I fixed that.  Note that the copyright is InovaDevelopment

andy-maier:
Karl,
I have re-established my changes as of commit 5a3b6a5, based upon your latest update.

That commit also contains a few more changes that I thought were necessary. The commit message has a detailed listing of all changes. Some of them change the external interface, so please review carefully.

Also, I have made the change for the improved logic for the `last_request/reply_len` attributes of `WBEMConnection` in commit ede1a96.

KSchopmeyer:
Made a number of minor changes including:
1. Fixed a couple of places where tests failed because still using req instead of request in name
2. Fixed yaml to reflect new names for request, reply
3. Added size tests to a number of yaml tests.  We do not test all but now test about 10 types.
4. Added test of statistics to the test_client tests.  Originally they were just testing the values in cim_operations.  Now they test the values in statistics also.
5. Fixed couple of things in documentation.
6. Test with run_cimoperations and found one issue, the last_operation_time is NULL if the iter test fails with invalid parameter (ex. non-integer MaxObjectCount). That is sort of documented but I extended run_cimoperations to cover the issue.  Ran run_cimoperations with stats on with no problems.

KSchopmeyer:
Hit an error with python 2.6 in test_statistics.py
        print("\n\nTest print of repr() for a small statistics:")
        print("================")
        print(repr(statistics))
        print("================")
        print("\nTest print of formatted() for the same statistics:")
        print("================")
>       print(statistics.formatted())
testsuite/test_statistics.py:270: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pywbem/_statistics.py:523: in formatted
    ret += stats.formatted()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = OperationStatistic(name='EnumerateInstances', count=3, exception_count=0, avg_...00, avg_reply_len=26666.666666666668, min_reply_len=20000, max_reply_len=35000)

    def formatted(self):
        """
            Return a formatted one-line string with the statistics values for this
            operation.   

            The returned string can be used with the formatted header string
            defined in :attr:`~pywbem.OperationStatistic.formatted_header`.   

            For an example, see :meth:`pywbem.Statistics.formatted`.
            """
        return ('{:5d} {:5d} {:7.3f} {:7.3f} {:7.3f} {:6.0f} {:6.0f} {:6.0f} '
                '{:8.0f} {:8.0f} {:8.0f} {}\n'.
                format(self.count, self.exception_count,
                       self.avg_time, self.min_time, self.max_time,
                       self.avg_request_len, self.min_request_len,
                       self.max_request_len, self.avg_reply_len,
                       self.min_reply_len, self.max_reply_len,
>                      self.name))
E       ValueError: zero length field name in format

pywbem/_statistics.py:374: ValueError

andy-maier:
Thanks for fixing run_cim_operations!

Seems there is still a flake8 error, and then we should be ready to go.

andy-maier:
```
pywbem/_statistics.py:368:18: E127 continuation line over-indented for visual indent

1     E127 continuation line over-indented for visual indent
```

KSchopmeyer:
+1 think we finally beat this one into the ground

andy-maier:
+1

andy-maier:
This PR addressed issue #671 (not 761).

PR #678, opened 2017-03-21
Fixes #674: Add a new class to the test.mof file

### LINKED ISSUE #674: Define a class in the standard tests mof that includes all properties
Today the standard test mof includes a very simple set of classes that test basic properties and
associations.

I propose that we should have at least one more class in this mof, a Class and some instances that
test all the property types..

KSchopmeyer:
This pr adds a new class to test.mof that defines properties with all of the DMTF CIM Data types.  

There is actually a compile issue between what will compile with pegasus compiler and with pywbem compiler that I have not figured out.  Turns out that the issue is in the the pegasus compiler, not the pywbem compiler.  I had written a comment as # blah... instead of // blah...
and the pywbem compiler caught this but the pegasus compiler simply ignored the statement.

BUT I do not consider that part of this pr since that is in the existing test.mof classes, not the new class


KSchopmeyer:
+1

andy-maier:
+1

Issue #676, opened 2017-03-10
Recorder issues


KSchopmeyer:
Several issues with the recorder as follows:

1. indent output yaml by 4 to be more consistent with normal use. Include issue # 609 in this, set indent to 4 (currently 2)
2. Issue where open/pull does not correctly return context info.
It returns:
```
pullresult:
        - true
        - null
```
Where is should be in the form

```
        pullresult:
            context:
                - blahblah
                - root/cimv2
            eos: false
```

3. (New problem) Recorder does not write utf-8 in python 2.7.  Fixed for now by modifying the open in both run_cim_operations.py and test_recorder.py to use codecs.open(...) with python 2.

4.  (New problem)  Issue with recorder writing and then testing some datetime properties.  Results cannot be compared. Produces an exception stating that naive and timezone aware cannot be compared. Working solution in test_recorder.py is to insure that all datetimes are timezone aware.  See issue #698

5. Exception from run_cimoperations.py, Representer cannot handle datetime.
Actually, the representer cannot handle the python timedelta type.  Datetime appears to pass through
correctly.  The solutions are a) find some way to bypass that particular type, or b) extend the representer to handle timedelta correctly NOTE: This issue moved to issue#702

```
======================================================================
ERROR: test_non_empty_arrays (__main__.InvokeMethod)
Call with non-empty arrays
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testsuite/run_cim_operations.py", line 2359, in test_non_empty_arrays
    RefArray=[name, name])
  File "testsuite/run_cim_operations.py", line 169, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 6961, in InvokeMethod
    self.operation_recorder.record_staged()
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 352, in record_staged
    self.record(pwargs, pwresult, httpreq, httpresp)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 514, in record
    default_flow_style=False)
  File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/yaml/__init__.py", line 218, in safe_dump
    return dump_all([data], stream, Dumper=SafeDumper, **kwds)
[...]

PR #681, opened 2017-03-23
Fixes #680: Move CIM Operations methods from except/else to except/finally

### LINKED ISSUE #680: Move of cimoperations else, except to else, finally fails
TODO add example

KSchopmeyer:
1. Moved cim_operations from using except/else pattern for the
request methods to using except/finally.  This eliminates one of
the recorder calls. It was applied to all of the request methods.  This meant
setting the exc and return value to None at the beginning of each method but
reduces the recorder call to one in the finally clause.

2. Fixed part of issue#676, returns named ret where it should be
ret_dict because I needed the fix to really check the change to use
finally.  I crossed this one off the comments in issue #676

3. Fixed an issue in run_cim_operations.py that was causing a test
failure. It was not tied to this fix but was a) a case where pegasus returns an invokemethod of 1 and we were testing for zero and 2 cases where we have difference between instancenames in the existence of the host name and we did not set the flag to ignore that difference.

Big WHOOPS by Karl.  I recreated the changes that thought I was having problems with before as part of the statistics pr and it all works fine.  I have no idea what the problem was.  Note that I also tested with the recorder on in run_cimoperations.py and that worked also except for the problem in 2 above that was fixed and the datetime representer problem.

Note that the testclient does not test with the recorder so that any issues with the fact that we changed that code would only be seen by the run_cimoperations.py test.  Also, I think we should review to be sure that all of the operations get an exception test in testclient.  But that is separate issue.

I have no idea why I was having the problem before but this is much more satisfying.  I was definitely getting getting a double exception when I did this as part of the statistics pr but there must have been some other issue in my code that I did not take the time to sort out that day.  This is being done to prepare for other calls that occur at the end of a request function (ex. get class) so that they need to be in the code only once.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -33,6 +33,11 @@ Enhancements
 * Added the MOF compiler API to the ``pywbem`` namespace. For compatibility, it
   is still available in the ``pywbem.mof_compiler`` namespace. See issue #634.
 
+* Modify the pattern used for cim_operation.py request methods from using
+  except/else to use except/finally to reduce number of places code like
+  the recorder call and future statistics, log, etc. calls have to be included.
+  See issue #680
+
 Bug fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
+1

andy-maier:
+1

PR #684, opened 2017-04-02
Fixes #677, fixes #679: Compiler inst alias and cim_obj doc missing alias keybinding definition.

### LINKED ISSUE #677: cim_obj documentation
def tocimobj does not document the ValueError exceptions it returns.

### LINKED ISSUE #679: Problem creating association instances with pywbem compiler
Documentation of possible problem.  I am documenting this for now while I explore the issue further.  The same mof appears to be correctly compiled by the pegasus mof compiler in that the member of collection class includes complete instance paths in the reference properties.

Having problems compiling the test mof memberofcollection instances with the pywbem mof compiler in that it appears to be creating the instances without the keys in the reference properties as shown below.  It is like the alias mechanism is not really working for some reason:

```
pywbemcli> instance enumerate Pywbem_MemberOfPersonCollection
instance of PyWBEM_MemberOfPersonCollection {
    Member = root/cimv2:PyWBEM_Person;
    Collection = root/cimv2:PyWBEM_PersonCollection;
};

instance of PyWBEM_MemberOfPersonCollection {
    Member = root/cimv2:PyWBEM_Person;
    Collection = root/cimv2:PyWBEM_PersonCollection;
};

instance of PyWBEM_MemberOfPersonCollection {
    Member = root/cimv2:PyWBEM_Person;
    Collection = root/cimv2:PyWBEM_PersonCollection;
};
```
The create instance mof is as follows.
```
POST /cimom HTTP/1.1
Host: localhost:5988
Accept-Encoding: identity
Content-type: application/xml; charset="utf-8"
Content-length: 1047
PegasusAuthorization: Local "kschopmeyer"
CIMOperation: MethodCall
CIMMethod: CreateInstance
CIMObject: root/cimv2

<?xml version="1.0" encoding="utf-8" ?>
<CIM CIMVERSION="2.0" DTDVERSION="2.0"><MESSAGE ID="1001" PROTOCOLVERSION="1.0">
<SIMPLEREQ>
<IMETHODCALL NAME="CreateInstance">
<LOCALNAMESPACEPATH>
<NAMESPACE NAME="root"/>
<NAMESPACE NAME="cimv2"/>
</LOCALNAMESPACEPATH>
<IPARAMVALUE NAME="NewInstance">
<INSTANCE CLASSNAME="PyWBEM_MemberOfPersonCollection">
<PROPERTY.REFERENCE NAME="Member" PROPAGATED="true" REFERENCECLASS="CIM_ManagedElement">
<VALUE.REFERENCE><LOCALINSTANCEPATH><LOCALNAMESPACEPATH>
<NAMESPACE NAME="root"/>
<NAMESPACE NAME="cimv2"/></LOCALNAMESPACEPATH>
<INSTANCENAME CLASSNAME="PyWBEM_Person"/></LOCALINSTANCEPATH>
</VALUE.REFERENCE>
</PROPERTY.REFERENCE>
<PROPERTY.REFERENCE NAME="Collection" PROPAGATED="true" REFERENCECLASS="CIM_Collection">
<VALUE.REFERENCE><LOCALINSTANCEPATH><LOCALNAMESPACEPATH><NAMESPACE NAME="root"/><NAMESPACE NAME="cimv2"/>
</LOCALNAMESPACEPATH><INSTANCENAME CLASSNAME="PyWBEM_PersonCollection"/></LOCALINSTANCEPATH></VALUE.REFERENCE></PROPERTY.REFERENCE></INSTANCE>
</IPARAMVALUE></IMETHODCALL></SIMPLEREQ></MESSAGE></CIM>
```

KSchopmeyer:
1. Added keybindings to instance alias in compiler. The compiler not produces keybinding in the path and the alias table for instance definitions that include alias.  Before in only produced the class component of the path. That meant that other instances that used the alias did not get a complete instance alias.

2. Added test for instance alias - Added test for this in test_mof_comp

3. Fixed documentation issue in cim_obj.py

### DIFF
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -72,6 +72,7 @@ from .cim_constants import CIM_ERR_NOT_FOUND, CIM_ERR_FAILED, \
     CIM_ERR_INVALID_SUPERCLASS, CIM_ERR_INVALID_PARAMETER, \
     CIM_ERR_NOT_SUPPORTED, CIM_ERR_INVALID_CLASS, _statuscode2string
 from .exceptions import Error, CIMError
+from .cim_types import atomic_to_cim_xml
 
 __all__ = ['MOFParseError', 'MOFWBEMConnection', 'MOFCompiler',
            'BaseRepositoryConnection']
@@ -1594,6 +1595,8 @@ def p_instanceDeclaration(p):
             raise
     path = CIMInstanceName(cname, namespace=ns)
     inst = CIMInstance(cname, qualifiers=quals, path=path)
+    keybindings = NocaseDict()   # dictionary to build kb if alias exists
+
     for prop in props:
         pname = prop[1]
         pval = prop[2]
@@ -1619,6 +1622,10 @@ def p_instanceDeclaration(p):
             pprop.qualifiers = NocaseDict(None)
             pprop.value = tocimobj(cprop.type, pval)
             inst.properties[pname] = pprop
+            # if alias and this is key property, add keybinding
+            if alias and 'key' in cprop.qualifiers:
+                keybindings[pname] = atomic_to_cim_xml(pprop.value)
+
         except ValueError as ve:
             ce = CIMError(CIM_ERR_INVALID_PARAMETER,
                           'Invalid value for property %s: %s' %
@@ -1627,7 +1634,10 @@ def p_instanceDeclaration(p):
             raise ce
 
     if alias:
+        if keybindings:
+            inst.path.keybindings = keybindings
         p.parser.aliases[alias] = inst.path
+
     p[0] = inst
 
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -61,6 +61,16 @@ Bug fixes
 
 * Fixed names of Python development packages for SLES/OpenSUSE.
 
+* Fixed issue in mof_compiler where instance aliases were incomplete. They
+  only included the class component so that if they were used in the definition
+  of other instances (ex. to define an association where a reference property
+  was the aliased instance, the reference path was incomplete.) This is now
+  a path with keybindings.  Note: It is the responsibility of the user to
+  make these instances complete (i.e. with all key properties) see issue #679
+
+* Correct documentation issue in cim_obj (Exceptions definition missing).
+  See issue #677
+
 Build, test, quality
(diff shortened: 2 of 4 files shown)


pywbem/mof_compiler.py
```diff
@@ -1627,7 +1633,10 @@ def p_instanceDeclaration(p):
             raise ce
 
     if alias:
+        if keybindings:
+            inst.path.keybindings = keybindings
```
andy-maier: Here we are setting the path attribute of the instance we are building up. We need to figure out whether we want that. The logic is now: If an instance declaration in MOF has an alias definition, the compiled `CIMInstance` object will have its path set. otherwise, it will not have its path set. It seems to me that this is not particularly consistent.

pywbem/mof_compiler.py
```diff
@@ -1611,14 +1613,18 @@ def p_instanceDeclaration(p):
...
-
         try:
             # build instance property from class property but without
             # qualifiers, default value,
             pprop = cprop.copy()
             pprop.qualifiers = NocaseDict(None)
             pprop.value = tocimobj(cprop.type, pval)
             inst.properties[pname] = pprop
+            # if alias and this is key property, add keybinding
+            keybindings = NocaseDict()
```
andy-maier: The `keybindings` dictionary should probably be initialized before the loop through the properties.

andy-maier:
-1, see comments

pywbem/mof_compiler.py
KSchopmeyer: Actually the it always has the path set.

on line 1596 it first creates the path and then the new instance, putting the path it created into the new instance.

We are clearly wildely inconsistent here in that the only time this path is really correct is when the alias is set.  Otherwise it is only the name component of the path.

But the only change we are really making is to add the keybindings if the alias is set.

pywbem/mof_compiler.py
KSchopmeyer: Agreed an moved.

KSchopmeyer:
The sad part of all this is that it appears nobody was using alias before and even when we used it in our own test mof, we did not see the error.

pywbem/mof_compiler.py
andy-maier: Ok, I agree with the change.

andy-maier:
+1

PR #685, opened 2017-04-08
Fixes #642: Move from README.md to README.rst and extend the README

### LINKED ISSUE #642: Readme.md needs update
1.  Reflect the current proposed date for next release when we figure it out

2. We can significantly improve the description.  I found that out when I tried to use Readme.md to tell some people what pywbem really was.

KSchopmeyer:
This also adds a diagram. I am not married to this diagram and suggestions welcome

It extends the description, etc. 

NOTE: I thought I had set up th pr before but the branch was there but no pr.  Not sure why

Removed the README.md


KSchopmeyer:
Error is
0.05s$ make build

echo '# file GENERATED by makefile, do NOT edit' >MANIFEST.in

echo 'pywbem/LICENSE.txt README.md INSTALL.md *.py pywbem/*.py ' | xargs -r -n 1 echo include >>MANIFEST.in

Done: Created manifest input file: MANIFEST.in

make: *** No rule to make target `README.md', needed by `dist/pywbem-0.11.0.dev0-py2.py3-none-any.whl'.  Stop.

andy-maier:
I pushed a commit that addresses all of my comments, except for these:

* A rebase is still needed.
* Comments on the picture:
  - We may want to remove the future area entirely.
  - Remove the versions that indicate when a feature was added, and show the features as of the current release.
  - There is a mix of general terms (e.g. `WBEM Server Class`) and class names (e.g. `WBEMConnection`). Stay consistent. One option would be to use the exact class name followed by `class`, e.g. `WBEMServer class`, or for `CIMObjects` insert a space so it becomes a general term. Another option would be to use the names of the APIs we use in the documentation, minus `API` (e.g. `WBEM Server`). For the WBEM server, this option bears the potential for misunderstandings, however.
  - The picture shows some layering, but in case of the MOFCompiler class that layering is not correct. It should really be in its own box on top of the box with WBEMConnection.
  - The picture should show the existing two command line utilities as separate boxes on top of the right lower boxes.
  - Change `PyWBEM Client` to `Pywbem Client`. We seem to be simplifying the lexical case of the word everywhere now.
  - There is an inconsistency in the text in the big arrows at the bottom, between the left and right arrow. I suggest for the text:
    - Left arrow. `CIM-XML Operations`
    - Right arrow. `CIM-XML Indications`

KSchopmeyer:
+1  There is one issue in that the picture does not scale as it should.  When viewed in restview, the scaling works but it shows up as a full size on git.  Fixing that will require more research.

andy-maier:
Some more comments on the new picture:
* Insert space into `WBEMConnection`, for consistency with the other components shown
* Insert space into `CIMObjects` because there is no such class.
* Change "CIM-XML Requests and Responses" to "CIM-XML Operations", because that is the term we use in the DMTF specs.
* Be consistent between "Pywbem Client" and "(pywbemtools repository)".

andy-maier:
Karl, I have set this one back to update needed, since you said you were going to address my comments on the picture.

KSchopmeyer:
+1  I hope we are back in sync and up to date now.

I fixed the picture and went through the text again.  Added some more to the text,

1. More links to specific parts of documentation, specifically subscriptions, indications, wbemserver. 
2. Extend doc to indicate pypi vs current master versions for documentation
3. Added ref to presentations since those are sort of hidden today.
4. Add section on Contributions and Next Release sections to readme.

I still hate the fact that the picture cannot be scaled but that appears to be a limitation of the restructuredTxt resolver. I was wrong. httipie diagrams have same problem in their README.rst. Github tools limitation since they are doing the rendering, not your local browser.  When I use restview locally it scales the diagram correctly.

andy-maier:
Karl,
the mof_compiler and wbemcli tools have disappeared from the picture now. Was that intentional?

andy-maier:
Also, it appears the merging of the statistics PR caused a need to rebase this one.

KSchopmeyer:
It did and for some reason it was a bear to integrate.  the README.rst got all tangled up in itself.

Done and I added one sentence to the README pointing to the directions issue for 0.11.0

+1

andy-maier:
Karl, now the blue layer with "WBEM Connections, CIM Objects" disappeared.

andy-maier:
Karl, the blue layer with "WBEM Connections, CIM Objects" disappeared and needs to be re-established..

KSchopmeyer:
Fixed the diagram.  Wierd. That box was in the popwerpoint diagram group and locally it shows up when I do png back into powerpoint but when I write the picture to png file, it is not there.  Went back and ungrouped/regrouped complete diagraml and now it works and that blue box is in the result.

andy-maier:
+1

PR #688, opened 2017-04-24
Improved testcases for CIMInstanceName with different key binding order.

andy-maier:
Please review.

Details:
- Added a test that attempts to establish two `CIMInstanceName` objects with different iteration order of their key bindings, and then tests for equality.
- Removed the sorting of key bindings in `CIMInstanceName.__repr__()`.


andy-maier:
It appears that the new test is able to construct keybindings with different iteration order for Python 2.7 and 3.6, but not for 3.4 and 3.5. Apparently, Python 3.3 introduced a randomized hash seed, which probably causes that. Python 3.6 introduced that normal dicts maintain the order of insertion as the iteration order, that's why it works again there.

This is not a problem, it just means that on python 3.4 and 3.5, the test has a somewhat lower quality because it does not prove that the equal comparison also works for objects with differently ordered key bindings.

andy-maier:
+1

KSchopmeyer:
+1

PR #689, opened 2017-04-24
Pylint config changes.

andy-maier:
Please review.

Details:
- Set number of parallel pylint jobs to 4 for speedup.
- Disabled full pylint report because it added large amounts of source code.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -26,7 +26,7 @@ persistent=yes
 load-plugins=
 
 # Use multiple processes to speed up Pylint.
-jobs=1
+jobs=4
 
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
I tested and did not find much difference between using one job and 4 jobs and about the same (a few seconds) when running up to 8 jobs.   But there was some improvement.

I agree with the second one.  It is very useful locally but gains nothing on ci tests during commit, etc.

+1

andy-maier:
I also did not see much improvcement with multiple jobs, but the second change helped me also when looking at CI results, because it significantly reduced the size of the output.

PR #693, opened 2017-04-25
Fixes #630: Add --version (-V) to argparse for wbemcli and mof_compiler

### LINKED ISSUE #630: Minor but we have overlooked for long time. Should all clis have --version option
We should add the --version option to wbemcli and mof_compiler. That is the linux way.

KSchopmeyer:
Added a statement for version to all of the current cli tools


andy-maier:
+1

KSchopmeyer:
+1

PR #703, opened 2017-05-05
Fixes #670: Add documentation about case sensitivity

### LINKED ISSUE #670: WBEM Server get_selected_profiles. Compare is case sensitive
This test is case sensitive.  Trying to compare profile name array to find the Array profile returns no profile.  Not sure whether this is correct or not.  At the least this should be documented in the _server documentation.

DISCUSSION: Should the test be case sensitive?

This appears to be correct in that these are property values and property values are case sensitive.

The one change that might be logical is the documentation to document that the tests are case sensitive.

KSchopmeyer:
I did this one just to retest travis.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -317,13 +317,16 @@ class WBEMServer(object):
         Parameters:
 
             profile_org (:term:`string`) or None: the `RegisteredOrganization`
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1

PR #705, opened 2017-05-24
Fixed the Travis issue with installing swig.

andy-maier:
This is a quick fix for the issue that Travis fails since its Ubuntu update in early may, when "make develop" tries to install swig. For unknown reasons, installing it directly in the Travis control file works fine, and that's what this change does.

A follow-on change is supposed to provide the "real fix".


KSchopmeyer:
Passes all tests above except python 2.6 where we found another issue.

+1

andy-maier:
The Sphinx installation on py26 issue also needs to be solved. PR #706 does that, and includes the commit from this PR. Abandoning this PR here.

PR #710, opened 2017-06-14
Fixes #709: Correct code error in wbemcli.py

### LINKED ISSUE #709: wbemcli error in args for enable statistics
We set  a parameter of the argparser item for statistics incorrectly in the parser definition. Since this has not been released it is minor but must be fixed.  Used word verbose instead of enable_stats on dest=

KSchopmeyer:
Correct dest error. Add display of enable_stats for connection


andy-maier:
+1

PR #714, opened 2017-06-23
Fixes #312: Clarified MOF escaping in internal docstring of mofstr().

### LINKED ISSUE #312: DSP0004 issue: Escaping issue with \ in tomof.
Issue in mofstr with enclosed single quote. tommof output drops the escaped single quote mark that is in at least description strings in mof.  It completely drops the single quote. See example below.

This is an issue with the mofstr function in that it does nothing special at all with single quotes in strings.   I need to double check but I believe that DSP0004 states that they are escaped.

The original mof was:

```
Description (
"Indicates the PowerSupply\'s input frequency range "
"measured by the associated sensor. Range 1, 2 or both "
"can be specified using the values 2, 3 or 4, "
"respectively." ),
```

The mofoutput is"

```
Description (
"Indicates the PowerSupplys input frequency range measured by the "
"associated sensor. Range 1, 2 or both can be specified using the "
"values 2, 3 or 4, respectively."),
```

But even stranger, the compiler produces the following error:

```
[Deprecated { "No value"},
Description (
"Indicates the PowerSupplys input frequency range measured by the "
"associated sensor. Range 1, 2 or both can be specified us
```

Edit: Andy edited for proper formatting of MOF.

andy-maier:
Only changes in an internal docstring, to clarify what the function does.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -662,37 +662,37 @@ def mofstr(strvalue, indent=MOF_INDENT, maxline=MAX_MOF_LINE):
 
     DSP0004 defines MOF escape sequences for several more characters, but it
     does not require their use in MOF. For example, it is valid for a MOF
-    string constant to contain the characters U+000D (newline) or U+0009
-    (horizontal tab).
+    string constant to contain the (unescaped) characters U+000D (newline) or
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
+1

PR #715, opened 2017-06-23
Fixes #698: Made CIMDateTime always timezone-aware, re-enabled test_cim_types.py

### LINKED ISSUE #698: CIMDateTime does not enforce datetime awareness
The definition of the dtarg in CIMDateTIme includes the following in its documentation:

```          
 * A :class:`py:datetime.datetime` object must be timezone-aware
              (see :class:`~pywbem.MinutesFromUTC`) and will result in a point
              in time.
```
However the code does not include a test to insure awareness (i.e. existence of tzinfo) although the documentation says that the tzone must be part of the constructor. We have bunch of examples in our tests where the tzone is not there.

One issue is that if you use the Pywbem CIMDateTime compare function to compare naive and aware date times it generates a TypeError exception rather than returning a False indicating that they are not equal.

The other issue is that the documentation and code to not match.

Found this a recorder test where the datetime naive definition input with the request was compared with a datetime aware version produced by yaml dump/load which results in a TypeError exception: niave and aware datetimes cannot be compared)

It would be a real incompatibility to force the change to insure that tzinfo was part of the interface but this leaves us open to the TypeError exception when simply comparing two CIMDateTime objects.

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -72,6 +72,11 @@ Enhancements
 
 * Added unit test for recorder. See issue #676
 
+* Ensured that `CIMDateTime` objects for point in time values are
+  timezone-aware when supplied with a timezone-naive `datetime` object.
+  This does not change the behavior, but increases code clarity.
+  Clarified that in the documentation of  `CIMDateTime`. See issue #698.
+
 Bug fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
+1

PR #716, opened 2017-07-02
Fixes #687: Implement handling of WBEMServerResponseTime

### LINKED ISSUE #687: Add processing of WBEMServerResponseTime
The DMTF Specification provides for an optional response header 'WBEMServerResponseTime" that, if provided defines a measure in microseconds of the elapsed time required by the server to process the request and create a response.  We should extend pywbem to handle this and include it as part of the statistical information provided via the statistics processing in the client.

This is defined in section 6.3.15 of DSP0200 v 1.4.0

Issue #276 is a duplicate of this issue and has been closed as a duplicate.

KSchopmeyer:
Extend pywbem to get WBEMServerResponseTime from server and include as new attribute/property in WBEMConnection (last_server_response_time) and new statistic in statistics.

1. Extend cim_http.py to capture optional WBEM_ServerResponseTime header
and convert from microseconds to seconds.
2.Extend cim_operations to add new attribute _server_response_time and
new property server_response_time that holds either None or the server
response time in seconds.
3. Extend statistics to include this statistic and display it in reports
if it is provided by the wbem server.
4. Extend run_cimoperations.py to display the last_server_response_time
if it exists.

Adds entry to changes.rst

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -80,6 +80,16 @@ Enhancements
 * Extend the documentation to list support for specific non-specification
   features of some WBEM servers. Issue #653.
 
+* Extend cim_http.py, cim_operations.py, _statistics.py to handle optional
+  WBEMServerResponseTime header from WBEMServer.  This HTTP header reports
+  the server time in microseconds from request to response in the operation
+  response.  The extension adds the WBEMConnection property
+  last_server_response_time and places the time from the server into the
+  attribute for this property.
+  It also passes server_response_time to statistics so that max/min/avg are
+  maintained.  See issue # 687.
+
+
 Bug fixes
 ^^^^^^^^^
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -86,6 +86,12 @@ Enhancements
   response.  The extension adds the WBEMConnection property
   last_server_response_time and places the time from the server into the
   attribute for this property.
+
+* Extend pywbem to handle optional WBEMServerResponseTime header from a
+  WBEM server.  This HTTP header reports the server time in microseconds from
+  request to response in the operation response.  The extension adds the
+  WBEMConnection property `last_server_response_time` and places the time from
+  the server into the attribute for this property.
   It also passes server_response_time to statistics so that max/min/avg are
   maintained.  See issue # 687.
 
--- a/pywbem/_statistics.py
+++ b/pywbem/_statistics.py
@@ -324,7 +324,7 @@ class OperationStatistic(object):
             processing the operation.
 
           server_time (:class:`py:bool`)
-            Time in microseconds that the server optionally returns to the
+            Time in seconds that the server optionally returns to the
             client in the HTTP response defining the time from when the
             server received the request to when it started sending the
             response. If None, there is no time from the server.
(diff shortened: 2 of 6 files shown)


KSchopmeyer:
The following is the formattet table output with and without wbemserver information

```
================
Statistics (times in seconds, lengths in Bytes):
Count Excep            Time                 RequestLen            ReplyLen        Operation
        Cnt     Avg     Min     Max    Avg    Min    Max    Avg      Min      Max
    3     0   0.234   0.100   0.400   1233   1000   1500  26667    20000    35000 EnumerateInstances
    1     0   0.100   0.100   0.100   1200   1200   1200  22000    22000    22000 EnumerateInstanceNames
================

Statistics (times in seconds, lengths in Bytes):
Count Excep            Time               ServerTime              RequestLen               ReplyLen         Operation
        Cnt     Avg     Min     Max     Avg     Min     Max    Avg    Min    Max      Avg      Min      Max
    3     0   0.234   0.100   0.400   0.233   0.100   0.400   1233   1000   1500    26667    20000    35000 EnumerateInstances
    1     0   0.100   0.100   0.100   0.100   0.100   0.100   1200   1200   1200    22000    22000    22000 EnumerateInstanceNames
================
```

docs/changes.rst
```diff
@@ -77,6 +77,15 @@ Enhancements
   This does not change the behavior, but increases code clarity.
   Clarified that in the documentation of  `CIMDateTime`. See issue #698.
 
+* Extend cim_http.py, cim_operations.py, _statistics.py to handle optional
```
andy-maier: In the change history, I don't think we should talk about our internal modules. Describe it from a user perspective, instead.

docs/changes.rst
```diff
@@ -77,6 +77,15 @@ Enhancements
...
+  WBEMServerResponseTime header from WBEMServer.  This HTTP header reports
```
andy-maier: "WBEMServer" -> "WBEM server".

andy-maier:
-1, see my (minor) comments.

docs/changes.rst
KSchopmeyer: And yet another one. Done

KSchopmeyer:
We are having problems with python 2.6 and the test_wbemcli.py tests.
```
        if not expected_stderr:
            self.assertEqual(std_err, "",
                             '%s stderr not empty. returned %s'
>                            % (test_name, std_err))
E           AssertionError: test_help stderr not empty. returned /home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/coverage/config.py:29: DeprecationWarning: You passed a bytestring as `filenames`. This will not work on Python 3. Use `cp.read_file()` or switch to using Unicode strings across the board.
E             return configparser.RawConfigParser.read(self, filename, **kwargs)
```

Note that I force warnings off for the command by exporting PYTHONWARNINGS=""

```
        command = 'export PYTHONWARNINGS="" && %s' % command
```
 but this one still gets through. This cured an earlier problem with a warning from one of the imports so the code should be working. 

It also generates the following pytest warning message:

```
WC1 /home/travis/build/pywbem/pywbem/testsuite/test_recorder.py cannot collect test class 'TestClientRecorder' because it has a __init__ constructor
```

andy-maier:
+1

KSchopmeyer:
Andy. Did you mean to approve this one since I have the issue with python 2.6 testing.  Note that I found that coverage did have problems with their imports in version 4.0 that was causing exactly this warning but I cannot explain why a) setting the PYTHONWARNING flag to ignore does not seem to work and b) why it occurs only on python 2.6 when the issue they had was tied to python 2.6 and python 2.7

PR #717, opened 2017-07-02
Fixes #653 Add documenation to document non-specific server features

### LINKED ISSUE #653: There is special code for the openwbem server but no way to test or knowledge about why
This code is located in cim_http.py, wbem_request where pywbem is processing the response and looks for a response  header ' Server' that contains the 'openwbem' attriibute.  This appears to be part of authorization challange for OpenWBEM that includes  a special nonce but whose use in the response processing is limited to responses with that specific response header.

 Since the response header 'Server' is a special header, I propose that we simply ignore that fact that this code exists but that we cannot test it.

The only thing we might do is to make a comment in the documentation about special code for the openwebem server.

Since this is open source and we are in a minor version, there is no way this could be removed.

However, I assume that if we move to a different http library we will have to seriously consider if we should move this processing forward since it is 1-00+ lines of code and we have no way of testing it.

KSchopmeyer:
Adds section to intro to document a) not-spec features, b) level of testing against servers.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -78,7 +78,7 @@ Enhancements
   Clarified that in the documentation of  `CIMDateTime`. See issue #698.
 
 * Extend the documentation to list support for specific non-specification
-  features of some WBEMServers. Issue #653.
+  features of some WBEM servers. Issue #653.
 
 Bug fixes
 ^^^^^^^^^
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -77,6 +77,9 @@ Enhancements
   This does not change the behavior, but increases code clarity.
   Clarified that in the documentation of  `CIMDateTime`. See issue #698.
 
+* Extend the documentation to list support for specific non-specification
+  features of some WBEMServers. Issue #653.
+
 Bug fixes
 ^^^^^^^^^
 
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -391,3 +391,52 @@ Configuration variables
 .. automodule:: pywbem.config
       :members:
 
+.. _`WBEM Servers`:
+
+WBEM servers
+------------
+
+Pywbem supports communication with any WBEMServer that conforms to the
+DMTF specifications. See section  :ref:`Standards conformance`.
+
+Server Specific Features
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are some specific non-specification features that are included in
+pywbem for support of WBEM server specific features including:
+
+1. Openwbem server - includes support for an authentication extension
+(OWLocal password-less local authorization) that is part of the openwbem server.
+
+2. OpenPegasus - includes support for the special interop namespace that may
+be used in some OpenPegasus implements `root/PG_InterOp`. Most implementations
+have moved to supporting the standard namespace (`interop`, 'root\interop') but
+for backward compatibility, this old interop namespace name was included in
+the table of namespaces that are searched by the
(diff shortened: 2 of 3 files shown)


docs/intro.rst
```diff
@@ -391,3 +391,52 @@ Configuration variables
 .. automodule:: pywbem.config
       :members:
 
+.. _`WBEM Servers`:
+
+WBEM servers
+------------
+
+Pywbem supports communication with any WBEMServer that conforms to the
```
andy-maier: "WBEMServer" -> "WBEM server"

docs/intro.rst
```diff
@@ -391,3 +391,52 @@ Configuration variables
...
+DMTF specifications. See section  :ref:`Standards conformance`.
+
+Server Specific Features
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are some specific non-specification features that are included in
+pywbem for support of WBEM server specific features including:
+
+1. Openwbem server - includes support for an authentication extension
+(OWLocal password-less local authorization) that is part of the openwbem server.
```
andy-maier: "openwbem" -> "Openwbem"

docs/intro.rst
```diff
@@ -391,3 +391,52 @@ Configuration variables
...
+
+2. OpenPegasus - includes support for the special interop namespace that may
+be used in some OpenPegasus implements `root/PG_InterOp`. Most implementations
+have moved to supporting the standard namespace (`interop`, 'root\interop') but
```
andy-maier: Use forward slash: "root/interop"

docs/intro.rst
```diff
@@ -391,3 +391,52 @@ Configuration variables
...
+for backward compatibility, this old interop namespace name was included in
+the table of namespaces that are searched by the
+`WBEMServer.get_interop_namespace()` method).
+
+3. OpenPegasus - Supports a mixed-case attribute name `EmbeddedObject` in xml
```
andy-maier: "xml" -> "XML"

docs/intro.rst
```diff
@@ -391,3 +391,52 @@ Configuration variables
...
+
+* OpenPegasus PegasusUDSConnection
+* OpenWbem - OpenWBEMUDSConnection
+* SFCB(Small Footprint Broker) - SFCBUDSConnection
+
+WBEM Server Testing
+^^^^^^^^^^^^^^^^^^^
+
+Today the pywbem project tests primarily against current versions of the
+OpenPegasus WBEM Server because that server is available to the project.
```
andy-maier: "WBEM Server" -> "WBEM server"

andy-maier:
-1, see my comments

docs/changes.rst
```diff
@@ -77,6 +77,9 @@ Enhancements
   This does not change the behavior, but increases code clarity.
   Clarified that in the documentation of  `CIMDateTime`. See issue #698.
 
+* Extend the documentation to list support for specific non-specification
+  features of some WBEMServers. Issue #653.
```
andy-maier: "WBEMServers" -> "WBEM servers"

docs/intro.rst
KSchopmeyer: Done

docs/intro.rst
KSchopmeyer: Done

docs/intro.rst
KSchopmeyer: Done

PR #720, opened 2017-07-13
Fixes #700: Fix issue with recorder creating non-text files

### LINKED ISSUE #700: Recorder does not write utf8 data in python 2.7
In python 2.7, a simple file open will not create a file in which utf-8 characters can be written. The following error occurs

  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/_recorder.py", line 518, in record
    self._fp.write(data)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 2204: ordinal not in range(128)

Strictly this is not our issue since we assign the responsibility for file open to the user. Thus, for example, run_cim_operations must create a utf-8 writable file pointer with the open meaning that it must use something like rather than the built-in open:

    codecs.open(...)

The simplest solutions is just to document the requirement in the recorder documentation. 

The alternate is to provide a file opener in recorder.py or cim_operations.py that covers the issue.

DISCUSSION: Alternate solutions

For the moment I have put code into run_cimoperations.py and test_recorder.py to use the codecs.

KSchopmeyer:
1. Documented that if an already open file is supplied to the
TestClientRecorder, it must be a text file.

2. Since creating text files is messy in python 2/3 we added a static
method (open_file) to the recorderthat will open the file and return the
handle.

ex:
```
	fp = TestFileRecorder.open_file('yamloutputfile', 'a')
```
will open a test file for append.

We did NOT add any check to be sure that the file handle provided to the
TestClientRecorder __init__ function is, in fact defined as a text file.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -148,6 +148,12 @@ Bug fixes
 * Correct Error in run_cimoperations with use of namespace in iter... function
   See issue #718. This was a test code issue. No changes to the iter
   operations.
+
+* Correct issue with Recorder creating non-text files.  This issue
+  Documents the requirement for text files and also adds a static
+  method to force creation of the recorder output as a text file.
+  See issue # 700
+
   
 
 Build, test, quality
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -290,6 +290,7 @@ class BaseOperationRecorder(object):
             recorder = TestClientRecorder(open_file('recorder.log')
         """
         if six.PY2:
+            # Open with codecs to define text mode
             return codecs.open(filename, mode=file_mode, encoding='utf-8')
 
         return open(filename, file_mode, encoding='utf8')
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -37,6 +37,9 @@ from .cim_obj import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
 from .cim_types import CIMInt, CIMFloat, CIMDateTime
 from .exceptions import CIMError
 
+if six.PY2:
+    import codecs  # pylint: disable=wrong-import-order
+
 __all__ = ['BaseOperationRecorder', 'TestClientRecorder',
            'OpArgs', 'OpResult', 'HttpRequest', 'HttpResponse']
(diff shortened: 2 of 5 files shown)


KSchopmeyer:
# compare all properties returned against original
        self.assertEqual(len(pd), len(NewInstance.properties))
        for pn, pv in pd.items():
            prop = CIMProperty(pn, pv['value'], type=pv['type'],
                               array_size=pv['array_size'],
                               propagated=pv['propagated'],
                               is_array=pv['is_array'],
                               reference_class=pv['reference_class'],
                               qualifiers=pv['qualifiers'],
                               embedded_object=pv['embedded_object'])
    
            # temp display while sorting out datetime issues.
            # if pv['type'] == 'datetime':
            #    print('datetime prop name:%s orig:%s new:%s' %
            #          (pn, NewInstance.properties[pn], prop))
    
            self.assertEqual(NewInstance.properties[pn], prop,
                             'Property compare failed orig %s, recreated %s' %
>                            (NewInstance.properties[pn], prop))
E           AssertionError: CIMPr[21 chars]ue='H�m', type='string', reference_class=None,[113 chars]({})) != CIMPr[21 chars]ue='Häm', type='string', reference_class=None[114 chars]({})) : Property compare failed orig CIMProperty(name='S2', value='H�m', type='string', reference_class=None, embedded_object=None, is_array=False, ...), recreated CIMProperty(name='S2', value='Häm', type='string', reference_class=None, embedded_object=None, is_array=False, ...)
testsuite\test_recorder.py:418: AssertionError
============================== warnings summary ===============================
testsuite/test_recorder.py::TestClientRecorder

KSchopmeyer:
I fixed by forcing reads of the file (both in test_client.py and test_recorder.py) to be utf-8. Note that we used two different techniques (codecs in recorder and io in test_client and test_recorder) this is because the general wisdom is that the codecs are faster (rumour).

Andy: Please review this one since we are playing with fire with codecs, io, and utf-8

KSchopmeyer:
+1

andy-maier:
+1. I am squashing the two commits on merge, because the second one fixes issues in the first one.

PR #722, opened 2017-07-14
Fixes #719: Fixed py.test skip warning about TestClientRecorder.

### LINKED ISSUE #719: Warning message in result of travis 2.6 and 2.7 tests.
WC1 /home/travis/build/pywbem/pywbem/testsuite/test_recorder.py cannot collect test class 'TestClientRecorder' because it has a __init__ constructor

COMMENT: This warning occurs in the python 2.7 tests also.  Since it is only a warning and does not stop the tests or cause an error, I propose we write an issue and ignore for now.

Per Andy: The issue is that you cannot have a __init__ in your class if you use pytest.

I am confused here.  1 we are using unittest in this test code. Second, there is no __init__ in any of the test code in this module.

Comment on what py.test does.
```
The documentation for py.test says that py.test implements the following standard test discovery:

    collection starts from the initial command line arguments which may be directories, filenames or test ids. recurse into directories, unless they match norecursedirs
       -  test_*.py or *_test.py files, imported by their package name.
       - Test prefixed test classes (without an __init__ method) [<-- notice this one here]
       - test_ prefixed test functions or methods are test items

So it's not that the constructor isn't needed, py.test just ignores classes that have a constructor. There is also a guide for changing the standard test discovery.
```

andy-maier:


andy-maier:
I verified (in Travis run on py26) that the skip warning about TestClientRecorder no longer shows up.

PR #731, opened 2017-07-26
Fixes #598: Fixed issue with ref class check in CIMProperty.

### LINKED ISSUE #598: reference instances return REFERENCECLASS property in test_client

andy-maier:
**Note:** This PR is on top of PR #738.

This PR is now complete and is ready for review and merge.

@KSchopmeyer: Karl, if you still want to extend the testcases, it would probably be easier if you did that in a separate PR on top of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ install:
   - pip list
   - make develop
   - pip install python-coveralls
+# Update coverage because the version defined for python-coveralls breaks
+# some of our tests (they specify a specific version) see pr #734 where at
+# least on python 2.6, the test_wbemcli.py code catches a deprecation warning
+# from coverage 4.03 or earlier.
+  - pip install -U coverage
   - pip list
 
 # commands to run builds & tests
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -44,18 +44,18 @@ Enhancements
   request/reply size information on server requests and replies. The detailed
   information is available in WBEMConnection for operation execution time
   and request/reply content size at the end of each operation.
-  
+
   When statistics gathering is enabled, the information is placed into the
   Statistics class where min/max/avg information is available for each
   operation type.
   Statistics gathering is enabled if the WBEMConnection attribute
   `enable_stats` is `True`.
-  
+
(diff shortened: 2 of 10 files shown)


KSchopmeyer:
Not yet sure what problem is but this is dying ig GetAllTypes1 test.  It also fails ReferenceInstances1, the onethat is marked ignore_test but the error is in Test_client iteslf, not in cim_obj.py.  Will look forther later tonight

KSchopmeyer:
The test is dying in test_client apparently OpenReferenceInstancePaths1. 

Changing the tupleparser affects nothing.

However, restoring the original code in cim_obj.py causes test_client to succeed (not that this is without the ReferencesInstances1 test which is the one that was failing originally)

andy-maier:
We found out that testcase `GetAllTypes1` in `test_client/getalltypes.yaml` was missing to specify `reference_class` attributes for its `PRef1` and `PRef2` properties. They had previously be inferred from the classname of the CIMInstancePath, but that was exactly the fix of this PR to no longer do that. So they now need to be specified explicitly in the testcase.

andy-maier:
I made the fix that we discussed in the call, and made it in such a way that one reference property returns the REFERENCECLASS attribute, and the other does not.

I also rebased this PR on top of PR #738 to circumvent the recent flake8 issue.

PR #735, opened 2017-07-26
Fixes #734: Fix issue with old version of coverage

### LINKED ISSUE #734: Coverage generates deprecation errors because old version installed
The environment loads an old version of coverage as part of dependency installation.  This version generates a number of warning messages that interfere with other tests, specifically the wbemcli test in another pr.  Any of the 4.0 versions are have the problem

This forces the correct version of coverage.

KSchopmeyer:
This fixes issue where coveralls forced installation of old version of
coverage that causes problems with other tests because of deprecation
warnings it issues.  The problem was fixed after version 4.03 of
coverage
We had to directly change .travis.yml to reinstall/update coverage after
the installation in the setup and reinstallation as a coveralls
dependency with an old version.

We added the version to setup.py but that is mostly for documentation.

Added comment to changes.rst

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ install:
   - pip list
   - make develop
   - pip install python-coveralls
+# Update coverage because the version defined for python-coveralls breaks
+# some of our tests (they specify a specific version) see pr #734 where at
+# least on python 2.6, the test_wbemcli.py code catches a deprecation warning
+# from coverage 4.03 or earlier.
+  - pip install -U coverage
   - pip list
 
 # commands to run builds & tests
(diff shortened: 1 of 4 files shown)


andy-maier:
The reason this one fails is that the new flake8 3.4.1 stumbles across the unused `import multiprocessing ` in `setup.py` on line 53, and does not seem to honor the disabling of that error on the same line anymore. Don't know why it no longer honors that.

andy-maier:
The flake8 failure can be circumvented by rebasing this PR on the commit of PR #738.

KSchopmeyer:
cherrypicked the flake8 fix.

KSchopmeyer:
+1

PR #736, opened 2017-07-26
Fixes #569: Add test tool for wbemcli script.

### LINKED ISSUE #569: Add a concept for unit testing the wbemcli and mof_compiler scripts.
The problem is that there is significant code in these scripts but it is not tested in unit tests and therefore is negative hit on coverage as well as being open to undiscovered bugs.

Generally some of the possible solutions include:

1. Move much of the wbemcli code back into testable functions within the pywbem library and leave only a thin shell within the wbemcli file itself.  Several web sites propose this.  Thus, for example, the parser creater could become a function or method, the argparse execution another.  Then we could create unit tests that would simply create the argparser and execute various cmdline options against it testing the results.  Further we could isolate a number of the functions within wbemcli and make them testable with unittests.

Then wbemcli itself becomes a much smaller script and a large part of the code is unittested.

Here are a couple of examples:

   http://stackoverflow.com/questions/12809970/python-unittest-test-a-script-with-command-line-args

   http://dustinrcollins.com/testing-python-command-line-apps

2. Write a unittest that executes the wbemcli as a subprocess and tests results. This is probably fairly limited because it would either have to have mock support to really test much.  We could use it as a regression test for the help, for example but pretty much everything else would have to have a mock component.

3. Use the scripts extension to run wbemcli and test various components.  Thus, if we agree to implement the scripts option (and maybe extend it) we might be able to use this by itself to test at least some of the functionality of wbemcli but not sure how we tie it into overall testingat this point.  You could, for example have test scripts that tested the various cmdline options and did a pass/fail but never passed control on to the interactive console (This would mean making the opts available to the test script).

4. Extend our unittest concepts to be able to run scripts as part of the unittests.

KSchopmeyer:
Adds a unit test that executes wbemcli and tests responses. This is primarily the structure. There are only a couple of tests added.  We have a number of other tests to add, largely through the pr's for logger.

**This pr requires prs 735, 734 and 732 before it will execute correctly.** NOTE: Used cherrypick, etc. to achieve that so that the script now works and passes tests.

This script uses a map of multiple test definitions and expected results and dynamically generates the tests against wbemcli.

The submission includes a wbemcli script that will cause wbemcli to close after each test that is passes to wbemcli as part of the call with the -s option.

Note that this tool cannot actually execute any  tests against a server since there is no server so the tests are primarily setup, configuration, and errors in input parameters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ install:
   - pip list
   - make develop
   - pip install python-coveralls
+# Update coverage because the version defined for python-coveralls breaks
+# some of our tests (they specify a specific version) see pr #734 where at
+# least on python 2.6, the test_wbemcli.py code catches a deprecation warning
+# from coverage 4.03 or earlier.
+  - pip install -U coverage
   - pip list
 
 # commands to run builds & tests
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -95,6 +95,9 @@ Enhancements
   It also passes server_response_time to statistics so that max/min/avg are
   maintained.  See issue # 687.
 
+* Add test for wbemcli script that will execute the script and test
+  results. issue #569
+
 
 Bug fixes
 ^^^^^^^^^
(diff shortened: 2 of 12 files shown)


andy-maier:
The flake8 failure can be circumvented by rebasing this PR on the commit of PR #738.

PR #737, opened 2017-07-28
Ks/#691 part1 define logger

KSchopmeyer:


KSchopmeyer:
This one also is tied to pr #738. Will wait for commit to update.

KSchopmeyer:
**Discussion:**  I I originally created the logger options with 3 parameters for each named logger.

Destination: file, stderr, etc.
LogLevel:   Python levels: ... DEBUG, INFO
detailLevel: this was the one that defines how much of each http response, etc. is logged.  Right now that is either 'min' or nothing

I think that was a mistake.

1. For loggers like the http logger, everything is just at DEBUG level.  So until we add other loggers, this characteristics really does not matter.
2. We can accomplish the same thing by using log level.  We log the full http response and operation response at DEBUG level.  At INFO level we log only a specified length(xxx bytes of http and operation responses).  We could even extend to log only errors (exceptions, etc.) and the next level down (either WARNING or ERROR.

That would mean eliminating that middle component of the logging parameters and tie the whole thing into just the existing logger components.

The code change is minimal for this and I think it is more understandable.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
+# Copyright 2017 InovaDevelopment Inc. All Rights Reserved.
+
```
andy-maier: Remove empty line

pywbem/__init__.py
```diff
@@ -45,6 +45,7 @@
 from ._recorder import *  # noqa: F403,F401
 from .config import *  # noqa: F403,F401
 from ._statistics import *  # noqa: F403,F401
+from ._logging import *  # noqa: F403,F401
```
andy-maier: General comment (not for here): Mention the change in the change history.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+      format_string = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+      logging.basicConfig(format=format_string, level=logging.DEBUG)
+"""
+from __future__ import absolute_import
+
+import logging
+import six
+
+# from ._constants import API_LOGGER_NAME
+API_LOGGER_NAME = 'pywbem.api'
```
andy-maier: The API logger was not added. Remove this constant.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+"""
+This package supports logging using the standard Python :mod:`py:logging`
+module. The logging support for multiple :class:`~py:logging.Logger` objects
+(named loggers):
+
+Currently there are two defined in this code:
```
andy-maier: Remove "Currently". It is as defined right now, and we don't need to play that down.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+* 'pywbem.ops' for user-issued calls to pywbem CIM OperationsAPI functions, at
+  the info level. Internal calls to API functions are not logged.
+
+* 'pywbem.http' for http level information, at the info level.
```
andy-maier: What does "at the info level" mean? If it means that these loggers produce log entries at info level or higher levels, we should say that in more specific words. This comment applies to both loggers.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+"""
+This package supports logging using the standard Python :mod:`py:logging`
+module. The logging support for multiple :class:`~py:logging.Logger` objects
+(named loggers):
+
+Currently there are two defined in this code:
+
+* 'pywbem.ops' for user-issued calls to pywbem CIM OperationsAPI functions, at
```
andy-maier: On "pywbem CIM OperationsAPI functions": This is a bit awkward. I suggest replacing it with:
```
pywbem API functions that drive WBEM operations (see :ref:`WBEM operations`).
```

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+  the info level. Internal calls to API functions are not logged.
+
+* 'pywbem.http' for http level information, at the info level.
+
+In addition, there may be loggers for pywbem modules with the module name, for
+situations like errors or warnings.
```
andy-maier: There are no such loggers today. Remove the paragraph.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+All these loggers have a null-handler (see :class:`~py:logging.NullHandler`)
```
andy-maier: Add "initially" after "All these loggers".

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+If you want to change the default log message format, use
+:meth:`~py:logging.Handler.setFormatter`. Its ``form`` parameter is a format
+string with %-style placeholders for the log record attributes (see Python
+section :ref:`py:logrecord-attributes`).
+
+Examples:
+
+* To output log records for one of the defined named loggers, either the
+  PywbemLogger.create_loggers or the Pywbem.create_logger can be used.
```
andy-maier: `PywbemLogger.create_loggers` should be a reference to the method: `:meth:'PywbemLoggers.create_loggers'`. (and this would have caught the missing `s`).
Similar for `Pywbem.create_logger`.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+  create_loggers creates one or more loggers from a string input that
```
andy-maier: Change `create_loggers` to a method reference.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+  defines the component name and characteristics of each logger.  This allows
+  other tools like clis to create the pywbem known logs with minimal work
+  from command line or config file input.
+
+  create_logger creates one or more loggers from parameter inputs that
```
andy-maier: Change `create_logger` to a method reference.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+section :ref:`py:logrecord-attributes`).
+
+Examples:
+
+* To output log records for one of the defined named loggers, either the
+  PywbemLogger.create_loggers or the Pywbem.create_logger can be used.
+
+  create_loggers creates one or more loggers from a string input that
+  defines the component name and characteristics of each logger.  This allows
+  other tools like clis to create the pywbem known logs with minimal work
```
andy-maier: clis -> CLIs

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+  from command line or config file input.
+
+  create_logger creates one or more loggers from parameter inputs that
+  define the component name and characteristics of each logger.
+
+* Also, loggers can be created by calling the logging  methods directly as
```
andy-maier: I would not say that this example creates loggers, but it configures the pywbem loggers.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+   in this example.To output the log records for all cim operations to
```
andy-maier: 1. insert a space after the sentence.
2. "cim operations" -> "WBEM operations".

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+"""
+from __future__ import absolute_import
+
+import logging
+import six
+
+# from ._constants import API_LOGGER_NAME
+API_LOGGER_NAME = 'pywbem.api'
+
+#: Name of logger for operantion call methods and their responses.
```
andy-maier: 1. Typo. "operantion"
2. I would change the description to: "Name of logger for logging user-issued calls to pywbem API functions that drive WBEM operations." to be consistent with how this logger is described further up.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+#:
+LOG_OPS_CALLS_NAME = 'pywbem.ops'
+
+#: Name of logger for http information
+#:
+LOG_HTTP_NAME = 'pywbem.http'
+
+#: pywbem logger names. The following are the logger names that pywbem
+# recognizes.
```
andy-maier: `#` -> `#:` and remove the empty line after the comment, to make the doc generator process this correctly.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+PYWBEM_LOG_COMPONENTS = ['ops', 'http', 'all']
```
andy-maier: The prefix `PYWBEM_` is not really needed. The constant is scoped to the pywbem namespace anyway.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+#:
+LOG_OPS_CALLS_NAME = 'pywbem.ops'
+
+#: Name of logger for http information
+#:
+LOG_HTTP_NAME = 'pywbem.http'
+
+#: pywbem logger names. The following are the logger names that pywbem
+# recognizes.
+
```
andy-maier: Add external description (i.e. with `#:`)

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+PYWBEM_LOG_COMPONENTS = ['ops', 'http', 'all']
+
+# possible log output destinations
+LOG_DESTINATIONS = ['file', 'stderr', 'none']
+
+LOG_DETAIL_LEVELS = ['all', 'min']
```
andy-maier: Add external description (i.e. with `#:`)

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+DEFAULT_LOG_DETAIL_LEVEL = 'min'
```
andy-maier: Add external description (i.e. with `#:`)

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+DEFAULT_LOG_LEVEL = 'debug'
```
andy-maier: Add external description (i.e. with `#:`)

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+LOG_LEVELS = ['error', 'warning', 'info', 'debug']
```
andy-maier: Add external description (i.e. with `#:`)

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+        Setup a single named logger with the characteristics defined on input.
+
+        This function can be used to set up all of the named loggers used by
+        pywbem.
+
+        Parameters:
+          log_compnent (:term:`string`):
+           The name of the logger. It must be one of the
+           names defined in PYWBEM_LOG_COMPONENTS. Subsequent logging users
+           will use this name to reference the parameters of this logger
```
andy-maier: The last sentence does not make sense to me. The logger is referenced by its name, not by its log component.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+                      log_filename=None,
+                      log_detail_level='all', log_level='debug'):
+        """
+        Setup a single named logger with the characteristics defined on input.
+
+        This function can be used to set up all of the named loggers used by
+        pywbem.
+
+        Parameters:
+          log_compnent (:term:`string`):
```
andy-maier: Typo: `log_compnent`

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+    @classmethod
+    def reset(cls):
+        """Reset the logger dictionary. Used primarily in unittests"""
+        cls.loggers = {}
+
+    @classmethod
+    def create_loggers(cls, input_str, log_filename=None):
+        """
+        Create the pywbem loggers defined by the input string in the following
+        format and place in a class level dictionary in this class:
```
andy-maier: The first sentence of this method description is inconsistent with the first sentence of the description of the `create_logger` method. Make it consistent.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+      PywbemLoggers.create_logger - Creates a single logger fromthe
+      separate pywbem logging attributes.
+
+    """
+    __metaclass__ = MetaPywbemLoggers
+
+    @classmethod
+    def reset(cls):
+        """Reset the logger dictionary. Used primarily in unittests"""
```
andy-maier: This method has some built-in inconsistency, because it resets the dictionary but does not reset the logger objects.
Why does a unit test need to reset this dictionary? Is it because of the singleton design? There must be a better way to do that instead of providing a documented public method for that purpose.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+    @classmethod
+    def create_loggers(cls, input_str, log_filename=None):
+        """
+        Create the pywbem loggers defined by the input string in the following
+        format and place in a class level dictionary in this class:
+
+        Parameters:
+          input_str (:term:`string`) that specifies the logger definitions:
+            as follows:
+
```
andy-maier: Make formatting use monospace font.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+            log_spec[,log_spec]
+
+            log_spec := log_comp['=' [dest][":"[detail_level][":"[log_level]]]]
+
+            where:
+                comp (str) must be one of PYWBEM_LOG_COMPONENTS
```
andy-maier: `comp` -> `log_comp`

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+                log_level (str) must be one of LOG_LEVELS above
+                detail_level (str) must be one of LOG_DETAIL_LEVELS
+                dest (str) must be one of LOG_DESTINATIONS
+
+          log_filename (:term:`string`)
+            Optional string that defines the filename for output of logs
+            if the dest type is `file`
+
+        Exceptions:
+          ValueError if any of the parameters are invalid
```
andy-maier: An example section would be good, here.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Parameters:
+          log_compnent (:term:`string`):
+           The name of the logger. It must be one of the
+           names defined in PYWBEM_LOG_COMPONENTS. Subsequent logging users
+           will use this name to reference the parameters of this logger
+
+          log_dest (:term:`string`):
+            String defining the destination for this log. It must be one of the
+            destinations defined in LOG_DESTINATIONS
```
andy-maier: Change `LOG_DESTINATIONS` to be a reference that generates a link to it.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+          log_filename (:term:`string`):
+            Filename to use as logging file if the log destination is `file`.
+            Ignored if log destination is not `file`. The default is
+            DEFAULT_LOG_FILENAME defined in config.py
```
andy-maier: Remove "in config.py" and change `DEFAULT_LOG_FILENAME` to be a reference that generates a link to it.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+          log_detail_level (:term:`string`):
+            String defining the level of detail for log output. This is
+            optional.
+            Not all of the loggers use this attribute.  Thus, the ops and http
+            loggers used a defined detail level and always set logging to that
+            level.
```
andy-maier: Are the last two sentences above a statement about the default behavior? If so, state that. If not, I don't understand what this is supposed to mean.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+          log_level (:term:`string`):
+            String defining the log level. This is optional.  Not all of the
+            loggers use this attribute.  Thus, the ops and http loggers used a
+            defined level and always set logging to that level.
```
andy-maier: Are the last two sentences above a statement about the default behavior? If so, state that. If not, I don't understand what this is supposed to mean.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+            destinations defined in LOG_DESTINATIONS
+
+          log_filename (:term:`string`):
+            Filename to use as logging file if the log destination is `file`.
+            Ignored if log destination is not `file`. The default is
+            DEFAULT_LOG_FILENAME defined in config.py
+
+          log_detail_level (:term:`string`):
+            String defining the level of detail for log output. This is
+            optional.
```
andy-maier: Add a reference to the constant defining the default.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+            Not all of the loggers use this attribute.  Thus, the ops and http
+            loggers used a defined detail level and always set logging to that
+            level.
+
+          log_level (:term:`string`):
+            String defining the log level. This is optional.  Not all of the
```
andy-maier: Add a reference to the constant defining the default.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+            loggers use this attribute.  Thus, the ops and http loggers used a
+            defined level and always set logging to that level.
+
+        Return:
+            None
```
andy-maier: If `None` is returned, remove the entire Return section.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Exceptions:
+            ValueError - Input cannot be mapped to a defined logger or
+            component.
+            No named logger is created.
```
andy-maier: The description of this exception is confusing (particularly the part that talks about mapping to a defined logger). This exception always means that input parameters were invalid.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        if log_dest not in LOG_DESTINATIONS:
+            raise ValueError('Invalid log destination %s. Not in list %s' %
+                             (log_dest, LOG_DESTINATIONS))
```
andy-maier: Instead of "Not in list %s", I would write: "Valid log destinations are: %s"

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+            raise ValueError('Invalid log detail %s. Not in list %s' %
+                             (log_detail_level, LOG_DETAIL_LEVELS))
+
+        # If destinations or components is all, recurse for all destinations
+        if log_dest == 'all':
+            for dest in LOG_DESTINATIONS:
+                cls.create_logger(log_component, dest,
+                                  log_filename=log_filename,
+                                  log_detail_level=log_detail_level,
+                                  log_level=log_level)
```
andy-maier: The statements between lines 264 and 269 should be removed, because there is no destination "all".

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+        elif log_component == 'all':
+            for comp in PYWBEM_LOG_COMPONENTS:
+                if comp != 'all':
+                    cls.create_logger(comp, log_dest=log_dest,
+                                      log_filename=log_filename,
+                                      log_detail_level=log_detail_level,
+                                      log_level=log_level)
+
+        # Otherwise process results of any recursive calls above
+        else:
```
andy-maier: Te checks between lines 245 and 262 can really be moved to here. In their current position, they are needlessly performed one extra time in case of components = all, plus it is more logical to have then here, and maybe some of them can be merged with some of the other checks here.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+                                        log_dest, log_filename)
+
+    @classmethod
+    def get_logger(cls, logger_name):
+        """
+            Get the name of a logger that has been defined.
+        """
+        if logger_name in cls.loggers:
+            return cls.loggers[logger_name][0]
+        return None
```
andy-maier: This whole `get_logger()` method can be eliminated if the users use `get_logger_info()`.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+    @classmethod
+    def get_logger_info(cls, logger_name):
+        """Get information about a logger by name"""
```
andy-maier: Add documentation about the return value. Specify the input parameter in the documentation.

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+        if logger_name in cls.loggers:
+            t = cls.loggers[logger_name]
+            return 'comp=%s dest=%s detail=%s log_level=%s ' \
+                   'file=%s' % (logger_name, t[3], t[0], t[2], t[3])
+        return('%s not defined.' % logger_name)
```
andy-maier: I suggest that the return value of this method is changed to be a tuple, and not a string. If the logger is not found, let it raise KeyError (as a byproduct of accessing the dict without check).

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+
+    A :class:`~py:logging.NullHandler` handler is added to the logger if it
+    does not have any handlers yet. This prevents the propagation of log
+    requests up the Python logger hierarchy, and therefore causes this package
+    to be silent by default.
+    """
+    logger = logging.getLogger(name)
+    if not logger.handlers:
+        logger.addHandler(logging.NullHandler())
+    return logger
```
andy-maier: If the recommended approach is to use `PywbemLoggers.create_logger()` to create a logger, why do we have the `get_logger()` function in addition?

pywbem/config.py
```diff
@@ -60,3 +61,11 @@
 #: operations.
 
 DEFAULT_ITER_MAXOBJECTCOUNT = 1000
+
+#: Maximum log entry size. An integer that sets the maximum size of each log
+#: entry if the log_detail_level flag is set.
+
```
andy-maier: Remove empty line.

pywbem/config.py
```diff
@@ -60,3 +61,11 @@
...
+DEFAULT_MAX_LOG_ENTRY_SIZE = 1000
```
andy-maier: If this is a default max, how can it be changed? Is it not rather just the max?

pywbem/config.py
```diff
@@ -60,3 +61,11 @@
...
+
+#: Set default log destination.  None means that there will be no logging
+DEFAULT_LOG_DESTINATION = 'file'
```
andy-maier: Why do we have `DEFAULT_LOG_DESTINATION` here in `config.py`, but `DEFAULT_LOG_DETAIL_LEVEL` and `DEFAULT_LOG_DETAIL_LEVEL` are in `_logger.py`?

pywbem/__init__.py
KSchopmeyer: I was holding the change history change for the last of the parts of this change rather than piece parting it in.

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done. Also remove that usage in the documentation

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: DONE.

pywbem/_logging.py
KSchopmeyer: Done. Made one more minor change to WBEMConnection methods that drive WBEM...

pywbem/_logging.py
KSchopmeyer: Gone

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Discarded whole paragraph since we really do depend on the named loggers being created by
the PywbemLoggers to capture the log detail.

pywbem/_logging.py
KSchopmeyer: 1. Done. 2. Done

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+import six
+
+# from ._constants import API_LOGGER_NAME
+API_LOGGER_NAME = 'pywbem.api'
+
+#: Name of logger for operantion call methods and their responses.
+#:
+LOG_OPS_CALLS_NAME = 'pywbem.ops'
+
+#: Name of logger for http information
```
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Good catch on this one

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+LOG_DESTINATIONS = ['file', 'stderr', 'none']
+
+LOG_DETAIL_LEVELS = ['all', 'min']
+
+DEFAULT_LOG_DETAIL_LEVEL = 'min'
+DEFAULT_LOG_LEVEL = 'debug'
+
+LOG_LEVELS = ['error', 'warning', 'info', 'debug']
+
+__all__ = ['PywbemLoggers', 'LOG_DESTINATIONS', 'LOG_LEVELS',
```
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Good. Done.

pywbem/_logging.py
KSchopmeyer: done

pywbem/_logging.py
```diff
@@ -0,0 +1,402 @@
...
+#: Name of logger for http information
+#:
+LOG_HTTP_NAME = 'pywbem.http'
+
+#: pywbem logger names. The following are the logger names that pywbem
+# recognizes.
+
+PYWBEM_LOG_COMPONENTS = ['ops', 'http', 'all']
+
+# possible log output destinations
```
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done:

pywbem/_logging.py
KSchopmeyer: DONE

pywbem/_logging.py
KSchopmeyer: DONE

pywbem/_logging.py
KSchopmeyer: Changed to state that "Used to create the named logger by
           prepending with the roon logger name 'pywbem'."

pywbem/_logging.py
KSchopmeyer: Done

pywbem/config.py
KSchopmeyer: Sloppy coding. It should all be here.  I believe that the DEFAULT_LOG_SIZE should also be in logging and not config so moved it.

pywbem/config.py
KSchopmeyer: First, moved it to _logging.py.  Second, right now we have no way to allow a user to c hange this so it is really the max so I changed the name.

pywbem/config.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done.  Also changed so it returns None if no logger defined.  Note that this needs anther review
to be consistent with changes to some of these functions.

pywbem/_logging.py
KSchopmeyer: Agreed and done.

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Agreed. Done to all uses.

pywbem/_logging.py
KSchopmeyer: Input contains an invalid log destination, log level,  or log detail level.
            No named logger is configured.

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Done. Recheck all of these descriptions for same thing

pywbem/__init__.py
KSchopmeyer: I am putting the part of this issue tied to logging component into this pr and we will do the rest with the next pr

pywbem/_logging.py
KSchopmeyer: Tried to bring the two into line.

pywbem/_logging.py
KSchopmeyer: I agree.  I am removing the function and just doing it in tests with internal knowledge of the implementation.

pywbem/_logging.py
KSchopmeyer: Done.

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: DISCUSSION:  right now we fix each logger level.  I think we should use this as the level at which we log.  That is only documentation change here but I will have to change the recorder to get the log level from here rather than fixing it.

pywbem/_logging.py
KSchopmeyer: Dropped the whole statement.  We will use this as the log level for the logs.

pywbem/_logging.py
KSchopmeyer: TODO

pywbem/_logging.py
KSchopmeyer: TODO. Code changed. Need to clean up documentation

pywbem/_logging.py
KSchopmeyer: Done I think

pywbem/_logging.py
KSchopmeyer: Done but comments not clear yet

pywbem/_logging.py
KSchopmeyer: Done. moved all the checks below the else:

KSchopmeyer:
I have made all changes (and commented on the changes) except:

1. The refs to the lists of allowed strings (I have a question about those.

2. The issue about get_logger in general

3. Should we be doing create_loggers or configure loggers

Finally, I want to review the tests to be sure we are really covered.

I am committing since we are closing up here at plugfest and driving to Reno in few minute  and I want this recorded in git for tomorrow morning

pywbem/_logging.py
```diff
@@ -0,0 +1,407 @@
...
+        in a class level dictionary in this class.
+
+        This function can be used to set up all of the named loggers used by
+        pywbem.
+
+        Parameters:
+          log_component (:term:`string`):
+           The name of the logger. It must be one of the
+           names defined in LOG_COMPONENTS. Used to create the named logger by
+           prepending with the roon logger name 'pywbem'.
```
andy-maier: 1. Change "named logger" to "logger name".
2. Typo "roon". Beyond that, the term "root logger" in python is used for the logger named "". that is not the case here. So I suggest "with the roon logger name 'pywbem'." -> "with the prefix `pywbem.`."

pywbem/_logging.py
```diff
@@ -0,0 +1,407 @@
...
+            assert isinstance(level, int)
+            if level is None:
+                raise ValueError('Invalid log level %s specified. Must be one '
+                                 'of %s.' % (log_level, LOG_LEVELS))
+
+            # create named logger
+            if handler:
+                handler.setFormatter(logging.Formatter(format_string))
+                logger = logging.getLogger(logger_name)
+                logger.addHandler(handler)
```
andy-maier: Using `addHandler() ` adds a second handler if this method is invoked a second time for the same log_component. This should be replaced with whatever is needed to replace the existing handler.

pywbem/_logging.py
```diff
@@ -0,0 +1,407 @@
...
+        for name, value in six.iteritems(results):
+            cls.create_logger(name, log_dest=value[0],
+                              log_detail_level=value[1],
+                              log_level=value[2],
+                              log_filename=log_filename)
+
+    @classmethod
+    def create_logger(cls, log_component, log_dest='stderr',
+                      log_filename=None,
+                      log_detail_level='all', log_level='debug'):
```
andy-maier: The name of the `create_logger()` method should be changed to e.g. `get_logger()` because it really does what Python `logging.getLogger()` does: Return the existing logger if it exists, or create a new logger otherwise.

docs/client.rst
```diff
@@ -289,6 +290,20 @@ Statistics
    :members:
    :special-members: __str__
 
+.. _`Logging`:
+
+Logging
+-------
+
+.. automodule:: pywbem._logging
+
```
andy-maier: Add `:members:` below the automodule statement, to make the global constants show up.

When I did that, I got the documentation of the `PywbemLoggers` class twice. So maybe the autoclass statement for that class needs to be removed? You need to try that out.

KSchopmeyer: DONE. That works.

pywbem/_logging.py
KSchopmeyer: Done

pywbem/_logging.py
KSchopmeyer: Added code to remove any handlers before adding the new one.

pywbem/_logging.py
KSchopmeyer: I have a problem generating the ref for links to the variables defined in this module. 

DISCUSSION: how do we generate the anchor for entities that are defined by #: comment???

pywbem/_logging.py
KSchopmeyer: See comment above about generating refs to constants.

KSchopmeyer:
I still have the followign issues:

1. the anchor for #: variables

2. The create_logger vs get_logger and logger_name vs log_component.

KSchopmeyer:
We agreed that we have issues with this design as follows:

1. We do not need the log_level at all since it is just a flag and does not really set anything.  The log writers are at the debug level for http and ops. All this does is set the log level into the named loggers which simply turns these logs off or on.  Thus, we move from the strings in wbemcli that are (ex. http=file,min,debug) to http=file,min and the loglevel. It always debug that turns on the logging for that component.

2. This is a completely separate log setup mechanism whereas the logging is used only in WBEMConnection so this is too complex for the problem.  We really want to simply set the loggers in wbemconnection based on some definition of each of the loggers used by wbem connection and not get involved with this separate class, config, etc.

3. Andy noted that we really want this to be part of a complete logging structure so that if logging is turned on in general, these logs could get invoked at some predefined level (ex. debug).  However, that means that the log recorder is always enabled rather than only being enabled when it is in the list of active loggers.  That can be very heavy overhead so needs more study than we have time for right now.  The idea is that you would get logging of http and ops if you turned on logging at the pywbem level even if you did not know anything about these particular logs so that this became part of the general structure of log names.  Thus anybody who set a logger to log 'pywbem' would get some level of http and ops logging.

4. We want to move any of the processing of log definition strings (ex. the https=file,min,debug into some separate module (ex. toolssupport) that could at some point be moved into another repository rather than be part of the core logging for pywbemus .

5. Pull the definition of these logs into wbemconnection rather than have a separate PywbemLoggers dictionary that contains information needed by WBEMConnection.  The we would remove the enable_log init attribute from logging and add something like (set_log(component, detail_level)) to enable logs at that level.  Note that there is a core inconsistency here between this and (3) above in controlling logging with knowledge or without detailed knowledge.

We propose to do minor cleanup to this, and commit as experimental and make the next set of changes after we finish and commit this so we have logging in wbemcli.  We will change the wbemcli part so that we drop the debug part, mark this code experimental and commit

KSchopmeyer:
Updated this to add experimental to the classes/functions in _logging and the changes.rst.  I did NOT remove the loglevel from create_logger or create_loggers. It has the proper default so if not used will not cause problems.  I am assuming that we can commit this and am going to rebase it into the part2 code.

I like where we want to go now but lets get this in and then revamp.

KSchopmeyer:
Based on this being experimental and our agreement that I will update it

+1

PR #738, opened 2017-07-28
Circumvented issue with flake8 3.4.0.

andy-maier:
This circumvents the issue with flake8 3.4.0 that causes some of the other PRs to fail.


KSchopmeyer:
+1

PR #743, opened 2017-08-24
Fixes #691: Part 3 Add logging to wbemcli and to tests for wbemcli

### LINKED ISSUE #691: Add ability to log selected data on operations
Today the user can get back some information after a wbem operations (cim_operation method) has executed including 1. last request/reply, last time, size.

However, all of this information is lost when the user initiates the next operation.  It would be logical to add the functionality to selectively log this information.  Since Python provides a complete and extensively used logging tool, it would be logical to extend pywbem to actually provide the logging within pywbem itself.

Add a mechanism to enable python logging of selected data on operations.  It appears that the two areas of primary interest for logging are:

1. HTTP requests and responses - This would log the http for all requests and responses
2. CIM Operation method calls and responses. -This would log the parameters of each of the wbemconnection methods that result in wbem requests and also the responses at the level they are returned to the user

Since this can represent a lot of data we need to sort out a scheme that allows the user to be very selective in what is logged.

Questions for the design:

1. Log things as they happen.  Thus one log each for pywbemcall, httpsent, rttp_response, pywbem_response vs. log the complete results as a single log entry.  I propose that we log as things happen so that complete logging on a single request/response would be 4 log entries. 

CONCLUSION: We agree that we log as they occur

2. Log with just a single variable on/off vs. ability to enable log separately for the request/response info vs the http info.  I propose that we at least provide ability to enable the request/response level from the http level logging.
It may be logical to set upper size on responses although we never did with Pegasus and it worked well.  It would be good to have a switch to include or exclude the payload.

3. Selectively log certain methods - Right now this seems like overkill.

4. Attach the logging to the record mechanism since it is already gathering information  for each operation or create a parallel logger. I am prototyping using the recorder but there are alternatives such as decorators that would execute the logging.

This replaces part of issue # 569 that set requirement for testing wemcli and cmdline mof compiler

KSchopmeyer:
This pr extends wbemcli to allow defining logs for an execution of wbemcli

This is the final part of the fixes to implement issue#691

This pr requires the prs for part1(to configure logging) and part2(to generate log output) of issue#691 as a prerequisite. I have not cherry picked them until we have basic agreement on those prs


andy-maier:
@KSchopmeyer Rebasing this branch on master solves the conflicts automatically, but all tests in  `test_wbemcli.py` fail.

PR #744, opened 2017-08-25
Fixes #62: Add documentation on limitations of pywbem compiler.

### LINKED ISSUE #62: Cannot compile mof with property named indication
Not sure this is an error but the compiler keyword list which means that anything that is a keyword cannot be a property name should be reviewed against spec requirement

We agreed that the only change is a notation in the documentation.

KSchopmeyer:
1. Added comment about issue with property/etc. names that are also
compiler keywords.

2. Minor edits to the mof compiler doc to clarify some phrases.

3. Changed some section names in the rst documentation to separate
the mof compiler class from the mof compiler cmd line.

4. Added note to changes.rst

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -172,6 +172,9 @@ Bug fixes
   According to DSP0201, reference_class is the declared class, which can be
   a superclass of the  creation class of the referenced instance.
   This is related to issue #598
+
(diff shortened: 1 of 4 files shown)


andy-maier:
+1

PR #745, opened 2017-08-25
Fixes #741: Move documentation on operation recorder

### LINKED ISSUE #741: Docs section "Operation recording" is hard to find
It is currently at heading level 3 under "WBEM Operations", and will only be found if people happen to look at the left hand TOC menu.

We discussed to upgrade it by one heading level, and move it close to the "Statistics" section.

KSchopmeyer:
Moved documentation on operation recorder in client.rst so it has its
own section and kept the old subsection just as info about usage.

Added entry to changes.rst


andy-maier:
+1

PR #747, opened 2017-09-01
Fixes #349: Added class path to CIMClass

### LINKED ISSUE #349: Add class path to CIMClass
issue #339 clarified that we have an issue with the documentation/api on association responses in that the documentation defined the return of a class but the code/api returns a tuple. 

Since this knowledge of the path of the class is important, should it actually be part of the class itself?

andy-maier:
Ready for review and merge.

One particular review point would be that the CIM-XML and MOF representations of a class returned by the respective methods of `CIMClass` do not allow representing the class path we now added to `CIMClass`. That is not necessarily a problem, but it is inconsistent.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -106,6 +106,17 @@ Enhancements
 * Clarify documentation on wbem operation recorder in client.rst. see
   issue #741
 
+* Added an optional class path to the `CIMClass` class, as a convenience for
+  the user in order so that `CIMClass` objects are self-contained w.r.t. their
+  path. The class path is set in `CIMClass` objects returned by the `GetClass`,
+  `EmumerateClasses`, and the class-level `Associators` and `References`
+  operations. The path is added purely on the client side, based on existing
+  information returned from WBEM server. This change does therefore not affect
+  the interactions with WBEM servers at all.  issue #349.
+
+* Added a ``host`` property to ``WBEMConnection`` which contains the host:port
+  component of the WBEM server's URL.  This helps addressing issue #349.
+
 
 Bug fixes
 ^^^^^^^^^
@@ -183,7 +194,7 @@ Bug fixes
 
 * Modify mof_compiler documentation to indication issues with property
   names that are compiler keywords. See issue #62.
-  
+
 * Correct issue where dependency pip installs end up with old version
   of coverage package. This old version generates unwanted deprecation
   messages that are fixed after version 4.03. This requires a change to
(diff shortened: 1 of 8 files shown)


docs/changes.rst
```diff
@@ -106,6 +106,12 @@ Enhancements
 * Clarify documentation on wbem operation recorder in client.rst. see
   issue #741
 
+* Added an optional class path to the `CIMClass` class. The class path is
+  set in CIMClass objects returned by the GetClass, EmumerateClasses, and the
+  class-level Associators and References operations.  issue #349.
+
+* Added a ``host`` property to ``WBEMConnection`` which contains the host:port
+  component of the WBEM server's URL.  This helps addressing issue #349.
```
KSchopmeyer: I would add the comment about this not reflecting any change to the interaction with the host here.

pywbem/cim_obj.py
```diff
@@ -1932,6 +1948,7 @@ def tocimxml(self):
         as an instance of an appropriate subclass of :term:`Element`.
 
         The returned CIM-XML representation is consistent with :term:`DSP0201`.
+        It does not represent the class path of this class.
```
KSchopmeyer: Should the word be represent or include here. I would have used include. Also might add "since that is local to pywbem"

pywbem/cim_obj.py
```diff
@@ -1946,6 +1963,7 @@ def tocimxmlstr(self, indent=None):
         :class:`~pywbem.CIMClass` object, as a :term:`unicode string`.
 
         The returned CIM-XML representation is consistent with :term:`DSP0201`.
+        It does not represent the class path of this class.
```
KSchopmeyer: Same as above about representation, etc.

pywbem/cim_obj.py
```diff
@@ -1970,6 +1988,8 @@ def tomof(self):
         Return a :term:`unicode string` that is a MOF fragment with the
         class definition represented by the :class:`~pywbem.CIMClass`
         object.
+
+        The result does not represent the class path of this class.
```
KSchopmeyer: Same comment about representation, etc.

KSchopmeyer:
Still learning how to use review.  I added a bunch of single comments and a review. I am happy with code but think we need to explain somewhere what this does for the api user or they will probably just ignore it.

docs/changes.rst
andy-maier: Done

pywbem/cim_obj.py
andy-maier: Done - changed to "include"

pywbem/cim_obj.py
andy-maier: Done - changed to "include"

PR #748, opened 2017-09-01
Fixes #673: pywbem_os_setup.sh script for installing prereq. OS packages

### LINKED ISSUE #673: Easier approach for installing prerequisite OS level packages
The issue is that users(!) of Pywbem today do not get the prerequisite OS level packages installed if they do a `pip install pywbem` - that installation fails if the prerequisite OS level packages are not installed.

Pywbem today provides `python setup.py install_os` to install prerequisite OS level packages, but that requires to check out the source repository, and that approach is not really ideal for a user of Pywbem.

Ideally, we want to have some mechanism that installs the prerequisite OS level packages that is easily useable by a user of Pywbem, without having to deal with the source repo of Pywbem.

The pywbemcli project would be a first user of such a mechanism, but any user-written project could also use it.

andy-maier:
Ready for review.
It would be good to test the script on some Linux distributions.
For details, refer to the commit message.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -61,60 +61,42 @@ Installation
 ------------
 
 Pywbem is a pure Python package that can be installed from PyPi or from
-its `repository`_ by the usual means for installing Python packages.
+its `Git repository`_ by the usual means for installing Python packages.
 
-.. _repository: https://github.com/pywbem/pywbem
+.. _Git repository: https://github.com/pywbem/pywbem
 
 Some of the Python packages used by pywbem have dependencies on operating
-system packages. As of pywbem v0.8, the setup script has support for installing
-prerequisite operating system packages via a new ``install_os`` command for the
-setup script. Alternatively, you can establish these prerequisites yourself.
-For details, see `Prerequisite operating system packages`_.
+system packages. These operating system packages must be installed before the
+pywbem Python package can be installed. For details, see
+`Prerequisite operating system packages`_.
 
-The following examples show different ways to install pywbem. They all ensure
-that prerequisite Python packages are installed as needed:
+Pywbem also has dependencies on other Python packages. These Python packages
+will be automatically installed when installing pywbem using any of the
+approaches shown below:
 
-1. Install the latest released version, assuming the OS-level prerequisites are
-   already established::
+* Install the latest released version of pywbem (assuming the OS-level
+  prerequisites are already established):
 
-       $ pip install pywbem
+  .. code-block:: bash
 
-2. Install the latest released version and also install (or display) any
-   OS-level prerequisites::
+      $ pip install pywbem
 
-       $ pip download --no-deps --no-binary :all: pywbem
-       $ tar -xzf pywbem-*.tar.gz
+* Install the latest development version of pywbem (assuming the OS-level
+  prerequisites are already established):
 
-   or, when using ``pip`` versions before 8.0.0, you have to know the pywbem
-   version to download::
+  .. code-block:: bash
 
-       $ curl -L https://github.com/pywbem/pywbem/archive/v0.9.0.tar.gz |tar -xz
-
-   Followed in both cases by these commands::
-
-       $ cd pywbem-*
-       $ python setup.py install_os install
-
-3. Install a particular branch from the Git repository::
-
-       $ pip install git+https://github.com/pywbem/pywbem.git@<branch-name>
+      $ pip install git+https://github.com/pywbem/pywbem.git@master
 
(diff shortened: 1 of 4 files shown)


docs/intro.rst
```diff
@@ -134,32 +119,78 @@ by importing the package into Python::
...
-Prerequisite operating system packages can be installed by the pywbem setup
-script using a new command ``install_os``::
+Some of the Python packages used by pywbem have dependencies on operating
+system packages. These operating system packages must be installed before the
+pywbem Python package can be installed. This section describes multiple
+approaches how that can be done, and documents these packages.
+
+* Approach with ``pywbem_os_setup.sh`` script:
+
+  This is the recommended approach, as of pywbem 0.11.
```
KSchopmeyer: Suggest it would be clearer if we created subsections for each pywbem release where this differs.  0.11.0, 0.8.1, pre 0.8.1

docs/intro.rst
```diff
@@ -134,32 +119,78 @@ by importing the package into Python::
...
+    archive (``*.whl``) does not have that capability.
+
+  * Extract the pywbem source archive and run the ``install_os`` command of
+    ``setup.py``:
 
-    $ python setup.py install_os
+    .. code-block:: bash
 
-For a number of well-known Linux distributions, the setup script will install
-the packages using the respective installer program (e.g. ``yum`` on RHEL). For
```
KSchopmeyer: I am just getting into this now, but Fedora is in process of changing from yum to dnf as installer. Hit this in tests on vagrant saturday.

docs/intro.rst
```diff
@@ -134,32 +119,78 @@ by importing the package into Python::
...
 
-This command also downloads and builds ``swig`` and installs its build
-prerequisites if it cannot be installed as an operating system package in the
-required version.
+Both approaches use the ``sudo`` command under the covers, so your userid needs
+to be authorized accordingly, but you don't need to run the script using
+``sudo``.
 
-The second example in the previous section shows how to use the new setup
-command ``install_os``. It is really simple to be used and avoids the manual
```
KSchopmeyer: Drop work really. Think the phrase should be "It is simple to use and avoids"

docs/intro.rst
```diff
@@ -134,32 +119,78 @@ by importing the package into Python::
...
-procedure for establishing the operating system packages as described in the
-remainder of this section.
+Both approaches also download and build ``swig`` and install its build
```
KSchopmeyer: add which is a prerequisite for m2crypto.

pywbem_os_setup.sh
```diff
@@ -0,0 +1,122 @@
...
+    if [[ "$distro_id" == "enterpriseenterprise" ]]; then
+        distro_id="oracle"  # Oracle Enterprise Linux
+    fi
+    if [[ "$distro_id" == "redhatenterpriseworkstation" ]]; then
+        distro_id="rhel"  # RHEL 6, 7 Workstation
+    fi
+    if [[ "$distro_id" == "redhatenterpriseserver" ]]; then
+        distro_id="rhel"  # RHEL 6, 7 Server
+    fi
+  else
```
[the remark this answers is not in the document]
andy-maier: Both DONE

docs/intro.rst
andy-maier: DONE: Added support for dnf installer.

docs/intro.rst
andy-maier: IGNORED - this is a comment on removed text.

docs/intro.rst
andy-maier: DONE

docs/intro.rst
andy-maier: DONE

andy-maier:
Karl, I have addressed your comments, and have updated the script significantly. Please re-test it.

andy-maier:
I have improved the documentation and error message to state that the `distro´ package is needed.

andy-maier:
Rebased and squashed the commits.

KSchopmeyer:
Wed morning.  Test with ubuntu on vagrant.  Required to load before pywbem_os... script runs at least when not using a virtual env.  When after a vagrant destroy/vagrant up I first created a virtual machine and then ran pywbem_os... it did not fail however.

KSchopmeyer:
Note Fedora also requires sudo pip install distro to run pywbem_os_... without virtual env.

KSchopmeyer:
We agreed that we do not want to install distrib automatically if not in a virtual env so Karl's comment above can be ignored.

PR #786, opened 2017-10-02
pylint cleanup

KSchopmeyer:
Removed a number of issues found by pylint including:

1. The issue with unnecessary else. Changed most of those to use the
ternary form

2. The len(condition) issue which can be generally replaced with 

    if blah:
for cases were something must exist in iterable and 

    if not blah:
if the interable should be empty or None.

3. Couple of pylint messages that pylint is now marking as obsolete.

### DIFF
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -1156,7 +1156,7 @@ def parse_property_reference(tup_tree):
 
     value = list_of_matching(tup_tree, ['VALUE.REFERENCE'])
 
-    if value is None or not value:
+    if not value:
         value = None
     elif len(value) == 1:
         value = value[0]
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -96,10 +96,7 @@ def filter_tuples(list_):
     In a tupletree, tuples correspond to XML elements.  Useful for
     stripping out whitespace data in a child list."""
 
-    if list_ is None:
-        return []
-    else:
-        return [x for x in list_ if isinstance(x, tuple)]
+    return [] if list_ is None else [x for x in list_ if isinstance(x, tuple)]
 
 
 def pcdata(tup_tree):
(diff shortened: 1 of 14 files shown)


pywbem/tupleparse.py
```diff
@@ -1159,7 +1156,7 @@ def parse_property_reference(tup_tree):
 
     value = list_of_matching(tup_tree, ['VALUE.REFERENCE'])
 
-    if value is None or len(value) == 0:
+    if value is None or not value:
```
andy-maier: Line 1159 could be simplified to:
```
if not value:
```

testsuite/run_cim_operations.py
```diff
@@ -331,7 +320,7 @@ def assertInstancesValid(self, instances, includes_path=True,
 
             if includes_path:
                 self.assertTrue(isinstance(instance.path, CIMInstanceName))
-                self.assertTrue(len(instance.path.namespace) > 0)
+                self.assertTrue(instance.path.namespace, 'Includes namespace')
```
andy-maier: The assertion message should say what the issue is when the assertion fails, not what it asserts.
I.e. change 'Includes namespace' to 'Namespace missing in instance path'.
Apart from that, is such a message really useful given that the test output shows the source line with the assertion?

testsuite/run_cim_operations.py
```diff
@@ -1150,7 +1138,7 @@ def test_close_invalid(self):
     def test_get_onebyone(self):
         """Get instances with MaxObjectCount = 1).
         This test is subject to differences between the pull and non
-        pull sequence because it is getting live instances from the
+    gy s    pull sequence because it is getting live instances from the
```
andy-maier: Remove spurious text: "gy s" ?

testsuite/run_cim_operations.py
KSchopmeyer: Fixed

testsuite/run_cim_operations.py
KSchopmeyer: Fixed

KSchopmeyer:
Comments fixed.

PR #803, opened 2017-10-06
Fixes #779: test_statistics displays instead of testing results

### LINKED ISSUE #779: remove print output from test_statistics
this currently displays the table outputs.  That is:

a. not really testing the results.
b. wasting output since we do not really look at it.

Remove the displays and replace with test of results.

KSchopmeyer:
Replaced display code with tests of the generated output.


KSchopmeyer:
Fails regex match test in windows

PR #807, opened 2017-10-08
Improved checks and conversions when initializing and modifying CIM objects

andy-maier:
Ready for review and merge.

This is the second part of migrating from public attributes to getter/setter properties for the CIM object classes. Part 1 was in PR #784.

One review point is whether the following new functions should be added to the public API as proposed in the PR:
```
cimvalue()
CIMInstanceName.from_wbem_uri()
CIMClassName.from_wbem_uri()

**KS:** I vote yes to adding these methods.
```


KSchopmeyer:
I really do not know how to review this in detail. Just too many changes.  It does pass all the tests we have including running against OpenPegasus.

andy-maier:
Karl and I talked about this today, and we concluded to merge this PR (after the change list formatting error is fixed). Specifically, we agree to add the three new functions (see initial description, above), and we agree that the incompatible changes as stated in the change list are acceptable.

PR #811, opened 2017-10-24
Fixes #808: Fix more pylint issues

### LINKED ISSUE #808: Clean up missing string and other similar pylint issues
We agreed that the goal was to have documentation on all classes, functions, methods and thereby eliminate the missing string pylint warning.

Since there are currently 90 such warnings this ranks as the highest count warning besided the fixeme/todo one.

KSchopmeyer:
This change cleans up or hides several more minor pylint issues.  

Note that there is one change in the tests where it removes a one line test because that test appears to do nothing except cause a lint (unused variable).  See test_exceptions.py line 198

I consider this an extension of issue #808, cleaning up pylint issues.

One change was to remove several disable=redefined-variable-type comments because apparently that particular message is no longer in pylint.

Modifies some methods to staticmethod in unittest_extensions to get rid of pylint warnings.

There was one change where the pylint message is defined before and after the line of code once iwht the text and the second time with the warning code.  For some reason the first was not working and I used the code the second time because it was shorter.

### DIFF
--- a/os_setup.py
+++ b/os_setup.py
@@ -174,6 +174,7 @@ else:
     string_types = str,             # noqa: F821
     text_type = str                 # noqa: F821
     binary_type = bytes             # noqa: F821
+    # pylint: disable=import-error
     from xmlrpc.client import ServerProxy as xmlrpc_ServerProxy
 # pylint: enable=invalid-name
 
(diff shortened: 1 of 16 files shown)


KSchopmeyer:
+1 but Andy please review, in particular the line I took out of one of the tests in case you had any future plans for tha.

andy-maier:
Other than that, I'm ok with removing the one line from the test file, and have no further comments.

KSchopmeyer:
Fixed the ones I messed and a couple more in test_mof_compiler.  At this point we are down to a couple of pylints that are not TODO or XXX and those are in run_cim_operations.py. I will have another issue to clean up run_cim_operations.py

+1

PR #813, opened 2017-10-27
Correct WBEMConnection._get_rslt_params method and update unit tests

ssOleg:
From Andy: This PR intends to straighten out the conditions for raising CIM_ERR_INVALID_PARAMETER. Discussion is needed.

KS: Agreed.

Karl created issue # 844 to document this issue and therefore, a new pr, #845  The code changed somewhat in the new code and there is now a complete test for success/failures for this function

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -1945,28 +1945,25 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         rtn_objects = []
         end_of_sequence = False
         enumeration_context = None
-        valid_result = False
+        sequence = None
+        context = None
 
         for p in result:
             if p[0] == 'EndOfSequence':
-                if p[2] is None:
-                    valid_result = True
-                elif isinstance(p[2], six.string_types) and \
+                sequence = True
+                if isinstance(p[2], six.string_types) and \
                     p[2].lower() in ['true', 'false']:  # noqa: E125
-                    valid_result = True if p[2].lower() == 'true' else False
-                    end_of_sequence = valid_result
+                    end_of_sequence = True if p[2].lower() == 'true' else False
 
             elif p[0] == 'EnumerationContext':
-                if p[2] is None:
-                    valid_result = True
+                context = True
                 if isinstance(p[2], six.string_types):
                     enumeration_context = p[2]
-                    valid_result = True
 
             elif p[0] == "IRETURNVALUE":
                 rtn_objects = p[2]
 
-        if not valid_result:
+        if not sequence and not context:
             raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence "
                            "or EnumerationContext required")
 
--- a/testsuite/test_cim_operations.py
+++ b/testsuite/test_cim_operations.py
@@ -157,5 +157,23 @@ class test_create_connection(unittest.TestCase):
             pass
 
 
+class Test_get_rslt_params(unittest.TestCase):
+    """Test WBEMConnection._get_rslt_params method."""
+    def test_iterability(self):
+        from pywbem import WBEMConnection
+
+        result = [
+            (u'IRETURNVALUE', {}, None),
+            (u'EnumerationContext', None, u'context----------------'),
+            (u'EndOfSequence', None, u'FALSE')
+        ]
+
+        result = WBEMConnection._get_rslt_params(result, 'namespace')
+        self.assertEqual(
+            result,
+            (None, False, (u'context----------------', 'namespace'))
+        )
+
+
 if __name__ == '__main__':
     unittest.main()


andy-maier:
Hi Oleg. Thanks a lot for this PR! It is great to see that users of pywbem contribute.

We need to discuss the PR, though.

I definitely like the additional testcase. It is a good starting point for being further extended to cover more cases, when we move this module from unittest to pytest.

As far as the code change in `_get_rslt_params()` goes, I assume that your intention was to straighten out the conditions under which `CIM_ERR_INVALID_PARAMETER` is raised, to something that better matches what the exception message says (i.e. that at least one of the two output parameters must be provided). The change causes it to no longer be raised when the `EndOfSequence` output parameter is specified with a value of `FALSE`. I agree that specifying it with FALSE should not be treated much different than assuming a default of FALSE.

However, I think there is a larger problem with the code in this function (both before and after your change). DSP0200 states that both output parameters are required (The `OPTIONAL` marker is missing on these parameters in the operation declaration of the pull operations). Consequently, it does not specify default values to be assumed by a client if they are omitted in the response. However, the code (with and without your change) tolerates either one or the other output parameter to be omitted (but not both) and implements a default of "NULL" for a missing `EnumerationContext` parameter, and a default of `FALSE` for a missing `EndOfSequence` parameter.

**Discussion points:**
* Do known WBEM servers implement pull operations such that these output parameters can be omitted? (Oleg: If you have input on this, that would be welcome. Otherwise, Karl will provide what Pegasus does)
* How do we want the client to behave w.r.t. omitted output parameters?
* Do we need to raise an issue against DSP0200 because it does not allow NULL for the `EnumerationContext` parameter in the operation declarations, but talks about it being NULL in the description of `EnumerationContext` in 5.4.2.24.2 "Common Parameters for the Open Operations"?

andy-maier:
Karl will run some tests against Pegasus.

KSchopmeyer:
Both traces and the code establish that OpenPegasus inserts both EOS and EnumerationContext entities into all open/pull responses. When EOS is true, it forces the context value to empty string (not null).  It always sets either a True or False into EOS.

Thus, the following example of an open/pull sequence from OpenPegasus where we printed the call to imethodcall, and the call to _get_rslt_params:

    imethodcall methodname OpenEnumerateInstancePaths, ns root/cimv2, rqd True params  {'FilterQuery': None, 'FilterQueryLanguage': None, 'ClassName': CIMClassName(classname=u'CIM_ComputerSystem', namespace=None, host=None), 'ContinueOnError': None, 'MaxObjectCount': 1, 'OperationTimeout': None}

    get_rslt_params [(u'IRETURNVALUE', {}, [CIMInstanceName(classname=u'PG_ComputerSystem', keybindings=NocaseDict({u'CreationClassName': u'PG_ComputerSystem', u'Name': u'sheldon.snia.net'}), namespace=u'root/cimv2', host=u'sheldon.snia.net')]), (u'EndOfSequence', None, u'FALSE'), (u'EnumerationContext', None, u'500005')]

    imethodcall methodname PullInstancePaths, ns root/cimv2, rqd True params {'EnumerationContext': u'500005', 'MaxObjectCount': 1}

    get_rslt_params [(u'EndOfSequence', None, u'TRUE'), (u'EnumerationContext', None, '')]


However, we are not really sure that this applies to ALL servers since this is noted simply by the non-existence of OPTIONAL marked on the parameter definitions (something that some server writer may have missed) and there may be some confusion in the text

I would propose that the real goal here for the pywbem code is:

1. If EOS is True, enumeration context is unimportant since it carries no information whether it has NULL, a value, or does not even exist.

2. If EOS is False, EnumerationContext is important and a real value is required with the parameter. An EOS of False with no EnumerationContext entity or even NULL in EnumerationContext must be an error.  What that value is will be resolved by the next step when the client tries to use the EnumerationContext.

3. If either EOS or EnumerationContext exist, it is clearly an error. The client does not have enough information to continue

 - So the real questions here are whether we should force this client to adher exactly to the spec or to simply to be as loose as possible while insuring that we get good information back to the user? My vote is for code that works, but not rigid adherence to the spec.

- Should we file a mantis bug on the spec? In that case I propose that we do file the bug whether it ever gets fixed or not.

andy-maier:
Karl, on your two questions:

> *  So the real questions here are whether we should force this client to adher exactly to the spec or to simply to be as loose as possible while insuring that we get good information back to the user? My vote is for code that works, but not rigid adherence to the spec.
I second your vote. However, the description of that logic in your comment has an issue: item 3 should be "neither" instead of "either"? plus it misses the case where EC is present and EOS is omitted, defaulting to False.

> * Should we file a mantis bug on the spec? In that case I propose that we do file the bug whether it ever gets fixed or not.
I think we should file an issue.

andy-maier:
Looking through DSP0200 for output parameters, the only ones are in pull operations. Mostly, these are EC and EOS, which are always required, and there is QueryResultClass in OpenQueryInstances that is optional and may be NULL.

I found an issue in the description of QueryResultClass in OpenQueryInstances, which requires the server to set the parameter to NULL, I.e. to have it present with a vbalue of NULL. However, the "optional" marker also allows it to be omitted. i suggest to file another Mantis issue that proposes to change the text to also allow QueryResultClass to be omitted, and that adds a default value of NULL to it (anything that is optional needs to have a default).

andy-maier:
PR #845 will fix the error this PR fixed, and does a few more things (e.g. more tests). We will close this PR.

PR #819, opened 2017-10-30
Fixes #818: Add tests for the iter operations

### LINKED ISSUE #818: There is no unit testing for the IterOperations
Because the IterOperations involve multiple lower level operations we did not include them in testclient.

Therefore there is no real unittesting for these operations except against OpenPegasus in the run... tests.

We need to create a unit test for these methods.

KSchopmeyer:
Done but rebased from pr #833 to pick up issues we found in WBEMConnection

Adds the file testsuite/test_itermethods.py that executes tests using
pytest on  the iter... methods in WBEMConnection since these
methods were not really tested with the test_client.py tests and only the
main paths were tested with run_cimoperations.py

Implements tests for:

1. Force use of original method including parametrized input parameters 
2. Force use of new open method with parametrized input parameters
3. Use open and return and then use open/pull and return
4. Fail if force to original method and that returns error.
5. Test for invalid maxobjectcount since this one is specifically tested in the iter methods

Tests we do not have:
1. We do not have a test for close.
2. We do not have failure tests for a number of the other input parameters because they are part of the lower level functions we are mocking including:
    - filter without filter language
3. We have a value test on OperationTimeout but are not testing that.  

This test mocks all of the lower level operations (ex. enumerateInstances,
open/pullInstancesWithPath, etc. so this can be run as a unit test.

Overall this brings coverage up to about 76%.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -38,6 +38,17 @@ Enhancements
 Bug fixes
 ^^^^^^^^^
 
+* Fixed issue in iterReferenceNames and IterAssociatiorNames where it was
+  not passing the IncludeQualifiers input parameter to the OpenReferenceNames.
+  This should not have been a significant issue since in general qualifiers
+  are not parts of instances.  Also changed code in IterQueryInstances were
+  parameters that are required by the called ExecQuery and OpenQueryInstances
+  were defined as NamedArguments where since they are required, the Name
+  component is not required.  This should not change operations except that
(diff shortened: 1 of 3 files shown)


andy-maier:
I assume this is no longer WIP but ready to merge?

andy-maier:
Karl will move the things that remain to be done into separate issues.

PR #845, opened 2017-11-10
Fixes #844: Fix issue with _get_rslt_params documented in issue #844

### LINKED ISSUE #844: Issue with logic in CIM_operations _get_rslt_params
This method has an issue whereby it can ignore a problem in an open or pull response where the EOS is False but the EnumerationContext does not exist. 

This was filed directly as a pr by a user without filing an issue. See pr #813 for the original issue, the discussion, etc.

We will keep that discussion in the pr at least for the moment.

KSchopmeyer:
Fixes for andy's comments done.

Removed WIP status

1. Fixed the issue in the function noted in pr #813.  However we fixed
it with different code than the original pr.

2. Created a complete test for this function.

3. Converted  other tests in test_operation.py to pytest.

5. Added note to changes.rst

If we agree to this, we should also close pr #813

This code is ready for review.  I have one more test I would like to add but the complexity of the parametrize is defeating me right now.  Note that this appears to be a fairly complete test and also actually found an error in one of the test_client YAML files (had TFALSE where should have been FALSE).

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -1948,33 +1948,34 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         rtn_objects = []
         end_of_sequence = False
         enumeration_context = None
-        valid_result = False
+        end_of_sequence_found = False
+        enumeration_context_found = False
 
         for p in result:
             if p[0] == 'EndOfSequence':
-                if p[2] is None:
-                    valid_result = True
-                elif isinstance(p[2], six.string_types) and \
-                    p[2].lower() in ['true', 'false']:  # noqa: E125
-                    valid_result = True if p[2].lower() == 'true' else False
-                    end_of_sequence = valid_result
+                if isinstance(p[2], six.string_types):
+                    if p[2].lower() in ['true', 'false']:  # noqa: E125
+                        end_of_sequence = True if p[2].lower() == 'true' \
+                            else False
+                        end_of_sequence_found = True
+                    else:
+                        raise CIMError(CIM_ERR_INVALID_PARAMETER, 'Invalid '
+                                       'value %s on EndOfSequence' % p[2])
 
             elif p[0] == 'EnumerationContext':
-                if p[2] is None:
-                    valid_result = True
                 if isinstance(p[2], six.string_types):
                     enumeration_context = p[2]
-                    valid_result = True
+                    enumeration_context_found = True
 
             elif p[0] == "IRETURNVALUE":
                 rtn_objects = p[2]
 
-        if not valid_result:
+        if not end_of_sequence_found and not enumeration_context_found:
             raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence "
                            "or EnumerationContext required")
 
-        # convert enum context if eos is True
-        # Otherwise, returns tuple of enumeration context and namespace
+        # Drop enumeration_context if eos True
+        # Returns tuple of enumeration context and namespace
         rtn_ctxt = None if end_of_sequence else (enumeration_context,
                                                  namespace)
         return (rtn_objects, end_of_sequence, rtn_ctxt)
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -1948,8 +1948,8 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         rtn_objects = []
         end_of_sequence = False
         enumeration_context = None
-        end_of_sequence_found = False
-        enumeration_context_found = False
+        end_of_sequence_found = False  # flag True if found and valid value
+        enumeration_context_found = False  # flag True if ec tuple found
         for p in result:
             if p[0] == 'EndOfSequence':
                 if isinstance(p[2], six.string_types):
@@ -1962,19 +1962,19 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
                                        'value %s on EndOfSequence' % p[2])
 
             elif p[0] == 'EnumerationContext':
+                enumeration_context_found = True
                 if isinstance(p[2], six.string_types):
                     enumeration_context = p[2]
-                    enumeration_context_found = True
 
             elif p[0] == "IRETURNVALUE":
                 rtn_objects = p[2]
 
         if not end_of_sequence_found and not enumeration_context_found:
             raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence "
-                           "or EnumerationContext required")
+                           "or EnumerationContext required in response.")
         if not end_of_sequence and enumeration_context is None:
             raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence False"
-                           "and invalid EnumerationContext found")
+                           "and EnumerationContext Null value or missing.")
 
         # Drop enumeration_context if eos True
         # Returns tuple of enumeration context and namespace
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -1950,7 +1950,6 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         enumeration_context = None
         end_of_sequence_found = False
         enumeration_context_found = False
-
         for p in result:
             if p[0] == 'EndOfSequence':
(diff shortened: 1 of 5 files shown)


pywbem/cim_operations.py
```diff
@@ -1952,33 +1952,31 @@ def _get_rslt_params(result, namespace):
...
-                if p[2] is None:
-                    valid_result = True
-                elif isinstance(p[2], six.string_types) and \
+                if isinstance(p[2], six.string_types) and \
                     p[2].lower() in ['true', 'false']:  # noqa: E125
-                    valid_result = True if p[2].lower() == 'true' else False
-                    end_of_sequence = valid_result
+                    end_of_sequence = True if p[2].lower() == 'true' else False
+                    if end_of_sequence is True:
+                        valid_result = True
```
andy-maier: This logic renders EndOfSequence=FALSE an invalid result. I don't think that is what we want, particularly because we now require both parameters in the result.

pywbem/cim_operations.py
```diff
@@ -1952,33 +1952,31 @@ def _get_rslt_params(result, namespace):
...
-                    valid_result = True
                 if isinstance(p[2], six.string_types):
                     enumeration_context = p[2]
                     valid_result = True
 
             elif p[0] == "IRETURNVALUE":
                 rtn_objects = p[2]
 
-        if not valid_result:
+        if valid_result:
```
andy-maier: The previous way of going through the logic is more Pythonic:
```
if error:
    raise exc
good path
```

andy-maier:
On 4.: I don't know why moving this test into a separate file would eliminate the warning you intend to eliminate, once you process all files in a `make test` run.

Because py.test matches test classes and their test methods by name (and not by superclass like `unittest.main()` does), and because it does not know what arguments to pass to an object of the test class, it raises the warning about a class that matched as a test class when it has `__init__()`.

The way to get rid of that warning is to no longer make `TestClientRecorder` match the py.test pattern. 

Because that class is part of the public API of pywbem, changing its name would be an incompatible change, so while we could probably do that, it would not be nice.

A better way would be to change the name of the class *in the namespace of the test module*, in `testsuite/test_cim_operations.py`, by renaming it on import:
```
from pywbem._recorder import TestClientRecorder as _TestClientRecorder

# Change all usages of TestClientRecorder to _TestClientRecorder.
```
I verified that this eliminates the warning.

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python
```
andy-maier: I don't think we should move this into a new test file. See my explanation in the conversation of this PR.

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
...
+        self._run_single(b'<V>a\xF1\x80\xFFbc</V>', False)
+        # 4-byte sequence with incorrect 3rd byte that is an correct new start:
+        self._run_single(b'<V>a\xF1\x80\xC2\x81c</V>', False)
+
+
+class TestGetRsltParams(object):
+    """Test WBEMConnection._get_rslt_params method."""
+
+    @pytest.mark.parametrize(
+        'irval, ec, eos, eosr', [[None, u'contextblah', u'False', False],
```
andy-maier: I usually use an suffix of `_exp` for expected values, to make that easily recognizable. Suggest to rename `eosr` to `eos_exp`, and further down in `test_pass()`, to rename `ecs` to `ec_exp`.

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
...
+        """Test good combinations of EndOfSequence and EnumerationContext"""
+        result = [
+            (u'IRETURNVALUE', {}, irval),
+            (u'EnumerationContext', None, ec),
+            (u'EndOfSequence', None, eos)
+        ]
+        # pylint: disable=protected-access
+        result = WBEMConnection._get_rslt_params(result, 'namespace')
+        # the _get_rslt_params method sets context to None if eos True
+        ecs = None if eosr is True else (ec, 'namespace')
```
andy-maier: `eosr is True` can be simplified to `eosr`.

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
...
+        """Test EndOfSequence and EnumerationContext variations that should
+           cause exception."""
+        result = [
+            (u'IRETURNVALUE', {}, irval),
+            (u'EnumerationContext', None, ec),
+            (u'EndOfSequence', None, eos)
+        ]
+        with pytest.raises(CIMError):
+            # pylint: disable=protected-access
+            result = WBEMConnection._get_rslt_params(result, 'namespace')
```
andy-maier: I suggest to also assert the status code of the `CIMError` that is raised:
```
        with pytest.raises(CIMError) as exc_info:
            result = . . .
        exc = exc_info.value
        assert exc.status_code_name == 'CIM_ERR_. . .'
```

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @staticmethod
+    def test_missing():
+        """Test both Enum context and EndOfSequence completely missing"""
+        result = [
+            (u'IRETURNVALUE', {}, {})
+        ]
+        with pytest.raises(CIMError):
+            # pylint: disable=protected-access
+            result = WBEMConnection._get_rslt_params(result, 'namespace')
```
andy-maier: I suggest to also assert the status code of the `CIMError` that is raised.

testsuite/test_cim_operations2.py
```diff
@@ -0,0 +1,142 @@
...
+                                 [None, "", u'True', True],
+                                 [None, u'contextblah', u'True', True]]
+    )
+    @staticmethod
+    def test_pass(irval, ec, eos, eosr):
+        """Test good combinations of EndOfSequence and EnumerationContext"""
+        result = [
+            (u'IRETURNVALUE', {}, irval),
+            (u'EnumerationContext', None, ec),
+            (u'EndOfSequence', None, eos)
```
andy-maier: Your test always has both elements. Because the tested function allows one of the elements to be missing and has defaults, that should also be tested.
I suggest to extend the structure of the parametrization data for that. Maybe the `result` list that you have in the function should be in the parametrization data, plus `ec_exp` and `eos_exp` for the expected values.

andy-maier:
I have made a comment on PR #813 that we should discuss.

pywbem/cim_operations.py
```diff
@@ -1952,33 +1952,31 @@ def _get_rslt_params(result, namespace):
...
         for p in result:
             if p[0] == 'EndOfSequence':
-                if p[2] is None:
-                    valid_result = True
-                elif isinstance(p[2], six.string_types) and \
+                if isinstance(p[2], six.string_types) and \
                     p[2].lower() in ['true', 'false']:  # noqa: E125
-                    valid_result = True if p[2].lower() == 'true' else False
-                    end_of_sequence = valid_result
+                    end_of_sequence = True if p[2].lower() == 'true' else False
```
andy-maier: It would be good to raise an error if the value is invalid. Right now, we translate any invalid value to False.

testsuite/test_cim_operations2.py
KSchopmeyer: Done.

KSchopmeyer:
Andy. I think we are up to date with this one and I am ready to commit.

andy-maier:
Comments on the code as of today:

1. In pywbem/cim_operations.py around line 1970 there is:
    ```
        if not end_of_sequence_found and not enumeration_context_found:
            raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence "
                           "or EnumerationContext required")
        if not end_of_sequence and enumeration_context is None:
            raise CIMError(CIM_ERR_INVALID_PARAMETER, "EndOfSequence False"
                           "and invalid EnumerationContext found")
    ```
   The second check is pretty much the same as the first check, because the code sets enumeration_context_found to True when enumeration_context is set to the value in the response. The only difference is when the value in the response was Null, but I think if we wanted to check for specifically that then the test should test exactly for that (i.e. `if enumeration_context_found and enumeration_context is None`) and the exception message should say specifically that.

PR #856, opened 2017-12-01
Fixes #852: Improved the description of the ModifyInstance operation.

### LINKED ISSUE #852: Text for WBEMConnection.ModifiyInstance is simplistic
I think that part of the definition of actions for WBEMConnection.CIM_ModifyInstance are confusing in particular relative to the ModifiedInstance and PropertyList.

The current statement is:
```
          ModifiedInstance (:class:`~pywbem.CIMInstance`):
            A representation of the modified instance, also indicating its
            instance path.

            The `path` attribute of this object identifies the instance to be
            modified. Its `keybindings` attribute is required. If its
            `namespace` attribute is `None`, the default namespace of the
            connection will be used. Its `host` attribute will be ignored.

            The `classname` attribute of the instance path and the `classname`
            attribute of the instance must specify the same class name.

            The properties defined in this object specify the new property
            values for the instance to be modified. Missing properties
            (relative to the class declaration) and properties provided with
            a value of `None` will be set to NULL.

            Typically, this object has been retrieved by other operations,
            such as :meth:`~pywbem.WBEMConnection.GetInstance`.
```

The property list is treated as an independent description.  Therefore it is easy for the user
to not merge the two ideas.

I propose the following change to the paragraph that starts with "The properties defined ..."
```
            The properties defined in this object specify the new property
            values for the instance to be modified within the constrants
            set by the PropertyList parameter. Missing properties  (relative to the
            class declaration) and properties provided with a value of `None`
            will be set to the class default value (including NULL) or ifht
            property does not have a default value, there is no designated new value
            for the property. The server will reject any request that
            modifies key properties.
```

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -6364,7 +6364,8 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
 
         The `PropertyList` parameter determines the set of properties that are
         designated to be modified. The WBEM server shall reject modification
-        requests for key properties.
+        requests for key properties and for properties that are not exposed
+        by the creation class of the target instance.
 
         Pywbem sends the property values provided in the `ModifiedInstance`
         parameter to the WBEM server as provided; it does not add any default
@@ -6448,9 +6449,9 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
             This parameter defines which properties are designated to be
             modified.
 
-            It is an iterable specifying the names of the properties, or a
-            string that specifies a single property name. In all cases, the
-            property names are matched case insensitively.
+            This parameter is an iterable specifying the names of the
+            properties, or a string that specifies a single property name. In
+            all cases, the property names are matched case insensitively.
             The specified properties are designated to be modified. Properties
             not specified are not designated to be modified.
 
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -6359,53 +6359,53 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         (see :term:`DSP0200`). See :ref:`WBEM operations` for a list of all
         methods performing such operations.
 
-        The properties provided in the `ModifiedInstance` parameter specify
-        the new property values for the instance to be modified.
-
         The `PropertyList` parameter determines the set of properties that are
-        designated to be modified. The WBEM server shall reject modification
-        requests for key properties and for properties that are not exposed
-        by the creation class of the target instance.
+        designated to be modified (see its description for details).
+
+        The properties provided in the `ModifiedInstance` parameter specify
+        the new property values for the properties that are designated to be
+        modified.
 
         Pywbem sends the property values provided in the `ModifiedInstance`
         parameter to the WBEM server as provided; it does not add any default
         values for properties not provided but designated to be modified, nor
         does it reduce the properties by those not designated to be modified.
 
-        DSP0200 defines the following rules for how the WBEM server modifies
-        the property values of the target instance:
+        The properties that are actually modified by the WBEM server as a result
+        of this operation depend on a number of things:
 
-        * The designated new value for properties (that are designated to be
-          modified) shall be the value specified in the `ModifiedInstance`
-          parameter if it specifies a value (including `None`), else the
-          class-defined default value for the property if the class defines a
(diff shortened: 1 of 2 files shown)


pywbem/cim_operations.py
```diff
@@ -6399,12 +6445,23 @@ def ModifyInstance(self, ModifiedInstance, IncludeQualifiers=None,
             cannot rely on it being implemented by WBEM servers.
 
           PropertyList (:term:`string` or :term:`py:iterable` of :term:`string`):
-            An iterable specifying the names of the properties (or a string
-            that defines a single property) to be modified (case independent).
+            This parameter defines which properties are designated to be
+            modified.
 
-            An empty iterable indicates to modify no properties.
+            It is an iterable specifying the names of the properties, or a
```
KSchopmeyer: Suggest that 'It is an iterable..'. be changed to The PropertyList is ...

andy-maier: I don't like "The {parameter-name} is ...". I have changed it to "This parameter is ...", consistent with the sentence before the one you commented on.

pywbem/cim_operations.py
```diff
@@ -6359,6 +6359,53 @@ def ModifyInstance(self, ModifiedInstance, IncludeQualifiers=None,
         (see :term:`DSP0200`). See :ref:`WBEM operations` for a list of all
         methods performing such operations.
 
+        The properties provided in the `ModifiedInstance` parameter specify
+        the new property values for the instance to be modified.
+
+        The `PropertyList` parameter determines the set of properties that are
+        designated to be modified. The WBEM server shall reject modification
+        requests for key properties.
```
KSchopmeyer: add 'or if properties are in the propertylist that are invalid for the target instance'

andy-maier: "invalid for target instance" leaves some wiggle room,and this is a shall-rule. I prefer "not exposed by the creation class of the target instance."
DONE.

pywbem/cim_operations.py
```diff
@@ -6359,6 +6359,53 @@ def ModifyInstance(self, ModifiedInstance, IncludeQualifiers=None,
...
+          class-defined default value for the property if the class defines a
+          default value (including `None`). Otherwise, the property does not
+          have a designated new value.
+
+        * The properties are modified as follows:
+
+          * For properties designated to be modified that have a designated new
+            value, the WBEM server shall either update the property to that
+            value, or reject the request.
+          * For properties designated to be modified that do not have a
```
KSchopmeyer: This one is so general that it means nothing but that is pretty much what the spec says. However, we make the statement on line 6389 and then just below we also state that "Note that futher determinations about modifications of property values may be stated at the CIM modeling ... which is really part of the same thought.

andy-maier: The sad truth is that the actual way to understand when a property can be expected to be changed is paved with all kinds of flaky rules. Pywbem being a client interface, my intention is to describe what a client needs to know to get to an understanding. I did not find an approach for that that does not allude to all these things. And no, I don't want to assume they have read and understood the various specs involved. Let's see whether we can agree on the goal and then discuss what to have in the text.

andy-maier: We agreed to reduce this description to just what matters to the client, not repeating all the rules that need to be considered for someone implementing the server side. The text should include a recommendation to specify a propertylist with just the properties that are intended to get a new value.

andy-maier: The following sentence in the ModifyInstance description in DSP0200 is difficult:

"If PropertyList is NULL, the properties of ModifiedInstance with values different from the current values in the instance are designated to be modified."

The difficulty is that the part about values different from current values is difficult to understand, not possible to implement reasonably in a CIMOM, and most likely not implemented in any actually existing provider. Removing this part of the sentence does not seem to change anything in the rules, because making all properties provided in the ModifiedInstance parameter those that are designated to be changed still protects the target instance from getting property values set back to their class defined default values. Which is probably what the motivation was for that sentence.

Proposal for changed sentence:

"If PropertyList is NULL, the properties of ModifiedInstance are designated to be modified."

Karl agreed to take this to the DMTF as a Mantis issue.

andy-maier:
I have updated this PR to apply the second round of review comments. Please review again.

KSchopmeyer:
+1 OK

PR #881, opened 2017-12-15
Fixes #782: Made use_pull_operations attribute a read-only property

### LINKED ISSUE #782: Issues with use_pull_operations attribute
The description of the `use_pull_operations` instance attribute of `WBEMConnection` states:

* Indicates whether the client will attempt the use of pull operations in any Iter…() methods. However, each operation has its own internal equivalent flag based on this flag to allow for WBEM servers that implement only some of the pull operations.
  This variable is initially set from the same-named input argument. If it is initially None, it is modified when Iter…() methods are invoked, dependent on the support for pull operations in the WBEM server.

  After the first call , the variable will be either True or False and indicating whether pull operations will be used in this and subsequent Iter…() methods.
  **Update:** On the last statement: This applies to the non-public instance attributes for each operation type, but not to the `use_pull_operations` instance attribute.

Looking at the code, there are these issues:
* This instance variable is not changed as the connection learns about supported pull operations.
* It is writeable because it is simply a public instance attribute, but writing to it has no effect.

We need to discuss what we want to do with this variable, as long as the Iter...() methods are still experimental:
* Do we want an indication of what was learned?
* Do we want the users to be able to reset the automatic detection behavior?
* Do we want the users to be able to change the preference once the connection object exists?

andy-maier:
This is now ready for review and merge, as the discussion point in issue #782 has been resolved.

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -596,22 +596,6 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
       operation_recorder (:class:`~pywbem.BaseOperationRecorder`):
         This attribute from version 0.10.0 has become a propererty in
         0.11.0. and as a property is marked **deprecated**
-
-      use_pull_operations (:class:`py:bool`):
-        Indicates whether the client will attempt the use of pull operations in
-        any `Iter...()` methods. However, each operation has its own internal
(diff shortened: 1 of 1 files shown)


andy-maier:
Changed as proposed.

PR #887, opened 2017-12-20
Fixes #883: Dropped duplicate logging symbols from config module, Improved logging docs

### LINKED ISSUE #883: Log related symbols defined twice
We define the following public symbols (and export them into the `pywbem` namespace):

In pywbem/config.py:
```
DEFAULT_MAX_LOG_ENTRY_SIZE = 1000
DEFAULT_LOG_DESTINATION = 'none'
```

In pywbem/_logging.py:
```
MAX_LOG_ENTRY_SIZE = 1000
DEFAULT_LOG_DESTINATION = 'none'
```

For the first one, we need to agree on one name.

For both symbols, we need to discuss whether we want them to show up in the "Configuration variables" section of the docs (where we say that a user can modify these values) vs in the logging sectin (where such a statement is not made).

**Discussion needed**

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/_logging.py
+++ b/pywbem/_logging.py
@@ -41,20 +41,23 @@ by pywbem:
   logging is at the INFO log level.
 
 To output log records for one of the defined named loggers, either
-:meth:`~pywbem.PywbemLoggers.create_loggers` or the
-:meth:`~pywbem.PywbemLoggers.create_logger` should be used to define the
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
Note that branch has conflicts.  I approve of code changes

PR #905, opened 2017-12-24
Fixes #899: Deprecated CIMInstance.property_list

### LINKED ISSUE #899: Deprecate property_list attribute of CIMInstance
The `CIMInstance` class has an input parameter and settable attribute named `property_list` that acts as a filter for the properties of the instance. It is not clear how this would be useful in any way. Note that the property_list parameter of some WBEM operations returns `CIMInstance` objects that have the specified set of properties, without having a need for a a further property list filter on that object.

The approach would be to deprecate the `property_list` input parameter and attribute, after double checking that pywbem itself does not use it.

andy-maier:
**Note:** This PR is on top of PR #886.

Ready for review and merge, except that issue #899 still contains a discussion point.
This PR has become somewhat larger than intended, plus it caused some Todos to be added that I'd like to leave in for now, and resolve later (because they drive some more changes like converting test cases from unittest2 to pytest, or adding expected warnings to testcase definitions).

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -3224,8 +3224,13 @@ class CIMProperty(_CIMComparisonMixin):
     @qualifiers.setter
     def qualifiers(self, qualifiers):
         """Setter method; for a description see the getter method."""
+        # We make sure that the dictionary is a NocaseDict object, and that the
+        # property values are CIMQualifier objects:
         # pylint: disable=attribute-defined-outside-init
-        self._qualifiers = NocaseDict(qualifiers)
+        self._qualifiers = NocaseDict()
+        if qualifiers:
+            for key, value in qualifiers.items():
+                self._qualifiers[key] = _cim_qualifier(key, value)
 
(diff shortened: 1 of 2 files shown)


pywbem/cim_obj.py
```diff
@@ -1702,13 +1706,20 @@ def property_list(self):
...
+        Set only the desired properties on the object, instead of working with
+        this property filter.
         """
         return self._property_list
 
     @property_list.setter
     def property_list(self, property_list):
         """Setter method; for a description see the getter method."""
         if property_list is not None:
+            warnings.warn("The 'property_list' init parameter and attribute "
```
KSchopmeyer: I attached a note to the dependent pr.  Should we note that the correct way to use a property list to control modify wbem operations is through the property list on the operation???

andy-maier: No. That is too far away, here we are in context of constructing a CIMInstance object. It may not even be the result of an Enumerate operation.

PR #907, opened 2017-12-25
Improvements in issuing Python warnings

andy-maier:
**Note:** This PR is on top of PR #905.
For details, see the commit message.
Ready for review and merge.

PR #909, opened 2017-12-27
Fixed some more Python warnings

andy-maier:
**Note:** This PR is on top of PR #907.
Ready for review and merge.

### DIFF
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -310,6 +310,8 @@ class BaseOperationRecorder(object):
         issue where the file should be opened in text mode but that is
         done differently in Python 2 and Python 3.
 
+        The returned file-like object must be closed by the caller.
+
         Parameters:
 
           filename(:term:`string`):
@@ -320,10 +322,18 @@ class BaseOperationRecorder(object):
             existing file.  if 'a' is used, the data is appended to any
             existing file.
 
+        Returns:
+
+          File-like object.
+
         Example::
 
-            recorder = TestClientRecorder(
-                BaseOperationRecorder.open_file('recorder.log'))
+            recorder = TestClientRecorder(...)
+            recorder_file = recorder.open_file('recorder.log')
+
+            . . . # Perform WBEM operations using the recorder
+
+            recorder_file.close()
         """
         if six.PY2:
             # Open with codecs to define text mode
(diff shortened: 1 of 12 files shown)


pywbem/_recorder.py
```diff
@@ -320,10 +322,18 @@ def open_file(filename, file_mode='w'):
             existing file.  if 'a' is used, the data is appended to any
             existing file.
 
+        Returns:
+
+          File-like object.
```
KSchopmeyer: What is  a file-like object

pywbem/cim_obj.py
```diff
@@ -561,7 +565,8 @@ def _makequalifiers(qualifiers, indent):
     """
     if not qualifiers:
         return ''
-    qual_list = [q.tomof(indent + 2) for q in sorted(qualifiers.values())]
+    qual_list = [qualifiers[qn].tomof(indent + 2)
```
KSchopmeyer: Not arguing but what is the advantage of the replacement?

testsuite/test_mof_compiler.py
```diff
@@ -1427,35 +1451,27 @@ def test_mof_schema_roundtrip(self):
 
         repo2 = self.mofcomp2.handle
 
-        print('Start recompile file= %s' % mofout_filename)
-
+        # print('Start recompile file= %s' % mofout_filename)
```
KSchopmeyer: My only comment here is that this is a long delay for a single test.  If you are watching you begin to wonder. The start recompile gave a clue that it was really running

pywbem/_recorder.py
andy-maier: "File-like object" is a Python term for an object that behaves like a file (e.g. read, write, close). Here is one article about it, but not 100% exhaustive, I guess: https://stackoverflow.com/questions/4359495/what-is-exactly-a-file-like-object-in-python

pywbem/cim_obj.py
andy-maier: The sorting was previously done on the dictionary values, i.e. on a list. That uses the ordering comparisons between the list items, i.e. CIMQualifierDeclaration objects. These objects issue a deprecation warning if ordering comparisons are used on them. The change sorts on the dictionary keys and therefore avoids the deprecation warning.

testsuite/test_mof_compiler.py
andy-maier: I think it is more useful to have a clean output. py.test shows a dot for each test, and I don't think there is anything wrong is the next dot takes a minute or so to appear.

PR #912, opened 2017-12-27
Fixes #903: Fixes and cleanup for tocimxml()

### LINKED ISSUE #903: Order of bool and int tests in tocimxml()
The `tocimxml()` function in cim_obj.py has this TODO:
```
    # TODO: Verify whether this is a bug to have this test after the one for
    #       int. Bool is a subtype of int, so bool probably matches in the test
    #       above.
```
In Python, `bool` is a subtype of `int`, so from that perspective the test for `int` in that function would also match for a `bool` value. However, the situation is more complex, because dependent on how the function is called, not all types may be passed in, and it may end up not being a bug but rather redundant code that needs some cleanup.

Therefore, the situation of how the function is called first needs to be clarified, and then the code needs to be fixed or cleaned up as needed.

Also, add test cases for `tocimxml()` that cover the situation.

andy-maier:
**Note:** This PR is on top of PR #911.
This PR grew somewhat larger than the original issue described in #903. The changes belong together, though.
For details, see the (one) commit message added by this PR.
Ready for review and merge.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -78,7 +78,7 @@ from . import cim_xml
 from .config import DEBUG_WARNING_ORIGIN
 from .cim_types import _CIMComparisonMixin, type_from_name, cimtype, \
     atomic_to_cim_xml, CIMType, CIMDateTime, Uint8, Sint8, Uint16, Sint16, \
-    Uint32, Sint32, Uint64, Sint64, Real32, Real64
+    Uint32, Sint32, Uint64, Sint64, Real32, Real64, number_types
 
 if six.PY2:
     # pylint: disable=wrong-import-order
@@ -756,6 +756,11 @@ def _stacklevel_above_module(mod_name):
 def _cim_keybinding(key, value):
     """
     Return a keybinding value, from dict item input (key+value).
+
+    The returned value will be a CIM-typed value, except if it was provided as
+    Python number type (in which case it will remain that type).
+
+    Invalid types or values cause TypeError or ValueError to be raised.
     """
 
     if key is None:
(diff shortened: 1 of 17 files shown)


pywbem/cim_obj.py
```diff
@@ -778,20 +783,17 @@ def _cim_keybinding(key, value):
...
-        warnings.warn(msg, DeprecationWarning,
-                      stacklevel=_stacklevel_above_module(__name__))
+
+        # Ideally, we won't accept keybinding values specified as Python number
+        # typed values, but require a CIM data type (e.g. Uint32 or Real32).
+        # However, there are two reasons for continuing to allow that:
+        # * It was allowed in earlier versions of pywbem
+        # * When parsing the (untyped) WBEM URI of an instance path, we get
+        #   int or float values without size, and don't have any way to
+        #   get hold of the size information.
```
KSchopmeyer: Change 'get hold of' to 'access'

pywbem/cim_obj.py
```diff
@@ -778,20 +783,17 @@ def _cim_keybinding(key, value):
...
         # cannot use isinstance() for this test.
-        msg = "Setting a CIMInstanceName keybinding to a %s value is " \
-            "deprecated; use a CIM data type (e.g. Uint32) instead." % \
-            builtin_type(value)
-        if DEBUG_WARNING_ORIGIN:
-            msg += "\nTraceback:\n" + ''.join(traceback.format_stack())
-        warnings.warn(msg, DeprecationWarning,
-                      stacklevel=_stacklevel_above_module(__name__))
+
+        # Ideally, we won't accept keybinding values specified as Python number
```
KSchopmeyer: Picky comment but should we use something other than we, i.e. pywbem???

pywbem/cim_obj.py
```diff
@@ -5840,7 +5822,7 @@ def cimvalue(value, type):
 
     if type is None:
         # The following may raise TypeError or ValueError. This enforces that
-        # Python integer or float values cannot be provided without specifing
+        # Python number typed values cannot be provided without specifing
```
KSchopmeyer: Change word in line above from enforces to ensures

pywbem/cim_types.py
```diff
@@ -855,38 +860,42 @@ def type_from_name(type_name):
...
+    The returned CIM-XML string is ready for use as the text of a CIM-XML
+    'VALUE' element.
 
     Parameters:
 
-      obj (atomic scalar CIM typed value):
-        The CIM typed value, including `None`. Must be a scalar (not an array).
-        Must be an atomic type (i.e. those listed in :ref:`CIM data types`,
-        except any :ref:`CIM objects`).
+      obj (:term:`CIM data type`, :term:`number`, :class:`py:datetime`):
```
KSchopmeyer: Suggest build the word atomic back in just to help explain why we call it atomic... in the function call.  To be hones I am not sure why we use the word atomic

pywbem/cim_types.py
```diff
@@ -855,38 +860,42 @@ def type_from_name(type_name):
...
     from .cim_obj import _ensure_unicode  # due to cycles
-    if isinstance(obj, bool):
-        return u"true" if obj else u"false"
-    elif isinstance(obj, CIMDateTime):
+    if obj is None:
+        return obj
+    elif isinstance(obj, six.string_types):
+        return _ensure_unicode(obj)
+    elif isinstance(obj, bool):
+        return u'TRUE' if obj else u'FALSE'
```
KSchopmeyer: Question.  Was there any real issue with the lower case true false before.  I agree with the change but just wonder if we fixed any hidden issue

pywbem/cim_obj.py
andy-maier: DONE. Changed to "pywbem".

pywbem/cim_obj.py
andy-maier: DONE

pywbem/cim_obj.py
andy-maier: DONE

pywbem/cim_types.py
andy-maier: Me neither. I think we inherited that as part of the public API.
DONE.

pywbem/cim_types.py
andy-maier: DONE
DSP0201 says in 5.3.3.1.4 "Boolean Values":

* These values shall be treated as case-insensitive by CIM-XML consumers. CIM-XML producers should use upper case.

So the change just follows that recommendation, but for consumers that meet the requirement, it should not be a change. However, now that you said it, I have added an according statement to the (already existing) item in the change list (Cleanup section).

andy-maier:
All comments resolved as proposed, or answered. Considering this one approved.

PR #923, opened 2017-12-29
Fixes #898: Removed limitations on from_wbem_uri()

### LINKED ISSUE #898: CIMInstanceName.from_wbem_uri() does not handle nested double quotes
Example:
```
EX_ComposedOf.composer="EX_SampleClass.label1=9921,label2=\"SampleLabel\"",component="EX_SampleClass.label1=0121,label2=\"Component\""
```
Note that this example is not a case of nested references, but instead an association whose key properties reference a class that has a string typed key property. So the issue is very real and must be addressed.

andy-maier:
Ready for review and merge.

Particular review points worth mentioning:
* Extensions to WBEM URI format (see docstrings of the two `from_wbem_uri()` methods), particularly the support for omitting the leading slash in "local" WBEM URIs.
* DeprecationWarnings that are issued in the two `from_wbem_uri()` methods.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -71,6 +71,7 @@ import inspect
 import warnings
 from copy import copy
 import traceback
+import re
 
 import six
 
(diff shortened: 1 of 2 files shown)


pywbem/cim_obj.py
```diff
@@ -1391,86 +1576,111 @@ def from_wbem_uri(wbem_uri):
...
+          values that are not surrounded by double quotes, but issues a
+          :exc:`py:UserWarning`.
+
+        * DSP0207 does not allow the special float values INF, -INF, and NaN
+          in WBEM URIs (according to realValue in DSP0004). However, the
+          CIM-XML protocol supports representation of these special values,
+          so to be on the safe side, pywbem supports these special values as
+          keybindings in WBEM URIs.
+
+        Note that keybindings that are references are supported, recursively.
```
KSchopmeyer: No real reason for the "Note that" phrases.  I tend to do the same thing but realize later that not really necessary

pywbem/cim_obj.py
```diff
@@ -1382,6 +1435,138 @@ def tocimxmlstr(self, indent=None):
...
+            # Therefore, the realValue format has been checked explicitly,
+            # and an error in the float() function is not expected.
+
+            # Note that the special values 'INF', '-INF', and 'NAN' are
+            # also covered by Python float().
+
+            cimval = float(val)
+            return cimval
+
+        # The only remaining CIM type now is:
```
KSchopmeyer: This phrase is confusing.  Took a while to realise that what I think you meant was that all CIM Types are covered???

andy-maier: DONE: Changed to "At this point, all CIM types have been processed, except:"

pywbem/cim_obj.py
andy-maier: DONE: Removed "Note that" in this place and a few lines down.

PR #924, opened 2017-12-29
Different fix for Travis repo cloning to save resources

andy-maier:


andy-maier:
We agreed on today's call to just increase the depth and continue with a shallow clone, to save resources. Since this PR works, I'll merge it.

PR #933, opened 2018-01-02
Fixes #928: Made local WBEM URIs compliant to DSP0207

### LINKED ISSUE #928: WBEM URIs generated by str() miss a leading slash
The WBEM URI format defined in DSP0207 defines certain components in the WBEM URI as optional. A WBEM URI that omits the optional namespace type (URI scheme) and authority components then still has a leading slash before the namespace component, e.g. `/root/cimv2:CIM_Class.k=1`.

However, the WBEM URIs created by pywbem in `CIMInstanceName.__str__()` and `CIMClassName.__str__()` do not have that leading slash when the host component is `None`, they create `root/cimv2:CIM_Class.k=1` or `root/cimv2:CIM_Class`.

Because the WBEM URI is not used anywhere in MOF or in the CIM-XML protocol, we likely can fix the format in pywbem without breaking anything between the client library's API and the WBEM server.

When we fix it, the creating of the `CIMObject` header field value for instance paths which is currently done by using `CIMInstanceName.__str__()` must be adjusted, because for that header field value, the leading slash must not be present (see DSP0200).

andy-maier:
**Note:** This PR contains changes that cause the generated WBEM URIs to be conformant to DSP0207, including the leading slash for local WBEM URIs. Issue #938 attempts to confirm whether DSP0207 is intended as written, so we should wait for the outcome of that clarification before merging.

Otherwise, ready for review. 


andy-maier:
Rebasing after merging the underlying PRs, for easier reviewability.

PR #935, opened 2018-01-03
Improvements and fixes in CIM object's tomof(), part 1

### LINKED ISSUE #900: Cover embedded objects in CIMProperty.tomof()
This issue was triggered by the following TODO in `CIMProperty.tomof()` (in `cim_obj.py`):

W:3355, 0: TODO ks 8/16 do special formatting for this so output (fixme)

1. Add testcases for instance property values and class property default values that are embedded classes & instances.
2. Fix any issues.

andy-maier:
This is part 1 of a number of updates to the `tomof()` methods of the CIM object classes.
This PR was originally intended to address issue #900, but that will now be done by a subsequent part.
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -123,6 +123,16 @@ Incompatible changes
 * The global `tocimxml()` function now raises `TypeError` if it cannot
   convert the input value. Previously, it raised `ValueError`.
 
+* The `CIMQualifierDeclaration.tomof()` method now generates the flavor
+  keywords only if the `tosubclass` and `overridable` attributes are set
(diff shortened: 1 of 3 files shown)


andy-maier:
**Discussion points:**

* Flavors in `CIMQualifierDeclaration`: They have an officially documented state of `None` which exists for historical reasons, but does not really make sense. For example, the old `tomof()` method before this change defaulted them. See the new unit testcases for the difference. Should they not better be defaulted when creating the object?

andy-maier:
We agreed that we would go forward with this amount of changes and call it a stage 1.

PR #939, opened 2018-01-03
Fixes #936, association class with alias breaks mof_compiler

### LINKED ISSUE #936: mof compiler fails with build of assoc with alias
The changes made in pr 903 (changes to atomic_to_cim_xml) causes a failure in the mof compiler when building aliases for instances with reference properties.   The code causing the problem is below.

        try:
            # build instance property from class property but without
            # qualifiers, default value,
            pprop = cprop.copy()
            pprop.qualifiers = NocaseDict(None)
            pprop.value = tocimobj(cprop.type, pval)
            inst.properties[pname] = pprop
            # if alias and this is key property, add keybinding
            if alias and 'key' in cprop.qualifiers:
                keybindings[pname] = atomic_to_cim_xml(pprop.value)

        except ValueError as ve:
            ce = CIMError(CIM_ERR_INVALID_PARAMETER,
                          'Invalid value for property %s: %s' %
                          (pname, str(ve)))
            ce.file_line = (p.parser.file, p.lexer.lineno)
            raise ce

Apparently CIMInstance name was considered as a valid input to atomic before but is not after the change in the pr

The offending mof is probably driven by the class mof below where the two references are keys

        [Association, Description(" Family gathers person to family.") ]
        class TST_MemberOfFamilyCollection {
           [key] TST_Person ref family;
           [key] TST_Person ref member;
        };

and the following instance that defines an alias for this:

        instance of TST_MemberOfFamilyCollection as $SofiMember
        {
            family = $Family1;
            member = $Sofi;
        };

KSchopmeyer:
Changed compiler to let CIMinstanceName stringify keybinding values.

Added test for compile of association instances with reference property
keys which is what was broken.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -236,6 +236,10 @@ Bug fixes
   the return value of CIM methods, by raising `ValueError` if
   `CIMMethod.return_value` is initialized or set to "reference".
 
+* Fixed issue introduced in mof_compiler when atomic_to_cimxml was cleaned up
+  that did not allow using alias with some association classes.  Also
+  added test for this issue. See issue #936
+
 Cleanup
 ^^^^^^^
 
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
That was a one line change.

PR #955, opened 2018-01-09
Fixes #941: (1) Preserved order when iterating child elements of CIM objects

### LINKED ISSUE #941: CIM objects don't preserve order of child elements
The CIM objects do not currently preserve the order of their child elements.

For example, the `properties` attribute of `CIMClass` is a `NocaseDict` dictionary which does not preserve order of its items.

One undesired consequence of that is that MOF output of CIM objects using the `tomof()` methods does not have a defined order of such child elements.

Note that issue #951 covers this topic for qualifiers, and for those, a different approach than preservation may be needed.

**Discussion points:**

* When we preserve order of child elements, how do we want the equality comparison of parent elements to behave: Should two parent elements that have equal sets of child elements but as a list in different order, be equal or not?
* Do we want keybindings in WBEM URI strings to be in the preserved order, or in sorted order?
* Do we want the key names in keybindings in WBEM URI strings to be in the preserved lexical case, or in lower case?
* How do we want the equality comparison for WBEM URI strings of instance paths to behave: Should the order and lexical case of the keybinding names matter or not?

andy-maier:
**Note:** This PR contains a change for the pywbem_mock testcases (in a separate commit).
Ready for review and merge.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -78,6 +78,242 @@ This documentation uses a few special terms to refer to Python types:
    CIM object
       one of the types listed in :ref:`CIM objects`.
 
+   keybindings input object
+      a Python object used as input for initializing an ordered list of
+      keybindings in a parent object (i.e. a :class:`~pywbem.CIMInstanceName`
+      object).
+
+      `None` will result in an an empty list of keybindings.
+
+      Otherwise, the type of the input object must be one of:
+
+      * iterable of :class:`~pywbem.CIMProperty`
+      * iterable of tuple(key, value)
+      * :class:`~py:collections.OrderedDict` with key and value
+      * :class:`py:dict` with key and value (will not preserve order)
+
+      with the following definitions for key and value:
+
+      * key (:term:`string`):
+        Keybinding name.
+
+        Must not be `None`.
+
+        The lexical case of the string is preserved. Object comparison
+        and hash value calculation are performed case-insensitively.
+
+      * value (:term:`CIM data type` or :term:`number` or :class:`~pywbem.CIMProperty`):
+        Keybinding value.
+
+        If specified as :term:`CIM data type` or :term:`number`, the provided
+        object will be stored unchanged as the keybinding value.
+
+        If specified as a :class:`~pywbem.CIMProperty` object, its `name`
+        attribute must match the key (case insensitively), and a copy of its
+        value (a :term:`CIM data type`) will be stored as the keybinding value.
+
+        `None` for the keybinding value will be stored unchanged.
+
+      The order of keybindings in the parent object is preserved if the input
+      object is an iterable or a :class:`~py:collections.OrderedDict` object,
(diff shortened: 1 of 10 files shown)


KSchopmeyer:
I am getting one other issue in run_cimoperations.py in an embedded instance test.


    FAIL: test_receive (__main__.PegasusTestEmbeddedInstance)
      File "testsuite/run_cim_operations.py", line 4158, in test_receive
        r"embeddedInst = \"instance of Test_CLITestEmbedded1 {")
      File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/testsuite/unittest_extensions.py", line 38, 
    in assertRegexpContains
        "%s" % (text, pattern, msg))
    AssertionError: String text does not contain regexp pattern
        text:    u'instance of Test_CLITestEmbeddedClass {\n   embeddedInst =\n      "instance of 
    Test_CLITestEmbedded1 {\\n   Id = \\"100\\";\\n   name = "\n      "\\"Ronald\\";\\n   comment = 
  \\"Instance 
  created by provider.\\";\\n};\\n";\n};\n'
      pattern: 'embeddedInst = \\"instance of Test_CLITestEmbedded1 {'
  None

andy-maier:
A change from PR #973 for this PR:

On 2: This is actually on cim_obj.py line 1754, and the try path is used for iterables, while the except path is used for dictionaries. I admit that adding a comment pointing that out would be a good idea.

andy-maier:
On the issue when running run_cim_operations.py:

This is caused by the fact that `tomof()` on instances now breaks property values into a new line if they don't fit into the max line length. See these examples:

```
$ python -c "import pywbem; ie=pywbem.CIMInstance('CE',{'PE': 'a'}); i=pywbem.CIMInstance('C',{'P': 'a', 'EmbeddedInstanceProperty': ie}); print(i.tomof())"
instance of C {
   EmbeddedInstanceProperty = "instance of CE {\n   PE = \"a\";\n};\n";
   P = "a";
};

$ python -c "import pywbem; ie=pywbem.CIMInstance('CE',{'PE': 'a'}); i=pywbem.CIMInstance('C',{'P': 'a', 'EmbeddedInstancePropertyWithLongName': ie}); print(i.tomof())"
instance of C {
   P = "a";
   EmbeddedInstancePropertyWithLongName =
      "instance of CE {\n   PE = \"a\";\n};\n";
};

$ python -c "import pywbem; ie=pywbem.CIMInstance('CE',{'InnerPropertyWithLongName': 'a'}); i=pywbem.CIMInstance('C',{'P': 'a', 'EmbeddedInstanceProperty': ie}); print(i.tomof())"
instance of C {
   EmbeddedInstanceProperty =
      "instance of CE {\n   InnerPropertyWithLongName = \"a\";\n};\n";
   P = "a";
};
```

This needs to be fixed in run_cim_operations.py, for example by replacing NL with space before matching. Sorry I did not catch that. I'll change that.

PR #961, opened 2018-01-12
(1) Testing: Added test_function decorator

andy-maier:
Ready for review and merge. This change is used by the hashing support (PR #960).

The testcase-table driven testing requires to write the same kind of code over and over again, that 
handles warnings and expected exceptions. In the four execution paths, the code to be tested must be coded each time separately, so that it shows up four times.

This change puts all the boilerplate code into a decorator named 'test_function' in a new module 'pytest_extensions.py'.

To demonstrate the new decorator, the testcases for testing equality for NocaseDict has been converted to use the new concept.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -18,6 +18,7 @@ mock>=2.0.0
 httpretty>=0.8.14
 lxml>=4.0.0
 requests>=2.12.4
+decorator>=4.0.11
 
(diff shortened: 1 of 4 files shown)


andy-maier:
I have updated the docstring of the `test_function` decorator to add a proper parameter description, and an example.

PR #963, opened 2018-01-12
Fixes #871: InvokeMethod mock support

### LINKED ISSUE #871: Add invokemethod to pywbem_mock
We expect to do the initial commit without the implementation of InvokeMethod.  This issue documents that and serves as a reminder to add the InvokeMethod functionality.

We expect that this will be added as a callback so that the user will process the invokemethod as if it was a server operation and generate output parameters and a return value.

KSchopmeyer:
InvokeMethod implementation.

Ready for review and merge

Includes:

1. subscribe_method - Method that adds method subscription to repository by (namespace, classname, methodname, call_back_method_name)

2. Implementation of InvokeMethod itself (i.e. _mock_call mock) That gets the  method subscription info from the repo and calls the method to be invoked.

3. definition of the input params and return for the method to be invoked. 
 Input parameters to the callback are:  
      <handle of repo>, MethodName, LocalObject, Namespace, Input_Params.

The callback executor returns
    (return_value, Dict of OutputParameters)

4. Modification to cim_operations.py.InvokeMethod to move some low level code from InvokeMethod to _methodcall so that the response to InvokeMethod from _methodcall is simply the tuple of (return_value, output_params_dictionary).  That way the implementer of the callback simply creates the return_value and
the dictionary of output params exactly as defined in the return definition of WBEMConnection.InvokeMethod.

5. Expansion of mockspport.rst to include InvokeMethod

The limitation of this approach is that we are not really considering the concept of the provider as a contained entity with its own responder and method definitions but are attaching the invokemethod to the repository itself.  

Ont the other hand it is simple, allows execution of InvokeMethod by the client and can use data in the repository to respond and can also modify the repository.

### DIFF
--- a/docs/mocksupport.rst
+++ b/docs/mocksupport.rst
@@ -357,24 +357,13 @@ output parameters.
 
 The user must create a function that will be executed as a callback for each
 method that is to be tested. This method is attached to the mock repository
-through :meth:`~pywbem_mock.FakedWBEMConnection.subscribe_method` which
+through :meth:`~pywbem_mock.FakedWBEMConnection.add_method_callback` which
 defines the namespace, class, and methodname for the InvokeMethod and the
 callback method to be executed.
 
 
 This user defined callback method should have the signature defined in
-:meth:`~pywbem_mock.FakedWBEMConnection.subscribe_method`.
-
-Generally callback input and output are as follows:
-
-- **conn**: The handle to the mock repository
-
-- **objectname**: The CIMClassName or CIMInstanceName from the InvokeMethod
-  call including the execution namespace.
-
-- **methodname**: The methodname to execute from the client InvokeMethod call
-
-- **params**: The Params input parameter from the client InvokeMethod call.
+:meth:`~pywbem_mock.FakedWBEMConnection.add_method_callback`.
 
 The callback is expected to return a tuple consiting of ReturnValue and
 a `NocaseDict` containing any output parameters. The following is an example
@@ -427,7 +416,7 @@ of the definition and execution of InvokeMethod.
 
     # Subscribe the defined Class and method. In this case it is for the
     # default namespace.
-    conn.subscribe_method('TST_Class', 'Method1',
+    conn.add_method_callback('TST_Class', 'Method1',
                           self.method1_callback)
 
     # execute the InvokeMethod as a static (class level) method.
--- a/docs/mocksupport.rst
+++ b/docs/mocksupport.rst
@@ -313,7 +313,7 @@ instance requests.  At the same time, they allow mock repositories that
 would be used to test only instance methods to be built without the
 corresponding classes. They include the following methods:
 
-- AssociatorNames: Behaves like
+- **AssociatorNames**: Behaves like
   :meth:`~pywbem.WBEMConnection.AssociatorNames`. If a classname is specified
   the source, target, and association classes must be in the repository. If
   an instance target is specified, correct results are returned if classes
@@ -321,15 +321,15 @@ corresponding classes. They include the following methods:
   set (the classes in the repository are ignored so do not need to be
   installed.)
 
-- Associators: Behaves like
+- **Associators**: Behaves like
   :meth:`~pywbem.WBEMConnection.Associators` See AssociatorNames above for
   limitations
 
-- ReferenceNames: Behaves like
+- **ReferenceNames**: Behaves like
   :meth:`~pywbem.WBEMConnection.ReferenceNames` See AssociatorNames above
   for limitations
 
-- References: Behaves like
+- **References**: Behaves like
   :meth:`~pywbem.WBEMConnection.References` See AssociatorNames above for
   limitations.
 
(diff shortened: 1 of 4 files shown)


docs/mocksupport.rst
```diff
@@ -340,17 +340,99 @@ Server InvokeMethod operation method
...
+method that is to be tested. This method is attached to the mock repository
+through :meth:`~pywbem_mock.FakedWBEMConnection.subscribe_method` which
+defines the namespace, class, and methodname for the InvokeMethod and the
+callback method to be executed.
+
+
+This user defined callback method should have the signature defined in
+:meth:`~pywbem_mock.FakedWBEMConnection.subscribe_method`.
+
+Generally callback input and output are as follows:
```
andy-maier: I don't think we need to have a brief form of the callback interface at this point. The pointer to the description should be sufficient here.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
...
+    def subscribe_method(self, classname, methodname, method_callback,
+                         namespace=None,):
+        """
+        Add a callback for a method that will be called by the client when
+        InvokeMethod is executed.  This method does no real validation of
+        the classname, methodname or method_callback in put parameters.
+
+          Parameters:
+            classname (:term:`string`):
+                The classname for the method defined by methodname.
```
andy-maier: Clarify in the description whether classname must be:
(1) The CIM class that defines the method.
(2) The creation class of CIM instances or the CIM class against which the method will be invoked.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
                 except KeyError:
                     self.qualifiers[namespace] = NocaseDict({qual.name: qual})
             else:
-                assert False, 'add_cimobjects. %s not valid.' % type(obj)
+                assert False, 'Object to add_cimobjects. %s invalid type' \
+                              % type(obj)
+
+    def subscribe_method(self, classname, methodname, method_callback,
```
andy-maier: I don't like the name `subscribe_method()`, it is too close to indication subscription. In the mock support, we have used "add" for CIM objects, and what we add here is a "method callback" (at least if we go with the corresponding parameter name). So how about renaming this method to `add_method_callback()`?

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
...
+            methodname (:term: `string`):
+                Name of the method to be executed.
+
+            method_callback (:term:`callable`):
+                The user-defined function or method that well be called when by
+                FakeWBEMConnection when the InvokeMethod for the defined
+                class and methodname is received.  This method will be called
+                with the parameters defined on the InvokeMethod call
+
+                The callback method defined by the user is expected to comply
```
andy-maier: It is possible to define a global function that defines and describes the callback interface. That function can then be referenced here, and the type of the method_callback parameter would be the function name.

An example of how that is done, is in `pywbem/_listener.py`, where `callback_interface()` defines the interface of the callback function and `WBEMListener.add_callback()` adds a callback function.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
...
+                    Each iterated item represents a single input parameter
+                    for the CIM method and must be a ``tuple(name, value)``,
+                    with these tuple items:
+
+                    * name (:term:`string`):
+                      Parameter name (case independent)
+                    * value (:term:`CIM data type`):
+                      Parameter value
+
+                Return:
```
andy-maier: Change to `Returns`, for consistency.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
...
+
+                          * key (:term:`unicode string`):
+                            Parameter name, preserving its lexical case
+                          * value (:term:`CIM data type`):
+                            Parameter value
+
+                      This is the same format as defined in
+                      :meth:`~pywbem.WBEMConnection.InvokeMethod` Return.
+
+                Exceptions:
```
andy-maier: Change to `Raises`, for consistency.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -449,7 +450,117 @@ def add_cimobjects(self, objects, namespace=None):
...
+                    * value (:term:`CIM data type`):
+                      Parameter value
+
+                Return:
+                  * The user written callback method must return a tuple
+                    consisting of
+                      * ReturnValue (:term:`CIM data type`):
+                          Return value from the executed method.
+
+                      * outparams (`NocaseDict`_):
```
andy-maier: So far, users cannot create `NocaseDict` objects because that symbol is not public. They only get `NocaseDict` objects generated by pywbem. This design however, requires that they be able to generate `NocaseDict` objects.

**Discussion Point:** Should we make `NocaseDict` public so that users can create objects of that type, or should we change the design e.g. to list(tuple(name,value))?

andy-maier:
I have made comments.

In addition, it seems you have made the mock related commit of PR #955 part of this PR. This makes reviewing it more complex. You don't need to do that at all, or if you want to do it, then cherry-pick that commit into this PR but keep the commit separate in this PR.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Agreed.

docs/mocksupport.rst
KSchopmeyer: Done

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Good catch, In fact, I defined to be exactly that the class defined for the method and do not allow for subclassing for the method call. That will also have to change.

KSchopmeyer:
Do not force use of of NocaseDict on response. Agreed.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Agreed. See the listener as an example

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Agreed

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Agreed

KSchopmeyer: Done

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Done

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: We agreed to not make NocaseDict public and to redefine this interface and map to NocaseDict internally.

andy-maier:
Karl wants to make another small change on this one -> setting update needed.

PR #973, opened 2018-01-15
(3) Make tomof() generate child elements in preserved order

andy-maier:
**Note:** This PR is on top of PR #967.
Ready for review and merge.


andy-maier:
On the previous comment:

On 1: This is a NocaseDict method, so it is consistent between Py2 and Py3.
On 2: This is actually on cim_obj.py line 1754, and the try path is used for iterables, while the except path is used for dictionaries. I admit that adding a comment pointing that out would be a good idea.
On 3: Same as 1.

PR #977, opened 2018-01-16
Removed warnings in Sphinx doc build; Improved operation_recorder docs

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -160,7 +160,8 @@ then you do not have the Python.h header file available.
 The installation of pywbem with OS-level prereqs (see :ref:`Installation`)
 installs the necessary Python SDK package for C/C++ (or displays its package
 name). On RHEL, the missing package is `python-dev`.
-For more details, see :ref:`Prerequisite operating system packages`.
+For more details, see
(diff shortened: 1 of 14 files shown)


andy-maier:
Changes in the second commit:

* On your item 1: The "prereq" section does not need to be executed and is just for info. On Linux and OS-X, these packages are installed as part of `make develop`, and on Windows, the "dev for windows" section does that. I added explanatory text about the nature of that section, and nevertheless moved it up one level to make it appear in the TOC, but I left it at its position after "dev env".

* On your item 2: The section on Windows is only needed because there are so many manual things to do there. For Linux and OS-X, nothing comparable needs to be done. Added an according sentence to the item that points to the Windows section.

While reading the section, I made some more changes:

* Moved the step for setting up the virtual python env after the step for cloning the repo.

* Added text that python and OS-level packages are installed to the respective steps.

* Renamed the "Development environment on Windows" section to "Manual setup on Windows".

* In the "Manual setup on Windows" section, added a sentence to collect the `set`commands into a batch script.

KSchopmeyer:
Approved

Issue #701, opened 2017-05-04
Recorder-created CreateInstance fails test_client tests


KSchopmeyer:
A test example we created in testsuite/testclient (CreateInstance3) that creates an instance with many properties and that\
was automatically generated from the recorder fails it test.  The problem is confusion in the http over properties.

The first question is that the Expected is NOT what was defined in the test and things go down hill from that point. 

The test is committed as part of pr #696 so that this can be duplicated and the test case is currently
marked ignore_test

The result is as follows:

```
AssertionError: XML is not as expected in Unexpected CIM-XML payload in HTTP request: Expected:
  <CIM CIMVERSION="2.0" DTDVE 
        <IMETHODCALL NAME="CreateInstance">
          <LOCALNAMESPACEPATH>
            <NAMESPACE NAME="root"></NAMESPACE>
            <NAMESPACE NAME="cimv2"></NAMESPACE>
          </LOCALNAMESPACEPATH>
          <IPARAMVALUE NAME="NewInstance">
            <INSTANCE CLASSNAME="PyWBEM_AllTypes">
              <PROPERTY.ARRAY NAME="arrayReal32" TYPE="real32">
                <VALUE.ARRAY>
                  <VALUE>4.20000000E+01</VALUE>
                  <VALUE>4.44290000E+03</VALUE>
                </VALUE.ARRAY>
              </PROPERTY.ARRAY>
              <PROPERTY.ARRAY NAME="arraySint8" TYPE="sint8">
                <VALUE.ARRAY>
                  <VALUE>0</VALUE>
                  <VALUE>-1</VALUE>
                  <VALUE>44</VALUE>
                  <VALUE>127</VALUE>
                </VALUE.ARRAY>
              </PROPERTY.ARRAY>
              <PROPERTY.ARRAY NAME="arrayUint32" TYPE="uint32">
                <VALUE.ARRAY>
                  <VALUE>42</VALUE>
                  <VALUE>99</VALUE>
                </VALUE.ARRAY>
              </PROPERTY.ARRAY>
              <PROPERTY NAME="scalUint8" TYPE="uint8">
                <VALUE>42</VALUE>
              </PROPERTY>
            </INSTANCE>
          </IPARAMVALUE>
        </IMETHODCALL>
      </SIMPLEREQ>
    </MESSAGE>
  </CIM>

Got:
  <CIM CIMVERSION="2.0" DTDVERSION="2.0">
[...]

PR #987, opened 2018-01-18
Issue #701: (2b) Fixed CreateInstance3 testclient test case and adjusted run_cim_ops

andy-maier:
**Note:** This PR is on top of PR #955.
Ready for review and merge.

Note: This is a partial fix for issue #701, that fixes the testcase. It does not necessarily address why the testcase was created incorrectly by the TestClientRecorder.


andy-maier:
Removed the review needed label, since you approved this one.

Issue #950, opened 2018-01-08
InvokeMethod does not support input parameters that pass instances as embedded objects


andy-maier:
The `WBEMConnection.InvokeMethod()` method does not support parameters that are embedded instances passed as embedded objects (i.e. in CIM-XML terms, setting embedded_object="object" and passing an INSTANCE, or in MOF terms, setting the qualifier EmbeddedObject(True) on the parameter.

Reason is that the parameters are specified as a tuple name, value, where the value is a CIMClass or CIMInstance object, and the embedded_instance attribute is implied from that class type.

The problem only exists for input parameters, but not for output parameters or method return values, because for those, the embedded_obejct attribute in the CIM-XML response determines the way the instances are represented, and in both cases ("object" and "instance"), the returned CIM instance is always handed byck to the caller of InvokeMethod as a CIMInstance object.

This deficiency is not a huge problem, because CIM instances are recommended to be passed as embedded instance (i.e. in CIM-XML terms, setting embedded_object="instance" and passing an INSTANCE, and that is supported. However, it is technically possible that users define a method with a parameter that sets EmbeddedObject(True) for the purpose of passing an instance.

The difference between method parameters and properties in pywbem is that property values can be represented using CIMProperty (and thus, its embedded_object attribute can be set to "object", but for method parameters, a corresponding CIM object type does not exist (CIMParameter is currently documented to only allow representing the parameter declaration).

andy-maier:
Options to address this issue:
1. Extend CIMParameter to also allow representing parameter values, and allow them to be passed to InvokeMethod as an alternative to the tuple(name,value).
2. Introduce a new class CIMParameterValue for the same purpose.

KSchopmeyer:
Karl agrees with option 1

andy-maier:
Another discussion point is the use of CIMParameter for output parameters. Needs discussion.

andy-maier:
On the discussion point: We agreed to implement option iii. and to document the situation with returning embedded objects in output parameters.

andy-maier:
Summarizing the work for this issue, following the decisions outlined in previous comments:
* Extend `CIMParameter` to also allow representing parameter values, by undeprecating its `value` attribute (it was deprecated in pywbem 0.9), adding a new `embedded_object` attribute, and documenting the use for parameter values.
* Extend `InvokeMethod()` to be able to get input parameters also as `CIMParameter` objects, in addition to the existing tuple(name,value).
[...]

PR #988, opened 2018-01-18
Issue #950, part 1: Extended CIMParameter to represent parameter values

andy-maier:
This is part 1, which just extends CIMParameter as needed. Part 2 will be separate and will exploit that in InvokeMethod.
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -219,11 +219,11 @@ Enhancements
   via a new `operation_recorders` read-only property (Issue #976).
 
 * Extended `CIMParameter` to represent CIM parameter values in method
-  invocations. As part of that, undeprecated its `value` property and
-  added an `embedded_object` property. Extended the testcases accordingly.
-  Added an `as_value` argument to `CIMParameter.tocimxml()` and `tocimxmlstr()`
-  to allow control over whether the object is interpreted as a value or as a
-  declaration. (Issue #950).
+  invocations. As part of that, removed the deprecation from its `value`
+  property and added an `embedded_object` property. Extended the testcases
+  accordingly. Added an `as_value` argument to `CIMParameter.tocimxml()`
+  and to `tocimxmlstr()` to allow control over whether the object is
+  interpreted as a value or as a declaration. (Issue #950).
 
 * Docs: Improved the descriptions of CIM objects and their attributes to
   describe how the attributes are used to determine object equality and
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -218,6 +218,13 @@ Enhancements
 * Added support for retrieving the operation recorders of a connection
   via a new `operation_recorders` read-only property (Issue #976).
 
+* Extended `CIMParameter` to represent CIM parameter values in method
+  invocations. As part of that, undeprecated its `value` property and
```
KSchopmeyer: change undeprecated to removed deprecation from

pywbem/cim_obj.py
```diff
@@ -5211,7 +5213,25 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+      * The `value` attribute is the actual value of the parameter.
+      * Qualifiers are not allowed.
+
+    For parameter declarations in method declarations:
+
+      * The `value` attribute is ignored.
+      * Qualifiers are allowed.
+
+    Scalar (=non-array) parameters and items in array parameters may have a
+    value of NULL (= `None`), any primitive CIM data type, reference type, and
```
KSchopmeyer: Should last word in sentence be or rather than any?

pywbem/cim_obj.py
```diff
@@ -5298,13 +5320,23 @@ def __init__(self, name, type, reference_class=None, is_array=None,
...
+
+          embedded_object (:term:`string`):
+            A string value indicating the kind of embedded object represented
+            by the parameter value. Has no meaning for parameter declarations.
+
+            For details about the possible values, see the corresponding
+            attribute.
+
+            `None` means that the value is unspecified, causing the same-named
+            attribute in the ``CIMParameter`` object to be inferred. An
```
KSchopmeyer: I would guess that the reader would ask at this point what we mean by 'inferred'

pywbem/cim_obj.py
KSchopmeyer: clarify comment.  Should last work on line be or rather than and

docs/changes.rst
andy-maier: Done.

pywbem/cim_obj.py
andy-maier: Done.

pywbem/cim_obj.py
andy-maier: DONE - clarified that it is inferred from the parameter value.

PR #994, opened 2018-01-19
Fixes #992: Fixed the use of hard coded value limits in ValueMapping

### LINKED ISSUE #992: ValueMapping: Make hard coded value limits dependent on integer data type
The ValueMapping class in _server.py module currently assumes hard coded value limits of 0 and 32767, that cause misinterpretation of open value ranges (e.g. "..6" or "6..") whose open side reaches to the min or max limit of the data type.

These TODOs describe that:

W:963, 0: TODO: Change to min(cimtype) once issue #268 is solved. (fixme)
W:974, 0: TODO: Change to max(cimtype) once issue #268 is solved. (fixme)

The solution is to use the min and max values that can now be gathered from the element type, instead of the hard coded values.

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -330,6 +330,17 @@ Bug fixes
   their `class_origin` attribute case-insensitively. If set, it contains a CIM
   class name. Previously, that attribute was compared case-sensitively.
 
+* Fixed the use of hard coded value limits in the `ValueMapping` class
+  for open ranges of the ValueMap qualifier, by making them dependent on
+  the data type of the qualified element. This only affected elements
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
My only comment is that it is too bad we put that functionality for value mapping in _server, and not somewhere general.  Have had several users who I know immediatly started using it in other contexts.

andy-maier:
TODO (maybe in another PR): Make sure the docs for Valuemapping are at the top level (and not under WBEMServer), and make sure we have a notebook for it.

PR #996, opened 2018-01-19
Fixes #993: Made cimtype() and type_from_name() conversion functions part of public API

### LINKED ISSUE #993: Make type_from_name() and cimtype() part of the API
This issue was created to resolve this TODO in _server.py:

W:1014, 0: TODO: We should make type_from_name() and cimtype() part of the API (fixme)

These functions are defined in cim_types.py, and _server.py being a high level user of the core API should not have to use internal functions.

We agreed that these functions should be made part of the public API.

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -225,6 +225,15 @@ Enhancements
   and to `tocimxmlstr()` to allow control over whether the object is
   interpreted as a value or as a declaration. (Issue #950).
 
+* Added a new conversion function to the public API: `cimtype()` takes a CIM
+  data typed value (e.g. `Uint8(42)`) and returns the CIM data type name for
+  it (e.g. "uint8"). Previously, this was an internal function (Issue #993).
+
+* Added a new conversion function to the public API: `type_from_name()` takes
+  a CIM data type name (e.g. "uint8") and returns the Python type representing
+  that CIM data type (e.g. `Uint8`). Previously, this was an internal
+  function (Issue #993).
+
 * Docs: Improved the descriptions of CIM objects and their attributes to
   describe how the attributes are used to determine object equality and
   the hash value of the object.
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -225,6 +225,14 @@ Enhancements
   and to `tocimxmlstr()` to allow control over whether the object is
   interpreted as a value or as a declaration. (Issue #950).
 
+* Added a new conversion function: `cimtype()` takes a CIM data typed value
```
KSchopmeyer: shouldn't we say Added a new conversion function to the public API?

pywbem/cim_types.py
```diff
@@ -872,33 +875,33 @@ def type_from_name(type_name):
...
     :class:`~pywbem.Uint8`.
 
     For CIM data type names ``"string"`` and ``"char16"``, the
     :term:`unicode string` type is returned (Unicode strings are the preferred
     representation for these CIM data types).
 
-    The returned type can be used to 'cast' a value to that type in most cases,
-    except for "string" and "char16" typed values:
+    The returned type can be used as a constructor from a differently typed
+    input value, in many cases. Notable limitations are:
```
KSchopmeyer: Remove the comma  value, in. M change many to some.

docs/changes.rst
KSchopmeyer: That cements what we are doing in the first sentence.  Applies to the following also.

andy-maier: DONE. Both changed as proposed.

pywbem/cim_types.py
andy-maier: Comma: DONE.
"in many cases": They are really many cases for which the statement is true: All CIM integer types from an integer; all CIM real types from integer and float; CIMDateTime from Python datetime and timedelta and string in datetime format; CIM string and char16 from byte string except for the limitation, CIM reference from string in WBEM URI format and from CIMInstanceName.

andy-maier:
Resolved the comments as stated, and considering this one approved since you considered these minor changes.

PR #1000, opened 2018-01-21
Fixes #255: Strip lead/trailing slashes from namespace; Fix namespace in OpenQueryInstances

### LINKED ISSUE #255: ConnectionError: Illegal value for NAMESPACE.NAME attribute for "/interop"
One of the standard names defined in DSP1033 for the Interop namespace is `"/interop"`, i.e. with a leading slash.

OpenPegasus causes this to be raised in PyWBEM after returning a CIMError:

```
  File "<stdin>", line 1, in <module>
  File "pywbem/server/__init__.py", line 136, in determine_interop_ns
    self._conn.EnumerateInstanceNames(test_classname, namespace=ns)
  File "pywbem/cim_operations.py", line 1117, in EnumerateInstanceNames
    **extra)
  File "pywbem/cim_operations.py", line 670, in _imethodcall
    timeout=self.timeout)
  File "pywbem/cim_http.py", line 673, in wbem_request
    raise ConnectionError(exc_str)
pywbem.exceptions.ConnectionError: CIMError: request-not-valid, PGErrorDetail: Semantic error: on line 2: Illegal value for NAMESPACE.NAME attribute
```

The reason for that seems to be that the PyWBEM client splits this namespace name as usual at the slash character, producing two NAMESPACE elements in the CIM-XML payload (this is from `conn.last_request`):

```
    <SIMPLEREQ>
      <IMETHODCALL NAME="EnumerateInstanceNames">
        <LOCALNAMESPACEPATH>
          <NAMESPACE NAME=""/>
          <NAMESPACE NAME="interop"/>
        </LOCALNAMESPACEPATH>
        <IPARAMVALUE NAME="ClassName">
          <CLASSNAME NAME="CIM_Namespace"/>
        </IPARAMVALUE>
      </IMETHODCALL>
    </SIMPLEREQ>
  </MESSAGE>
</CIM>
```

I think my first question would be: Given that `"/interop"` is one of the official Interop namespace names, how did we think it would be represented in CIM-XML?

In other words, do we have a problem in the standard, or a bug in OP?
This is with OP 2.12.0 on RHEL 6.7.

andy-maier:
Ready for review and merge.


andy-maier:
We agreed to merge this now and fix any issues later.

PR #1003, opened 2018-01-22
Fixes #950: Extended InvokeMethod() to accept iterable of CIMParameter as input parms

### LINKED ISSUE #950: InvokeMethod does not support input parameters that pass instances as embedded objects

andy-maier:
Ready for review and merge.
This one would also benefit from a real world test.


andy-maier:
Merging to be able to do the change to the method callback interface we discussed as part of PR #1004.

PR #1004, opened 2018-01-23
Fix invoke method interface. 

KSchopmeyer:
I misinterpreted the Params vs params options in invoke method and did not incorporate both Params and params into fake  InvokeMethods cleanly.  I fact, we really do not have an example that shows the various options for calling InvokeMethod  to make it easy for the client writer but that is a different question.

This cleans up the InvokeMethod and callback interface and adds more tests.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -91,8 +91,8 @@ def _display(dest, text):
         f.close()
 
 
-def method_callback_interface(conn, objectname, methodname, params):
-    # pylint: disable=unused-argument
+def method_callback_interface(conn, objectname, methodname, Params, **params):
+    # pylint: disable=unused-argument, invalid-name
     """
     **Experimental:** *New in pywbem 0.12 as experimental.*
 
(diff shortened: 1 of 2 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -91,8 +91,8 @@ def _display(dest, text):
         f.close()
 
 
-def method_callback_interface(conn, objectname, methodname, params):
-    # pylint: disable=unused-argument
+def method_callback_interface(conn, objectname, methodname, Params, **params):
+    # pylint: disable=unused-argument, invalid-name
```
andy-maier: Having both a Params argument and kwargs on the callback interface is really a historical relic on InvokeMethod that we did not simplify for compatibility reasons.

Technically, we could have just a single param argument on the callback interface, that is an iterable of CIMParameter objects. Please think about whether it is important to represent the exact interface of InvokeMethod on the callback (which requires the user to first merge the two lists) vs. simplifying it.

KSchopmeyer: Let’s go for the single interface in the callback method. I assume that it should be the iterable if CIMParameter and we will merge the two inputs. However. Lets get his fix in and also the CIM paramount fix and then I will modify.

andy-maier: ok. merging it.

KSchopmeyer: Tks.

PR #1006, opened 2018-01-23
Add test compile and display with unicode mof. 

KSchopmeyer:
Fixes  the display_repository to output unicode for python 2. Also removed a couple
of worthless TODOs.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -77,12 +77,17 @@ OUTPUT_FORMATS = ['mof', 'xml', 'repr']
 # document our behavior in relation to the spec.
 
 
-def _display(dest, text):
+def _uprint(dest, text):
     """
     Display to dest defined by text. This function appends the data
     in text to the file defined by dest or to stdout of dest is None
     """
-    text = text.encode("utf-8", errors='replace')
+    # replace errors since there are issues with some windows versions.
+    if sys.version_info[0] == 2 and sys.version_info[1] == 6:
+        text = text.encode("utf-8")  # python 2.6 does not support errors=
+    else:
+        text = text.encode("utf-8", errors='replace')
+
     if not dest:
         print(text)
     else:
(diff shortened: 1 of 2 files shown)


andy-maier:
On the faailure shown in the Appveyor run for py3: See the solutions in this article: https://stackoverflow.com/questions/14630288/unicodeencodeerror-charmap-codec-cant-encode-character-maps-to-undefined

KSchopmeyer:
Found a solution that seems to pass the test but it is really a hack.  Look at the function _uprint.  Better ideas would be readily accepted.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -77,12 +77,17 @@
...
 
-def _display(dest, text):
+def _uprint(dest, text):
     """
     Display to dest defined by text. This function appends the data
     in text to the file defined by dest or to stdout of dest is None
     """
-    # TODO make this unicode file for python 2
+    # replace errors since there are issues with some windows versions.
+    if sys.version_info[0] == 2 and sys.version_info[1] == 6:
```
andy-maier: You could use `sys.version_info[0:2] == (2, 6)` instead. Works even with `<` and `>`.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -77,12 +77,17 @@
...
+        text = text.encode("utf-8")  # python 2.6 does not support errors=
+    else:
+        text = text.encode("utf-8", errors='replace')
```
andy-maier: Do you even need the replace? After all, you are encoding to UTF-8, which can represent all Unicode characters.

testsuite/test_wbemconnection_mock.py
```diff
@@ -1031,6 +1032,46 @@ def test_compile_qualifiers(self, conn, ns, tst_qualifiers_mof):
...
+            Flavor(DisableOverride, ToSubclass);
+        Qualifier Description : string = null,
+            Scope(any),
+            Flavor(EnableOverride, ToSubclass, Translatable);
+        """
+
+        cmof = u"""
+        class CIM_Foo {
+                [Key,
+                 Description ("Å \u0420\u043e\u0441\u0441\u0438\u044f"
```
andy-maier: Please avoid having unicode characters directly in the source. Use `\u212b` instead.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -77,12 +77,17 @@
...
+
     if not dest:
         print(text)
     else:
```
andy-maier: If you use `with open(...)`, it closes the file again and you don't need the `fp.close()`.

andy-maier:
Karl,
PR #1011 addresses all of my comments to this change. You could for example cherry-pick it into your branch.

andy-maier:
We agreed that PR #1011 should be used as is, so we can just merge this PR and then merge the other PR.

PR #1018, opened 2018-01-28
(5) Fixes, more tests and some cleanup for CIM-XML tupleparser

andy-maier:
**Note:** This PR is on top of PR #1037.
For details, see the commit message (of only the last commit).
For status of tests that are still missing, see PR #1021.
Ready for review and merge.

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -3337,7 +3337,8 @@ class ClassOperations(ClientClassTest):
                                                type='uint8')})
 
         # force propagated False for all properties
-        # Effictive V0.12.0 this is required.
+        # Effective V 0.12.0 propagated must be set to compare with
+        # info returned from server.
         for p in test_class.properties:
             test_class.properties[p].propagated = False
 
@@ -3593,10 +3594,11 @@ class ModifyClass(ClassOperations):
                 print('NOTE: This server/namespace does not support '
                       'CreateClass since it returned INVALID_CLASS')
 
-        # Now modify the original class by adding a new property
+        # Modify the original class by adding a new property
         new_property = CIMProperty('Str2', None,
                                    type='string',
-                                   is_array=False)
+                                   is_array=False,
+                                   propagated=False)
 
         test_class.properties['Str2'] = new_property
 
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -3337,6 +3337,7 @@ class ClassOperations(ClientClassTest):
                                                type='uint8')})
 
         # force propagated False for all properties
+        # Effictive V0.12.0 this is required.
         for p in test_class.properties:
             test_class.properties[p].propagated = False
 
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -2276,7 +2276,6 @@ class CreateInstance(ClientTest):
                 ('arrayTimeDelta', [td, td]),
             ])
 
-
         tst_instance.path = CIMInstanceName(
             'PyWBEM_AllTypes', {'InstanceId': instance_id})
 
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -2244,7 +2244,7 @@ class CreateInstance(ClientTest):
 
         tst_instance = CIMInstance(
             'PyWBEM_AllTypes',
-            value=[
+            properties=[
                 ('InstanceId', instance_id),
                 ('scalBool', True),
                 ('scalUint8', Uint8(42)),
@@ -2275,6 +2275,7 @@ class CreateInstance(ClientTest):
                 ('arrayDateTime', [dt, dt]),
                 ('arrayTimeDelta', [td, td]),
             ])
+
 
         tst_instance.path = CIMInstanceName(
             'PyWBEM_AllTypes', {'InstanceId': instance_id})
@@ -3335,6 +3336,10 @@ class ClassOperations(ClientClassTest):
                         'MyUint8': CIMProperty('MyUint8', Uint8(99),
                                                type='uint8')})
 
+        # force propagated False for all properties
+        for p in test_class.properties:
+            test_class.properties[p].propagated = False
+
         return test_class
 
     def create_class(self):
@@ -3392,6 +3397,8 @@ class ClassOperations(ClientClassTest):
                                                    is_array=True),
                         'MyStr': CIMProperty('MyStr', 'This is a test',
                                              type='string')})
+        for p in test_class.properties:
+            test_class.properties[p].propagated = False
         return test_class
 
 


andy-maier:
**Discussion point:** Should we go ahead with this PR and do the remainder of the work in a subsequent PR?

KSchopmeyer:
Karl agrees. Lets merge as it stands and continue with other issues. Karl will review

pywbem/tupleparse.py
```diff
@@ -103,15 +105,20 @@ def filter_tuples(list_):
...
+    The child nodes must be text nodes (no element nodes).
+    """
     for inst in tup_tree[2]:
         if not isinstance(inst, six.string_types):
             raise ParseError("Element %r has unexpected child elements: %r"
                              "(allowed is only text content)" %
                              (name(tup_tree), inst))
-    return ''.join(tup_tree[2])
+    data = u''.join(tup_tree[2])
+    assert isinstance(data, six.text_type)
```
KSchopmeyer: Question? Leave this assert in permanently???

pywbem/tupleparse.py
```diff
@@ -949,7 +1024,7 @@ def parse_instance(tup_tree):
             %ClassName;>
     """
 
-    check_node(tup_tree, 'INSTANCE', ['CLASSNAME'],
+    check_node(tup_tree, 'INSTANCE', ['CLASSNAME'], [],
```
KSchopmeyer: Karl. Check why the empty list entry

pywbem/tupleparse.py
andy-maier: My vote is yes. On my end, there is a small remaining uncertainty about the guarantee that the sax parser always returns unicode, and if pywbem wants to guarantee that from that point on, we need to either ensure that ourselves by check&conversion (e.g. using `_ensure_unicode()`) or by verifying it with an assertion.

About the uncertainty: The docs about [ContentHandler.characters()](https://docs.python.org/2/library/xml.sax.handler.html#xml.sax.handler.ContentHandler.characters) state: "content may be a Unicode string or a byte string; the expat reader module produces always Unicode strings." but I did not find a statement that the expat reader is used. This is implemented in C, so I did not venture to have a look.

pywbem/tupleparse.py
andy-maier: This is parameter was missing and adding it is a fix.
It specifies the allowed optional attributes, so before the change, the list of child elements was in that argument position and was interpreted as optional attributes. And the argument position for the list of child elements was unspecified, so it defaulted to None which means any child element was allowed.

KSchopmeyer:
Ouch. we hit some failures in running against pegasus.  I am trying to sort these out now. These errors are in places where we normally do no get errors, creation of an instance, one invoke method, class create/delete test

KSchopmeyer:
Here are the reported errors for the first few tests. Some of the others include very large display so not posting yet.

```

ERROR: test_pywbem_AllTypes (__main__.CreateInstance)
Test Creation of an instance of PyWBEM_AllTypes.
  File "testsuite/run_cim_operations.py", line 2276, in test_pywbem_AllTypes
    ('arrayTimeDelta', [td, td]),
TypeError: __init__() got an unexpected keyword argument 'value'

ERROR: test_all_param_types (__main__.InvokeMethod)
Call invoke method with all possible parameter types
  File "testsuite/run_cim_operations.py", line 2452, in test_all_param_types
    Ref=name)
  File "testsuite/run_cim_operations.py", line 186, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 7440, in InvokeMethod
    result = self._methodcall(MethodName, ObjectName, Params, **params)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1278, in _methodcall
    for n, v, t, eo in ptuples]
ValueError: need more than 2 values to unpack

ERROR: test_new_params_arg (__main__.InvokeMethod)
Call with new Params arg
  File "testsuite/run_cim_operations.py", line 2526, in test_new_params_arg
    Beer=Uint16(4))
  File "testsuite/run_cim_operations.py", line 186, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 7440, in InvokeMethod
    result = self._methodcall(MethodName, ObjectName, Params, **params)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1278, in _methodcall
    for n, v, t, eo in ptuples]
ValueError: need more than 2 values to unpack

ERROR: test_non_empty_arrays (__main__.InvokeMethod)
Call with non-empty arrays
  File "testsuite/run_cim_operations.py", line 2511, in test_non_empty_arrays
    RefArray=[name, name])
  File "testsuite/run_cim_operations.py", line 186, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 7440, in InvokeMethod
    result = self._methodcall(MethodName, ObjectName, Params, **params)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1278, in _methodcall
    for n, v, t, eo in ptuples]
ValueError: need more than 2 values to unpack

ERROR: test_one_param_types (__main__.InvokeMethod)
Call invoke method with all possible parameter types
  File "testsuite/run_cim_operations.py", line 2481, in test_one_param_types
    Ref=name)
  File "testsuite/run_cim_operations.py", line 186, in cimcall
    result = fn(*pargs, **kwargs)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 7440, in InvokeMethod
    result = self._methodcall(MethodName, ObjectName, Params, **params)
  File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py", line 1278, in _methodcall
    for n, v, t, eo in ptuples]
ValueError: need more than 2 values to unpack

```

andy-maier:
The error "ValueError: need more than 2 values to unpack" in `_methodcall()` is a programming error that crept in when the ordering of parameters was established, and went on undetected because none of our unit tests used `InvokeMethod()` with passing the CIM parameter values as keyword parms (`**params`)

andy-maier:
The error "TypeError: __init__() got an unexpected keyword argument 'value'" i `test_pywbem_AllTypes()` happens because `CIMInstance` does not have a ctor parameter named `value`. Pass the properties as `properties=[...]`.

andy-maier:
Karl, the last push includes the fix in PR #1037 (for the error "ValueError: need more than 2 values to unpack" in `_methodcall()`).

--> Please fix the other error in run_cim_operations.py as described above, and re-run it.

KSchopmeyer:
On tests that create class we typically do the following:

1. Build the class using CIMClass
2. Use this to do CreateClass
3. do GetClass to confirm that the server created the correct class
4. Compare the original Class used to call CreateClass with the server returned GetClass.

The issue is simply that in the original CreateClass, the propagated attribute on each property is None but whatever the server returns results in propagated=False.  When we do the compare, it fails because these attributes are different.

That accounts for about 3 failures.  

The

andy-maier:
On the error with `propagated` in `CIMProperty`: This is a consequence of adding support for propagated in a recent PR. The change in that PR now takes the PROPAGATED=false information returned from the server and updates the `propagated` attribute of the CIMProperty it hands back to the caller as the operation's result. Previously, that attribute was on its default `None`. I was facing the same discrepancy in the tupleparse testcases, and my solution was to create any CIMProperty objects that are used for comparing actual results with, with an explicit `propagated=false` input.

KSchopmeyer:
We can do that for our tests to pass the tests. I wonder what the happens when we create classes with the mofcompiler.  I will check into that since we do not have a test to add a class to the pegasus serveor.

I have a quick fix for that one, just sets propagated False after creating the instances.  This may be logical since it implies that the definer actually knows that the properties are being defined in this instance.

NOTE: This does not occur on CIMMethods because the default for propagated is False.

**DISCUSSION:** Why the discrepancy between methods and properties for the default value of propagated??  Should we consider changing the default to False for Properties???

You want me to commit the changes or here are the 'major'  :-) patches. to change values to properties and force propagated False

--- /tmp/YfF1Q6_run_cim_operations.py
+++ /home/kschopmeyer/pywbem/githubpywbem/pywbem/testsuite/run_cim_operations.py
@@ -2244,7 +2244,7 @@
 
         tst_instance = CIMInstance(
             'PyWBEM_AllTypes',
-            value=[
+            properties=[
                 ('InstanceId', instance_id),
                 ('scalBool', True),
                 ('scalUint8', Uint8(42)),
@@ -2276,6 +2276,7 @@
                 ('arrayTimeDelta', [td, td]),
             ])
 
+
         tst_instance.path = CIMInstanceName(
             'PyWBEM_AllTypes', {'InstanceId': instance_id})
 
@@ -3334,6 +3335,10 @@
                                                   type='string'),
                         'MyUint8': CIMProperty('MyUint8', Uint8(99),
                                                type='uint8')})
+
+        # force propagated False for all properties
+        for p in test_class.properties:
+            test_class.properties[p].propagated = False
 
         return test_class
 
@@ -3392,6 +3397,8 @@
                                                    is_array=True),
                         'MyStr': CIMProperty('MyStr', 'This is a test',
                                              type='string')})
+        for p in test_class.properties:
+            test_class.properties[p].propagated = False
         return test_class

andy-maier:
On the discrepancy of the default values for propagated: That's historical probably just oversight in an area that is not used much. I'm not sure which default males more sense: False is more practical, but None is more correct since it means the information is not known.

andy-maier:
Summary of the behavior of the current code in pywbem (master branch as of 2018-02-04), for the `propagated` attribute and the flavor-related attributes of `CIMProperty`, `CIMMethod`, `CIMQualifier` and `CIMQualifierDeclaration`:
* For CIM objects created by pywbem users: If an attribute is not specified to the ctor, it defaults to `None` meaning "information not available" in all cases with the following exceptions:
  - `CIMMethod.propagated` defaults to `False`.
* For CIM objects returned from CIM operations: If an attribute is not specified in CIM-XML, the corresponding attribute of the CIM object will be set to its DSP0201-defined default value in all cases with the following exceptions:
  - flavor-related qualifiers in `CIMQualifier` default to `None`.
  - flavor-related qualifiers in `CIMQualifierDeclaration` all default to `False` (!!).

**Discussion point:** I think the exceptions described in the summary above should be considered issues that are to be fixed.

pywbem/tupleparse.py
KSchopmeyer: AGREED

andy-maier:
Karl, I'll fix the flake8 issue with your changes when I'm at home, in 30 min or so.

KSchopmeyer:
I updated again to set translatable=True on the property description qualifier I created and to set propagated on both class and property.  Note that this does not affect the server since it is supposed to ignore propagated but I set them to matches what comes back.

PR #1036, opened 2018-02-02
Fixes #1069: Cleanup TODOs in FakeWBEMConnection; Added from_instance()

### LINKED ISSUE #1069: Add from_instance method to CIMInstanceName
Add a method to cim_obj:CIMInstanceName to create a CIMInstanceName from an existing class and
instance.

This uses the class to define the key properties and the instance to get values for the keys.

The api for this method is

def from_instance(class, instance, namespace=None,, host=None, strict=False):

Note that this is driven by need for this functionality in the pywbem_mock package.  The issue is documented seperately here. The code is submitted as part of pr #1023

KSchopmeyer:
Updated with comments from Friday 9 Feb.  Note that I made one change that we had not discussed. The previous from_instance defaulted to getting property value from class if strict=false and no property in instance.  Removed the code so it simply does not put that property into the keybindings.
DISCUSSION if you would prefer the original.

Cleaned up tests for get_class and enumerateclasses to more throughly test the lo, iq, and ico attributes.
This and some other test changes to get rid of TODOs  sort of bumped up the test count from 650 to 900. 

1. Modified logging

2. Removed the write to file from mofcompiler errors.  Now writes error
information to stdout.

3. Modified the compiler error message format. New format is:

Syntax error:NoFile:3:24
            Qualifier Association : boolean = false,
                Scope(associations),
                      ^^^^^^^^^^^^
Old format did not put context on separate lines and only showed line
number if there was a file name.  i.e. Showed nothing for String input
to compiler.

Note that the display message still shows error as Syntax error but the
interrupt is MOFParaseError and that the MOFParseError __str__ output
does not format the context.

4.  Moved function to create CIMNamespaceName from Class and instance to
method of cim_obj.CIMInstanceName and created tests. Now static
method named from_instance(...)

5. Remove some TODOs that were not revalent any more.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -2552,21 +2552,25 @@ class CIMInstanceName(_CIMComparisonMixin):
         :class:`~pywbem.CIMInstanceName` from the class key properties and
         instance key property values.
 
-        If the `strict` parameter is True, and a property value does not exist
-        in the instance the values are retrieved from the class default_value.
+        If the `strict` parameter is False, and a property value does not exist
+        in the `instance` that entry is not included in the constructed
+        CIMInstanceName
 
-        If the `strict` parameter is False all key properties in the `class_`
+        If the `strict` parameter is True all key properties in the `class_`
         must exist in the `instance` or a ValueError exception is raised.
 
         Parameters:
 
-            class_ (:class:`~pywbem.CIMClass`):
+            `class_` (:class:`~pywbem.CIMClass`):
                 CIMClass from which the list of key properties will be
                 retrieved. The class MUST contain all of the key properties
                 that are required to create the CIMInstanceName. Thus, for
                 example, if the class were retrieved from a server, generally,
                 the request `LocalOnly` parameter should be False to assure all
-                superclass properties are retrieved.
+                superclass properties are retrieved.  It need not be the
+                same class as the instance. Thus, it could be a
+                superclass if the key properties were the same for both the
+                instance and class
 
             instance (:class:`~pywbem.CIMInstance`):
                 CIMInstance containing the key property values.
@@ -2578,8 +2582,8 @@ class CIMInstanceName(_CIMComparisonMixin):
                 Host name to include in created CIMInstanceName or None.
 
             strict (:class:`py:bool`):
-                If True, all key properties are required in instance. Default
-                False.
+                If True, all key properties in the class must have values in
+                the  instance. Default False.
 
         Returns:
             :class:`~pywbem.CIMInstanceName`:
@@ -2588,16 +2592,11 @@ class CIMInstanceName(_CIMComparisonMixin):
                 `instance` parameter.
 
         Raises:
-          ValueError: Class and instance names do not match or the
-            strict attribute is True and key property does not exist in
-            the instance.
+          ValueError: The strict attribute is True and key property does not
+          exist in the instance.
 
         """
         keybindings = NocaseDict()
-        if class_.classname != instance.classname:
-            raise ValueError('Class name %s does not match instance '
-                             'classname %s' %
-                             (class_.classname, instance.classname))
 
         for prop in class_.properties:
             if 'key' in class_.properties[prop].qualifiers:
@@ -2610,9 +2609,6 @@ class CIMInstanceName(_CIMComparisonMixin):
                         raise ValueError('Key property %r of class %r '
                                          'missing in instance.' %
                                          (pname, class_.classname))
-                    else:
-                        default_value = class_.properties[prop].value
-                        keybindings[pname] = default_value
 
         return CIMInstanceName(class_.classname,
                                keybindings,
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -2548,38 +2548,47 @@ class CIMInstanceName(_CIMComparisonMixin):
     def from_instance(class_, instance, namespace=None, host=None,
                       strict=False):
         """
-        Given a class and corresponding instance, create and return the
-        CIMInstanceName from the class key properties and instance key
-        property values.
+        Given a class and corresponding instance, create and return a
+        :class:`~pywbem.CIMInstanceName` from the class key properties and
+        instance key property values.
 
-        If a property value does not exist in the instance it gets the
-        value from the class default_value unless strict is set. If strict
-        is True, all key properties must exist in the instance
+        If the `strict` parameter is True, and a property value does not exist
+        in the instance the values are retrieved from the class default_value.
+
+        If the `strict` parameter is False all key properties in the `class_`
+        must exist in the `instance` or a ValueError exception is raised.
 
         Parameters:
 
-            class_ (:class:~pywbem.CIMClass`):
-                CIMClass from which the key properties will be retrieved.
+            class_ (:class:`~pywbem.CIMClass`):
+                CIMClass from which the list of key properties will be
+                retrieved. The class MUST contain all of the key properties
+                that are required to create the CIMInstanceName. Thus, for
+                example, if the class were retrieved from a server, generally,
+                the request `LocalOnly` parameter should be False to assure all
+                superclass properties are retrieved.
 
-            instance: (:class:~pywbem.CIMInstance`):
+            instance (:class:`~pywbem.CIMInstance`):
                 CIMInstance containing the key property values.
 
-            namespace: (:term:``string`):
+            namespace (:term:`string`):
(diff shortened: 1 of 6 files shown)


testsuite/test_cim_obj.py
```diff
@@ -2411,7 +2411,7 @@ class Test_CIMInstanceName_from_instance(object):
             ),
             dict(
                 classname='CIM_Foo',
-                keybindings={('P1', 'Ham'), ('P2', 'Cheese')}
+                keybindings={'P1': 'Ham', 'P2': 'Cheese'}
```
andy-maier: If maintaining the order of keybinding items is important, this line should be changed to:
```
keybindings=[('P1', 'Ham'), ('P2', 'Cheese')]
```

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+        If a property value does not exist in the instance it gets the
+        value from the class default_value unless strict is set. If strict
+        is True, all key properties must exist in the instance
+
+        Parameters:
+
+            class_ (:class:~pywbem.CIMClass`):
+                CIMClass from which the key properties will be retrieved.
+
+            instance: (:class:~pywbem.CIMInstance`):
```
andy-maier: The colon directly after the parameter name is incorrect. Occurs multiple times.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+                CIMInstance containing the key property values.
+
+            namespace: (:term:``string`):
```
andy-maier: Duplicated back-quote. Occurs one more time.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
 
         return _ensure_unicode(''.join(ret))
 
+    @staticmethod
+    def from_instance(class_, instance, namespace=None, host=None,
+                      strict=False):
+        """
+        Given a class and corresponding instance, create and return the
+        CIMInstanceName from the class key properties and instance key
+        property values.
```
andy-maier: It is sufficient that the class is a superclass of the creation class of the instance, as long as it has all key properties defined.
Actually, that is a requirement we need to state explicitly, because it may be that the `CIMClass` object was created with `LocalOnly=True` and therefore has only its additional property definitions, which may not even include any key property.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+
+        If a property value does not exist in the instance it gets the
+        value from the class default_value unless strict is set. If strict
+        is True, all key properties must exist in the instance
+
+        Parameters:
+
+            class_ (:class:~pywbem.CIMClass`):
```
andy-maier: Missing back-quote.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+
+            host:  (:term:``string`):
+                Host name to include in created CIMInstanceName or None.
+
+            strict: (:class:`py:bool`):
+                If True, all key properties are required in instance. Default
+                False
+
+        Returns:
+            :class:`CIMInstanceName` built from the key properties in the
```
andy-maier: This link to the object probably does not resolve in the generated documentation. Add `~pywbem.`.
Also, insert a colon after the link - this will cause the type to be recognized by Sphinx.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+            class using the key property values in the instance.
+
+        Raises:
+            ValueError: Class and instance names do not match or the
+            strict attribute is True and key property does not exist in
+            the instance.
```
andy-maier: Most likely, the first line of this description paragraph needs to be outdented by 2 from the rest of the lines, to make Sphinx generate one paragraph out of this. Look at the generated documentation.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
...
+            if 'key' in class_.properties[prop].qualifiers:
+                pname = class_.properties[prop].name  # get original name
+
+                if prop in instance:
+                    keybindings[pname] = instance[prop]
+                else:
+                    if strict:
+                        raise ValueError('Key Property %s in class %s '
+                                         'but not in instance.' %
+                                         (pname, class_.classname))
```
andy-maier: I don't like this error message; it seems to suggest that a key property can stand on its own and can be in a class or not. How about: "Key property %r of class %r is missing in the instance.". Always use `%r` instead of `%s` for CIM names, because unicode characters will be properly escaped that way, if needed.

pywbem/cim_obj.py
```diff
@@ -2511,6 +2511,72 @@ def to_wbem_uri(self, format='standard'):
 
         return _ensure_unicode(''.join(ret))
 
+    @staticmethod
+    def from_instance(class_, instance, namespace=None, host=None,
+                      strict=False):
```
andy-maier: A change log entry is missing for this new method.

testsuite/test_cim_obj.py
```diff
@@ -2343,7 +2345,230 @@ def test_CIMInstanceName_from_wbem_uri(
...
+    testcases = [
+        # Testcases for CIMInstanceName.from_instance().
+        # Each testcase has these items:
+        # * desc: Short testcase description.
+        # * cls_kwargs: Dict with Attributes from which test class is
+        #   constructed
+        # * inst_kwargs: Dict with Attributes from which test inst is
+        #   constructed
+        # * exp_result: Dict of all expected attributes of resulting
+        #     CIMInstanceName without host and namespace if expected to suceed.
```
andy-maier: I'd prefer if we could also add namespace and host to the test input, and to the expected CIMInstanceName.

testsuite/test_cim_obj.py
```diff
@@ -2343,7 +2345,230 @@ def test_CIMInstanceName_from_wbem_uri(
...
+        #     Exception type, if expected to fail.
+        # * strict: Value of strict attribute on from_instance call
+        # * condition: Condition for testcase to run.
```
andy-maier: I suggest to add the capability to express an expected exception type, and expected warnings.

testsuite/test_cim_obj.py
KSchopmeyer: Done

pywbem/cim_obj.py

pywbem/cim_obj.py

pywbem/cim_obj.py

pywbem/cim_obj.py
KSchopmeyer: Done and note about discussion.

pywbem/cim_obj.py
KSchopmeyer: I added an issue for the new method as documentation since it is a general enhancement and added that to title of the pr.

Done

pywbem/cim_obj.py
KSchopmeyer: **DISCUSSION:** the general solution to logging, etc. of CIMNames.  One question is whether logging is generally protective of issues with unicode in name displays.

pywbem/cim_obj.py
KSchopmeyer: We cannot know that at this time since we do not have access to a repository. Therefore, we either ignore the issue about the relation between the classs except for documentation or keep the existing test which expects the classname to be the same as the instancename.

**DISCUSSION:** Should we drop the test for classname matching classname in instance??

andy-maier: I think we should drop the test.
And yes, I meant the requirement only to be documented, but not to be implemented as a check.

andy-maier:
**Discussion points from the comments:**

* For ` CIMInstance.Name.from_instance()`: What requirement do we document exists for the class so that it has the key properties the function wants to determine, and what checks related to that are made in the code, if any.
* In error messages, log entries etc: Use of `%s` vs. `%r`.

KSchopmeyer:
Lets force strict mode for now but leave the compatible code.

Remove the class match from from_instance and just be sure the documentation covers the issue.

No changes to the logging %s logging output for the moment.  See issue #1072

KSchopmeyer:
Add todo about breaking out repository as separate file.

andy-maier:
Approved and rebased.

PR #1045, opened 2018-02-04
Fixes #1031: (7) Added support for ARRAYSIZE to PROPERTY.ARRAY in tupleparser; test cases

### LINKED ISSUE #1031: array size on property arrays is ignored in tupleparser
In `parse_property_array()`, ArraySize is tolerated but ignored.

andy-maier:
**Note:** This PR is on top of PR #1029.
For details, see the commit message (of only the last commit).
Ready for review and merge.

### DIFF
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -80,10 +80,6 @@ representation of CIM in XML by having the following properties:
 from __future__ import absolute_import
 import re
 import six
-try:
-    from collections import OrderedDict
-except ImportError:
-    from ordereddict import OrderedDict
 
 from .cim_obj import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
     CIMProperty, CIMMethod, CIMParameter, CIMQualifier, \
@@ -1122,7 +1118,7 @@ def parse_qualifier_declaration(tup_tree):
     except KeyError:
         is_array = False
 
-    array_size = attr.get('ARRAYSIZE', None)
+    array_size = attrl.get('ARRAYSIZE', None)
     if array_size is not None:
         # TODO 2/18 AM #1044: Clarify if hex support is needed.
         array_size = int(array_size)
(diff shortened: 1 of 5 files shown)


pywbem/tupletree.py
```diff
@@ -41,9 +41,9 @@
 
 The NAME is the name of the element.
 
-The ATTRS are a name-value hash of element attributes.
+The ATTRS are a name-value dictionary of element attributes, not pres. order.
```
KSchopmeyer: what does pres. mean

KSchopmeyer: ahha.  preserving.  Lets spell it out.

pywbem/tupleparse.py
```diff
@@ -1201,7 +1232,7 @@ def parse_property(tup_tree):
                 'EMBEDDEDOBJECT'],
                ['QUALIFIER', 'VALUE'])
 
-    quals = {}
+    quals = OrderedDict()
```
KSchopmeyer: Why OrderedDict rather than NocaseDict???

andy-maier: The dict of the input arg is not used anyway; its items are put into a NocaseDict by the implementation of CIMProperty, so OrderedDict was the slightly cheaper alternative.

However, looking at the code again after your comment, it occurred to me that we now can pass in the list returned from list_matching() directly. Like we already do for CIMClass.

I'll change that.

pywbem/tupletree.py
andy-maier: Was too long for the line. But you are right, we should spell it out.

pywbem/tupleparse.py
andy-maier: DONE

pywbem/tupletree.py
andy-maier: DONE

PR #1070, opened 2018-02-07
Fixes #1038: Rejecting void method return types in tupleparser

### LINKED ISSUE #1038: Handling of METHOD without TYPE in tupleparser
In DSP0201, the `TYPE` attribute of `METHOD` is optional and omitting it is described to mean a void return type for the method. However, void return types of methods are not supported in DSP0004.

In pywbem, the tupleparser defines `TYPE` as an optional attribute, following the DSP0201 definition. However, if the attribute is actually omitted, the tupleparser cannot create the corresponding `CIMMethod` object from the CIM-XML because its ctor rejects `return_type=None`, following the DSP0004 definition. This results in a `ValueError` exception being raised.

Options:
* Support void methods in CIM-XML as defined in DSP0201. This would need to raise an error when trying to create MOF, because it is not representable in MOF.
* Do not support void methods, consistently.

See also issue #1050 which requests clarification in DSP0201.

**Discussion point:** We need to clarify how we want to handle void methods. Are there WBEM servers (or clients) that support void methods?

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -1388,9 +1388,14 @@ def parse_method(tup_tree):
 
     qualifiers = list_of_matching(tup_tree, ['QUALIFIER'])
 
-    # TODO #1038: Clarify how to deal with omitted TYPE of METHOD
+    return_type = attrl.get('TYPE', None)
+    if not return_type:
+        raise ParseError("Element %r missing attribute 'TYPE' (a void method "
+                         "return type is not supported in CIM)" %
+                         name(tup_tree))
+
     return CIMMethod(attrl['NAME'],
-                     return_type=attrl.get('TYPE', None),
+                     return_type=return_type,
                      parameters=parameters,
                      qualifiers=qualifiers,
                      class_origin=attrl.get('CLASSORIGIN', None),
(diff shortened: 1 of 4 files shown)


pywbem/tupleparse.py
```diff
@@ -1394,14 +1394,14 @@ def parse_method(tup_tree):
...
     qualifiers = list_of_matching(tup_tree, ['QUALIFIER'])
 
-    # TODO 2/18 AM #1038: Clarify how to deal with omitted TYPE of METHOD
-    # In DSP0201, TYPE is optional and omitting it means a void return
-    # type. That is not supported in DSP0004.
-    # This code here fails when creating the CIMMethod object with an omitted
-    # TYPE attribute.
+    return_type = attrl.get('TYPE', None)
+    if not return_type:
+        raise ParseError("Element %r misses attribute 'TYPE' (a void method "
```
KSchopmeyer: Change misses to missing

andy-maier: Done.

andy-maier:
Addressed comments and rebased.

PR #1091, opened 2018-02-19
Cleaned up run_cimoperations.py PyWBEMListenerClass

KSchopmeyer:
1. Built try blocks around all methods that create listeners so we are
certain to always destroy listeners at the end of each test.
2. Build context manager around most uses of Subscription manager to
automatically clean up if we do not clean up manually.
3. Move a number of function to common methods that were being repeated
in individual methos.

The enumerateinstances with deep inheritance test that fails was confirmed
as a server problem in that 1. Instance return includes properties from
superclass that it should not include. 2. Pegasus trace shows correct
value for de and lo getting to server.  Change message slightly to
state that it is just a warning.

There appears to be an issue in OpenPegasus when you set up multiples of exactly the same
subscription.  The provider appears to deliver exactly the correct number of indications to IndicationService but between IndicationService (which tests the received indication against the filter) and IndicationHandler component Pegasus sometimes multiplies the number of indications actually delivered to the user.  A trace, for example shows 2 indications (one from each indication generation provider call (we set the test to deliver one indication and quit)) delivered to indication service but there were 4 or more calls within the IndicationHandler.  It is like the filter review process is getting confused when there are multiples of the same filter. Note that our filter is also very simple (Select * from <classname>)
ACTION: File this as potential Pegasus bug.

### DIFF
--- a/testsuite/run_cim_operations.py
+++ b/testsuite/run_cim_operations.py
@@ -15,6 +15,8 @@ from __future__ import absolute_import
 
 # pylint: disable=missing-docstring,superfluous-parens,no-self-use
 import sys
+import re
+from socket import getfqdn
 import os.path
 import threading
 import types
@@ -23,6 +25,7 @@ import unittest
 from getpass import getpass
 import warnings
 import time
+import traceback
 from six.moves.urllib.parse import urlparse
 import six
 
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Andy. I am confused by these conflicts.  This looks like an issue in git diffing.  It appears that they simply cannot sort out the fact that the change I made in these places was to indent the code.

KSchopmeyer:
Rebased after you had made more committs and this time the rebase noted the issue in run_cimoperation.py and I fixed it.

Not sure why the earlier issue but it is clean now.

andy-maier:
Karl, when you do a `git rebase -i master`, delete the commit with title "Test: Renamed pywbem assertion methods to avoid unittest conflicts.", and then resolve the remaining conflict.

PR #1101, opened 2018-02-21
Remove the logging functionality from pywbem_mock

KSchopmeyer:
Completely remove the logging from the mocker since it is little
different from the cim_operations api logger and that logger can be used
to accomplish effectively the same thing. The only real difference
between the two is that that small bits of code that emulates
_imethodcall and _method_call return mapping from pure object to tuples
of data for each response.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -33,7 +33,6 @@ from __future__ import absolute_import, print_function
 import copy
 import uuid
 import time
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Andy. This thing has been stuck in appveyor for at least 18 hours now.

andy-maier:
After trying myself without success, I sent a mail to the Appveyor support tonight and they had it fixed within 3 minutes. Great support!

Issue #990, opened 2018-01-19
MOF compiler does not roll back changes to qualifier declarations


andy-maier:
This issue resulted from the following TODO in mof_compiler.py:

W:2216, 0: TODO: We want rollback to do something with qualifiers? (fixme)

The conclusion in a discussion about this was that we want changes to qualifier declarations as well.

When implementing that, we need to make sure we only roll back the changes we did, and only exactly the change we did. Remember, qualifiers have only a set operation that performs both modification and creation. That means, the undo action for set qualifier cannot be figured out from the input parameters alone, but it also needs the prior state.

andy-maier:
We agreed to move this beyond 0.12.

andy-maier:
We discussed the value of this and concluded that it is unimportant enough to not spend time on it. Closed as wont fix.

PR #1108, opened 2018-02-23
Docs: Simplified description of utility commands

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/utilities.rst
+++ b/docs/utilities.rst
@@ -14,12 +14,12 @@ The following commands are provided:
 
 * :ref:`mof_compiler`
 
-  A MOF compiler that takes MOF files as input and creates, updates or
-  removes CIM instances, classes or qualifier types in a CIM repository.
+  A MOF compiler that takes MOF files as input and updates the CIM repository
+  of a WBEM server with the result.
 
 * :ref:`wbemcli`
 
-  A WBEM command line interface that provides an interactive Python
+  A WBEM client in the form of a shell that provides an interactive Python
   environment for issuing WBEM operations to a WBEM server.
 
 .. _`mof_compiler`:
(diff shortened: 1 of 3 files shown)


docs/utilities.rst
```diff
@@ -14,8 +14,8 @@ The following commands are provided:
 
 * :ref:`mof_compiler`
 
-  A MOF compiler that takes MOF files as input and creates, updates or
-  removes CIM instances, classes or qualifier types in a CIM repository.
+  A MOF compiler that takes MOF files as input and updates the CIM repository
+  of a WBEM server with the result.
```
KSchopmeyer: My one argument with this is that the CIM repository may or may not be part of a server.  It may also be a standalone repository for other usage such as documentation.

docs/utilities.rst
```diff
@@ -27,46 +27,42 @@ The following commands are provided:
...
-The ``mof_compiler`` command is a command line interface to the pywbem MOF
-compiler. It compiles MOF files, and updates the repository of a WBEM server
-with the result.
+The ``mof_compiler`` command compiles MOF files and updates the CIM repository
+of a WBEM server with the result.
 
-The MOF compiler can also be invoked from programs via the
-:ref:`MOF compiler API`.
+If the compiler fails, any changes made to the CIM repository in the WBEM server
+as part of the current compilation are rolled back.
```
KSchopmeyer: Am I missing something? Did we change the qual decls to roll back?

docs/utilities.rst
```diff
@@ -27,46 +27,42 @@ The following commands are provided:
...
 
 wbemcli
 -------
 
-The ``wbemcli`` command is a WBEM client command line interface (CLI). It is
-implemented as an interactive shell.
-
-The WBEM client CLI does not have an external API on its own; it is for the
-most part a consumer of the :ref:`WBEM client library API`.
+The ``wbemcli`` command is a WBEM command line interface that provides an
```
KSchopmeyer: What is a WBEM command line interface?  Shouldn't it be pywbem command line interface?

docs/utilities.rst
andy-maier: Karl, what you say is true for the MOF compiler API, but this section is about the mof_compiler utility command and as I understand it, it supports only the CIM repository in a WBEM server, and no user-pluggable repository like the API does. Please double check.

docs/utilities.rst
andy-maier: Good catch ... we should mention that as a limitation.

docs/utilities.rst
andy-maier: How about "WBEM client"?

docs/utilities.rst
andy-maier: DONE.
Added: " A limitation is that changes to qualifier declarations are not yet rolled back (see issue #990)."

docs/utilities.rst
andy-maier: DONE. The description of the wbemcli command has been changed consistently to:
"a WBEM client in the form of a shell that provides an interactive Python environment for issuing WBEM operations to a WBEM server."

andy-maier:
Updated the PR with resolutions to all of the comments, as shown above, with the exception for the one comment about the description of the mof_compiler command that is a **discussion point:**

* Docs: "A MOF compiler that takes MOF files as input and updates the CIM repository of a WBEM server with the result."
  Karl: My one argument with this is that the CIM repository may or may not be part of a server. It may also be a standalone repository for other usage such as documentation.
  Andy: Karl, what you say is true for the MOF compiler API, but this section is about the mof_compiler utility command and as I understand it, it supports only the CIM repository in a WBEM server, and no user-pluggable repository like the API does. Please double check.

docs/utilities.rst
KSchopmeyer: You are right but even that is too limited.  It really just sends messages to create/delete classes/instances/qualifierdeclarations in a WEMServer defined by the script input parameters.

Normally this is the cim repository but it could also be a provider.

KSchopmeyer: But what you said is sufficient.

andy-maier: Ok, then I consider the discussion resolved and the PR approved.

PR #1129, opened 2018-03-21
Docs: Editorial changes in Security chapter

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/cim_http.py
+++ b/pywbem/cim_http.py
@@ -124,7 +124,7 @@ DEFAULT_PORT_HTTP = 5988        # default port for http
 DEFAULT_PORT_HTTPS = 5989       # default port for https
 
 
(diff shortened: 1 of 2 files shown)


pywbem/cim_http.py
```diff
@@ -231,19 +231,19 @@ def timer_expired(self):
 
 
 def parse_url(url, allow_defaults=True):
-    """Return a tuple of ``(host, port, ssl)`` from the URL specified in the
-    ``url`` parameter.
+    """Return a tuple (host, port, ssl) from the URL specified in the
```
KSchopmeyer: Why no ` around host, etc. here but there is below?

andy-maier: I did not go through the whole file because none of its functions is part of the external API. However, I did now and made it consistent.

Regarding the quotes: That was simply not consistent. I added one back-quote back in.

PR #1131, opened 2018-03-21
Docs: Fixes and improvements in WBEM server chapter

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -71,7 +71,7 @@ Example output:
     WBEM server URL:
       http://0.0.0.0
     Brand:
-      pegasus
+      OpenPegasus
     Version:
       2.12.0
     Interop namespace:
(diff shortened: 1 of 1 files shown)


pywbem/_server.py
```diff
@@ -266,7 +264,7 @@ def brand(self):
     @property
     def version(self):
         """
-        Version of the WBEM server, as a :term:`string`. `None`, if the version
+        :term:`string`: Version of the WBEM server. `None`, if the version
```
KSchopmeyer: Suggestion.  Where do we get the version???

andy-maier: We parse it out of the Description property of the CIM_ObjectManager instance, if it contains a pattern `r".+ *Version *([^ ]+)"`

PR #1134, opened 2018-03-23
Docs: Editorial changes for WBEM listener chapter

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -571,16 +571,16 @@ class WBEMListener(object):
         Parameters:
 
           host (:term:`string`):
-            IP address or host name this listener can be reached at.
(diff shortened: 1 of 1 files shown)


pywbem/_listener.py
```diff
@@ -702,7 +704,7 @@ def http_port(self):
     @property
     def https_port(self):
         """
-        The HTTPS port this listener can be reached at, as an :term:`integer`.
+        :term:`integer`: The HTTPS port this listener can be reached at.
```
KSchopmeyer: You gotta love english and split infinitives.  Change to The HTTPS port at which this listener can be reached.

andy-maier: Changed as proposed, in all 6 occurrences.

PR #1137, opened 2018-03-23
Fixes #1136: Various fixes and test case improvements for generating CIM-XML

### LINKED ISSUE #1136: Invalid CIM-XML for embedded instance and named instance
Issue #1123 has brought up an additional issue in pywbem: It creates incorrect CIM-XML for embedded instances and for named instances:

* According to DSP0201, an embedded instance must be represented as an INSTANCE element. Pywbem represents that as a VALUE.NAMEDINSTANCE element.
* Pywbem represents the instance path in a VALUE.NAMEDINSTANCE element as a LOCALINSTANCEPATH child element, when according to DSP0201, only INSTANCENAME is allowed.

For more details, see issue #1123.

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -2042,15 +2042,33 @@ class CIMInstanceName(_CIMComparisonMixin):
         return self.keybindings.iteritems()
 
     # pylint: disable=too-many-branches
-    def tocimxml(self):
+    def tocimxml(self, ignore_host=False, ignore_namespace=False):
         """
         Return the CIM-XML representation of the
         :class:`~pywbem.CIMInstanceName` object,
         as an instance of an appropriate subclass of :term:`Element`.
 
-        The returned CIM-XML representation is consistent with :term:`DSP0201`.
+        Parameters:
 
-        The order of keybindings is preserved.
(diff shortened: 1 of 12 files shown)


pywbem/cim_obj.py
```diff
@@ -4799,63 +4866,78 @@ def tocimxml(self):
...
 
-        The order of qualifiers is preserved.
+        The order of qualifiers in the returned CIM-XML representation is
+        preserved from the :class:`~pywbem.CIMProperty` object.
         """
 
+        qualifiers = [q.tocimxml() for q in self.qualifiers.values()]
+
         if self.is_array:
+            assert self.type != 'reference'
```
KSchopmeyer: Isn't this a real function change? I have not yet looked but is it true that we disallow arrays of references.

KSchopmeyer:
Ran run_cim_operations with no issues several times.

pywbem/cim_obj.py
andy-maier: Yes, DSP0004 disallows arrays of reference properties (it does allow parameters that are arrays of references, though).

PR #1139, opened 2018-03-27
Add comment on docstring of CIMInstanceName from_string method

KSchopmeyer:
Add to documentation that IncludeQualifiers should be True.

This is just a cautionary note since the setting in a server is implementation dependent if you do not set this and that could mean creating path with no keybindings.


pywbem/cim_obj.py
```diff
@@ -2571,7 +2571,8 @@ def from_instance(class_, instance, namespace=None, host=None,
                 object. Thus, for example, if the class were retrieved from a
                 server, generally, the `LocalOnly` parameter in the request
                 should be `False` to assure that all superclass properties are
-                retrieved.
+                retrieved and the IncludeQualifiers should be set to `True`
```
andy-maier: Use backquotes around "IncludeQualifiers", and insert "parameter" after it, consistent with the text for "LocalOnly".

PR #1140, opened 2018-03-27
Add unit test for WBEMserver class based on pywbem_mock.

KSchopmeyer:
Creates a new test module, test_serverclass.py that executes a unit test on the WBEMServer class using pywbem_mock.

This creates a number of instances and executes the WBEMServer methods against the mock. Note that in order to do this I created a method in the test code inst_from_class that creates instances given the class (which must be in the repository) and a dictionary of propertyname/propertyvalue that fills out each instance.

I also created a method that builds the class repository in the mock wbem server given only the classes that will be used in the test (the mof compiler finds all super classes).  This reduces the repository load time from multiple seconds to milliseconds since It only installs about 10 classes for this test.

Right now tests all of the calls for brands, namespaces, profiles except for get_central_instance but only for correct responses, not for error responses. Without the get_central class it gives me about 60% coverage for that module.

It will take a bit longer to add the test cases for get_central class so there is about a day more work on this to make it complete.

The inst_from_class (I created this one after Andy's first comments) method works for these tests but I have a couple of things in that I need to generalize so that it matches the pattern of options on GetInstance better and I think I can drop the strict as applied to property values.  It really means nothing since setting the value validates the value from the PropertyValues dictionary.  The only use for strict is for creating the path.

I never get it right the first time :-)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -404,6 +404,10 @@ Enhancements
   consistency with ignoring the path on
   `CIMInstance.tocimxml()/tocimxmlstr()` (as part of fixing issue #1136).
 
+* Added unit test for WBEMServer class using pywbem_mock.  See the file
+  testsuite/test_wbemserverclass.py.  This test is incomplete today but tests
+  most of the main paths.
+
 Bug fixes
 ^^^^^^^^^
 
--- a/testsuite/test_wbemserverclass.py
+++ b/testsuite/test_wbemserverclass.py
@@ -29,8 +29,10 @@ import os
 import pytest
 
 from pywbem import WBEMServer, ValueMapping, CIMInstance, CIMInstanceName
+from pywbem.cim_obj import NocaseDict
 
 from pywbem_mock import FakedWBEMConnection
+
 from dmtf_mof_schema_def import install_dmtf_schema, SCHEMA_MOF_DIR
 
 TEST_SCHEMA = os.path.join(SCHEMA_MOF_DIR, 'test_schema1.mof')
@@ -60,6 +62,8 @@ class BaseMethodsForTests(object):
             #pragma include ("Interop/CIM_RegisteredProfile.mof")
             #pragma include ("Interop/CIM_Namespace.mof")
             #pragma include ("Interop/CIM_ObjectManager.mof")
+            #pragma include ("Interop/CIM_ElementConformsToProfile.mof")
+            #pragma include ("Interop/CIM_ReferencedProfile.mof")
             """
         # See issue 1138.
         # conn.compile_mof_str(class_list, namespace=namespace,
@@ -74,16 +78,21 @@ class BaseMethodsForTests(object):
             os.remove(TEST_SCHEMA)
         return
 
-    def inst_from_class(self, klass, namespace=None, PropertyValues=None,
-                        IncludeNullProperties=True,
-                        IncludePath=True, strict=False,):
+    def inst_from_class(self, klass, namespace=None,
+                        property_values=None,
+                        include_null_properties=True,
+                        include_path=True, strict=False,
+                        include_class_origin=False):
         """
         Build a new CIMInstance from the input CIMClass using the
-        PropertyValues dictionary to complete properties and the other
+        property_values dictionary to complete properties and the other
         parameters to filter properties, validate the properties, and
         optionally set the path component of the CIMInstance.  If any of the
         properties in the class have default values, those values are passed
-        to the instance unless overridden by the PropertyValues dictionary.
+        to the instance unless overridden by the property_values dictionary.
+        No CIMProperty qualifiers are included in the created instance and the
+        `class_origin` attribute is transfered from the class only if the
+        `include_class_origin` parameter is True
 
         Parameters:
           klass (:class:`pywbem:CIMClass`)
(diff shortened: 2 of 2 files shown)


testsuite/test_serverclass.py
```diff
@@ -0,0 +1,267 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+#
+# Author: Karl  Schopmeyer <inovadevelopment.com>
+#
+
+"""
+Test for the _server class in pywbem that uses the mock support to exercise
```
andy-maier: There is no "_server" class. Either this test module as the scope of the "_server module" or the "WBEMServer class". I suppose it is the former.

testsuite/test_serverclass.py
```diff
@@ -0,0 +1,267 @@
...
+        conn = FakedWBEMConnection()
+        system_name = '"Mock Test_server_Class"'
+
+        self.build_instances(conn, tst_namespace, system_name)
+
+        server = WBEMServer(conn)
+
+        for iname in conn.EnumerateInstanceNames("CIM_Namespace",
+                                                 namespace=tst_namespace):
+            print('CIM_Namespace name %s' % iname)
```
andy-maier: The test should be silent if it succeeds.

testsuite/test_serverclass.py
```diff
@@ -0,0 +1,267 @@
...
+            `None` if the default_namespace is to be used.
+
+          PropertyList (string or iterable of string)
+            Limits the properties retrieved witht the class and therefore
+            the properties in the resulting instance.
+
+          PropertyValues (dictionary):
+            Dictionary containing nane/value pairs where the names are the
+            names of properties in the class and the properties are the
+            property values to be set into the instance.
```
andy-maier: The description leaves it open what happens when PropertyList and PropertyValues do not specify the same set of properties. If we think this is always the same set, PropertyList could be eliminated as a parameter and calculated internally as the set of keys of PropertyValues.

testsuite/test_serverclass.py
```diff
@@ -0,0 +1,267 @@
...
+
+          IncludeNullProperties (:class:`py:bool`:):
+            Determines if properties with Null values are included in the
+            instance.
```
andy-maier: The description leaves it open as to whether this condition is applied as an additional restriction to the set of properties in PropertyList, vs. adding to them.

testsuite/test_serverclass.py
```diff
@@ -0,0 +1,267 @@
...
+
+          strict (:class:`py:bool`:):
+            If `True` the property type in PropertyValue must exactly match
+            the property type in the class. Also, if IncludePath is `True` the
+            created instance is tested to assure that all key properties from
+            the class are in the instance so a complete path can be buil
+
+            If not `True` the value is inserted into the property whatever the
+            type
```
andy-maier: non-strict apparently provides for the capability to set up mocked instances with property values that have CIM types that do not match the CIM types declared in the class. Is that the intention?

testsuite/test_serverclass.py
KSchopmeyer: Since the only thing in the_server module is the WBEMServer class it is probably more logical to make it the WBEMServer class

Changed the name of the test file.

testsuite/test_serverclass.py
KSchopmeyer: Agreed. Done

testsuite/test_serverclass.py
KSchopmeyer: 1. Modified the code so that we now have a inst_from_class which could at some point be inserted into CIMInstance as from_class and inst_from_classname which just gets the class from the mock repo and calls inst_from_class and is intended just for this test code.

Changed doc to state that if a property is in the PropertyValues but not in the class a ValueError is raised.

I don't think is is always the same set since the PropertyList would be a way to force elimination of properties from the instance even if they had a default value.

2. So now, PropertyList is probably just a way to clip off properties for some tests.  Actually not sure I have a use for it.  I states in the klass parameter definition of inst_from_class that the class should include the properties that the user wants in the instance and must contain the key properties if the include_path is True and strict is True

testsuite/test_serverclass.py
KSchopmeyer: That is now clear I think since I separated getting the class so that the instance is created from the class in inst_from_class which does not have the PropertyList parameter.

testsuite/test_serverclass.py
KSchopmeyer: Yes but I want to rethink that before we would consider moving this to the pywbem code.  I think I probably should build the CIMProperty with the value so that all the attributes of construction of a CIMProperty are exercised.  that would make use of CIMProperty complete construction tools and eliminate the the test I have which requires an exact match.  The current code works for the test but not if we were to move this to pywbem

testsuite/test_serverclass.py
KSchopmeyer: Changed name in doc string and also changed name to test_wbemserverclass.py

testsuite/test_serverclass.py
KSchopmeyer: Modified the code so that strict only applies to the path now.  It was my error that I was trying to double check what CIMInstance.value did.

KSchopmeyer:
Changes per comments incorporated and added changes.rst paragraph and clarified state of tests

PR #1142, opened 2018-03-29
Fix issue in pywbem_mock with incorrect error response to invokemethod

KSchopmeyer:
We need to return something like NOT_SUPPORTED when there are no methods
in the methods repo. The code was returning NOT_FOUND.  This was causing
issues with the get_central_instances test.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -45,8 +45,9 @@ import six
 from pywbem import WBEMConnection, CIMClass, CIMClassName, \
     CIMInstance, CIMInstanceName, CIMQualifierDeclaration, \
     CIMParameter, cimtype, CIMError, \
-    CIM_ERR_NOT_FOUND, CIM_ERR_FAILED, CIM_ERR_INVALID_SUPERCLASS, \
-    CIM_ERR_INVALID_PARAMETER, CIM_ERR_INVALID_CLASS, CIM_ERR_ALREADY_EXISTS, \
+    CIM_ERR_NOT_FOUND, CIM_ERR_METHOD_NOT_FOUND, CIM_ERR_FAILED, \
+    CIM_ERR_INVALID_SUPERCLASS, CIM_ERR_INVALID_PARAMETER, \
+    CIM_ERR_INVALID_CLASS, CIM_ERR_ALREADY_EXISTS, \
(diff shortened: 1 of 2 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -2913,9 +2917,21 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
...
+                    raise CIMError(CIM_ERR_NOT_SUPPORTED,
+                                   'Method %s in namespace %s not registered '
+                                   'in repository' %
+                                   (methodname, namespace))
+                else:
+                    raise
         except KeyError:
-            raise CIMError(CIM_ERR_NOT_FOUND, 'Method %s in namespace %s not '
-                                              'registered in repository' %
+            raise CIMError(CIM_ERR_NOT_SUPPORTED,
```
andy-maier: `self._get_methods_repo()` cannot raise `KeyError` as far as I can tell from the code. I know this code was there before this change, but nevertheless we should verify and simplify this.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -2913,9 +2917,21 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
         namespace = localobject.namespace
         try:
             methodsrepo = self._get_methods_repo(namespace)
+        except CIMError as ce:
+            # If invoke indicates method repo not built, create it.
+            if ce.status_code == CIM_ERR_INVALID_NAMESPACE:
+                if namespace in self.classes or \
+                        namespace in self.instances:
+                    raise CIMError(CIM_ERR_NOT_SUPPORTED,
```
andy-maier: Two issues: 

1. According to DSP0200, CIM_ERR_NOT_SUPPORTED in InvokeMethod means that the WBEM server generally does not support method invocations. That is not the case for this situation. DSP0200 offers CIM_ERR_METHOD_NOT_FOUND for the case where the class does not expose the method.
invocations.)
2. If we want to raise the error that the method does not exist on the class (which I understand the intention is), then this check for it is incorrect, IMO, and one of the checks further down is probably close.
3. The checks further down seem pretty convoluted, with error messages that have subtle nuances in what they say, and that don't tell clearly what's actually wrong.

KSchopmeyer:
Fixed Anyy's comments

PR #1154, opened 2018-03-30
Fixes #1138: Compile mof based on pragma include file from compile_string

### LINKED ISSUE #1138: MOF Compiler fails string compile with the dir search options
The mof_compiler compile_file search option lets the compiler search directories in an iterable of directories for dependent classes.  Thus, classes where the superclasses reference properties, etc.are not in the repo these directories are searched for these classes.

However, despite the fact that compile_string has access to the same search path iterable it does not work if the mof to be compiled is a string rather than a file

If fails at the first include because that is not considered a dependent class so the find_class algorithm is not implemented.

Thus, if I can do:

``

        mofcomp = MOFCompiler(MOFWBEMConnection(),
                              search_paths=search_paths,
                              verbose=verbose)
        mofcomp.compile_file(mof_file, namespace)

where mof_file is:

``

     #pragma locale ("en_US")
     #pragma include ("Interop/CIM_RegisteredProfile.mof")

I should also be able to do:

``

       mof_str = """
            #pragma locale ("en_US")
            #pragma include ("Interop/CIM_RegisteredProfile.mof")
            """
        mofcomp = MOFCompiler(MOFWBEMConnection(),
                              search_paths=search_paths,
                              verbose=verbose)
        mofcomp.compile_str(mof_str, namespace)

KSchopmeyer:
Fix issue where compile_string that starts with include filefails the compile, cannot find the classes in the include file.

I fixed the issue and added tests for compiling partial pragma include files both as a file and as a string.

However, as we noted in issue #1160, the dependency finder mechanism in the compiler is still broken as it will not figure out that EmbeddedInstance classes or references are not compiled and, will in fact successfully complete a pragma or class compile while with these dependent classes completly missing.

This pr does not test for those dependency but sets up tests we can extend easily with the pr for issue #1138.  

Shakespeare in our world(From MacBeth.   'Out dammed bugs! Out I say!'.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -177,7 +177,7 @@ Incompatible changes
   defined through a mof file containing  #pragma include statements.
   This precluded using a string to define the classes to include in
   a mof compile in a string and required that the include be a file.
-  See issue 1138
+  See issue #1138
 
 Deprecations
 ^^^^^^^^^^^^
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -173,6 +173,12 @@ Incompatible changes
   - The "fq" parameter of the filtering functions was renamed to "fs",
     for consistency with the query functions.
 
+* Fix issue where mof. compile_str could not compile mof that was
+  defined through a mof file containing  #pragma include statements.
+  This precluded using a string to define the classes to include in
+  a mof compile in a string and required that the include be a file.
+  See issue 1138
+
 Deprecations
 ^^^^^^^^^^^^
 
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -717,8 +717,9 @@ def p_compilerDirective(p):
     param = p[5]
     if directive == 'include':
         fname = param
-        if os.path.dirname(p.parser.file):
-            fname = os.path.dirname(p.parser.file) + '/' + fname
+        if p.parser.file:
+            if os.path.dirname(p.parser.file):
+                fname = os.path.dirname(p.parser.file) + '/' + fname
         p.parser.mofcomp.compile_file(fname, p.parser.handle.default_namespace)
     elif directive == 'namespace':
         p.parser.handle.default_namespace = param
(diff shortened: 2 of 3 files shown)


pywbem/mof_compiler.py
```diff
@@ -2312,8 +2313,16 @@ def __init__(self, handle, search_paths=None, verbose=False,
...
+            An iterable of path names of directories where compiler will search
+            for mof files to complete a compile operation.  The compiler
+            searches these paths (including subdirectories) for
+            mof in a number of cases including: 1) find a superclass that
+            is not in repository while compiling a class 2) Finding a
+            qualifier that is not in the repository (it looks for filenames
+            'qualifiers' and 'qualifiers_optional', 3) finding a dependent
+            class reference property and the EmbeddedInstanceQualifier)
+            Currently this works for superclasses but not for embedded
+            instance qualifier and reference properties/parameters.
```
andy-maier: "Currently this works for superclasses" in item 3 created some confusion on my part until I realized that this part of the sentence is a statement about item 1. I suggest to change this sentence to only talk about item 3, i.e. change it to something like "(not currently implemented)".

docs/changes.rst
```diff
@@ -173,6 +173,12 @@ Incompatible changes
   - The "fq" parameter of the filtering functions was renamed to "fs",
     for consistency with the query functions.
 
+* Fix issue where mof. compile_str could not compile mof that was
+  defined through a mof file containing  #pragma include statements.
+  This precluded using a string to define the classes to include in
+  a mof compile in a string and required that the include be a file.
+  See issue 1138
```
andy-maier: Add "#" in front of issue number, for consistency.

pywbem/mof_compiler.py
```diff
@@ -2430,13 +2439,16 @@ def compile_file(self, filename, ns):
...
             self.parser.log('Compiling file ' + filename)
 
-        f = open(filename, 'r')
-        mof = f.read()
-        f.close()
+        if not os.path.exists(filename):
+            filename = os.path.basename(filename)
+            filename = self.find_mof(filename[:-4].lower())
+            if filename is None:
+                raise ValueError('No such file: %s' % filename)
```
andy-maier: Before this change, the method raised IOError if the file could not be found. It now raises ValueError. This is an incompatible change for users of the MOF compiler API (e.g. our wbemcli Python script (which admittedly has an issue in handling any of these two exceptions). We should go back to raising IOError, IMO.

pywbem/mof_compiler.py
KSchopmeyer: OK.

docs/changes.rst
KSchopmeyer: DONE

PR #1169, opened 2018-04-04
WIP Fixes #1167 Add examples from _logging.py as tests. This found a couple

### LINKED ISSUE #1167: Add tests to logging that verify the documented examples
There are a few possibilities documented for configuring and activating the pywbem loggers, but we don't have test cases for them.

KSchopmeyer:
Added 3 tests that fairly closely duplicate the examples in _logging.py.  There are some differences
and those are documented in the doc string for each test.  Mostly the differences are to use file in
place of stderr so I could capture with log_capture.

NOTE: One of the tests is with basicconfig and that does output logs.

TODO: Create the last couple of tests for examples and do one with stderr rather than file.

This found a couple more issues in the code itself which were fixed.  NOTE: I also
added one item to the WBEMConnection __repr__ but that is
questionable.

code issues:
1. If log_name not provided we default, not fail
2. Not setting the class level detail_level dictionary in one case.
3. Couple of issues in the examples (used dest rather than log_dest, left  WBEMConnection without any url parameter
4. Summary log output not created for single object return.  It did summary for list but for single object output the whole object.
5. Used if in one place where should have been elif


andy-maier:
It seems to me this boils down to the question how the use of Python logging for the purpose of tracing (which is our use of log levels DEBUG in the client and INFO in the listener) should be handled. I think we rightfully want to be able to control this for each kind of tracing (i.e. api and http) separately.

Having no NullHandlers when our loggers are not configured by our config functions causes the root logger configuration to influence those pywbem loggers that are not configured. The question is whether we consider that a good thing or a bad thing.

andy-maier:
There is one more consideration w.r.t. NullHandlers: The pywbem listener issues log records at the ERROR level. That should probably not be prevented from propagating up. So if we add NullHandlers for the pywbem client, they would need to be on the "pywbem.api" and "pywbem.http" loggers, but not on the "pywbem" logger, because the "pywbem.listener" logger should propagate upwards.

KSchopmeyer:
Update.  I fixed the problem itself that showed up in test_5 by adding code to NOT log if the detail_level for a particular logger is None. That is logical I think. I don't like it since is is one more flag on the test for each log but it does work.  Note  from your issue #1181 that this test works probably because we build the repo(i.e create the FakedWBEMConnection) before we configure logging.

PR #1172, opened 2018-04-05
Fixes #1171: Fixes and improvements in logging code

### LINKED ISSUE #1171: Fixes and improvements in logging code
Items:
- Quality of exception messages
- Move WBEMConnection.configure_logger() method to become a function
- Small coding errors
- Setting `connection=False` in the configure functions? - No, use new function to deactivate

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/_logging.py
+++ b/pywbem/_logging.py
@@ -88,9 +88,8 @@ for logging through the following pywbem functions.
 These functions are the only mechanism for setting the detail level of a pywbem
 logger and for activating WBEM connection(s) for logging.
 
-* :meth:`~pywbem.WBEMConnection.configure_logger` - Configure the pywbem
-  loggers and optionally activate WBEM connections for logging and setting a
-  log detail level.
+* :func:`configure_logger` - Configure the pywbem loggers and optionally
+  activate WBEM connections for logging and setting a log detail level.
 
 * :func:`configure_loggers_from_string` - Configure the pywbem loggers and
   optionally activate WBEM connections for logging and setting a log detail
@@ -106,17 +105,15 @@ logger and for activating WBEM connection(s) for logging.
 Logging configuration examples
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Examples for using :meth:`~pywbem.WBEMConnection.configure_logger` for
-configuring pywbem loggers and for activating WBEM connections for logging:
+Examples for using :func:`configure_logger` for configuring pywbem loggers and
+for activating WBEM connections for logging:
 
 * Example: Configure the `'pywbem.api'` logger with detail level `'summary'`
   and output to stderr, and activate all subsequently created WBEM connections
   for logging::
 
-    WBEMConnection.configure_logger('api',
-                                    log_dest='stderr',
-                                    detail_level='summary',
-                                    connection=True)
+    configure_logger('api', log_dest='stderr', detail_level='summary',
+                     connection=True)
 
     # All of the following connections will log:
     conn1 = WBEMConnection(...)
@@ -126,11 +123,10 @@ configuring pywbem loggers and for activating WBEM connections for logging:
   and output to a file, and activate a single existing WBEM connection for
   logging::
 
-    conn = WBEMConnection()
+    conn = WBEMConnection(...)
 
-    WBEMConnection.configure_logger('all',
-                                    log_dest='file', log_filname='xxx.log',
-                                    connection=conn)
+    configure_logger('all', log_dest='file', log_filname='xxx.log',
+                     connection=conn)
 
 Examples for configuring the pywbem loggers using Python logging methods,
 and using the pywbem logging configuration functions only for setting the
@@ -147,9 +143,7 @@ detail level and for activating WBEM connections for logging:
 
     conn = WBEMConnection(...)
 
-    WBEMConnection.configure_logger('all',
-                                    detail_level='all',
-                                    connection=conn)
+    configure_logger('all', detail_level='all', connection=conn)
 
   TODO: This configures the pywbem loggers with a null handler and thus
   does not propagate up to the Python root logger. Clarify whether this
@@ -169,9 +163,7 @@ detail level and for activating WBEM connections for logging:
     handler.setFormatter(formatter)
     logger.addHandler(handler)
 
-    WBEMConnection.configure_logger('api',
-                                    detail_level='summary',
-                                    connection=True)
+    configure_logger('api', detail_level='summary', connection=True)
 
     # All of the following connections will log:
     conn1 = WBEMConnection(...)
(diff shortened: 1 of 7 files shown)


pywbem/_logging.py
```diff
@@ -32,157 +34,206 @@
...
-  for output of summary information for both "pywbem.api" and "pywbem.http"::
+    # All of the following connections will log:
+    conn1 = WBEMConnection(...)
+    conn2 = WBEMConnection(...)
+
+* Example: Configure all pywbem loggers with the default detail level (`'all'`)
+  and output to a file, and activate a single existing WBEM connection for
+  logging::
 
     conn = WBEMConnection()
```
KSchopmeyer: Put a url in this or the ... to indicate parameters.

pywbem/_logging.py
```diff
@@ -32,157 +34,206 @@
...
-  * :func:`configure_loggers_from_string` - Allows configuring and
-    activating the pywbem loggers from a single formatted string that defines
-    logger names, log destination, and detail level. This is most
-    useful in defining the loggers from a command line tool such as pywbemcli
-    so the complete logger configuration can be compressed into a single
-    command line string.
+Pywbem logging uses two Python :class:`~py:logging.Logger` objects which are
+termed `pywbem loggers`:
+
+* API logger (Python logger name: `'pywbem.api'`) - Logs API calls and returns,
```
KSchopmeyer: These names are only useful if the user uses the Python logger configure methods.

pywbem/cim_operations.py
```diff
@@ -1220,68 +1223,74 @@ def configure_logger(cls, name, log_dest=DEFAULT_LOG_DESTINATION,
...
+            available information up to that size.
 
-            If `None` it is set to the default detail level for that logger.
+            If `None`, the detail level configuration will not be changed.
 
           log_filename (:term:`string`):
-            Filename to use as logging file if the log destination is `file`.
-            Ignored if log destination is not `file`. This parameter is
-            required if the dest parameter is `file`.
+            Path name of the log file (required if the log destination is
```
KSchopmeyer: Not required.  That was my error. See pr for issue # 1167 that states that there is a default log file name.

pywbem/_recorder.py
```diff
@@ -594,6 +599,9 @@ def stage_wbem_connection(self, wbem_connection):
 
         if self.enabled:
             if self.api_detail_level or self.http_detail_level:
+
+                # import pdb; pdb.set_trace()
```
KSchopmeyer: Do you want to leave this in???

pywbem/cim_operations.py
```diff
@@ -1497,6 +1518,16 @@ def operation_recorder_stage_result(self, ret, exc):
             recorder.stage_pywbem_result(ret, exc)
             recorder.record_staged()
 
+    @classmethod
+    def future_logging_reset(cls):
+        """
+        Reset the activation of logging and the log detail levels for future
```
KSchopmeyer: This is a public method but is it intended for anything but our tests??

pywbem/cim_operations.py
```diff
@@ -1328,112 +1326,135 @@ def _configure_detail_level(cls, detail_level):
...
             else:
                 assert isinstance(connection, WBEMConnection)
-                local_dict = cls._log_config_dict.copy()
-                local_dict[logger_simple_name] = detail_level
 
-                # Insert recorder into list if not already there.
-                # pylint: disable=protected-access
-                recorder_found = False
+                # Activate logging for this existing connection by ensuring
+                # the connection has a log recorder
```
KSchopmeyer: I think I found an issue here that I fixed yesterday in the pr for issue #1167.  Will look later.

KSchopmeyer:
Made some comments but want to review again.

pywbem/_logging.py
andy-maier: I mentioned them exactly for that reason.

pywbem/_logging.py
andy-maier: DONE

pywbem/cim_operations.py
andy-maier: **Discussion point:** If we have a default log file name, we should do that for both configure functions, and we should not use the same constant for the path name of the file that we use for the log destination.

pywbem/_recorder.py
andy-maier: No. removed. :-)

pywbem/cim_operations.py
andy-maier: Right now the user can only activate logging for future connections, but has no means to deactivate it again. It is just cleaner if we provide a means to do that.

andy-maier:
The latest commit has resolutions to your comments, except for those marked discussion point.

pywbem/cim_operations.py
andy-maier: **Discussion point:** We should make sure that the two PRs can come together.

pywbem/cim_operations.py
```diff
@@ -1337,112 +1326,135 @@ def _configure_detail_level(cls, detail_level):
...
-                             'destinations are: %s' %
-                             (log_dest, LOG_DESTINATIONS))
 
-        # process dest parameter
+        msg_format = '%(asctime)s-%(name)s-%(message)s'
+
         if log_dest == 'stderr':
-            # Cannot use the stream parameter because not defined for
-            # python 2.6. Removed stream=sys.stderr
+            # TODO: stream=sys.stderr parameter is not defined in py26
```
andy-maier: stderr is the default - change TODO to a note

pywbem/cim_operations.py
```diff
@@ -1337,112 +1326,135 @@ def _configure_detail_level(cls, detail_level):
...
             if not log_filename:
-                raise ValueError('Filename required if log destination '
-                                 'is "file"')
-            # pylint: disable=redefined-variable-type
+                raise ValueError("Log filename is required if log destination "
+                                 "is 'file'")
             handler = logging.FileHandler(log_filename, encoding="UTF-8")
             handler.setFormatter(logging.Formatter(msg_format))
-        elif log_dest == 'none' or log_dest is None:
+        elif log_dest == 'none':
```
andy-maier: We should actually have a default of having a null handler, and don't need a means to explicitly set up the null handler.

pywbem/cim_operations.py
```diff
@@ -1506,6 +1518,16 @@ def operation_recorder_stage_result(self, ret, exc):
             recorder.stage_pywbem_result(ret, exc)
             recorder.record_staged()
 
+    @classmethod
+    def future_logging_reset(cls):
```
andy-maier: It seems this function is so out of place compared to the rest of the log config functions, that we are probably better off supporting the reset via the connection=False on configure_logger()

pywbem/cim_operations.py
andy-maier: DONE

pywbem/cim_operations.py
andy-maier: DONE - Removed the support for 'none' and added an according TODO for the best practices.

andy-maier:
Resolved all comments from Karl and from our call today, except the one for checking the best practices for using null handler on loggers of libraries, which was put down in the code as a TODO comment.

PR #1178, opened 2018-04-07
Fixes #1174 Renamed compile_mof_str() to compile_mof_string() in mock support

### LINKED ISSUE #1174: Change name of compile mof method in mocker
Currently the name for the compile string method is mocker is 

compile_mof_str

Would like to change it to compile_mof_string just to keep the full word string to be more consistent with what is in the mof compiler

compile_string

and 

compile_file

Would like to get this into 0.12.0 since this is the initial release.   

I am doing this only because I keep making the error when typing in the method name.

Note that we could even change to

compile_string

compile_file

To be completely consistent with the mof compiler.

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/mocksupport.rst
+++ b/docs/mocksupport.rst
@@ -115,7 +115,7 @@ executing WBEM operations on the faked connection:
 
     # Compile the MOF string and add its CIM objects to the default namespace
     # of the mock repository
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
There is one _str left.  Other than that OK

Issue #1181, opened 2018-04-08
FakedWBEMConnection cannot log


andy-maier:
The following sequence of python statements shows that FakedWBEMConnection cannot log:

```
>>> import pywbem
>>> import logging_tree as lt
>>> pywbem.configure_logger('api', connection=True)
>>> lt.printout()
<--""
   Level WARNING
   |
   o<--"BIO"
   |   Level NOTSET so inherits level WARNING
   |
   o<--"EVP"
   |   Level NOTSET so inherits level WARNING
   |
   o<--[M2Crypto]
   |   |
   |   o<--"M2Crypto.AuthCookie"
   |   |   Level NOTSET so inherits level WARNING
   |   |
   |   o<--[M2Crypto.SSL]
   |   |   |
   |   |   o<--"M2Crypto.SSL.Connection"
   |   |       Level NOTSET so inherits level WARNING
   |   |
   |   o<--"M2Crypto.X509"
   |       Level NOTSET so inherits level WARNING
   |
   o<--[pywbem]
   |   |
   |   o<--"pywbem.api"
   |       Level DEBUG
   |       Handler File '/media/share/Projects/pywbem/repos/pywbem/pywbem.log'
   |         Formatter fmt='%(asctime)s-%(name)s-%(message)s' datefmt=None
   |
   o<--"util"
       Level NOTSET so inherits level WARNING
>>> import pywbem_mock
>>> conn=pywbem_mock.FakedWBEMConnection()
Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/__init__.py", line 861, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
  File "pywbem_mock/_wbemconnection_mock.py", line 335, in __repr__
    (self.__class__.__name__, self.response_delay,
  File "pywbem_mock/_wbemconnection_mock.py", line 322, in response_delay
    return self._response_delay
AttributeError: 'FakedWBEMConnection' object has no attribute '_response_delay'
Logged from file _recorder.py, line 606
>>>
```

[...]

PR #1185, opened 2018-04-08
Fixes #1184 fix connection log detail_level

### LINKED ISSUE #1184: Logger for connection does not honor the detail_level=int
We are ignoring the detail_level=<int> definition for detail level when logging the connection.

KSchopmeyer:
Added code to honor all attributes of detail_level for connection log.  This initially failed because of issue # 1181.  So  I moved the _response_delay attribute set to before the superclass init and for me that eliminated the issue.

Sorry, originally created this from another branch which sort of made a mess of the files changed.

Adds some more tests, cleans up some pylint issues in test_recorder.py and cleans up the setup method (it was repeated in subclasses) for the log tests.

DISCUSSION: Is this a legitimate way to get around issue #1181 (just moving the object attribute to before the call to the superclass ctor?  If so, Just turn off discussion.

Andy: I think it is legitimate. Turning off discussion needed, and merging.

### DIFF
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -606,15 +606,21 @@ class LogOperationRecorder(BaseOperationRecorder):
             if self.api_detail_level is not None:
                 logger = self.apilogger
                 detail_level = self.api_detail_level
+                max_len = self.api_maxlen
             elif self.http_detail_level is not None:
                 logger = self.httplogger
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
Sorry for confusion on first push. I created this from pr #1169 because I needed the test class but that just sort of made the whole thing messy.

PR #1189, opened 2018-04-10
Fixes #1180 Changed default log_dest in config_logger() to None

### LINKED ISSUE #1180: Default log_dest on pywbem log config functions
In the code of the current master branch, the log_dest default is "file" because that seemed to be convenient if one wants to enable logging and uses the pywbem log config functions for both python logger configuration and pywbem activation.

However, it is inconvenient for users that want to use the python logging config functions and use the pywbem log config functions only for detail level and activation. It seems a bit weird that one needs to disable the default "file" to `None`, in that case.

**Discussion point:** Should we change the default for log_dest in the pywbem log config functions to `None`?

andy-maier:
For details, see the commit message. Note particularly, that the log_dest default was only changed for configure_logger() (not for the string version) and that the default log filename of that function was not changed to None.
Ready for review and merge.

### DIFF
--- a/pywbem/_logging.py
+++ b/pywbem/_logging.py
@@ -305,7 +305,7 @@ LOG_DETAIL_LEVELS = ['all', 'paths', 'summary']
 DEFAULT_LOG_DETAIL_LEVEL = 'all'
 
 
-def configure_logger(simple_name, log_dest=DEFAULT_LOG_DESTINATION,
+def configure_logger(simple_name, log_dest=None,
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
At this point I don't have an issue with forcing the user to set the log_dest on the string input.  The parser for the the string cannot handle a string of the form api=:summary so we would have to change that also.

Since the filename is not used unless it is a log_dest='file' that also does not seem to matter.

Issue #1186, opened 2018-04-09
Add NullHandlers or not?


andy-maier:
The current master branch does not have any NullHandlers.

See the (merged) PR #1169 for discussion items.

KSchopmeyer:
The question is whether my code that simply does not call the loggers is enough or we need to set NullLoggers for any that we do not use. 

I am conflicted right now but the current code does work for us. I propose we release as is and if we sort out issues we add NullHandlers.  That is opinion only but it is before noon so maybe partly rational.

KSchopmeyer:
My tendency right now is release as is since this is experimental.

andy-maier:
I did some more experiments, and found:
1. My understanding that the existence of a handler on a logger would prevent the upwards propagation was incorrect. I don't know where I got that from. Instead, what prevents upwards propagation is the "propagate" attribute on a logger being False and the level of a logger being more restricted than the log event. The [Logging Flow](https://docs.python.org/2/howto/logging.html#logging-flow) also makes that clear.
2. Having a NullHandler on a logger is still recommended in [Configuring logging for a library](https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library) and in several blog entries, but for a different reason: It suppresses the warning "No handlers could be found for logger ...". I have verified that that warning is only issued on Python 2, but not on Python 3 (at least not on 3.4+). Yet, the [Configuring logging for a library (py36)](https://docs.python.org/3.6/howto/logging.html#configuring-logging-for-a-library) still recommends adding a NullHandler to a library.

So having or not having a NullHandler is just related to whether or not that warning should be suppressed. I suggest adding a NullHandler on Python 2.

The other thing is, that in our config functions, when we add a handler that actually emits the log records to the "pywbem.<xxx>" loggers, this also does not prevent the log event from propagating up. So a root logger that is configured to output DEBUG level would also output our logs, even when the pywbem handler already outputs them. To prevent that, we could for example set the "propagate" attribute of the logger that has a handler active, to False. If we do that, then we should probably have a log config function that disables the logging. Right now, our log config functions can only be used to enable logging, or to not change logging, but not to disable logging.

andy-maier:
On releasing as experimental: That is true, but it is also true that we are currently redesigning it, so this is already the second attempt of an experimental logging functionality. I don't want to put a third exterimental attempt on, if I can avoid it.

andy-maier:
[...]

PR #1191, opened 2018-04-10
Add more log tests to confirm NullHandler dest=None changes

KSchopmeyer:
Added a couple more tests to test_recorder.py as part of reviewing the changes for issues #1180 and #1186.  Also changed couple of examples where we were writing a log file that was left over after test.


andy-maier:
Approved

PR #1192, opened 2018-04-10
Changes to docs for _logging.py and mock

KSchopmeyer:
In _logging removed example that uses basicconfig and expanded text on some of the
others

Minor changes to mock doc. Changed word module to
subpackage since pywbem_mock is not a module. From other
doc it appears that it can be called either a subfolder
or a subpackage.


andy-maier:
Approved.

PR #1196, opened 2018-04-11
Release 0.12.0

andy-maier:
This is the release PR for pywbem 0.12.0.
Approving it means you agree to release this level as pywbem 0.12.0.


andy-maier:
Addressed both comments in an update to the PR.

KSchopmeyer:
We agreed that we are ready to release 0.12.0 now.

PR #1214, opened 2018-04-12
Release 0.11.1

andy-maier:


andy-maier:
This was an attempt to release the 0.11.1 that we never released.
It fails with various errors on Appveyor and Travis that we fixed meanwhile in the 0.12 release, and in order to actually release it, we would have to identify and roll back these fixes.

It seems to me that this is not worth the effort, so I am closing this PR again.

Should there be a need to release 0.11.1 (e.g. because its fixes are considered crucial and 0.12 cannot be used for some reason), we can still do that at that point.

PR #1215, opened 2018-04-12
fixes #1153 Added support to ValueMapping for translating to binary and iterating

### LINKED ISSUE #1153: New valuemapping class method
While the toValues is the core capability of the ValueMapping class (i.e. the ability to get from mapvalue to value for display, etc.  I had to do the opposite for creating instances to intsert into the mock repository for the wbemserver test. I definined what I wanted in terms of the Value and wanted to create convert that to the map.

I note that there is no method for that.

I propose that we need a tovaluemap that takes a string parameter and returns either the corresponding ValueMap or an error if that string is not in the valuelist.

andy-maier:
Ready for review and merge.

I'd like to point out some possible review points:

* What should be the datatype of the integer values returned by `tobinary()` (e.g. `Uint16` as in the current PR or simply `int`)?
  **COMMENT:** Karl will go back to the test we created. Since we do not lose anything by returning the exact cim type of the element, we agree that this is the best solution.

* What should `tobinary()` return for the Values string of the unclaimed marker? (e.g. `None` as in the current PR, or some representation of all unclaimed integer values)?
  CONCLUSION: Returning the None is much simpler and seems to be the most logical.

NEW QUESTION; What about an enumeration option.   Andy will look at the possibility. DONE.

### DIFF
--- a/pywbem/_valuemapping.py
+++ b/pywbem/_valuemapping.py
@@ -15,10 +15,10 @@
 #
 
 """
-The :class:`~pywbem.ValueMapping` class supports translating the values of an
-integer-typed CIM element (e.g. property, method, or parameter) that is
-qualified with the `ValueMap` and `Values` qualifiers, to the corresponding
-value of the `Values` qualifier.
+The :class:`~pywbem.ValueMapping` class supports translating between the values
+of an integer-typed CIM element (e.g. property, method, or parameter) that is
+qualified with the `ValueMap` and `Values` qualifiers, and the corresponding
+values of the `Values` qualifier, in both directions.
 
 This class supports value ranges (e.g. ``"4..6"``) and the unclaimed marker
 (``".."``).
@@ -26,6 +26,10 @@ This class supports value ranges (e.g. ``"4..6"``) and the unclaimed marker
 
 import re
 import six
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
 
 from .cim_types import CIMInt, type_from_name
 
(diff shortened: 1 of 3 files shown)


andy-maier:
In the new commit, I added in iteration capability and changed the type used for the representation of ranges from list of lo,hi to tuple of lo,hi.

Please review again.

pywbem/_valuemapping.py
```diff
@@ -36,17 +36,18 @@ class ValueMapping(object):
...
     *New in pywbem 0.9 as experimental and finalized in 0.10.*
 
-    A utility class that supports translating the values of an integer-typed
-    CIM element (property, method, parameter) that is qualified with the
-    `ValueMap` and `Values` qualifiers, to the corresponding value of the
-    `Values` qualifier.
+    A utility class that supports translating between the values of an
+    integer-typed CIM element (property, method, parameter) that is qualified
+    with the `ValueMap` and `Values` qualifiers, and the corresponding values
+    of the `Values` qualifier, in both direcitons.
```
KSchopmeyer: spelling directions not direcitions

pywbem/_valuemapping.py
```diff
@@ -72,36 +73,74 @@ class CIM_Foo {
           };
 
       Assuming this class exists in a WBEM server, the following code will
-      create a value mapping for this property and will print a few property
-      values and their corresponding `Values` strings::
+      create a value mapping for this property and will look up the `Values`
```
KSchopmeyer: change will create  to will get the class from the server, create a value mapping.

pywbem/_valuemapping.py
```diff
@@ -72,36 +73,74 @@ class CIM_Foo {
...
+                    10: 'unclaimed'
+                    11: 'unclaimed'
+
+      Translating in the other direction is also of interest, for example when
+      processing values that are provided by humans in terms of the `Values`
+      strings, or when using the pywbem mock support and the test cases specify
+      property values for CIM instances in terms of the more human-readable
+      `Values` strings.
+
+      Again, assuming the class shown above exists in a WBEM server, the
```
KSchopmeyer: Again add  will get from server

pywbem/_valuemapping.py
andy-maier: DONE

pywbem/_valuemapping.py
andy-maier: DONE

pywbem/_valuemapping.py
andy-maier: DONE

andy-maier:
I seem to have screwed up what I pushed up before you did your review. It did not contain the iteration functionality, and not the fix for the typo that you also found.

The latest commit contains what I wanted to add, plus resolutions to your comments.

Please review the iteration functionality.

PR #1217, opened 2018-04-15
fixes #1160 Enhance MOF compiler and missing dependent classes resolution logic

### LINKED ISSUE #1160: MOF compiler support for finding REF/Embedded classes and not creating inconsistent repo without them
The mof_compiler includes support for finding dependent classes that is aimed at finding things like:

1. Classes that are dependent on the EmbeddedInstance qualifier but that the class has not been compiled already

2. Classes that where the defined superclass has not been compiled.

3. Reference properties where the ReferenceClass has not been defined.

This is part of the method p_mp_createClass which upon receiving an exception to a createclass tries to
fix issues with a)superclass not yet compiled, b) class for embeddedInstance not yet compiled, c) reference classes not yet compiled

However, while the code is in place, it depends on the mof_compiler actually creating an exception when the preconditions have not been met.  This works for superclasses but does not appear to work for either the embeddedInstance qualifier or for reference properties. 

We need to add exceptions for the following:

For the reference property need to add test for existence of the compiled reference class.
The following is an example:

def p_referenceDeclaration

```
        try:
            qual = self.classes[self.default_namespace][cname]
        except KeyError:
            if self.conn is None:
                raise CIMError(CIM_ERR_NOT_FOUND, cname)

```
For the EmbeddedInstanceQualifier:

Add a similar test to p_propertyDeclaration functions to test for EmbeddedInstance Qualifier  and then test for existence of the corresponding class

These exceptions would then be caught 

At least that is my best guess right now.

KSchopmeyer:
Review needed based on update with comments fixed.

Adds capability to find missing class dependencies(cases where a class is defined as a requirement in another class declaration to try to find the depend classes in the compile search path and compile them.
This includes finding reference classnames defined in reference properties and parameters (i.e CIM_ManagedElement REF parent) and the classes for embedded instances in the compile  search path
if they are not already in the repository.  This works with any WBEM server repository implementation that returns CIM_ERR_INVALID_PARAMETER or CIM_ERR_FAILED if the class defined in the dependency is not in the repository.

This pr extends MOFWBEMConnection to generate a CIM_ERR_INVALID_PARAMETER exception if
the class for a REF defined in a class being compiled does not exist in the MOFWBEMConnection repository or if the class defined for an EmbeddedInstance qualifier does not exist.

The mof_compiler catches these exceptions and tries to compile in the class by searching the compile search path for the class and compiling it.

Adds tests for cases of these conditions both correct and error (the REF or EmbeddedInstance class exists or does not exist) for the conditons in both properties and method parameters.

Fixes error in one of the test_mofs where the error actually existed (REF for nonexistent class) but was not being caught by the compiler before this pr.

Adds entry to changes.rst

DISCUSSION; This pr today DOES not add the requested test for circular search for missing class. I am still workin on that one

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -27,9 +27,20 @@ Released: not yet
 **Bug fixes:**
 
 * fix issue where wbemcli-help-txt was not being updated when wbemcli.py
-  changed.  Issue # 1205
+  changed.  (Issue #1205)
 
 **Enhancements:**
+
+* Extend MOFWBEMConnection to generate an exception if the compile of a
+  class with reference parameters or properties reference class is not in the
+  repository or if the class defined for an EmbeddedInstance qualifier is
+  not in the repository or. This makes use of the capability with the
+  mof compiler (extended slightly) to search the search path defined for the
+  compile for the corresponding missing classes if they are not in the
+  repository.  This means that the mof_compiler can be used to create complete
+  class repository builds without having to define specifically all dependent
+  classes for the classes defined for the build if those classes are in the
+  search path. (Issue #1160)
 
 **Cleanup**
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,16 +31,25 @@ Released: not yet
 
 **Enhancements:**
 
-* Extend MOFWBEMConnection to generate an exception if the compile of a
+* Extend pywbem MOF compiler to search for dependent classes including:
+  a) reference classes (classes defined in reference properties or parameters)
+  b)EmbeddedInstance qualifier classes if they are not compiled before the
+  classes that reference them is compiled. Previously the lack of these
+  dependent classes was ignored.  The compiler already searches for superclasses
+  if they are not compiled before their subclasses.
+
+  Extends MOFWBEMConnection to generate an exception if the compile of a
   class with reference parameters or properties reference class is not in the
   repository or if the class defined for an EmbeddedInstance qualifier is
-  not in the repository or. This makes use of the capability with the
-  mof compiler (extended slightly) to search the search path defined for the
-  compile for the corresponding missing classes if they are not in the
-  repository.  This means that the mof_compiler can be used to create complete
-  class repository builds without having to define specifically all dependent
-  classes for the classes defined for the build if those classes are in the
-  search path. (Issue #1160)
+  not in the repository.
+
+  This uses the capability in the MOF compiler to search the defined
+  search path for the missing classes if they are not in the repository.
+
+  This means that the mof_compiler can be used to create a complete class
+  repository builds without having to  specifically declare all dependent
+  classes for the classes the user needs in a repository if the mof for the
+  dependent classes in in the search path. (Issue #1160).
 
 **Cleanup**
 
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -599,31 +599,26 @@ def p_mp_createClass(p):
                     objects = list(cc.properties.values())
                     for meth in cc.methods.values():
                         objects += list(meth.parameters.values())
-                    # TODO the test for class in list does not work now that
-                    # we have made dep_classes a tuple with class and ce
-                    dep_classes = []
-                    dep_rslts = []
+
+                    dep_classes = NocaseDict()  # dict dep_class, ce
                     for obj in objects:
                         if obj.type not in ['reference', 'string']:
                             continue
                         if obj.type == 'reference':
-                            obj_ref = obj.reference_class.lower()
-                            if obj_ref not in dep_classes:
-                                dep_classes.append(obj_ref)
-                                dep_rslts.append((obj_ref, ce))
+                            if obj.reference_class not in dep_classes:
+                                dep_classes[obj.reference_class] = ce
                         elif obj.type == 'string':
                             try:
                                 embedded_inst = \
                                     obj.qualifiers['embeddedinstance']
                             except KeyError:
                                 continue
-                            embedded_inst_cln = embedded_inst.value.lower()
-                            if embedded_inst_cln not in dep_classes:
-                                dep_classes.append(embedded_inst_cln)
-                                dep_rslts.append((embedded_inst_cln, ce))
+                            if embedded_inst.value not in dep_classes:
+                                dep_classes[embedded_inst.value] = ce
 
                             continue
-                    for cln, err in dep_rslts:
+
+                    for cln, err in dep_classes.items():
                         if cln in p.parser.classnames[ns]:
                             continue
                         try:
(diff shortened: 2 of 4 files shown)


andy-maier:
On the discussion point about strict mode: I'm against introducing multiple modes. Currently, we leave the repository in an inconsistent state, and that should not be allowed in a non-strict mode.

docs/changes.rst
```diff
@@ -31,6 +31,17 @@ Released: not yet
...
+* Extend MOFWBEMConnection to generate an exception if the compile of a
+  class with reference parameters or properties reference class is not in the
+  repository or if the class defined for an EmbeddedInstance qualifier is
+  not in the repository or. This makes use of the capability with the
+  mof compiler (extended slightly) to search the search path defined for the
+  compile for the corresponding missing classes if they are not in the
+  repository.  This means that the mof_compiler can be used to create complete
+  class repository builds without having to define specifically all dependent
+  classes for the classes defined for the build if those classes are in the
+  search path. See issue # 1160
```
andy-maier: Please remove the space between "#" and the issue number. That allows for better searching.

pywbem/mof_compiler.py
```diff
@@ -599,42 +599,55 @@ def p_mp_createClass(p):
                     objects = list(cc.properties.values())
                     for meth in cc.methods.values():
                         objects += list(meth.parameters.values())
+                    # TODO the test for class in list does not work now that
+                    # we have made dep_classes a tuple with class and ce
```
andy-maier: * In the code, `dep_classes` is still a list, not a tuple as stated in the TODO text.
* There's something missing at the end of the TODO text.

pywbem/mof_compiler.py
```diff
@@ -599,42 +599,55 @@ def p_mp_createClass(p):
...
-                            p.parser.classnames[ns].append(klass)
+                            p.parser.classnames[ns].append(cln)
                         except CIMError:
-                            moffile = p.parser.mofcomp.find_mof(klass)
+                            moffile = p.parser.mofcomp.find_mof(cln)
                             if not moffile:
-                                raise
-                            p.parser.mofcomp.compile_file(moffile, ns)
-                            p.parser.classnames[ns].append(klass)
+                                raise err
```
andy-maier: I did not follow this path, but how does this case (no mof file) leave the loop of raising error / retrieving and compiling the class?

pywbem/mof_compiler.py
```diff
@@ -599,42 +599,55 @@ def p_mp_createClass(p):
...
+                            try:
+                                p.parser.mofcomp.compile_file(moffile, ns)
```
andy-maier: Hmm. I thought this function is just supposed to raise an error, and the compile is done elsewhere?

pywbem/mof_compiler.py
```diff
@@ -2178,6 +2191,41 @@ def CreateClass(self, *args, **kwargs):
...
+            objects += list(meth.parameters.values())
+
+        for obj in objects:
+            # Validate that reference_class exists in repo
+            if obj.type == 'reference':
+                try:
+                    self.GetClass(obj.reference_class, LocalOnly=True,
+                                  IncludeQualifiers=True)
+                except CIMError as ce:
+                    if ce.args[0] == CIM_ERR_NOT_FOUND:
```
andy-maier: Please access the CIM status via the `cim_status` attribute of the `CIMError` object.

pywbem/mof_compiler.py
```diff
@@ -2178,6 +2191,41 @@ def CreateClass(self, *args, **kwargs):
...
+                        ce.args = (CIM_ERR_INVALID_PARAMETER,
+                                   'Reference object "%s" of class "%s" '
+                                   'reference_class "%s" does not exist' %
```
andy-maier: Please improve the exception message to be more readable. How about: "The class referenced by property %r of class %r does not exist: %r".

pywbem/mof_compiler.py
```diff
@@ -2178,6 +2191,41 @@ def CreateClass(self, *args, **kwargs):
...
+                                   (obj.name, cc.classname,
+                                    obj.reference_class))
+                    raise
```
andy-maier: I'm not sure that modifying the existing exception and re-raising it is a good idea. It is more confusing than it helps (the traceback will point to GetClass but the exception text and status code only makes sense for CreateClass). Plus, I seem to recall that `args` was not modifiable in some Python version. But maybe I recall this incorrectly, and its items were not modifiable because it is a tuple.
Anyway, I suggest raising a new exception instead of modifying the existing one. Multiple occurrences of this pattern.

pywbem/mof_compiler.py
```diff
@@ -2178,6 +2191,41 @@ def CreateClass(self, *args, **kwargs):
...
+                    eiqualifier = obj.qualifiers['EmbeddedInstance']
+                    try:
+                        self.GetClass(eiqualifier.value, LocalOnly=True,
+                                      IncludeQualifiers=False)
+                    except CIMError as ce:
+                        if ce.args[0] == CIM_ERR_NOT_FOUND:
+                            ce.args = (CIM_ERR_INVALID_PARAMETER,
+                                       'EmbeddedInstance qualifier classname '
+                                       '"%s"  in object "%s" of class "%s" '
+                                       'does not exist.' %
```
andy-maier: This exception text could be improved, because at a first read, it is not clear whether the qualifier has no class name or the class does not exist. Thinking about it then makes it clear, but it could be better. Also, it might be helpful to mention the namespace.
How about: "Class %r specified by the EmbeddedInstance qualifier on element %r in class %r in namespace %r does not exist."

docs/changes.rst
KSchopmeyer: Fixed and pushed this fix.  Found another place with same issue

pywbem/mof_compiler.py
KSchopmeyer: OK Will do

pywbem/mof_compiler.py
KSchopmeyer: OK

pywbem/mof_compiler.py
KSchopmeyer: Why use %r. that may spray out extra data when all we want is the name. Otherwise no issue with message.

pywbem/mof_compiler.py
KSchopmeyer: Create issue to fix the ce.args.

pywbem/mof_compiler.py
KSchopmeyer: The %r puts quotes around the message and the string form does not do that.

KSchopmeyer: We need to create a new issue to assure that the above is correct and that we fix it.

KSchopmeyer: DONE

pywbem/mof_compiler.py
KSchopmeyer: DONE

pywbem/mof_compiler.py
KSchopmeyer: DONE. Revamped message and change %s to %r for template string

PR #1219, opened 2018-04-16
[WIP] Fixes #426: Migrated test_client yaml execution to pytest

### LINKED ISSUE #426: The  test_client is an all or nothing test and cannot be broken out by component
Right now the test_client is all or nothing.  It does not allow testing against individual yaml tests or files.

It would be good if a user could execute single tests.  This could be done by making this a dynamic unit test so that the testcases were dynamically created from the yaml file list or maybe other techniques.

Also, Andy noted that someone else in IBM noted that there are other solutions than yaml for achieving mock unit tests as we have done in test_client.  (KS) In retrospect this is actually logical I think.   The real question is whether using yaml for the test definition gains us anything that using a python destdefintion class or dictionaies would not against the fact that we would not have to  map from yaml to python.  I think this is more an issue of convience and simplicity than a major direction change.

I unilaterally set it to v 0.10.0

andy-maier:
TODO:
* Re-enable xml_*embed() function tests


KSchopmeyer:
I really like this, much cleaner. However will take some time to review in detail.

andy-maier:
Rebased.

testsuite/testclient/conftest.py
```diff
@@ -0,0 +1,900 @@
+"""
+Module for running the testclient .yaml test cases.
```
KSchopmeyer: Can we document how this all comes together between this and the yaml files.  We know what is happening but not sure anybody else would without a lot of work.  I.e this is diriven by the yaml files in dir testclient, etc.

QUESTION: Where did the name conftest in conftest.py come from???

testsuite/testclient/conftest.py
```diff
@@ -0,0 +1,900 @@
...
+
+    If the input value is not a valid type or list of valid types an
+    error is raised.
+
+    If the pywbem_object defines a valid type (ex. CIMInstance) or
+    list of items where the pywbem_object entry defines a valid type,
+    the type is evaluated with the remaining entries in the
+    dictionary and returned.
+
+    Examples for a testcase value (JSON object / Python dict):
```
KSchopmeyer: Suggest that this be at the module doc level since it is really the definition of the yaml

testsuite/testclient/conftest.py
KSchopmeyer: I have struggled more that normal trying to sort out how this all ties togther but part of that is that it is significantly different from the original code.

testsuite/testclient/conftest.py
```diff
@@ -0,0 +1,900 @@
...
+            import sys
+            self.exc = sys.exc_info()
+
+    def join(self, *args, **kwargs):
+        threading.Thread.join(self, *args, **kwargs)
+        if self.exc:
+            six.reraise(*self.exc)
+
+
+def patched_makefile(self, mode='r', bufsize=-1):
```
KSchopmeyer: Can we rename.  Took quite a while to sort out what makefile meant since the work Makefile has a very pecular usage.

testsuite/testclient/conftest.py
```diff
@@ -0,0 +1,900 @@
...
+
+    return self.fd
+
+
+# Monkey-patching httpretty to pass exception raised in callbacks
+# back to the caller.
+fakesock.socket.makefile = patched_makefile
+
+
+def pytest_collect_file(parent, path):
```
KSchopmeyer: That one took a while to sort out until I finally found it in the pytest documentation. Think it needs more explanation than "pytest hook"

KSchopmeyer: Since I assume you derived  this from 
https://docs.pytest.org/en/latest/example/nonpython.html#a-basic-example-for-specifying-tests-in-yaml-files

Lets point back to it.

testsuite/testclient/conftest.py
andy-maier: Will add an explanation. conftest.py is a hard coded name that pytest recognizes.

andy-maier: DONE. Documented the test approach, and the testcase YAML syntax.

testsuite/testclient/conftest.py
andy-maier: Added YAML syntax at module level, but this part is for representation of Python values and therefore only a small subset, so I left it here and the new module-level description references it.
DONE.

testsuite/testclient/conftest.py
andy-maier: I added a description on what it means, ands hope we can live with that. The original function is called makefile() and I think we should keep the connection, naming-wise.
DONE.

andy-maier:
The last commit addresses all comments. Since a lot of description text was added, please review again.

testsuite/testclient/conftest.py
KSchopmeyer: I am happy.  That one threw me until I dug far enought into pytest doc

testsuite/testclient/test_conftest.py
```diff
@@ -0,0 +1,60 @@
+"""
+Testcases for conftest.py
```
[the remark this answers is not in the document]
andy-maier: Made the description more verbose.

andy-maier:
With the one additional comment resolution, and Karl's statement that it is GOOD other than that, I consider this one approved.

PR #1233, opened 2018-04-24
Excluded .DS_Store files from git.

andy-maier:
On OS-X, the file finder creates hidden metafiles named .DS_Store for the
directories it opens. This change excludes them from the git repo.


andy-maier:
Rolled back into 0.12.6 via PR #1374.

Issue #1201, opened 2018-04-11
Migrate unit tests from unittest to py.test


andy-maier:
Test scripts that have partially moved to pytest:

* test_cim_obj.py
* test_nocasedict.py
* test_logging.py
* test_tupletree.py

Test scripts that have not moved to pytest at all yet:

* test_cim_xml.py
* test_client.py (high benefit)
* test_recorder.py
* test_mof_compiler.py
* test_cim_http.py
* test_indicationlistener.py
* test_statistics.py
* test_wbemcli.py

KSchopmeyer:
There is a log capture for pytest-logcapture. It might even be a bit more flexible in that it does not require complete compare.

andy-maier:
Issue #1645 has been created for the remaining migrations.

PR #1238, opened 2018-04-28
Issue #1201 Added improved test_function_new decorator

andy-maier:
Ready for review and merge.

### DIFF
--- a/testsuite/pytest_extensions.py
+++ b/testsuite/pytest_extensions.py
@@ -4,16 +4,37 @@ Extensions for pytest module.
 
 from __future__ import absolute_import
 
+import functools
+from collections import namedtuple
+import six
 import pytest
-from decorator import decorator
+if six.PY3:
+    # pylint: disable=no-name-in-module
+    from inspect import Signature, Parameter
+else:
+    from funcsigs import Signature, Parameter
+
+__all__ = ['simplified_test_function']
 
 
-__all__ = ['test_function']
+# Pytest determines the signature of the test function by unpacking any
+# wrapped functions (this is the default of the signature() function it
+# uses. We correct this behavior by setting the __signature__ attribute of
+# the wrapper function to its correct signature. To do that, we cannot use
+# signature() because its follow_wrapped parameter was introduced only in
+# Python 3.5. Instead, we build the signature manually.
+TESTFUNC_SIGNATURE = Signature(
+    parameters=[
+        Parameter('desc', Parameter.POSITIONAL_OR_KEYWORD),
+        Parameter('kwargs', Parameter.POSITIONAL_OR_KEYWORD),
+        Parameter('exp_exc_types', Parameter.POSITIONAL_OR_KEYWORD),
+        Parameter('exp_warn_types', Parameter.POSITIONAL_OR_KEYWORD),
+        Parameter('condition', Parameter.POSITIONAL_OR_KEYWORD),
+    ]
+)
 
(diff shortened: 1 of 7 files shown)


andy-maier:
The discussion of pytest issue pytest-dev/pytest#3435 revealed how this can be done. Implemented the suggestion.

testsuite/pytest_extensions.py
```diff
@@ -137,3 +161,146 @@ def test_CIMClass_equal(
 
             ret = None  # Debugging hint
     return ret
+
+
+def test_function_new(test_func):
```
KSchopmeyer: Since this is probably what will be the major test function decorator in the future, shouldn't we call it something else that _new.   Expanded, complete, 2. Not sure of name but new is just transitional as a name.

testsuite/pytest_extensions.py
```diff
@@ -137,3 +161,146 @@ def test_CIMClass_equal(
...
+
+    Parameters of the test function that is decorated:
+
+    * testcase (testcase_tuple): The testcase, as a named tuple.
+
+    * **kwargs: Keyword arguments for the test function.
+
+    Notes:
+
+    * Using this decorator together with the `pytest.mark.parametrize`
```
KSchopmeyer: How about when applying this decorator with param... this decorator must be inserted before the param... decorator

testsuite/test_nocasedict.py
```diff
@@ -374,23 +369,19 @@ def test_NocaseDict_init(
...
-    key = kwargs['key']
-    exp_value = kwargs['exp_value']
-
     # The code to be tested
     act_value = obj[key]
 
     # Verify that an exception raised in this function is not mistaken
     # to be the expected exception
-    assert exp_exc_types is None
+    assert testcase.exp_exc_types is None
```
KSchopmeyer: comment on line above.  'mistaken as the expected exception.

testsuite/test_nocasedict.py
```diff
@@ -517,23 +508,18 @@ def test_NocaseDict_getitem(
...
-    key = kwargs['key']
-    value = kwargs['value']
-
     # The code to be tested
     obj[key] = value
 
     # Verify that an exception raised in this function is not mistaken
     # to be the expected exception
-    assert exp_exc_types is None
+    assert testcase.exp_exc_types is None
```
KSchopmeyer: see above

testsuite/pytest_extensions.py
KSchopmeyer: Should be inserted after the param, not before.

testsuite/pytest_extensions.py
andy-maier: We agreed to remove the old function and to name zhe new function `simplified_test_function()`.

andy-maier: DONE

testsuite/pytest_extensions.py
andy-maier: DONE.

testsuite/test_nocasedict.py
andy-maier: DONE, plus changed verify to ensure

testsuite/test_nocasedict.py
andy-maier: DONE

andy-maier:
The latest commit addresses all comments

PR #1242, opened 2018-05-03
Fixed Travis issue on py26 with format of empty field error.

andy-maier:
Ready for review and merge.


andy-maier:
I added an entry in the change log, as proposed by Karl.

PR #1243, opened 2018-05-03
[WIP] Issue #1201 Migrated test_cim_obj from unittest to pytest

andy-maier:
TODO:
- Migrate the remaining unittest based functions
- Make the pytest based classes/functions consistent (see test_nocasedict.py)


testsuite/test_cim_obj.py
```diff
@@ -358,546 +350,627 @@ def runtest_dict(self, obj, exp_dict):
...
+        None, None, True
+    ),
 
-        # Keybinding tests
-        (
-            "Verify that keybinding with name None succeeds (since 0.12",
-            dict(
+    # Keybinding tests
+    (
+        "Verify that keybinding with name None succeeds (since 0.12",
```
KSchopmeyer: trailing )

testsuite/test_cim_obj.py
```diff
@@ -358,546 +350,627 @@ def runtest_dict(self, obj, exp_dict):
...
+        None, None, True
+    ),
 
-        # Namespace tests
-        (
-            "Verify that binary namespace is converted to unicode",
-            dict(
+    # Namespace tests
+    (
+        "Verify that binary namespace is converted to unicode",
```
KSchopmeyer: Isn't that byte namespace and not binary namespace?

testsuite/test_cim_obj.py
```diff
@@ -358,546 +350,627 @@ def runtest_dict(self, obj, exp_dict):
...
+        None, None, True
+    ),
 
-        # Host tests
-        (
-            "Verify that binary host is converted to unicode",
-            dict(
+    # Host tests
+    (
+        "Verify that binary host is converted to unicode",
```
KSchopmeyer: isn't that byte and not binary???

testsuite/test_cim_obj.py
```diff
@@ -2170,541 +2246,650 @@ def test_CIMInstanceName_tocimxmlstr(testcase,
...
-            None, CHECK_0_12_0
         ),
-        (
-            "namespace type upper case HTTP",
-            'HTTP://10.11.12.13:5989/root/cimv2:CIM_Foo.k1="v1"',
-            dict(
+        None, None, CHECK_0_12_0
+    ),
+    (
+        "namespace type upper case HTTP",
```
KSchopmeyer: Not sure of nomenclature but should that be scheme  rather than namespace type?  Same below

testsuite/test_cim_obj.py
andy-maier: done

testsuite/test_cim_obj.py
andy-maier: done - changed binary to bytes in all occurrences in this file, except when it was used for numbers in binary (= base 2) representations.

testsuite/test_cim_obj.py
andy-maier: see above

testsuite/test_cim_obj.py
andy-maier: DSP0207 uses the term "namespace type" so I used that too.

PR #1254, opened 2018-05-16
Fixes #1253 Fix issue with deepcopy of objects in _wbemconnection_mock

### LINKED ISSUE #1253: Because of #1251, pywbem_mock does not create complete copies
Pywbme_mock uses the cim_object copy methods to create copies of classes and instances in the repository for methods like getclass, getinstance.   Since these are not deepcopy methods, it actually returns references to things like the qualifiers.

These copies need to be changed to use deepcopy.

KSchopmeyer:
See commit for documentation.

This pr cherrypicks pr #1260 where we had a CIMError defined wrong in the Fake_InvokeMethod and with the deepcopy changes we hit that issue which caused the WBEMServer tests to fail.

NOTE: I think we want to put this in 0.12.3 also.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -66,6 +66,16 @@ Released: not yet
   (CIM_ERR_NOT_FOUND rather than CIM_ERR_METHOD_NOT_FOUND) when the
   class for a method is not defined in the methods repository. issue #1256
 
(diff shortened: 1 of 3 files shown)


andy-maier:
Rebased it and eliminated the duplicate commit for #1260.

PR #1255, opened 2018-05-16
Fixes #1251 Clarified and improved copy() methods of NocaseDict and CIM objects

### LINKED ISSUE #1251: cim_obj copy methods do not create completely new objects
I had problem in the following code in the mock:

cc = [class in repo].copy()  # trying to create copy of class in repo to return for GetClass

When I try to remove the qualifiers from the properties in copied class (cc) It actually removes them from the properties in the class in the repo.

The following code change fixed the issue. Changed the copy of the properties, methods, and
qualifiers NocaseDict to do a deepcopy.  I am not saying that this is the correct solution but it does fix the issue where what was in the copied class was references to the original <classinrepo> qualifiers

    def copy(self):
        """
        Return a copy of the :class:`~pywbem.CIMClass` object.
        """
        result = CIMClass(self.classname)
        result.properties = deepcopy(self.properties)
        result.methods = deepcopy(self.methods)
        result.superclass = self.superclass
        result.qualifiers = deepcopy(self.qualifiers)
        result.path = None if self.path is None else self.path.copy()

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,20 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Changed the `path` argument of `CIMInstance` to be deep copied, because it 
+  may be modified by setting properties. It was previously shallow copied
+  (and incorrectly documented as not being copied). This is only incompatible
+  if user code relies on the init method modifying the keybindings of its
+  `path` input argument. If user code relies on that, it is highly recommended
+  that you decouple such dependencies (Issue #1251).
+
+* Changed the `path` argument of `CIMClass` to be shallow copied, in order
+  to decouple the created object from its input arguments. It was previously
+  not copied but referenced. This is only incompatible if user code relies on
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -22,6 +22,20 @@ Released: not yet
...
+* Changed the `path` argument of `CIMInstance` to be deep copied, because it 
+  may be modified by setting properties. It was previously shallow copied
+  (and incorrectly documented as not being copied). This is only incompatible
+  if your code relies on the init method modifying the keybindings of its
+  `path` input argument. If your code relies on that, it is highly recommended
+  that you decouple such dependencies (Issue #1251).
+
+* Changed the `path` argument of `CIMClass` to be shallow copied, in order
+  to decouple the created object from its input arguments. It was previously
+  not copied but referenced. This is only incompatible if your code relies on
```
KSchopmeyer: change your code to user code

PR #1260, opened 2018-05-17
fixes #1256 Correct issue in pywbem_mock with incorrect CIMError return

### LINKED ISSUE #1256: mock invoke method returns incorrect CIMError
In one case, CIMError returns CIM_ERR_NOT_FOUND when it should be returning CIM_ERR_METHOD_NOT_FOUND

KSchopmeyer:
Fixes issue in Fake_invokemethod where we are returning
CIM_ERR_NOT_FOUND rather than CIM_ERR_METHOD_NOT_FOUND.  Also
reorganizes order of the tests to be more logical and adds one more test
to validate that the
class_origin is correct.  The problem occurs in WBEMServer where the
method that gets the central instances expects CIM_ERR_METHOD_NOT_FOUND.

This should also be in release 0.12.3

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -3043,30 +3043,33 @@ class FakedWBEMConnection(WBEMConnection):
                              local_only=False, include_qualifiers=True,
                              include_classorigin=True)
 
+        # Determine if method defined in classname defined in
+        # the classorigin of the method
         try:
-            target_class = cc.methods[methodname].class_origin
+            target_cln = cc.methods[methodname].class_origin
         except KeyError:
             raise CIMError(CIM_ERR_METHOD_NOT_FOUND, 'Method %s not found '
-                           'in %s class hiearchy' % (methodname,
-                                                     localobject.classname))
+                           'in class %s class' % (methodname,
+                                                  localobject.classname))
+        if target_cln != cc.classname:
+            # TODO FUTURE: add method to repo that allows privileged users
+            # direct access so we don't have to go through _get_class and can
+            # test classes directly in repo
+            tcc = self._get_class(target_cln, namespace,
+                                  local_only=False, include_qualifiers=True,
+                                  include_classorigin=True)
+            if methodname not in tcc.methods:
+                raise CIMError(CIM_ERR_METHOD_NOT_FOUND, 'Method %s not found '
+                               'in class %s' % (methodname, target_cln))
 
-        # test for target class in methods repo
+        # Test for target class in methods repo
         try:
-            methods = methodsrepo[target_class]
+            methods = methodsrepo[target_cln]
         except KeyError:
-            raise CIMError(CIM_ERR_NOT_FOUND, 'Class %s for Method %s in '
-                                              'namespace %s not '
-                                              'registered in methods '
-                                              'repository' %
+            raise CIMError(CIM_ERR_METHOD_NOT_FOUND,
+                           'Class %s for method %s in namespace %s not '
+                           'registered in methods repository' %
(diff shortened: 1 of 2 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -3043,31 +3043,34 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
...
+        if target_cln != cc.classname:
+            # TODO FUTURE: add method to repo that allows privileged users
+            # direct access so we don't have to go through _get_class and can
+            # test classes directly in repo
+            tcc = self._get_class(target_cln, namespace,
+                                  local_only=False, include_qualifiers=True,
+                                  include_classorigin=True)
+            if methodname not in tcc.methods:
+                raise CIMError(CIM_ERR_METHOD_NOT_FOUND, 'Method %s not found '
+                               'in class %s' % (methodname, target_cln))
```
andy-maier: I think it would be useful for understanding this situation that the exception message mentions that this is in the origin class of the method.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -3043,31 +3043,34 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
...
-            target_class = cc.methods[methodname].class_origin
+            target_cln = cc.methods[methodname].class_origin
         except KeyError:
             raise CIMError(CIM_ERR_METHOD_NOT_FOUND, 'Method %s not found '
-                           'in %s class hiearchy' % (methodname,
-                                                     localobject.classname))
-
-        # test for target class in methods repo
+                           'in class %s class' % (methodname,
+                                                  localobject.classname))
```
andy-maier: In "class %s class" there is one "class" too much.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -3043,31 +3043,34 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
...
-            methods = methodsrepo[target_class]
+            methods = methodsrepo[target_cln]
         except KeyError:
-            raise CIMError(CIM_ERR_NOT_FOUND, 'Class %s for Method %s in '
-                                              'namespace %s not '
-                                              'registered in methods '
-                                              'repository' %
+            raise CIMError(CIM_ERR_METHOD_NOT_FOUND,
+                           'Class %s for method %s in namespace %s not '
+                           'registered in methods repository' %
```
andy-maier: Can this ever happen? I think it can only happen when the implementation is incorrect. If that is the case, this should be an assertion error, not a CIMError.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: I agree and changed the messages that involve the origin class.

DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Actually I want to leave it in for now. In particular because this is the one I am hitting in the wbem_server test .  I am not sure why but originally it failed at the method level but once we did the the change to Deepcopy the error moved over to the the class not being there. To be honest, I did not look at why so far.  In any case I would like to get this in as it is and if necessary write another issue to clean up those few lines of code

KSchopmeyer:
I made fixes for two of your proposals and am asking that we not make the third one now since that is the fix we wanted to get into this pr.

pywbem_mock/_wbemconnection_mock.py
andy-maier: ok

PR #1269, opened 2018-05-25
Fixes #1268 Add input parameter to wbemcli to allow mocking a server

### LINKED ISSUE #1268: Add mock capability to wbemcli
A new option on wbemcli that allows it to execute the mock wbemconnection with a defined input file.

KSchopmeyer:
This adds a parameter --mock_server to the command line parameters of
wbemcli to allow wbemcli to be used with the pywbem_mock environment.

This also adds one new constructor parameter to FakedWBEMConnection,
timeout that the WBEMConnection can use to return the timeout exception
if this value is set and is of a lower value than the response_delay
time which would indicate that the server delayed longer than the client
timeout setting

It also adds a test of the wbemcli new parameter to test_wbemcli.py and
a simple mof file to be used in that test.

Adds initial tests. Adds tests for the new parameters which means two new files in testsuite. This DOES NOT convert test_wbemcli.py to pytest.

Adds entry under enhancements to changes.rst

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -1429,7 +1429,7 @@ class FakedWBEMConnection(WBEMConnection):
         Remove properties from an instance or class that aren't in the
         plist parameter
 
-        obj(:class:`~pywbem.CIMClass` or :class:`~pywbem.CIMClass):
+        obj(:class:`~pywbem.CIMClass` or :class:`~pywbem.CIMInstance):
             The class or instance from which properties are to be filtered
 
         property_list(list of :term:`string`):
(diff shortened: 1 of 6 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -1421,7 +1425,7 @@ def _filter_properties(obj, property_list):
         Remove properties from an instance or class that aren't in the
         plist parameter
 
-        obj(:class:`~pywbem.CIMClassName` or :class:`~pywbem.CIMClassName):
+        obj(:class:`~pywbem.CIMClass` or :class:`~pywbem.CIMClass):
             The class or instance from which properties are to be filtered
```
andy-maier: I assume this should have been CIMClass or CIMInstance?

KSchopmeyer: Yes. FIXED.

KSchopmeyer:
The bigger question might be a better way to incorporate python scripts than exec. That one does bother me and I am using it both here and in pywemcli.  It works but there must be a better way to get to plugins, etc.

andy-maier:
Merging this PR because the question of executing scripts is not part of its changes. If you want to follow up on that, open an issue and try to phrase the issues with using exec().

PR #1272, opened 2018-05-29
fixes #1271 - Issue with mock EnumerateClassNames and EnumerateClasses

### LINKED ISSUE #1271: pywbem_mock misnamed parameter request and missing test
Found two issues with pywbem_mock:

1. EnumerateClasses returns incorrect response when ClassName parameter is use.  It was named
   classname in the code and so was interpreted as None. This means we were returning extra classes
   from requests that had this parameter

2. We are missing a test for the EnumerateClasses and EnumerateClassNames methods to return an error if the ClassName is not in the namespace.  It was just returning an empty list

KSchopmeyer:
This fixes the issues documented in issue #1271 where there was a param
acquisition name that lower case instead of mixed case (Should have been
ClassName) and a test for invalid classname on EnumerateClassNames and
EnumerateClasses was mising.

This also removes two print statements that should not have been in the
code (printed out the error for the two methods that are not
implemented)
and fixed one pylint issue.

We added tests to cover the new code for ClassName not in repo.

Added note to changes.rst

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -1175,7 +1175,11 @@ class FakedWBEMConnection(WBEMConnection):
             list of strings with the names of all subclasses of `classname`.
 
         """
-        assert classname is None or isinstance(classname, six.string_types)
+        assert classname is None or isinstance(classname, (six.string_types,
+                                                           CIMClassName))
+
+        if isinstance(classname, CIMClassName):
+            classname = classname.classname
 
         # retrieve first level of subclasses for which classname is superclass
         rtn_classnames = [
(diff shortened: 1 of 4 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -1488,9 +1508,19 @@ def _fake_enumerateclassnames(self, namespace, **params):
 
             CIMError: CIM_ERR_INVALID_NAMESPACE if invalid namespace,
             CIMError: CIM_ERR_NOT_FOUND if Classname not found
+            CIMError: CIM_ERR_INVALID_CLASS if class that is basis for tes
+                      does not exist
```
andy-maier: See previous comments.

KSchopmeyer: See above.  I think this is fixed unless we want to put try block around the GetClass loop and output an exception about the repository integrity in place of the CIM_ERR_NOT_FOUND. I did not want to ignore that error if it ever occured but it could only occur if one function found the class and right after another did not find the same class in the repo.  

After all, this is a test tool.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -1455,13 +1459,29 @@ def _fake_enumerateclasses(self, namespace, **params):
...
         self._get_class_repo(namespace)
 
-        cns = self._get_subclass_names(
-            params.get('classname', None),
-            namespace,
-            params['DeepInheritance'])
+        classname = params.get('ClassName', None)
+        if classname:
+            assert(isinstance(classname, CIMClassName))
+            if not self._class_exists(classname.classname, namespace):
```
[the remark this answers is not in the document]
KSchopmeyer: The added test is correct and the expected error is INVALID_CLASS. However, we go back and to _get_class to get all the classes for the EnumerateClasses and if that methodreturns an error it is CIM_ERR_NOT_FOUND.  
To be completely correct, this would have to be some corruption in the repository since the get_subclass_names just found all those classnames in the same repository but it could still occur but since it is almost impossible  I did not protect against it with a try block around the get_class loop.  That is where the original CIM_ERR_NOT_FOUND exceptions comment came from.

PR #1273, opened 2018-05-31
Changes to messages in _wbemconnection_mock to use %r in place of

KSchopmeyer:
change all the raises CIMError in _wbemconnection_mock to use %r in place of %s for all classnames, property names, namespace names, etc.

This was per our discussion that this eliminates issues of possible exceptions if there are illegal unicode characters in the text of things like classnames, namespaces, etc.

DISCUSSION: Does the format command protect __str__ output so that in place of using the __repl__ we could just use format and still output the __str__ definitions??

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -1,5 +1,5 @@
 #
-# (C) Copyright 2018 InovaDevelopment.com
+# (C) Copyright 2018 InovaDevelopment.comn
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -2315,7 +2315,7 @@ class FakedWBEMConnection(WBEMConnection):
         the querylanguage and query defined
         """
         self._get_instance_repo(namespace)
-        raise CIMError(CIM_ERR_NOT_SUPPORTED, 'ExecQuery Not Implemented!')
+        raise CIMError(CIM_ERR_NOT_SUPPORTED, 'ExecQuery not implemented!')
 
     #####################################################################
     #
@@ -2819,7 +2819,7 @@ class FakedWBEMConnection(WBEMConnection):
         except KeyError:
             raise CIMError(CIM_ERR_INVALID_ENUMERATION_CONTEXT,
                            'EnumerationContext %r not found in mock server '
-                           'Enumeration contexts.' % context_id)
+                           'enumeration contexts.' % context_id)
 
         if context_data['pull_type'] != req_type:
             raise CIMError(CIM_ERR_INVALID_ENUMERATION_CONTEXT,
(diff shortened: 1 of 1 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -2816,13 +2818,13 @@ def _pull_response(self, namespace, req_type, **params):
             context_data = self.enumeration_contexts[context_id]
         except KeyError:
             raise CIMError(CIM_ERR_INVALID_ENUMERATION_CONTEXT,
-                           'EnumerationContext %s not found in mock server '
-                           'contexts.' % context_id)
+                           'EnumerationContext %r not found in mock server '
+                           'Enumeration contexts.' % context_id)
```
andy-maier: Use lower case for "enumeration contexts".

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -3029,11 +3031,11 @@ def _fake_closeenumeration(self, namespace, **params):
             # with the context.
             if context_data['namespace'] != namespace:
                 raise CIMError(CIM_ERR_INVALID_NAMESPACE,
-                               'Incorrect Namespace %s for CloseEnumeration %s '
-                               % (namespace, context_id))
+                               'Incorrect Namespace %r for CloseEnumeration '
```
andy-maier: Jse lower case "namespace"

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -3116,32 +3118,40 @@ def _fake_invokemethod(self, methodname, objectname, Params, **params):
...
+                               'objectname class %r' % (methodname, target_cln,
                                                         localobject.classname))
 
         # Test for target class in methods repo
         try:
             methods = methodsrepo[target_cln]
         except KeyError:
             raise CIMError(CIM_ERR_METHOD_NOT_FOUND,
-                           'Class %s for method %s in namespace %s not '
+                           'Class %r for Method %r in namespace %r not '
```
andy-maier: Use lower case "method"

andy-maier:
On the discussion point: Needs to be tried out on Py 2 and 3, e.g.
```
    cn = 'CIM_\u2018'
    s1 = "Class {}". format(cn)
    s2 = ÷"Class {!}". format(cn)
```

pywbem_mock/_wbemconnection_mock.py

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Done. Also change 'EnumerationContexts' to 'enumeraton contexts'

pywbem_mock/_wbemconnection_mock.py

KSchopmeyer:
Comments fixed.  Your test showed that in this case neither the format or the % generated exception for the define unicode.  However, python 2 does not understand u00c4 and python 3 does.  python 2 replaces output with  unicode sequence

    print("%r" % cn)
    'CIM_\\u00c4nderung'

However, lets not play with that right now.
However, I

andy-maier:
Reason that it did not work on py2 was probably that I forgot the u in front of the literals:

```
cn = u'...'
```

andy-maier:
Also, just in case U+00c4 is supported by the codepage that is used, the test should be repeated with a code point above 255. (but without trying to print it)

andy-maier:
Merging this one, leaving the repl vs str question for the future.

PR #1282, opened 2018-06-09
Fixes #1281 Improved exception messages for ValueMapping and WBEMServer

### LINKED ISSUE #1281: ValueMapping.toValue fails with bad CIMRegisteredProfileInstance
We found a place where an issue in a registered profile instance with the a server in the SMI test center causes a failure in the:

      get_selected_profiles(...) method and actually in the ValueMapping.tovalue.

The issue is that someone (CISCO in production code) generated a profile instance with the RegisteredOrganization property as a string value instead of a Uint16:

The profile instance has the following error:

(u'InstanceId', CIMProperty(name=u'InstanceId', value=u'CISCO:CISCO_RegisteredProfile.11.Profile Registration.1.0.0', type=u'string', reference_class=None, embedded_object=None, is_array=False, array_size=None, class_origin=None, propagated=False, qualifiers=NocaseDict([]))),
(u'RegisteredOrganization', CIMProperty(name=u'RegisteredOrganization', value=u'11', type=u'string', reference_class=None, embedded_object=None, is_array=False, array_size=None, class_origin=None, propagated=False, qualifiers=NocaseDict([]))),

In the RegisteredOrganization, the value is  string with numeric in it whereas it should be something like the following which is a Uint16

(u'InstanceID', CIMProperty(name=u'InstanceID', value=u'CISCO:SNIA:Profile Registration Profile:1.0.0:1.0.0', type=u'string', reference_class=None, embedded_object=None, is_array=False, array_size=None, class_origin=None, propagated=False, qualifiers=NocaseDict([]))),
(u'RegisteredOrganization', CIMProperty(name=u'RegisteredOrganization', value=Uint16(cimtype='uint16', minvalue=0, maxvalue=65535, 11), type=u'uint16', reference_class=None, embedded_object=None, is_array=False, array_size=None, class_origin=None, propagated=False, qualifiers=NocaseDict([]))),

The issue is not that it fails but that it fails at a very low level and gives the user little information so that
the user must dig through the Wbemserver class, the get_select_profiles, the 
statement
            inst_org = org_vm.tovalues(inst['RegisteredOrganization'])

to finally get down to the toValues where the TypeError is generated simply noting that the element_value is the incorrect type.

After putting in some checks, a traceback of the issue generates the following:

     File "/home/kschopmeyer/pywbem/smipyping/smicli/_cmd_explorer.py", line 257, in smi_version
    registered_org='SNIA', registered_name='SMI-S')
    File "/home/kschopmeyer/.virtualenvs/smipyping/local/lib/python2.7/site-packages/pywbem /_server.py", line 349, in get_selected_profiles
    inst_org = org_vm.tovalues(inst['RegisteredOrganization'])
    File "/home/kschopmeyer/.virtualenvs/smipyping/local/lib/python2.7/site-packages/pywbem  /_valuemapping.py", line 559, in tovalues
      "value %r" % (type(element_value), element_value))
  TypeError: Element value is not an integer type: <type 'unicode'>; value u'11'

NOTE: I already modified the exception in ValueMapping.toValues to add a second parameter on the exception (the element_value itself)

The question is whether we should:

1. Extend the message on the type error to at least reflect the actual element_value as I did.
2. Put try block around the code in WBEMServer that executes the tovalues to catch the exception and give better information.

           about line 350
            inst_org = org_vm.tovalues(inst['RegisteredOrganization'])
            inst_name = inst['RegisteredName']
            inst_version = inst['RegisteredVersion']

To catch these errors if the type of the property is incorrect.

Actually, the question may be more general,  In our code where we are actually accessing values from instances in a WBEM server, should we be doing an assert at a high level before trying ot use the data not to correct anything but to make the issue really visible.  ie. Error found in instanc ... from server... namespace ... property... invalid type. or something like that

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -152,7 +152,8 @@ class WBEMServer(object):
         """
         if not isinstance(conn, WBEMConnection):
             raise TypeError("conn argument of WBEMServer must be a "
-                            "WBEMConnection object")
+                            "WBEMConnection object, but has type: %s" %
+                            type(conn))
         self._conn = conn
         self._interop_ns = None
         self._namespaces = None
@@ -345,9 +346,25 @@ class WBEMServer(object):
                                            'RegisteredOrganization')
         rtn = []
         for inst in self.profiles:
-            inst_org = org_vm.tovalues(inst['RegisteredOrganization'])
-            inst_name = inst['RegisteredName']
-            inst_version = inst['RegisteredVersion']
+            try:
+                inst_org_value = inst['RegisteredOrganization']
+            except KeyError:
+                raise KeyError("CIM_RegisteredProfile instance in %r does not "
+                               "have a property 'RegisteredOrganization'" %
+                               self.interop_ns)
+            inst_org = org_vm.tovalues(inst_org_value)
+            try:
+                inst_name = inst['RegisteredName']
+            except KeyError:
+                raise KeyError("CIM_RegisteredProfile instance in %r does not "
+                               "have a property 'RegisteredName'" %
+                               self.interop_ns)
+            try:
+                inst_version = inst['RegisteredVersion']
+            except KeyError:
+                raise KeyError("CIM_RegisteredProfile instance in %r does not "
+                               "have a property 'RegisteredVersion'" %
+                               self.interop_ns)
 
             # pylint: disable=too-many-boolean-expressions
             if (registered_org is None or registered_org == inst_org) and \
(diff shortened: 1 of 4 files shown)


pywbem/_server.py
```diff
@@ -481,17 +501,23 @@ def get_central_instances(self, profile_path, central_class=None,
...
             ResultRole="Dependent")
         if not referencing_profile_paths:
-            raise ValueError("No referencing profile found")
+            raise ValueError("When attempting the scoping class methodology, "
+                             "no referencing profiles were found, for "
+                             "profile instance: %s" % profile_path)
         elif len(referencing_profile_paths) > 1:
-            raise ValueError("More than one referencing profile found")
+            raise ValueError("When attempting the scoping class methodology, "
+                             "more than one referencing profiles were found, "
```
KSchopmeyer: How about changing 'more than one' to 'multiple'

PR #1287, opened 2018-06-10
Added connection information to exceptions

andy-maier:
Ready for review and merge.
What is needed for this PR, is a test on a life server.

The first three commits implement having the connection information in the pywbem exceptions, plus some preparatory changes. Each of these three commits is separate logical step and is best reviewed separately. They were part of the first review.

Compared to the first review, I had to change the connection information that is added to the pywbem exceptions, from adding the entire WBEMConnection object, to adding just the connection ID string. Reason was that in cim_http.py, the connection ID was available but not the connection object, and I did not want to change that there. This change in the exception classes (to take the connection ID string instead of the WBEMConnection object) is done in the fourth commit in this PR.

The fifth commit changes the places that raise pywbem exceptions to pass the connection ID to the exceptions, including some changes that were needed in support of that such as changing a static method to become an instance method.

The following pywbem exceptions that are raised do not get the connection ID passed as part of this PR:

* Parse errors in CIM-XML responses received from a WBEM server are raised as pywbem.ParseError exceptions and they do not have the connection ID. Reason is that adding the connection ID there is difficult to impossible, because in tupleparse we have flat functions whose interface is determined by our use of PLY.
* Pywbem exceptions raised in the pywbem listener do not have the connection ID. Reason is that for the listener, we don't have a concept of a connection ID.

### DIFF
--- a/pywbem/exceptions.py
+++ b/pywbem/exceptions.py
@@ -34,30 +34,30 @@ class Error(Exception):
         """
         Parameters:
 
-          conn (:class:`~pywbem.WBEMConnection`): Must be a keyword argument.
-            Connection in whose context the error happened. Omitted or `None`
-            if the error did not happen in context of any connection, or if the
-            connection context was not known.
+          conn_id (:term:`connection id`): Must be a keyword argument.
+            Connection ID of the connection in whose context the error
+            happened. Omitted or `None` if the error did not happen in context
+            of any connection, or if the connection context was not known.
 
           Any other positional arguments or keyword arguments are passed to
           :exc:`py:Exception`.
         """
-        if 'conn' in kwargs:
-            conn = kwargs['conn']
-            del kwargs['conn']
+        if 'conn_id' in kwargs:
+            conn_id = kwargs['conn_id']
+            del kwargs['conn_id']
         else:
-            conn = None
+            conn_id = None
         super(Error, self).__init__(*args, **kwargs)
-        self._conn = conn
+        self._conn_id = conn_id
 
     @property
-    def conn(self):
+    def conn_id(self):
         """
-        :class:`~pywbem.WBEMConnection`: Connection in whose context
+        :term:`connection id`: Connection ID of the connection in whose context
         the error happened. `None` if the error did not happen in context
         of any connection, or if the connection context was not known.
         """
-        return self._conn
+        return self._conn_id
 
     @property
     def conn_str(self):
@@ -65,8 +65,7 @@ class Error(Exception):
         :term:`unicode string`: String that identifies the connection in
         exception messages.
         """
-        conn_id = self.conn.conn_id if self.conn else None
-        ret_str = "Connection id: %s" % conn_id
+        ret_str = "Connection id: %s" % self.conn_id
         return ret_str
 
     def __str__(self):
@@ -102,7 +101,7 @@ class HTTPError(Error):
 
     # pylint: disable=super-init-not-called
     def __init__(self, status, reason, cimerror=None, cimdetails=None,
-                 conn=None):
+                 conn_id=None):
         """
         Parameters:
 
@@ -123,9 +122,10 @@ class HTTPError(Error):
 
             Passing `None` will result in an empty dictionary.
 
-          conn (:class:`~pywbem.WBEMConnection`): Connection in whose context
-            the error happened. `None` if the error did not happen in context
-            of any connection, or if the connection context was not known.
+          conn_id (:term:`connection id`): Connection ID of the connection in
+            whose context the error happened. `None` if the error did not
+            happen in context of any connection, or if the connection context
+            was not known.
 
         :ivar args: A tuple (status, reason, cimerror, cimdetails) set from the
             corresponding init arguments.
@@ -133,7 +133,7 @@ class HTTPError(Error):
         if cimdetails is None:
             cimdetails = {}
         super(HTTPError, self).__init__(
-            status, reason, cimerror, cimdetails, conn=conn)
+            status, reason, cimerror, cimdetails, conn_id=conn_id)
 
     @property
     def status(self):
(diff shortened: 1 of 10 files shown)


pywbem/exceptions.py
```diff
@@ -78,10 +122,18 @@ def __init__(self, status, reason, cimerror=None, cimdetails=None):
             * Value: header field value (i.e. text message)
 
             Passing `None` will result in an empty dictionary.
+
+          conn (:class:`~pywbem.WBEMConnection`): Connection in whose context
+            the error happened. `None` if the error did not happen in context
+            of any connection, or if the connection context was not known.
+
+        :ivar args: A tuple (status, reason, cimerror, cimdetails) set from the
```
KSchopmeyer: I noticed that the args doc moved moved up to HTTPError at the same time you inserted the super... However, I do not see where above this in the hierarchy the args actually gets set.  I may be just missing it.

andy-maier: The args doc on HTTPError was missing, and adding it is a docs fix that is independent of all the other changes.

The args instance variable is set in the Python Exception class, and the previous implementation duplicated that into HTTPError (and into CIMError, fwiw). Now that our Error class has an init method, we can use the more logical approach of delegating that into the Python Exception class. Maybe that would have worked even without an init method in the Error class in between, but I never tried. In any case, the unit tests verify that the args instance variable is set.

KSchopmeyer:
Hit one issue it doing make clobber install check test in the test.

============================= test session starts ==============================
platform linux2 -- Python 2.7.6, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
rootdir: /home/kschopmeyer/pywbem/githubpywbem/pywbem, inifile:
plugins: cov-2.5.1
collected 0 items / 1 errors
Coverage.py warning: No data was collected.

==================================== ERRORS ====================================
______________________________ ERROR collecting  _______________________________
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/py/_path/common.py:377: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/py/_path/common.py:429: in gen
    for p in self.gen(subdir):
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/py/_path/common.py:419: in gen
    if p.check(dir=1) and (rec is None or rec(p))])
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/_pytest/main.py:511: in _recurse
    ihook = self.gethookproxy(path)
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/_pytest/main.py:413: in gethookproxy
    my_conftestmodules = pm._getconftestmodules(fspath)
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/_pytest/config/__init__.py:383: in _getconftestmodules
    mod = self._importconftest(conftestpath)
../../../.virtualenvs/pywbem27/local/lib/python2.7/site-packages/_pytest/config/__init__.py:412: in _importconftest
    raise ConftestImportFailure(conftestpath, sys.exc_info())
E   ConftestImportFailure: AttributeError("'module' object has no attribute 'SSLContext'",)
E     File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/_pytest/assertion/rewrite.py", line 216, in load_module
E       py.builtin.exec_(co, mod.__dict__)
E     File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/py/_builtin.py", line 221, in exec_
E       exec2(obj, globals, locals)
E     File "<string>", line 7, in exec2
E     File "/home/kschopmeyer/pywbem/githubpywbem/pywbem/testsuite/testclient/conftest.py", line 137, in <module>
E       import httpretty
E     File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/httpretty/__init__.py", line 28, in <module>
E       from . import core
E     File "/home/kschopmeyer/.virtualenvs/pywbem27/local/lib/python2.7/site-packages/httpretty/core.py", line 102, in <module>
E       old_sslcontext_wrap_socket = ssl.SSLContext.wrap_socket

**_NOTE: I repeated the test on the latest master and hit same issue._**

================================
The good news is that it passed several runs with run_cim_operations.  I am not picking up any extra errors.

 However, for some reason it takes a very long time to go through the subscriptions and indications so that there is something I don't understand.  When I say long time, I mean minutes instead of seconds.  Further, the whole slowdown when we start the subscription processing.

Note that this may not be part of this change since I retested with master and one time it operated at normal speed but on a later try it also slowed way down when the subscription tests started.



I think I am going to have to dig issue probably tomorrow afternoon. I am going to continue testing.

KSchopmeyer:
Simple tests with wbemcli worked fine.

andy-maier:
Karl, on your issue: The "check" and "test" targets depend on the "develop" target. If "develop" is not run, you can get these kinds of issues. Please try with "develop" before "check" and "test" and let me know whether the problem persists.

andy-maier:
**Discussion points:**

* Are we confident enough w.r.t. tests against real CIMOMs?
* We can actually add connection information also to tupleparse, by adding attributes to the parser objects. We do that already for some things. Should we add that?

KSchopmeyer:
We agreed to add connection info to tupleparse.

Karl will rerun tests again with clean virt env and be sure to make develop

andy-maier:
I am merging this and will add support for connection info in tupleparse in another PR.

andy-maier:
Just for the record: There were several misconceptions in what I stated further up in this issue:

1. The difficulty of adding cinnection info to tupleparse has been described as being caused by the use of PLY. That is not true; that argument applies to the mof_compiler module but not to tupleparse.
2. The solution for adding connection info to tupleparse has been suggested to be done by adding attributes to the parser objects. There are no parser objects in tupleparse (they are again in the mof_compiler module).

The solution for adding connection info to the tupleparse (and tupletree) modules is to put the module-global functions into a new class that stores the connection ID as an instance attribute. The solution for tupletree is to add the connection ID as an additional optional argument to the three relevant functions of that module.

PR #1293, opened 2018-06-21
Fixes #1188 Add CIMInstance.from_class() static method and tests

### LINKED ISSUE #1188: Add method to create instance from CIMClass object
The mocker can easily put classes into a mock repo with issue #1187.

We can now define a method that will create an instance from a pywbem CIMClass with a single statement.

Add a method that will create an instance from a CimClass object with serveral parameters on the charastics of the instance to be created.

KSchopmeyer:
Marked for review and merge.  See commit for details

Adds new method to CIMInstance, the static method instance_from class
and set of tests for this method.

This also removes the temporary version of this method in testsuite.

**DISCUSISON:** As it stands this is not guaranteed property name case independent unless the property_values dict is a NoCaseDict. RESOLVED AND FIXED


pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            `None` if the default_namespace is to be used.
+
+          property_values (:class:`py:dict` or `NocaseDict`_):
+            Dictionary containing name/value pairs where the names are the
+            names of properties in the class and the properties are the
+            property values to be set into the instance. The values must match
+            the type defined for the property in the class. If a property is in
+            the property_values dictionary but not in the class a ValueError
+            exception is raised. Property name case independence is only
+            honored if it is a NoCaseDict (`NocaseDict`_).
```
andy-maier: I don't think this requirement is acceptable when providing a case sensitive dict. How should one know what the right case is that ought to be used? The current implementation requires using the case of the properties in the klass object (and that is not even documented). But how does one know whether these are in sync with the MOF class definition?
I think this requirement should be eliminated, e.g. by rethinking the algorithm such that "in" tests are only done against NocaseDicts that already eixst, or if that cannot be achieved by copying the input dict into a NocaseDict.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+
+          include_null_properties (:class:`py:bool`):
+            Determines if properties not in the property_values dictionary are
+            included in the instance. Default is `True`.
+
+            If `True` they are included in the instance returned with any
+            value from the class.
```
andy-maier: I don't understand what "with any value from the class." means.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            names of properties in the class and the properties are the
+            property values to be set into the instance. The values must match
+            the type defined for the property in the class. If a property is in
+            the property_values dictionary but not in the class a ValueError
+            exception is raised. Property name case independence is only
+            honored if it is a NoCaseDict (`NocaseDict`_).
+
+          include_null_properties (:class:`py:bool`):
+            Determines if properties not in the property_values dictionary are
+            included in the instance. Default is `True`.
```
andy-maier: I'm not sure I like the parameter name "include_null_properties". Properties not in the property_values dictionary are missing in that dictionary, but are not NULL. So what really happens is that missing properties are set to NULL. How about "include_missing_properties"?

Also, we should state that these properties are set to NULL, regardless of what their default value is on the class. Maybe what we need is a choice between three options: Do not include them, Include them as NULL, include them with their class default value.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+
+            If `False` they are not included in the instance returned.
+
+         include_class_origin  (:class:`py:bool`):
+            Determines if ClassOrigin information is included in the returned
+            instance. Default is `False`.
+
+            If None or False, class origin information is not included.
+
+            If True, class origin information is included.
```
andy-maier: Use backquotes around None, False, True.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            the new instance.  If `strict` all key properties must be in the
+            instance. Default is `True`
+
+          strict (:class:`py:bool`:):
+            If `True` and `include_path` is set, all key properties must be in
+            the instance so that a correct path can be built. Default is
+            `False`.
+
+            If not `True` The path component is created even if not all
+            key properties are in the created instance.
```
andy-maier: Describe how the missing key properties are set in the CIMInstanceName object. Set to NULL? Missing?

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+
+        Returns:
+            Returns an instance with the defined properties and optionally
```
andy-maier: Add the returned type.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            the path set.  No qualifiers are included in the returned instance
+            and the existence of ClassOrigin depends on the
+            `include_class_origin` parameter. The value of each property is
+            either the value from the `property_values` dictionary, the
+            default_value from the class or Null(unless
+            `include_null_properties` is False). All other attributes of each
+            property are the same as the corresponding class property.
```
andy-maier: All of this is already described for each of the respective parameters. I don't think it shuld be repeated here.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2853,124 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            CIMClass from which the instance will be constructed.  This
+            class must include qualifiers and should include properties
+            from any superclasses to be sure it includes all properties
+            that are to be built into the instance. Properties may be
+            excluded from the instance by not including them in the `klass`
+            parameter.
+
+          namespace (:term:`string`):
+            Namespace in the WBEMConnection used to retrieve the class or
+            `None` if the default_namespace is to be used.
```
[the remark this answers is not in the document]
KSchopmeyer: Actually I was completly wrong in this description.  Rewrote.  It is the namespace to be included in the path if include_path parameter is True

pywbem/cim_obj.py
KSchopmeyer: I agree.  I propose that if the input is a NoCaseDict we use it, else map the input into a NoCaseDict.

pywbem/cim_obj.py
KSchopmeyer: Propose above changing name to include_all_properties.  Actually they are set to the value defined in the class be it the default value or Null if there is no default_value in the class.

pywbem/cim_obj.py
KSchopmeyer: DONE: Fixed. Fixed a couple of others also in cim_obj.py

pywbem/cim_obj.py
KSchopmeyer: DONE: They are missing.

pywbem/cim_obj.py
KSchopmeyer: Actually like include_missing_properties better.  Change made

pywbem/cim_obj.py
KSchopmeyer: Badly written.  I will rewrite but am also unhappy with the name.   Changed name of parameter to include_missing_properties.

pywbem/cim_obj.py
KSchopmeyer: DONE

pywbem/cim_obj.py

pywbem/cim_obj.py
KSchopmeyer: Note that this is actually in from_instance code

pywbem/cim_obj.py
KSchopmeyer: DONE Rewrote that completely

pywbem/cim_obj.py
KSchopmeyer: DONE: Added code to convert any other dict type to NocaseDict and added couple more tests

pywbem/cim_obj.py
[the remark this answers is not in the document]
KSchopmeyer: DONE

pywbem/cim_obj.py
```diff
@@ -2853,6 +2855,148 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            property is in the property_values dictionary but not in the class
+            a ValueError exception is raised.
+
+          include_missing_properties (:class:`py:bool`):
+            Determines if properties not in the `property_values` parameter are
+            included in the instance.
+
+            If `True` all properties from the class are included in the new
+            instance including those not defined in `property_values` parameter
+            with any value from the class. This is the `Default`.
```
andy-maier: You indicated you had improved the phrase "with any value from the class.", but it is still there. Given that you clarified that the class-defined default value or NULL is used, I suggest to replace that phrase with: "with the default value defined in the class if one is defined, or otherwise `None`".

andy-maier: On "This is the `Default`": Normally we don't repeat what the default is. Suggest to remove that sentence, or at least remove the backquotes.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2855,148 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+
+            If `False` only properties in the `property_values` parameter are
+            included in the new instance.
+
+         include_class_origin  (:class:`py:bool`):
+            Determines if class origin information from the class is included
+            in the returned instance. Default is `False`.
```
andy-maier: On "Default is `False`": Normally, we don't repeat the default of the method argument.

pywbem/cim_obj.py
```diff
@@ -2853,6 +2855,148 @@ def tomof(self, indent=0, maxline=MAX_MOF_LINE):
...
+            into the new instance based on properties in the new instance.  If
+            `strict` all key properties must be in the instance.
+
+          strict (:class:`py:bool`:):
+            This parameter is only used if the `include_path parameter is
+            `True`.
+
+            If `True` all key properties from the class must exist the instance
+            so that a correct path can be built. Default is `False` or an
+            ValueError exception is raised. This is the default.
```
andy-maier: On "This is the default": Normally, we don't repeat the default of the method argument.

KSchopmeyer: Removed

pywbem/cim_obj.py
KSchopmeyer: Removed sentence.

pywbem/cim_obj.py
KSchopmeyer: Removed sentence.

KSchopmeyer: DONE. That is clearer.

KSchopmeyer:
DISCUSSION; In retrospect I don't like the 'strict' param, it just adds confusion.  It would seem logical that if you are asking for this to create the path, you should have enough in the instance to actually create the correct path, not just a partial path.

Would it make sense to drop the strict and state that it will include a correct path or fail (i.e. always use strict))????

KSchopmeyer:
strict parameter of from_class dropped throughout the code that involves this new method. That includes cim_obj, test_cim_obj test=subscription, and wbemservermock.

Note that I added a couple of pylint defs and also reworded some of the from_class documentation for clarity and simpler wording.

PR #1296, opened 2018-06-21
Improved makefile in some areas

andy-maier:
Ready for review and merge.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,27 +1,25 @@
 # ------------------------------------------------------------------------------
-# Makefile for pywbem
+# Makefile for pybem repository of pywbem project
 #
-# Supported platforms for this makefile:
-#   Linux
-#   OS-X
-#   Windows (with CygWin, MinGW, etc.)
+# Supported OS platforms for this makefile:
+#     Linux (any distro)
+#     OS-X
+#     Windows with UNIX-like env such as CygWin (with Python in UNIX-like env)
+#     native Windows (with Python in Windows)
 #
-# Prerequisite commands for this makefile:
-#   make
-#   bash, sh, rm, mv, mkdir, echo
-#   find, xargs, grep, sed, tar
-#   python (Some active Python version, virtualenv is supported)
-#   pip (in the active Python environment)
-#   twine
-#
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -355,7 +374,7 @@ clobber: clean
 clean:
 	@echo "makefile: Removing temporary build products"
 	find . -name "*.pyc" -delete
-	sh -c "find . -name \"__pycache__\" |xargs rm -Rf __pycache__"
+	bash -c "find . -name \"__pycache__\" |xargs -n 1 rm -Rf"
```
KSchopmeyer: Do we really need bash.  There may be some platforms left over that still have sh as the default shell.

makefile
```diff
@@ -517,7 +536,7 @@ ifeq ($(PLATFORM),Windows)
 else
 	testsuite/test_uprint.sh
 endif
-	bash -c "set -o pipefail; PYTHONWARNINGS=default py.test --cov $(package_name) --cov $(mock_package_name) $(coverage_report) --cov-config coveragerc --ignore=attic --ignore=releases -s 2>&1 |tee $(test_tmp_file)"
+	bash -c "set -o pipefail; PYTHONWARNINGS=default py.test --cov $(package_name) --cov $(mock_package_name) $(coverage_report) --cov-config coveragerc $(pytest_opts) --ignore=attic --ignore=releases -s 2>&1 |tee $(test_tmp_file)"
```
KSchopmeyer: Same comment

makefile
andy-maier: We do depend on bash features in several of our .sh scripts, as well as in several places in makefiles (if conditions, pipefail, ...).

bash is the standard shell in Linux. OS-X has it, too. CygWin has it. There is more danger of not having Bourne shell on Linux than not having bash. If you know of any Linux platforms that don't have bash, post them here.

I checked and I did not find any other sh holdouts in pywbem or pywbemtools. If you know of any, post them here.

KSchopmeyer: Sounds good to me.  Forget my comment.

andy-maier:
Karl agreed in slack.

PR #1301, opened 2018-06-22
Fixes #1298 null value keybindings

### LINKED ISSUE #1298: Method from_instances allows Null values in CIMInstanceName
We do not test for None in key properties when build the instance path from the instance.

DSP0004 states that key properties should not be None.

Suggest adding test to confirm property value is not None and raise error if it is None.

KSchopmeyer:
This is ready for review and merge. See commit for details

### DIFF
--- a/pywbem/cim_obj.py
+++ b/pywbem/cim_obj.py
@@ -253,6 +253,7 @@ import six
 
 from . import cim_xml
 from .config import DEBUG_WARNING_ORIGIN, SEND_VALUE_NULL
+import pywbem.config
 from .cim_types import _CIMComparisonMixin, type_from_name, cimtype, \
     atomic_to_cim_xml, CIMType, CIMDateTime, Uint8, Sint8, Uint16, Sint16, \
     Uint32, Sint32, Uint64, Sint64, Real32, Real64, number_types, CIMInt, \
@@ -1160,6 +1161,8 @@ class CIMInstanceName(_CIMComparisonMixin):
         Raises:
 
           ValueError: An error in the provided argument values.
+          ValueError: A keybinding value is `None` and the config variable
+            IGNORE_NULL_KEY_VALUE is `False`
           TypeError: An error in the provided argument types.
         """  # noqa: E501
 
@@ -1208,6 +1211,8 @@ class CIMInstanceName(_CIMComparisonMixin):
           preserved.
 
         * value (:term:`CIM data type` or :term:`number`): Keybinding value.
+          If the config variable IGNORE_NULL_KEY_VALUE is True, `None` is
+          allowed as a key value.
 
         The order of keybindings in the instance path is preserved.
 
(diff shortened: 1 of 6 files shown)


andy-maier:
It seems to me there is a general issue that we (I) don't understand yet, and that applies to all config variables:

It seems that `pywbem.config.XXX` and `pywbem.XXX` are two different objects, even though the names from `pywbem.config` are imported into `pywbem`.

Given that we document to use the config variables in the `pywbem` namespace, but they are documented to be in the `pywbem.config` namespace, this creates a significant chance for users to run into that issue.

We need to do some basic research on this and find a general solution.

KSchopmeyer:
I agree. One more discussion item on the agenda.

KSchopmeyer:
We agreed to change the access methods for config. We will commit this change and andy will do the general change.

pywbem/cim_obj.py
```diff
@@ -253,6 +253,7 @@ class (the parent object) has a list of CIM properties, CIM methods and CIM
 
 from . import cim_xml
 from .config import DEBUG_WARNING_ORIGIN, SEND_VALUE_NULL
+import pywbem.config
```
andy-maier: This should import from `.`, not from `pywbem`, to make sure it remains a relative import. In other words, this would become:
```
from . import config
```
and your new config variable would be accessed as:

```
config.IGNORE_NULL_KEY_VALUE
```

pywbem/config.py
```diff
@@ -93,3 +93,17 @@
 #:
 #: *New in pywbem 0.12.*
 SEND_VALUE_NULL = True
+
+#: Backward compatibility option that controls creating CIMInstanceNames with
```
andy-maier: Instead of using plural on a class name, use a link on the class name:
```
... creating :class:`~pywbem.CIMInstanceName` objects with ...

File: pywbem/config.py [hunk 3]
```diff
@@ -93,3 +93,17 @@
...
+#: null values for keybindings.
+
+#: :term:`DSP0004`, section 7.9 specifically forbids key properties with
+#: no values but because pywbem has always allowed this, adding the code to
```
andy-maier: Instead of saying "with no values", I suggest to say "that are NULL".

pywbem/config.py
```diff
@@ -93,3 +93,17 @@
...
+#: disallow None as a keybinding value is an incompatible change.
+#:
+#: * True: Pywbem ignores None as a value when keybindings are defined.
```
andy-maier: It does not "ignore" None in this case, but "tolerates" it.

pywbem/config.py
```diff
@@ -93,3 +93,17 @@
...
+#: * False (default): Pywbem generates ValueError when keybindings are created
```
andy-maier: Use back quotes around True, False, None.

andy-maier: It does not "generate" ValueError, but "raises" ValueError. Also, use a link for ValueError.

pywbem/config.py
KSchopmeyer: DONE

pywbem/config.py
KSchopmeyer: DONE

pywbem/config.py
KSchopmeyer: DONE

PR #1308, opened 2018-07-15
Fixes #1302 Fixed issue with updating pywbem config variables.

### LINKED ISSUE #1302: Issue with updating pywbem config variables
The pywbem config variables are defined in the `pywbem.config` namespace and then imported by pywbem into the `pywbem` namespace. Pywbem documents that these config variables should be accessed (read and written) through the `pywbem` namespace. However, pywbem code checks them from the `pywbem.config` namespace.

This causes an update that is done by a pywbem user through the `pywbem` namespace, not to be visible in the `pywbem.config` namespace, so pywbem does not react to the user's change.

The root cause of this is that in Python, variables are not containers for values, but names that are bound to objects. Assigning a new value into the config variable causes that name to be bound to the object representing the new value.

The variable name in the `pywbem` namespace is a different name than the name in the `pywbem.config` namespace, so binding one of them to a new object does not change the name binding of the other one.

This can be seen in this example code:

```
>>> import pywbem
>>> id(pywbem.config.SEND_VALUE_NULL)
4319171424
>>> id(pywbem.SEND_VALUE_NULL)
4319171424
>>> pywbem.SEND_VALUE_NULL=1
>>> id(pywbem.SEND_VALUE_NULL)
140517924967032
>>> id(pywbem.config.SEND_VALUE_NULL)
4319171424
``` 

In this example, both names initially are bound to the same object. After the assignment, the assigned name is bound to the new object (with value 1), but the other name is still bound to the original object.

How to fix this:

The problem only happens if reading and writing of the config variables is done through different namespaces.

I verified, that in all cases, the pywbem implementation checks the config variables through the `pywbem.config` namespace.

However, the documentation states that despite the fact that the variables are shown (for doc build tool reasons) in the `pywbem.config` namespace, they should be accessed through the `pywbem` namespace.

If user code has actually attempted to set them through the `pywbem` namespace, that cannot have worked. So any user code that did succeed in changing the config variables must have done it through the `pywbem.config` namespace.

Therefore, the suggested solution is to no longer import the config variables into the `pywbem` namespace, and to change the documentation to recommend access through the `pywbem.config` namespace. That keeps working user code working, and removes the somewhat weird statement about access through `pywbem` when they are documented as being in `pywbem.config`.

Because the config variables get eliminated from the pywbem namespace, this change needs to be flagged as an incompatible change.

Also, ideally the unit test cases that depend on config variable settings should be extended to have variants with the different values for the config variables.

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -43,6 +43,14 @@ Released: not yet
   this property to `TRUE` will restore the original behavior but the NULL
   keybinding may be an issue with some WBEM servers.(Issue #1298)
 
+* The fix for issue #1302 removed the pywbem config variables from the
+  `pywbem` namespace. They are now only available via the `pywbem.config`
+  namespace. However, this change should not affect you because the
+  previously documented approach for changing them through the `pywbem`
+  namespace did not work, so if you changed the config variables
+  successfully, you have done that through the `pywbem.config` namespace
+  already.
+
 **Deprecations:**
 
 * Deprecated the `tocimobj()` function because it has some inconsistencies,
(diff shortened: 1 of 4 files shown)


andy-maier:
Yes, that's a good idea.

andy-maier:
Done, and also updated the change log.
Considering this one approved.

Issue #29, opened 2016-02-07
module mof_compiler depends on PG_InterOp ns name


KSchopmeyer:
It appears that this particular namespace name is hardcoded into the compiler.

See the function _create_ns on about line 216

andy-maier:
If I read this correctly, then this is code that attempts to support the `__Namespace` class in the `PG_InterOp` namespace for OpenPegasus, in case it does not support the `CIM_Namespace` class in the `Interop` namespace. The latter is also not quite complete. [DSP1033 1.1.0](http://www.dmtf.org/sites/default/files/standards/documents/DSP1033_1.1.0.pdf) defines for the name of the Interop namespace:
* Standard names are:
  - `interop` (preferred)
  - `/interop` (DEPRECATED)
  - `root/interop` (DEPRECATED)
  - `/root/interop` (DEPRECATED)
* Clients need to be prepared to deal with any one of these standard names.
* A WBEM server may expose Interop namespaces using additional implementation-defined names. This accommodates backwards compatibility of existing WBEM server implementations. Clients should use the standard names instead of such implementation-defined names.

By the way, I verified that the `mof_compiler` module has the only code in the PyWBEM client project that deals with the Interop namespace.

I think extending that code so that it supports all four standard names is something we should do in any case.

That code only special-cases for OpenPegasus. In any case, we can move that special case after the attempt to use the standard names.

The interesting question is can we remove that special code entirely. In which OpenPegasus version was the support for the standard class name and the standard Interop namespace names introduced?

andy-maier:
We agreed to use the standard interop namespace names first. There are still current OP versions that can be configured with only the PG_Interop name, so that cannot be dropped and should be tried after the standard names.

KSchopmeyer:
Consider removing the _Namespace support since it is deprecated.

KSchopmeyer:
Karl will review this to determine if still valid.

KSchopmeyer:
Yes this is still valid.  

1. A agree that starting with CIM_Namsepace is more logical and __namespace__ is only a final fallback position.

2. Why can't we use the code in wbemserver that finds namespaces and the interop namespace in place of this years old hack?  DISCUSSION

KSchopmeyer:
[...]

PR #1318, opened 2018-07-20
Added WBEMServer.create_namespace() and .cimom_inst.

andy-maier:
This PR addresses item 1 described in issue #29.
Ready for review and merge.
It would be good to run WBEMServer.create_namespace() against a life server.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -98,12 +98,13 @@ import re
 
 from .cim_constants import CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_CLASS, \
     CIM_ERR_METHOD_NOT_FOUND, CIM_ERR_METHOD_NOT_AVAILABLE, \
-    CIM_ERR_NOT_SUPPORTED, CIM_ERR_NOT_FOUND
+    CIM_ERR_NOT_SUPPORTED, CIM_ERR_NOT_FOUND, CIM_ERR_FAILED
 from .exceptions import CIMError
 from ._nocasedict import NocaseDict
-from .cim_obj import CIMInstanceName
+from .cim_obj import CIMInstanceName, CIMInstance
 from .cim_operations import WBEMConnection
 from ._valuemapping import ValueMapping
+from ._utils import _ensure_unicode
 
 __all__ = ['WBEMServer']
 
@@ -160,6 +161,7 @@ class WBEMServer(object):
         self._namespace_classname = None
         self._brand = None
         self._version = None
+        self._cimom_inst = None
         self._profiles = None
 
     def __repr__(self):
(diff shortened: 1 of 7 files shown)


pywbem/_server.py
```diff
@@ -299,6 +319,115 @@ def profiles(self):
...
+              `CIM_ObjectManager` instances.
+            CIMError: CIM_ERR_FAILED, Unexpected number of
+              central instances of WBEM Server profile.
+        """
+
+        std_namespace = _ensure_unicode(namespace.strip('/'))
+
+        # Try approach 1: Method defined in WBEM Server profile
+
+        ws_profiles = self.get_selected_profiles('DMTF', 'WBEM Server')
```
KSchopmeyer: As long as we are getting the WBEM_Server instance here through the profile, should we consider making that public or its access public because it will be used for other things than just creating namespace???

KSchopmeyer: Either the result or the method to get it.

andy-maier: No. Let's not do that before we understand the use cases. This is a MOF compiler API and the user supplies a target repository connection, so the user already has everything they need to set up their own WBEM_Server if they need one. Just saving a few operations is not worth the function creep.

andy-maier:
Karl, could you please run this against a real CIMOM?

andy-maier:
Rebased to resolve a conflict.

KSchopmeyer:
Right now this one eats OpenPegasus's lunch.  The server fails and will not restart. Core dumps on restart. The namespace is created but marked as a pegasus shared namespace. with a special flag on the new namespace (a directory with the name SRF).  I think that our issue is with some parameter in the new_instance of the create_instance.

KSchopmeyer:
Filed bug on OpenPegasus for this issue. bug number 10112 at http://bugzilla.openpegasus.org/show_bug.cgi?id=10112

KSchopmeyer:
I submitted a new pr based on this pr with proposal for fix and the test defined for run_cimoperations.py

andy-maier:
**Discussion point:** The current code, also with your fix, provides values for the five key properties of PG_Namespace besides just "Name". Should they be provided or not provided?

KSchopmeyer:
We agreed to do the two issues listed in discussions above:

1. We need to do a delete_namespace method so the user does not have to understand the details of the mechanism to create and delete namespaces.

2. We agreed that we need to update the list of namespaces if the user does the create_namespace to that the local list reflects the change.  This will apply to the delete_namespace also.

3. We agree that we will not delete if anything is in the namespace.

4. We will not supply those parameters on openpegasus or the others in accord with DSP0200.

andy-maier:
Updated the PR to include the fixes to the WBEMServer code from Karl's PR #1346, plus items 2 and 4 from Karl's comment further up.

andy-maier:
Items 1 and 3 from Karl's list above will be covered in another PR.

KSchopmeyer:
DISCUSSION: Despite what I read in the code, OpenPegasus requires all of the key properties to execute the create_instance in CIM_Namespace or PG_Namespace. I tested with each one individually NOT provided and in all cases failed with the CIM_ERR_INVALID_PARAMETER exception

Before I commit to requiring them, I would like to go through the Pegasus code more throughly since I am apparently missing something in the code

andy-maier:
I think the behavior you describe is clear enough to conclude that we need to provide these properties at least in case of OP.

Plus, according to what you said in the slack channel, OP then ignores the provided properties and re-creates them, which can result in a different path being returned. We should document that in pywbem even though we don't use the path today.

andy-maier:
I added a commit that now provides all keys, both for OP and for other CIMOMs.

KSchopmeyer:
I turned off discussion needed

KSchopmeyer:
Tested with the run_cimoperations from my proposal that creates namespaces and then deletes them and also test error when trying to create namespace that exists.  Both passed.

PR #1324, opened 2018-07-30
Added support for 'canonical' format in to_wbem_uri()

andy-maier:
For details, see the commit message.
Ready for review and merge.

PR #1325, opened 2018-07-31
Ignoring host for CIMobject format of WBEM URIs

andy-maier:
Note: This PR is on top of PR #1324.

Ready for review and merge.
It would be good to test this PR against a real WBEM server.


pywbem/cim_obj.py
```diff
@@ -3576,10 +3574,8 @@ def case(s):
 
         ret = []
 
-        # TODO: Ignore host for CIMObject format
-        # if self.host is not None and format not in ('cimobject'):
-        #     # The CIMObject format assumes there is no host component
-        if self.host is not None:
+        if self.host is not None and format not in ('cimobject'):
```
KSchopmeyer: Question. Is this the most efficient way to determine value of format?  I would not thought of this syntax

KSchopmeyer:
Note that it was just a single error in the test, not a complete failure to communicate but

pywbem/cim_obj.py
andy-maier: No, it is not the most efficient format. I was just doing this for consistency with the other tests, but I will change it to a simple comparison.

andy-maier:
Addressed the review comment and rebased.
The test is still needed.

KSchopmeyer:
Tested OK against OpenPegasus except that for some reason I am getting big delays in the subscription tests.  Looking at that now.

KSchopmeyer:
Runs cim_operations.py OK except that for some reason it is taking very long time to run subscription create/delete test.  I am going to consider this one ok. Will get more info and file separate bug on the timing issue.  Filed new issue #1360 just to capture data

Issue #1332, opened 2018-08-18
ExecQuery is not included in wbemcli


KSchopmeyer:
We completely for got the ExecQuery operation.

andy-maier:
Addressed by PR #1333.

PR #1333, opened 2018-08-18
Issue #1332 Add ExecQuery Operation to wbemcli as shortcut eqy.

KSchopmeyer:
Somewhere we completely forgot this operation.  I only added it because
I needed to test it since we are getting some strange results from
run_cimoperations.py in another pr and I wanted to confirm or discard
that some conclusions.

Note: There is no test since the test_wbemcli.py tests only initialization, etc. not the individual operations.

Add new entry to changes.doc


andy-maier:
Please rebase on master to pick up the fix for the failure on Appveyor.

andy-maier:
Approved the changes, pending the rebase.

PR #1346, opened 2018-08-20
Ks/proposed fixes create namespace for OpenPegasus and test

KSchopmeyer:
This is just a proposal for pegasus to allow the existing OpenPegasus to do create namespace as an extension to pr #1318.

It also includes the test I used in run_cimoperations.py to test the code.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -408,6 +408,8 @@ class WBEMServer(object):
 
         # Try approach 2: CreateInstance of CIM_Namespace
 
+        # Use specific OpenPegasus namespace class to account for issue
+        # in OpenPegasus. See OpenPegasus bug report
         if self.brand == "OpenPegasus":
             ns_inst = CIMInstance('PG_Namespace')
             ns_inst['SchemaUpdatesAllowed'] = True
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -98,12 +98,13 @@ import re
 
 from .cim_constants import CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_CLASS, \
     CIM_ERR_METHOD_NOT_FOUND, CIM_ERR_METHOD_NOT_AVAILABLE, \
-    CIM_ERR_NOT_SUPPORTED, CIM_ERR_NOT_FOUND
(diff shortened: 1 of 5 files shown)


pywbem/_server.py
```diff
@@ -424,7 +428,7 @@ def create_namespace(self, namespace):
         ns_inst['CreationClassName'] = 'CIM_Namespace'
         ns_inst['Name'] = std_namespace
 
-        self.conn.CreateInstance(ns_inst, namespace=self.interop_ns)
+        path = self.conn.CreateInstance(ns_inst, namespace=self.interop_ns)
```
andy-maier: The `path` is not used and will fail the flake8 check.

testsuite/run_cim_operations.py
```diff
@@ -4503,6 +4503,43 @@ def test_server_select_profiles(self):
         self.assertEqual(len(server.get_selected_profiles('blah', 'blah')), 0)
         self.assertEqual(len(server.get_selected_profiles('blah')), 0)
 
+    def test_create_namespace(self):
+        """Test the creat_namespace method."""
+
+        server = WBEMServer(self.conn)
+        namespaces = server.namespaces
+        interop = server.interop_ns
+        test_ns = "root/runcimoperationstestns"
```
andy-maier: I think given that root has this special role, we should also test creation of namespaces that do not begin with `root/`.

KSchopmeyer: Agreed.  In fact I want to add code that creates a qualifier decl to really establish that there is a working namespace.

The failure is in the tests, in particular I used PG_Interop and that is not mocked.

andy-maier:
The remaining errors are:
* `path`variable assigned but not used.
* The newly used "PG_Namespace" class is not in the mocked repository that is used by the test_wbemserverclass.py test cases. It should be added in the build_class_repo() method in wbemserver_mock.py.

andy-maier:
Karl, I have updated my PR #1318 to include the changes to the `pywbem/_server.py` file. It should now be complete w.r.t. the new create_namespace() method. The deletion method is left for a future PR.

You can now update this PR here to provide just the changes to the `run_cim_operations.py` file on top of my PR.

KSchopmeyer:
This one had gotten so confusing that I am closing it and creating a new one to incorporate the run_cimoperation changes.

PR #1366, opened 2018-09-02
Changed namespace handling in mock support to be explicit

andy-maier:
The commit message details all changes, explains the background, and should be read before reviewing.
Ready for review and merge.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -47,6 +47,7 @@ from pywbem import WBEMConnection, CIMClass, CIMClassName, \
     CIM_ERR_INVALID_CLASS, CIM_ERR_ALREADY_EXISTS, \
     CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_ENUMERATION_CONTEXT, \
     CIM_ERR_NOT_SUPPORTED, CIM_ERR_QUERY_LANGUAGE_NOT_SUPPORTED, \
+    CIM_ERR_NAMESPACE_NOT_EMPTY, \
     DEFAULT_NAMESPACE, MOFCompiler, MOFWBEMConnection
 from pywbem._nocasedict import NocaseDict
 from ._dmtf_cim_schema import DMTFCIMSchema
@@ -288,6 +289,14 @@ class FakedWBEMConnection(WBEMConnection):
             use_pull_operations=use_pull_operations,
             stats_enabled=stats_enabled, timeout=timeout)
 
+        # The namespaces that may exist in the mock repository. This is a
+        # dictionary where the key is the namespace name, and the value does
+        # not matter (this approach is used to ensure namespace names are
+        # treated case insensitively). This set of namespaces is used to
+        # control into which namespaces of the mock repository classes,
+        # qualifier types and instances may be added.
+        self.namespaces = NocaseDict({default_namespace: True})
+
         # The CIM classes in the mock repository.
         # This is a dictionary of dictionaries where the top level key is the
         # CIM namespace name and the keys for each sub-dictionary in a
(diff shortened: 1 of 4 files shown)


testsuite/test_wbemconnection_mock.py
```diff
@@ -1559,11 +1667,12 @@ def test_getclass(self, conn, tst_class):
...
         "ns, cln, tst_ns, exp_er", [
-            [None, 'CIM_Foo', None, None],
-            [None, 'cim_foo', None, None],
-            [None, 'CIM_Foo', 'root/blah', CIM_ERR_INVALID_NAMESPACE],
-            ['root/blah', 'CIM_Foo', 'root/blah', None],
-            [None, 'CIM_Foox', None, CIM_ERR_NOT_FOUND],
+            [DEFAULT_NAMESPACE, 'CIM_Foo', None, None],
+            [DEFAULT_NAMESPACE, 'cim_foo', None, None],
+            [DEFAULT_NAMESPACE, 'CIM_Foo', 'BadNamespace',
+             CIM_ERR_INVALID_NAMESPACE],
```
KSchopmeyer: Doesn't this actually change the test? Most of the tests were None and one was a namespace?  This has namespace DEFAULT_NAMESPACE for all of the tests. I have not looked elsewhere to see if the same question existsyet.

testsuite/test_wbemconnection_mock.py
```diff
@@ -2161,12 +2270,15 @@ class TestInstanceOperations(object):
 
     @pytest.mark.parametrize(
         "ns, cln, inst_id, tst_ns, exp_er", [
-            [None, 'CIM_Foo', 'CIM_Foo1', None, None],
-            ['root/blah', 'CIM_Foo', 'CIM_Foo1', 'root/blah', None],
-            ['blah', 'CIM_Foo', 'badid', 'blah', CIM_ERR_NOT_FOUND],
-            ['blah', 'CIM_Foo', 'CIM_Foo1', 'whoop',
+            [DEFAULT_NAMESPACE, 'CIM_Foo', 'CIM_Foo1', None, None],
```
KSchopmeyer: Original test was None for ns for this test.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -356,14 +378,91 @@ def __repr__(self):
     #
     ################################################################
 
+    def add_namespace(self, namespace):
+        """
+        Add a CIM namespace to the mock repository.
+
+        The namespace must not yet exist in the mock repository.
+
+        Note that the default connection namespace automatically exists in the
```
KSchopmeyer: This comment may require some more explanation.  I had to think about it for a moment.  Maybe something like "It is created when the _wbemconnection_mock is constructed."  Also might better refernce with conn.default_namespace.

Question. What happens if you do delete_namespace on the default conneciton namespace

testsuite/test_wbemconnection_mock.py
andy-maier: The `ns` parameter is used as the namespace in which the CIM objects are set up. The `tst_ns` parameter is the namespace against which the operation to be tested (GetClass in this case) is run. It does not make sense to set up CIM objects during test preparation n a namespace `None`, if that is not the purpose of the test.

testsuite/test_wbemconnection_mock.py
andy-maier: Same response as above.

pywbem_mock/_wbemconnection_mock.py
andy-maier: On  the comment: Changed that to: "Note that the default connection namespace is automatically added to the mock repository upon creation of this object."

On deleting the default connection namespace: There is no externally available `delete_namespace()` method on the `FakedWBEMConnection` class (only an internal `_remove_namespace()`). However, the faked DeleteInstance operation would remove it. I have added a check in the internal `_remove_namespace()` method that raises CIMError(CIM_ERR_NAMESPACE_NOT_EMPTY) if the default connection namespace is attempted to be removed, and a description of that behavior in the faked DeleteInstance operation.

andy-maier:
Comments resolved; some have been rejected, some addressed by an update.

PR #1367, opened 2018-09-02
Fixes #1356: Added WBEMServer.delete_namespace()

### LINKED ISSUE #1356: Add WBEMServer.delete_namespace()
With the introduction of `WBEMServer.create_namespace()` in PR #1318, there should also be a way provided to delete a namespace.

Non-empty namespaces must fail to be deleted.

No "delete-all" capability (that would delete non-empty namespaces with all their content) should be provided, due to the danger of such a capability.

andy-maier:
**Note: This PR is on top of PR #1366**
Ready for review and merge (review only the last commit).

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -98,7 +98,8 @@ import re
 
 from .cim_constants import CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_CLASS, \
     CIM_ERR_METHOD_NOT_FOUND, CIM_ERR_METHOD_NOT_AVAILABLE, \
-    CIM_ERR_NOT_SUPPORTED, CIM_ERR_NOT_FOUND, CIM_ERR_FAILED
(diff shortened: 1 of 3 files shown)


pywbem/_server.py
```diff
@@ -461,6 +482,100 @@ def create_namespace(self, namespace):
...
         return std_namespace
 
+    def delete_namespace(self, namespace):
+        """
+        Delete the specified CIM namespace in the WBEM server and
+        update this WBEMServer object to reflect the removed namespace
+        there.
+
+        The specified namespace must be empty (i.e. must not contain any
+        classes, instances, or qualifier types.
```
KSchopmeyer: is this qualifier types or qualifier declarations?

andy-maier: According to DSP0004, it is qualifier types. In pywbem, we are stuck with the class name `QualifierDeclaration`, so I believe we are using both.

Issue #1355, opened 2018-08-22
Enumeration operations in mock support treat class names case sensitively


andy-maier:
The fake enumeration operations in the pywbem mock support create a listof subclass names via `_get_class_list_enums()` and then check the desired classname using the `in` operator, which is case sensitive. This makes the whole classname comparisoin case sensitive in the operations.

The issue exists (at least) in the following `FakedWBEMConnection` methods:

* `_fake_enumerateinstances()`
* `_fake_enumerateinstancenames()`
* `_fake_enumerateclasses()`
* `_fake_enumerateclassnames()`

and probably also in the association operations, e.g. for those filter arguments that are class names, or for the class-level invocations.

andy-maier:
PR #1368 fixes this for 0.13.0.
It needs to be rolled back into 0.12.

andy-maier:
PR #1377 rolls back the fix into 0.12.6.

PR #1369, opened 2018-09-05
Fixes #477: Poor test in run_cim_operations

### LINKED ISSUE #477: run_cimOperations.py returns different property counts on one test
We are getting some issues between the number of properties returned in instances vs in classes in one of the run_cimoperation.py tests trying to test the deepinheritance on calls.    While I think this is an issue with the characteristics of the localonly (which is deprecated and therefore nobody knows exactly what results it produces, our only goal is really to just confirm that the deepinheritance parameter is pased to the server so we will probably just rewrite the test.

KSchopmeyer:
Ready for review except for python 2.7 failure in getting plugin

Fix test for enumerateinstances deepinheritance option.

Original test could not really test enumerateinstances DeepInheritance.
Created simpler test that just does EnumerateInstances on
CIM_ManagedElement and determines the number of properties.  Without
DeepInheritance, there should be no more properties than in class. With
DeepInhertance, Pegasus at least has many instance that have more
properties than CIM_Managed element so finds some instances with
property count gt the number of properties in CIM_ManagedElement.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -140,6 +140,10 @@ This version contains all fixes up to pywbem 0.12.4.
   possibility for the result of the ExecQuery operation.
   See issue #1347.
 
+* Fixed issue in run_cimoperations.py with test for deep inheritance on
+  EnumerateInstances. It was reporting confused result so we created a simpler
+  test. See issue #477.
+
 **Enhancements:**
 
 * Extend pywbem MOF compiler to search for dependent classes including:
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
DISCUSSION: Failureis with plugin for pytest and cov. on python 2.7. Same as pr #1355

KSchopmeyer:
Rebased and pushed.

PR #1383, opened 2018-09-12
Fixes #1380: Fixed issue of missing support for INSTANCE child elements of ERROR

### LINKED ISSUE #1380: ParseError on INSTANCE child of ERROR (was: Confusing exception for illegal / unspecified namespaces)
The second call to EnumerateInstances in the attached code generates an exception as it should, since there is no namespace specified -- either in the call directly or a default namespace in the WBEMConnection object.

The problem is that the exception message is horrible.  There is no mention of namespace at all, and the message itself is nonsensical to anyone but a core pywbem developer.

andy-maier:
Ready for review and merge.

Needs to be rolled back to 0.12.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -149,6 +149,11 @@ This version contains all fixes up to pywbem 0.12.4.
   enumerate operations, reference operations, and associator operations. This
   also adds a number of tests to validate that classnames. See issue #1355.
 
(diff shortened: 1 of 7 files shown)


andy-maier:
I would not start putting a returned error in the log. We don't do that currently, and the only change now is that there is more information coming back.

andy-maier:
PR #1385 rolls back the fix into 0.12.6

PR #1389, opened 2018-09-13
Removed ez_setup.py

andy-maier:
Ready for review and merge.


andy-maier:
It is not mentioned anywhere in the docs anymore since a long time. It is not visible for people using pip or when downloading the wheel archive from the Pypi downloads page.

However, it is visible in these ways:
* when cloning the repo
* when using the source tarball from the GitHub release page
* when using the source tarball from the Pypi downloads page

I think that warrants mentioning it in the change history.

PR #1397, opened 2018-09-20
[Review 4] Fixes #1072: Ensuring ASCII-only in exceptions, warnings and repr()

### LINKED ISSUE #1072: Use of %s vs %r for CIM names in exception messages and log entries
The current code uses sometimes `%s` and sometimes `%r` for CIM names in exception messages and log entries.

CIM names may contain Unicode characters, and `%r` will escape Unicode characters as needed, while `%s` relies on the ability of the target environment to represent Unicode characters. Seeing escaped unicode characters in exception messages and log entries is not so nice, but it always works. Given that exception messages and log entries are mostly for error determination, it is probably more useful to avoid secondary Unicode conversion issues even at the price of not so nicely readable text.

More investigation is needed to figure out the exact behavior of `%s` vs `%r` in the various environments.

andy-maier:
**Note: This PR is on top of PR #1403.**
For details, see the commit message (of the last commit).
Ready for review and merge.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -355,6 +355,13 @@ This version contains all fixes up to pywbem 0.12.4.
   is set to the built-in default namespace "root/cimv2", instead. Previously,
   that was done only when not specifying the `default_namespace` argument.
 
+* All exception and warning messages produced by pywbem now are guaranteed to
+  contain only ASCII characters. Unicode characters in the messages are
+  represented using an escape syntax such as `\\uXXXX` or `\\U00XXXXXX`.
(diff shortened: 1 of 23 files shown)


KSchopmeyer:
I could not find the changes.rst change to cover this.

Finally, does this mean that we really want to force all prints in the future to use format???

andy-maier:
I added a change log entry.

andy-maier:
As discussed in the call today, we would use new-style formatting for all future use. And also convert the old-style formatting to new-style in testsuite.

PR #1406, opened 2018-09-28
Fixes #1398: Clarified how to determine scoping class/path if not specified in profile

### LINKED ISSUE #1398: Describe how to identify scoping path in profiles
WBEMServer methods build on the concept of a scoping path, but only a minority of the mgmt profiles actually define it.

andy-maier:
Ready for review and merge.

COMMENT: Karl is sitting on approval for thisuntil we get issue #1411 sorted out.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -323,6 +323,119 @@ This documentation uses a few special terms to refer to Python types:
       collection. Therefore, a caller must be careful to not accidentally
       modify the provided :class:`~pywbem.CIMQualifier` objects.
 
+
+.. _`Profile advertisement methodologies`:
+
+Profile advertisement methodologies
+-----------------------------------
+
+This section briefly explains the profile advertisement methodologies defined
+by DMTF. A full description can be found in :term:`DSP1033`.
+
+These methodologies describe how a client can discover the central instances
+of a management profile. Discovering the central instances through a management
+profile is the recommended approach for clients, over simply enumerating a CIM
+class of choice. The reason is that this approach enables clients to work
+seamlessly with different server implementations, even when they have
+implemented a different set of management profiles.
+
+DMTF defines three profile advertisement methodologies in :term:`DSP1033`:
+
+* GetCentralInstances methodology (new in :term:`DSP1033` 1.1)
+* Central class methodology
+* Scoping class methodology
+
+At this point, the GetCentralInstances methodology has not widely been
+implemented, but pywbem supports it nevertheless.
+
+All three profile advertisement methodologies start from the
+`CIM_RegisteredProfile` instance that identifies the management profile, by
+means of registered organisation, registered name, and registered version.
+
+It is important to understand that the `CIM_RegisteredProfile` instance not
+only identifies the management profile, but represents a particular use of the
+management profile within its scoping profiles. For an autonomous profile,
+there are no scoping profiles, so in that case, there is only one use of the
+autonomous profile in a server. However, component profiles do have scoping
+profiles, and it is well possible that a component profile is used multiple
+times in a server, in different scoping contexts. If that is the case, and if
+discovery of central instances using any of the profile advertisement
+methodologies is supposed to work, then each such use of the profile needs to
+have its own separate `CIM_RegisteredProfile` instance, because each such
+use of the profile will also have its own separate set of central instances.
+
+Unfortunately, neither the DMTF standards nor the SMI-S standards are clear
+about that requirement, and so there are plenty of implementations that
+share a single `CIM_RegisteredProfile` instance identifying a particular
+component profile, for multiple distinct uses of the profile by its scoping
+profiles. In such a case, the profile advertisement methodologies will
+not be able to distinguish the distinct sets of central instances alone,
+and other means need to be used to distinguish them.
+
+It is also important to understand that the choice which profile advertisement
+methodology to implement, is done by the WBEM server side. Therefore, a WBEM
+client such as pywbem needs to support all methodologies and needs to try them
+one by one until one succeeds. Pywbem tries the three methodologies in the
+order listed above.
+
+In the *GetCentralInstances methodology*, the `CIM_RegisteredProfile` instance
+has a CIM method named `GetCentralInstances` that returns the instance paths
+of the central instances of the use of the profile.
+
+In the *central class methodology*, the `CIM_RegisteredProfile` instance
+is associated directly with the set of central instances of the use of the
+profile, via a `CIM_ElementConformsToProfile` association.
+
+In the *scoping class methodology*, the `CIM_RegisteredProfile` instance
+is not associated directly with the set of central instances of the use of the
+profile, but delegates that to its scoping profile.
+The client navigates up to the `CIM_RegisteredProfile` instance representing
+the (use of the) scoping profile, looks up its central instances, and
+from each of those, navigates down along the reversed scoping path to the
+central instances of the profile in question. The scoping path of a component
+profile describes the traversal across associations and ordinary classes from
+the central class to the scoping class of the profile. This profile
+advertisement methodology is obviously the most complex one of the three.
+
+Pywbem encapsulates the complexity and choice of these methodologies into
+a single invocation of an easy-to use method
+:meth:`~pywbem.WBEMServer.get_central_instances`.
+
+Profile implementations in a WBEM server are not entirely free when making a
+choice of which methodology to implement:
+
+* Autonomous profiles in a WBEM server must implement the central class
+  methodology, and may in addition implement the new GetCentralInstances
+  methodology.
+
+  Note that the scoping class methodology falls together with the
+  central class methodology for autonomous profiles, because their scoping
+  class is also their central class.
+
+* Component profiles in a WBEM server may implement the central class
+  methodology and the new GetCentralInstances methodology, and must support the
+  scoping class methodology.
+
+  Note that implementing the scoping class methodology in a WBEM server
+  requires implementing the classes and associations of the scoping path,
+  which are usually mandatory anyway. So while the scoping class methodology
+  is more complex to use for clients than the central class methodology, it is
+  easier to implement for servers.
+
+Use of the scoping class methodology by a client requires knowing the central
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
Andreas, I want to talk this one through with Mike Walker on Thursday.

andy-maier:
In addition, it would be good to test the function on an implementation of the scoping path.

andy-maier:
This is suspended until we have clarity about the profile registration implementations.

andy-maier:
We now have sufficient clarity on the profile registration implementations to continue with this. I have updated the PR accordingly and it is up for review again.

pywbem/_server.py
```diff
@@ -735,31 +735,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
-        two more items: The class name of the near end of that hop, and
-        the class name of the traversed association.
-        As a result, the class names of the central instances and scoping
-        instances are not part of the list.
-
-        Example for a 1-hop traversal:
+        All three methodologies start from the CIM_RegisteredProfile instance
+        referenced by the `profile_path` parameter.
+
+        Autonomous profiles implement the GetCentralInstances methodology
```
KSchopmeyer: Is this True,  I thought autonomous profiles must implement the Central instance methodology.  Or is this simply that the scoping methodology requires that autonomous profiles do the ECTP.  In any case the phrasing methodology vs. defining exactly what is implemented implies that the reader has knowledge of the PRP and we should probably not assume that for clients.

pywbem/_server.py
```diff
@@ -735,31 +735,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+        and/or the central class methodology.
+        Component profiles implement any one or more of these methodologies.
+
+        The scoping class methodology requires specifying the central class,
```
KSchopmeyer: Maybe this clarifies my statement above.

pywbem/_server.py
```diff
@@ -735,31 +735,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+        scoping class and scoping path defined by the management profile;
+        the other methodologies do not require these parameters.
+
+        :term:`DSP1001` requires that all management profiles specify a
+        central class. In addition, component profiles are required to specify
+        scoping class and scoping path. The scoping path is a traversal path
```
KSchopmeyer: What does traversal path mean?  Remember I am assuming that readers of this do not want to read the PRP

pywbem/_server.py
```diff
@@ -735,31 +735,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+        from the central class to the scoping class.
+
+        Older component profiles may not follow the requirement to specify
+        scoping class and scoping path. In that case, the scoping class and
+        scoping path can often be determined from the class diagram. Most of
```
KSchopmeyer: Should probably say class diagram in the specification for the profile.

pywbem/_server.py
KSchopmeyer: Or possibly  just state that implement the scoping profile and are required to have an ECTP to central instances

andy-maier: It is true that autonomous profiles must implement the central class methodology. The GetCentralInstances methodology is an optional addition. I.e. an autonomous profile does not have a choice to implement GetCentralInstances and then not to implement the central class methodology. I corrected that.

On your questions: The scoping profile of a component profile can be another component profile. In that case, it again has the usual choices.

pywbem/_server.py
andy-maier: I expanded the description so that it no longer requires reading PRP or PUG, and moved it into the Appendix as a new section "Profile advertisement methodologies".

pywbem/_server.py
andy-maier: I admit that this defines one term by using another one. In the new appendix section, changed that to: "The scoping path of a component profile describes the traversal across associations and other classes from the central class to the scoping class of the profile."

pywbem/_server.py
andy-maier: DONE

andy-maier:
Karl, I reflected all of your comments.

However, it would be good to give it another critical read.

andy-maier:
Karl, on your last comment above: Did you mean to ask for a change with that?

KSchopmeyer:
Sorry for any confusion.  I am happy as it is.  My comment was too general to do anything about.  I did sort of worry about the last statement in the doc

"Most of the time, CIM_System or CIM_ComputerSystem is the scoping class."

True, but we really don't want people guessing at this.  Could we change to Many.  I am getting picky here.  

I think this is really a good overall view.

Before you commit, Let me ask Mike W. if he would look at if you don't mind.

andy-maier:
I changed "Most of the time" to "Many times".

Also, on re-reading the section about what component profiles need to implement, I found the text deserves some improvements. It is now:

------
* Component profiles may implement the central class methodology and the new
  GetCentralInstances methodology, and must support the scoping class
  methodology. Note that the scoping class methodology requires implementing
  the classes and associations of the scoping path, which are usually
  mandatory anyway. So while the scoping class methodology is more complex to
  use for clients than the central class methodology, it is easier to implement
  for servers.
------

See the last commit which is only the delta.

docs/appendix.rst
```diff
@@ -323,6 +323,94 @@ This documentation uses a few special terms to refer to Python types:
...
+
+* GetCentralInstances methodology (new in :term:`DSP1033` 1.1)
+* Central class methodology
+* Scoping class methodology
+
+At this point, the GetCentralInstances methodology has not widely been
+implemented, but pywbem supports it nevertheless.
+
+All three profile advertisement methodologies start from the
+`CIM_RegisteredProfile` instance representing the implementation of the
```
KSchopmeyer: change the word implementation to "definition of a particular defined profile including organization, name, and version"

andy-maier: The RP instance does not represent the profile document (i.e.l the definition), but the profile implementation:
Two implementations -> Two sets of central instances -> Two RP instances required
We discussed this in the last call with Mike.

docs/appendix.rst
```diff
@@ -323,6 +323,94 @@ This documentation uses a few special terms to refer to Python types:
...
+
+* Autonomous profiles must implement the central class methodology, and
+  may in addition implement the new GetCentralInstances methodology.
+  Note that the scoping class methodology falls together with the
+  central class methodology for autonomous profiles, because their scoping
+  class is also their central class.
+
+* Component profiles may implement the central class methodology and the new
+  GetCentralInstances methodology, and must support the scoping class
+  methodology. Note that the scoping class methodology requires implementing
```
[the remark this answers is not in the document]
andy-maier: This paragraph is not about what the user needs to know, but about what he needs to implement, and specifically that the requirement to support the scoping path methodology does not cause any additional requirements.

docs/appendix.rst
```diff
@@ -323,6 +323,94 @@ This documentation uses a few special terms to refer to Python types:
...
+  the classes and associations of the scoping path, which are usually
+  mandatory anyway. So while the scoping class methodology is more complex to
+  use for clients than the central class methodology, it is easier to implement
+  for servers.
+
+Use of the scoping class methodology by a client requires knowing the central
```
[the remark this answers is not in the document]
andy-maier: It only says the same thing after your comment above is applied, but I challenged your comment above.

docs/appendix.rst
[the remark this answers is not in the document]
KSchopmeyer: Ahha.  It is about what the server implements (the classes and associations of the scoping path) and what the client needs to know to navigate that set of associations. So the client must implement the association traversal to navigate this path.  The way I read the original sentence it confused my about who the word implements applies to. Client implements association requests, server implements classes and associations.

docs/appendix.rst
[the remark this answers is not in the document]
KSchopmeyer: I have no argument with leaving this in place.

docs/appendix.rst
KSchopmeyer: Does it not represent both, a) the definition of the org/name/version of the profile and the link to the implementation.

KSchopmeyer:
As we read through those were the points on first read that we got confused and had to go back to be clarify for ourselves what was meant and what the client and server do. 

Hope this clarifies what we noted.

docs/appendix.rst
andy-maier: What do you mean with "definition"? There is definitely no definition of the profile in the sense that CIM instances would represent the profile specification, i.e. like the content of a MRP variant of the profile just in CIM. For example, there is not any representation of a requirement level of any profile element.

I would say the full version of the statement would be that the RP instance represents an implementation of the profile identified by its org/name/version. I can add "identified by its org/name/version" if that helps.

docs/appendix.rst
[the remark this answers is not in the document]
andy-maier: Ok, I'll change the paragraph to make it explicit that this is about the server implementation.

andy-maier:
Karl, I have changed the item about component profiles to make it more explicit that this is about the server side. Please let me know whether you have any remaining comments.

andy-maier:
Karl, I have updated the PR following our discussion in the call today. The last commit shows the delta. Please re-read the whole set of changes in the PR and let me know whether you approve.

pywbem/_server.py
```diff
@@ -121,10 +121,11 @@ class WBEMServer(object):
 
     It supports determining the Interop namespace of the server, all namespaces,
     its brand and version, the advertised management profiles and finally
-    allows to retrieve the central instances of a management profile with
-    one method invocation regardless of whether the profile implementation
-    chose the central or scoping class profile advertisement methodology
-    (see :term:`DSP1033`).
+    allows to retrieve the central instances of an implementation of a
```
KSchopmeyer: Change to retrieve to retrieving

andy-maier: done

andy-maier:
The new flake8 issues probably should be resolved by moving all fixtres into a conftest.py file. See here: https://stackoverflow.com/questions/43746413/how-do-i-get-flake8-to-work-with-f811-errors

PR #1407, opened 2018-09-28
Docs: Removed sphinx warnings; improved keyword args

andy-maier:
Ready for review and merge.

### DIFF
--- a/pywbem/exceptions.py
+++ b/pywbem/exceptions.py
@@ -28,7 +28,9 @@ __all__ = ['Error', 'ConnectionError', 'AuthError', 'HTTPError', 'TimeoutError',
 
 
 class Error(Exception):
-    """Base class for pywbem specific exceptions."""
+    """
+    Base class for pywbem specific exceptions.
+    """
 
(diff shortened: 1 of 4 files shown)


pywbem/exceptions.py
```diff
@@ -75,16 +80,20 @@ def __str__(self):
 
 
 class ConnectionError(Error):
-    """This exception indicates a problem with the connection to the WBEM
-    server. A retry may or may not succeed. Derived from
-    :exc:`~pywbem.Error`."""
+    """
+    This exception indicates a problem with the connection to the WBEM
+    server. A retry may or may not succeed.
+    Derived from :exc:`~pywbem.Error`.
```
KSchopmeyer: Do we really need the derived comment since that is clear from the class definition itself.  No big deal, just trying to remove redundant text.

KSchopmeyer: Same comment couple of times below

andy-maier: I added that because it is not shown in the documentation and in order to have a link that can be followed.

PR #1419, opened 2018-10-12
Fixes #1417: Docs: Added auto summary of methods and attributes

andy-maier:
Ready for review and merge.

PR #1420, opened 2018-10-14
Docs: Improvements for method and attribute lists and other imprs.

andy-maier:
This PR is a follow up to PR #1419 and makes sure that the one-line descriptions that are now in the automatically generated method and attribute tables, are meaningful.

Ready for review and merge.

### DIFF
--- a/pywbem/_nocasedict.py
+++ b/pywbem/_nocasedict.py
@@ -131,7 +131,7 @@ class NocaseDict(object):
         * Passing an OrderedDict or NocaseDict object as a single positional
           argument.
 
-        A UserWarning will be issued if the provided constructor arguments will
+        A UserWarning will be issued if the provided init parameters will
         cause the order of provided items not to be preserved when adding them
         to the new dictionary.
         """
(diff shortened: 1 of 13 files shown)


andy-maier:
I did this in order to have a clickable reference. I would even have created a link to the specific parameter, but don't know how that could be done. At least, the class name gets you very close to the description of the init parameters.

Of course, we could have the link text be "this class", if that helps.

andy-maier:
I did some research, and in most cases, we have:
* references to attributes in descriptions of init parameters: "... same-named attribute in the CIMClassName object ..."
* references to init parameters in descriptions of attributes: "... same-named constructor parameter ..."
The case you found is actually an exception from this pattern, I think (but did not fully verify).

Because attributes can be referenced directly, we could:
* change the first form to become: "... the XXX attribute ..."
* make sure that same-named init parameters are consistently referenced, either as "... same-named constructor parameter ..." or "... same-named init parameter ...". I like the latter form better.

andy-maier:
In the call today, we agreed on the following:
* For references to attributes in the same class: "... the [xxx] attribute ...", where [xxx] is a link to the attribute.
* For references to init parameters in the same class: "... same-named init parameter in [this class] ...", where [this class] is a link to the current class.

PR #1428, opened 2018-10-18
Improved exception messages in WBEMServer.get_central_instances()

andy-maier:
Ready for review and merge.


andy-maier:
Discussed this in our call today.

PR #1431, opened 2018-10-19
Fixed minimum package version of Sphinx and sphinx-git

andy-maier:
This fixes the recent build breaks in the manual-ci-run branch for minimum package versions.
Reason was Sphinx version too low.

Ready for review and merge.


KSchopmeyer:
Approved

Issue #1438, opened 2018-11-14
get_central_instances returns SAXParseexception


KSchopmeyer:
We receive the following response from at least one SMI server indicating that it has a problem processing a request for a method that does not exist in the server.

EMC3: PROFILE get_central_class: SNIA:Masking and Mapping   inst.path=interop:ECOM_RegisteredSubProfile.InstanceID="Masking and Mapping+1.3.0"

central_class=CIM_ControllerConfigurationService,scoping_class=CIM_ComputerSystem, scoping_path=['CIM_HostedService']
pywbem.Error: EMC3: SNIA:Masking and Mapping:1.3.0
inst_path=interop:ECOM_RegisteredSubProfile.InstanceID="Masking and Mapping+1.3.0"
server=WBEMServer(conn.url='https://10.1.134.185', brand='EMC CIM Server', version='2.7.3.10.0.15', ... )
Exception=SAXParseException raised when parsing CIM-XML response: :3:78: not well-formed (invalid token), Connection id: None

::

    LAST_REQUEST
    <CIM CIMVERSION="2.0" DTDVERSION="2.0">
      <MESSAGE ID="1001" PROTOCOLVERSION="1.0">
        <SIMPLEREQ>
         <METHODCALL NAME="GetCentralInstances">
            <LOCALINSTANCEPATH>
              <LOCALNAMESPACEPATH>
                <NAMESPACE NAME="interop"/>
              </LOCALNAMESPACEPATH>
             <INSTANCENAME CLASSNAME="ECOM_RegisteredSubProfile">
               <KEYBINDING NAME="InstanceID">
                 <KEYVALUE VALUETYPE="string">Masking and Mapping+1.3.0</KEYVALUE>
               </KEYBINDING>
             </INSTANCENAME>
            </LOCALINSTANCEPATH>
         </METHODCALL>
       </SIMPLEREQ>
     </MESSAGE>
    </CIM>

    LAST_REPLY
    None

andy-maier:
PR #1439 improves the exception message, so that the faulty XML string is included.

Please run again with that PR.

andy-maier:
SAXParseError "not well formed" is raised when the start element misses its trailing angle bracket, as in:

```
<ELEM</ELEM>
```

This case is now part of the new testcases added by PR #1439. However, I'm not sure this is exhaustive yet, so there may be other cases that also produce the same symptom.

KSchopmeyer:
[...]

PR #1439, opened 2018-11-15
Improved error info in ParseError raised by tupletree parser & testcases

andy-maier:
Ready for review and merge.

This PR improves the ParseError Karl experienced in the SMI-S testlab (see issue #1438).


andy-maier:
Karl, as discussed, please use this PR to re-run your tests.

KSchopmeyer:
It passed my tests and we found the problem with the server that started this. My only suggestion is that if the xml is short why not bypass the code to get a line and display it all.

andy-maier:
What is not optimal in the current code is that the exact position reported for the issue cannot easily be seen in the XML string. That will be even more difficult when multiple lines are shown. CIM-XML strings that are all in one line make that one line very long, which increases the complexity to display this in a reasonable manner.

In any case, I think we need to mark the reported position in the output. That suggests to me that we need to display the XML string in one line and a marker on the reported column position in a second line below it, like some compilers do.

For deciding how much of the XML string to show,  I can think of multiple approaches, but I think an important criteria is that the string should include to the left of the reported position at least the begin of the XML element containing the reported position. Not sure about a criteria for how much to include after the reported position, because for example if the XML is not well-formed it is difficult to find the matching end element.

That's how I arrived at the max 100 chars before and after approach. Given the structure of CIM-XML, it is very likely that the begin of the XML element is within the next 100 chars before the reported position. Exceptions are for example issues within embedded objects, within text of description qualifiers, or within long string property values.

andy-maier:
@KSchopmeyer Karl, I have extended the truncation from 100 to 500 chars before and after the indicated column position, or to 1000 chars if no position can be found. This should do most of what you proposed. In addition, there is now a marker line pointing to the failure position within the line. The marker is adjusted in case of left truncations. Last but not least, non-ASCII characters in the XML string are now handled; they are shown in their UTF-8 encoding where non-ASCII bytes are shown using the `\xNN` notation. If that happens, the marker position will not account for the extra characters, but that is only a small issue for a human reader.

Please run the tests again to see whether it is now better.

KSchopmeyer:
Reran test.  Results of the particular failure are:

::

    EMC3: PROFILE get_central_class: SNIA:Software   inst.path=interop:ECOM_RegisteredSubProfile.InstanceID="Software+1.6.0"
   central_class=CIM_SoftwareIdentity,scoping_class=CIM_System, scoping_path=['CIM_InstalledSoftwareIdentity']
    GET_CENTRAL_INSTANCES Exception ParseError:XML parsing error encountered in CIM-XML response: <unknown>:3:78: not well-formed (invalid token)
    Line 3 of XML string (as binary UTF-8 string):
'<SIMPLERSP><METHODRESPONSE NAME="GetCentralInstances"><RETURNVALUE PARAMTYPE="<ERROR CODE="1" DESCRIPTION="A general error occurred that is not covered by a more specific error code. (Unhandled Exception)" />'
                                                                              ^
    , Connection id: None


    LAST_REQUEST
        <SIMPLEREQ>
          <METHODCALL NAME="GetCentralInstances">
            <LOCALINSTANCEPATH>
              <LOCALNAMESPACEPATH>
                <NAMESPACE NAME="interop"/>
              </LOCALNAMESPACEPATH>
              <INSTANCENAME CLASSNAME="ECOM_RegisteredProfile">
                <KEYBINDING NAME="InstanceID">
                  <KEYVALUE VALUETYPE="string">Array+1.6.0</KEYVALUE>
                </KEYBINDING>
              </INSTANCENAME>
            </LOCALINSTANCEPATH>
          </METHODCALL>
        </SIMPLEREQ>
      </MESSAGE>
    </CIM>

    LAST_REPLY
   None

KSchopmeyer:
The full response  (copied from the issue where I patched out the line selector component) is:

::

    GET_CENTRAL_INSTANCES Exception ParseError:XML parsing error encountered in CIM-XML response: <unknown>:3:78: not well-formed (invalid token), XML string (line 3, column 78): b'<?xml version="1.0" encoding="utf-8" ?>\n<CIM CIMVERSION="2.0" DTDVERSION="2.0"><MESSAGE ID="1001" PROTOCOLVERSION="1.0">\n<SIMPLERSP><METHODRESPONSE NAME="GetCentralInstances"><RETURNVALUE PARAMTYPE="<ERROR CODE="1" DESCRIPTION="A general error occurred that is not covered by a more specific error code. (Unhandled Exception)" />\n</METHODRESPONSE></SIMPLERSP></MESSAGE></CIM>\n\n', Connection id: None

KSchopmeyer:
In this case the single line covers the issue and in a exception response this is what we want, small.

However, this might leave the user confused if the issue is complex or part of the overall context of the response.

Would it make sense to try to put the bad response into the last_response so that if needed the user could capture it all????. Right now the last_response is None.

For this issue, however, I am marking this one as approved by me.

andy-maier:
1. There still seems to be an error with a missing newline after `Line 3 of XML string (as binary UTF-8 string):`. Need to double check.
2. Putting the response into last_response is a great idea. Will have a look to see whether that is possible.

andy-maier:
On 1. I double checked the code and the newline is there. Is it possible that the newline got lost during copy&paste?

andy-maier:
On 2. If receiving the CIM-XML response data succeeded, the `last_raw_reply` property is immediately set to the bytes received. Then, the CIM-XML parsing takes place, and if that succeeds, the `last_reply` property is set to a pretty version of the CIM-XML. In other words, in case of any ParseError exceptions, the `last_raw_reply` property has the response.

andy-maier:
TODO: Add column information to exception message.

andy-maier:
The last commit adds the column information.

PR #1533, opened 2018-11-21
[WIP] Fixes #1411: Fix get_central_instance and add direction method

### LINKED ISSUE #1411: Antecedent/Dependent issue for CIM_RegisteredProfile issue with SMI profiles
SMI registered profiles provide a unique error in that they do not comply with DMTF in the use of the CIM_ReferencedProfile because a number of the current implementation available actually implement the class  xxx a aubclass of CIM_ReferencedProfile which which confuxed the use of Antecedent and Dependent roles.  Thus, any particular smi provider can have the roles of these properties reveresed making the definitions of things like getting autonomous profiles in the DMTF Profile usage guide impossible to use directly.

Today we do not directly include the ability to view autonomous/component profiles or other algoritims that use ReferencedProfile in pywbem. It is not a direct part of getCentralClass or get registeredprofiles.

Should we include algorithms for this and if so, should it include the code to figure out this know issue with the confused dependent and antecedent roles in pywbem or leave to higher level code.

KSchopmeyer:
This is a WIP.  Code in place.  Testing against SMI servers. Unit test in process but hit problem documented in #1540 (qualifier override in mocker). Created bypass to that problem for our tests.  I have to fix that problem in parallel with this code

 There are 3 servers in the test group that do not test correctly (NetAPP, NetAPP1, EMC1) but they may have real issues with the structure of the CIM_ReferencedProfiles.

See #1538 for issues found running against SMI servers with this code, rebase from #1538 and create_instances.py script which is in the try directory.  NOTE: The test can only be run in a closed environment so the server definitions are not included.

Finally, Extended unit tests to provide test of get_central_instances AND determine_reference_direction (That work now tests the central_class methodology and tests being added for scoping class methodology and determine_reference_direction.


andy-maier:
Some initial comments:

1. Make the new parameter `dmtf_reference_direction` of `get_central_instances()` consistent with the new function `determine_cimrefrence_direction()`, so that they are named consistently, and both represent snia vs dmtf in the same way (preferrably as a string like `determine_cimrefrence_direction()` returns it.
2. Typo in the function name `determine_cimrefrence_direction()`.
3. I believe the `count_associators()` method is so special that I don't see a value in making that a public method.

andy-maier:
Karl, you should rebase this PR on PR #1538 .

pywbem/_server.py
```diff
@@ -776,6 +806,14 @@ def get_central_instances(self, profile_path, central_class=None,
               :class:`~pywbem.CIMInstanceName`.
         """  # noqa: E501
         # pylint: enable=line-too-long
+
+        if reference_direction != ('dmtf', 'snia'):
```
andy-maier: I don't think this comparison works. What you want, is: 
```
if reference_direction not in ('dmtf', 'snia'):
```

pywbem/_server.py
```diff
@@ -776,6 +806,14 @@ def get_central_instances(self, profile_path, central_class=None,
...
+            ValueError('Direction parameter must be "dmtf" or "snia". '
+                       'Parameter "%s" is invalid.' % reference_direction)
```
andy-maier: Name the parameter correctly, it is named "reference_direction".
Also, use new style formatting.
Third, we have the `_format()` function that makes sure unicode formatting arguments are presented correctly.

Suggestion:
```
ValueError(
    _format("The reference_direction parameter must be 'dmtf' or 'snia', "
            "but is: {0!A}", reference_direction))
```

docs/changes.rst
```diff
@@ -395,6 +395,17 @@ This version contains all fixes up to pywbem 0.12.4.
...
+  `ServerWarning` is raised in cases when the WBEM server exhibits some
+  incorrect behavior that is tolerated by pywbem.
+
+* Added support for tolerating ill-formed XML in the CIM-XML response returned
+  by the server from the attempt to invoke the CIM method GetCentralInstances()
+  inside of `WBEMServer.get_central_instances()`. One server was found to
+  return such ill-formed XML. This now causes pywbem to issue a `ServerWarning`
+  and to continue with the next approach for determining the central instances.
+  See issue #1438.
+
```
andy-maier: An entry in the change log is missing

fixedassoc.py
```diff
@@ -0,0 +1,976 @@
+#!/usr/bin/env python
```
andy-maier: This script should go to the `try` directory, and possibly be consolidated with the other script.

central_instances.py
```diff
@@ -0,0 +1,1250 @@
+#!/usr/bin/env python
```
andy-maier: This script should go to the `try` directory, and possibly be consolidated with the other script.

pywbem/_server.py
```diff
@@ -762,6 +764,34 @@ def get_central_instances(self, profile_path, central_class=None,
             implementation supports only the scoping class methodology.
             `None` will cause the scoping class methodology not to be attempted.
 
+          reference_direction (:term:`string`):
+            Defines the usage of the ResultRole parameter in the association
```
andy-maier: I think it is easier to understand if we describe the difference in terms of the navigation direction across the CIM_ReferencedProfile association, i.e. navigating to the component profile in direction of Antecedent (DMTF) or Dependent (SNIA). The ResultRole parameter of the Associators operation does not need to be mentioned at all that way.

pywbem/_server.py
```diff
@@ -762,6 +764,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+            request that determines the relation between component registered
+            profiles.  The possible values are 'dmtf' and 'snia' the default is
+            'dmtf'. This parameter exists because of issues between DMTF and
+            SNIA in the the definition of CIM_ReferencedProfile and its
+            subclasses.
```
andy-maier: I would not talk about "issues", but about different definitions without judging it.

pywbem/_server.py
```diff
@@ -762,6 +764,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+
+            Generally the DMTF definition documented in DSP1033 and the
+            DMTF PRP (DSP1001) is that:
+
+            * Antecedent = referenced profile = component profile
+            * Dependent = referencing profile = autonomous profile
+
+            At the same time, for historical reasons, the SNIA SMI specification
+            is ambiguous in that the definition varies by document.
```
andy-maier: The SMI-S specs have only a minor inconsistency between the various pieces, which is that Figure 56 in the SMI-S PRP uses Antecedent / Dependent opposite to all other places. That can well be considered an error in an example diagram, i.e. in a non-normative part of the spec. Otherwise, everything else is consistent in itself, but opposite to DMTF.
Bottom line, I would not say it is "ambiguous", or "varies by document", but instead simply that the definition is different from DMTF. If you want, you can add "except for a non-normative diagram", but I would not mention that.

pywbem/_server.py
```diff
@@ -762,6 +764,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+
+            SMI-S PRP in book xxx specification is that:
```
andy-maier: Don't abbreviate PRP, but use its proper profile name "Profile Registration Profile".
It is in the "Common Profiles" book.

pywbem/_server.py
```diff
@@ -762,6 +764,34 @@ def get_central_instances(self, profile_path, central_class=None,
...
+
+            * Antecedent = autonomous profile
+            * Dependent = component (= sub) profile
+
+            To aid the user in defining this direction, the method
+            :meth:`~py:WBEMServer.determine_cimrefrence_direction` attempts to
```
andy-maier: Typo in `determine_cimrefrence_direction()`.
Also, I suggest to make the method name consistent with the parameter name of `WBEMServer.__init__()`, i.e. rename it to `determine_reference_direction()`.
Make that change here and on the actual method definition.

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
 
+    def determine_cimrefrence_direction(self, possible_autonomous_profiles=None,
+                                        possible_component_profiles=None):
+        # pylint: disable=line-too-long
+        """
+        Determine the antecedent/dependent direction for the
+        CIM_ReferencedProfile class for this server by testing possible
+        autonomous top level profiles for their direction. This method may be
+        required because there is an issue between implementations in the usage
+        of the dependent and antecedent reference properties. Typically SNIA
```
andy-maier: DMTF, not SNIA

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
         return total_ci_paths
 
+    def determine_cimrefrence_direction(self, possible_autonomous_profiles=None,
+                                        possible_component_profiles=None):
+        # pylint: disable=line-too-long
+        """
+        Determine the antecedent/dependent direction for the
+        CIM_ReferencedProfile class for this server by testing possible
+        autonomous top level profiles for their direction. This method may be
+        required because there is an issue between implementations in the usage
```
andy-maier: "may be required" to what?
Suggest instead: "may need to be used"

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
+            ValueError: If the test finds inconsistent results. This can include
+            the possible_autonomous_profiles list generating one definition and
+            the possible_component_profiles the opposite definition, or one of
+            the lists concluding that the possible direction could be both
+            'dmtf' and 'snia'because of inconsistencies in the instances of
+            CIM_ReferencedProfile
+        """  # noqa: E501
+        # pylint: enable=line-too-long
+
+        def _determine_type(profilepaths, v0_dict, v1_dict, autonomous):
```
andy-maier: Local function definitions should only be used if there is a reason, i.e. if they use local variables of the outer function. That does not seem to be the case here. So I suggest to make this a classmethod in the class.

pywbem/_server.py
```diff
@@ -1122,3 +1320,65 @@ def _determine_profiles(self):
...
+            # Remove host from responses since the host causes confusion
+            # with the enum of registered profile. Enums do not return host but
+            # the associator properties contain host
+            for ref_inst in ref_insts:
+                for prop in ref_inst.values():
+                    prop.host = None
+
+        except Error as er:
+            print('CIM_ReferencedProfile failed for server=%s exception=%s'
+                  % (self, er))
```
andy-maier: Use new style formatting

pywbem/_server.py
```diff
@@ -1122,3 +1320,65 @@ def _determine_profiles(self):
...
+                "CIM_ReferencedProfile", namespace=self.interop_ns)
+            # Remove host from responses since the host causes confusion
+            # with the enum of registered profile. Enums do not return host but
+            # the associator properties contain host
+            for ref_inst in ref_insts:
+                for prop in ref_inst.values():
+                    prop.host = None
+
+        except Error as er:
+            print('CIM_ReferencedProfile failed for server=%s exception=%s'
```
andy-maier: What about that class failed?

andy-maier: I don't think expanding the whole WBEMServer object is necessary in this message.
Second, the exception expands to some human readable text, so I don't think it should be behind a "=" sign.

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
+                return auto_dir_type
+            elif comp_dir_type:
+                return comp_dir_type
+
+        raise ValueError('Server: %s; Cannot determine '
+                         'CIM_ReferencedProfile direction. '
+                         'Autonomous and component tests do not match. '
+                         'auto_dir_type=%s, '
+                         'comp_dir_type=%s; ' %
+                         (self, auto_dir_type, comp_dir_type))
```
andy-maier: Use new style formatting.

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
+                dir_type = t[0]
+            elif v1_paths and not v0_paths:
+                dir_type = t[1]
+            else:
+                ps = 'possible %s' % ('autonomous' if autonomous else
+                                      'component')
+                raise ValueError("Cannot determine type. "
+                                 "determine_cimrefrence_direction shows "
+                                 "conflicts in %s. %s=%s %s=%s" %
+                                 (ps, t[0], v0_paths, t[1], v1_paths))
```
andy-maier: Use new style formatting.

docs/changes.rst
KSchopmeyer: log entry added

pywbem/_server.py
KSchopmeyer: fixed

pywbem/_server.py
KSchopmeyer: That is what I get for testing against the version in the test scripts. Fixed

docs/changes.rst
KSchopmeyer: Note that I added it under enhancements rather that bugs

pywbem/_server.py
KSchopmeyer: DONE

pywbem/_server.py
KSchopmeyer: good catch. Fixed

pywbem/_server.py
KSchopmeyer: DONE

pywbem/_server.py
KSchopmeyer: DONE

pywbem/_server.py
KSchopmeyer: Done

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
+            Dependent = component (= sub) profile
+
+        This method processes the CIM_RegisteredProfiles associatorNames
+        operation with ResultRole set to `Dependent` and then to `Antecedent` to
+        find any profile returns associations for one of the ResultRole
+        settings but not the other. If such an association can be found, that
+        determines the implementation of the antecedent/dependent direction by
+        matching the returns against possible_autonomous_profiles and
+        possible_component_profiles.  Either of the lists or both may be
+        provided. At least one of the lists must exist.
```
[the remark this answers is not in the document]
KSchopmeyer: Removed the last sentence

pywbem/_server.py
KSchopmeyer: made it staticmethod since it makes no use of anythin in the class.

pywbem/_server.py
KSchopmeyer: Good call, use the phrase 'navigation direction'

pywbem/_server.py
KSchopmeyer: made consistent and remove all use of cimreference in favor of reference

pywbem/_server.py
```diff
@@ -885,6 +943,146 @@ def get_central_instances(self, profile_path, central_class=None,
...
+            Dependent = referencing profile = autonomous profile
+
+        and many SMI implementations defined them with:
+
+            Antecedent = autonomous profile
+            Dependent = component (= sub) profile
+
+        This method processes the CIM_RegisteredProfiles associatorNames
+        operation with ResultRole set to `Dependent` and then to `Antecedent` to
+        find any profile returns associations for one of the ResultRole
```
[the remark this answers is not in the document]
KSchopmeyer: Rewrote the whole paragraph.

pywbem/_server.py
KSchopmeyer: I am uncomfortable with this one.  Possibilities are:

1. No try block, just let the error flow upward if CIMReferencedProfile enumerate results in exception. That means that there is no possibility of scoping information.

2. Use try block and try to pick out the types of error. Generally it could be 
    a. Class failed (that means no concept of scoping, etc.
    b. Enumeration returns NOTSUPPORTED. means that enumeration of this association is not supported.
    c. Other errors. Total failure somewhere.

However, the result is:
    a. Any error other than NOTSUPPORTED. complete failure
    b. NOTSUPPRORTED - we could go back to determining from executing associations which takes 10+++ times as long. (minutes rather than seconds)

NOTE: All the tested servers do return enumeration.

KSchopmeyer: TODO. See comment above about this whole exception.

KSchopmeyer: DISCUSSION ITEM

pywbem/_server.py
andy-maier: sorry, yes, that‘s what I meant

pywbem/_server.py
KSchopmeyer: DONE

pywbem/_server.py
KSchopmeyer: DONE

pywbem/_server.py
KSchopmeyer: DONE

andy-maier:
**Discussion item**: There is a discussion item, further up (search for DISCUSSION ITEM).

pywbem/_server.py
KSchopmeyer: We agreed to do the exception and put more context into the error here so that the user has overall operation context in the error.

pywbem/_server.py
KSchopmeyer: See above.

pywbem/_server.py
andy-maier: I want to implement a ServerError class for that.

fixedassoc.py
KSchopmeyer: Whoops put into the tools directoyr

KSchopmeyer: Fixed

pywbem/_server.py
KSchopmeyer: NOTE: This one is NOT done yet.  I need to change the message to reflect what might have happened.

KSchopmeyer:
Current test output, comparing the 3 techniques indicates that we have some differences I need to explore.

Three servers (NETAPP1, NetAPP, EMC1) do not respond with the same conclusion from the 3 tests and 11 servers respond with a match for the 3 techniques.

 ack "DETERMINE_DIRECTION ALGORITHMS" central_instancesfull.txt 
Server Brocade: DETERMINE_DIRECTION ALGORITHMS match
Server Hewlett_Packard3: DETERMINE_DIRECTION ALGORITHMS match
Server EMC3: DETERMINE_DIRECTION ALGORITHMS match
Server NetApp1: ERROR: DETERMINE_DIRECTION ALGORITHMS differ slow=snia fast=None prod=None
Server Brocade1: DETERMINE_DIRECTION ALGORITHMS match
Server Fujitsu2: DETERMINE_DIRECTION ALGORITHMS match
Server Dell/Compellent: DETERMINE_DIRECTION ALGORITHMS match
Server Fujitsu: DETERMINE_DIRECTION ALGORITHMS match
Server NetApp: ERROR: DETERMINE_DIRECTION ALGORITHMS differ slow=snia fast=None prod=None
Server EMC4: DETERMINE_DIRECTION ALGORITHMS match
Server Hewlett_Packard: DETERMINE_DIRECTION ALGORITHMS match
Server Fujitsu1: DETERMINE_DIRECTION ALGORITHMS match
Server Hewlett_Packard5: DETERMINE_DIRECTION ALGORITHMS match
Server EMC1: ERROR: DETERMINE_DIRECTION ALGORITHMS differ slow=snia fast=None prod=None

pywbem/_server.py
KSchopmeyer: print statement completely removed.

andy-maier:
Karl, I have split off the addition of the `reference_direction` parameter into a new PR #1556.
Also, I don't think we need the files:
* `fixedassoc.py` - This is investigation-only
* `try/central_instances.py` - This is investigation-only
* `try/profiledictionary.json` - The profile definitions are now covered in PR #1550.

My suggestion is to remove these three files from this PR, and to rebase this PR on top of PR #1556. 

That will reduce this PR to contain just the WBEMServer methods to automatically determine the reference direction.

andy-maier:
Update to my previous comment: As discussed on Slack, Karl wants to keep the central_instances.py script. So I suggest to keep this in the try directory, change it to use the profiles.json file introduced with PR #1550, and to have that script in a new PR.

KSchopmeyer:
We discussed this and now think that the direction can be determined simply by testing the top level wrapper registered profile that represents the initiative.  The current implementation was way to complex.
Therefore, there is no need for this code any more.  We will close this pr as Pursuing different issues.

Andy started a new issue #1617 to to clarify how the user uses this profile to confirm direction.

KSchopmeyer:
Closing

PR #1542, opened 2018-12-01
Fixes #1430: Added Interop namespace to set of all namespaces in WBEMServer, if missing

### LINKED ISSUE #1430: Testing with smi servers, issue with server namespaces
One server returns an interop namespace for interop_ns but does not include the interop namespace in the return for general namespaces.  I am not sure exactly what is returned from the server but  this is what shows up in the result use of the properties.

server.interop_ns      server.namespaces
interop                     /root/eternus

i.e the  interop does not show up in the list displayed for server.namespaces.

TODO: KS, Display exactly what the server returns.

Tests with a number of other servers show correct results.

andy-maier:
Ready for review and merge.

**How to test against a real server:**
* No other commits needed on top of the branch of `andy/add-interop-to-namespaces`
* Specifically, PR #1550 is not needed for this test.
* Have `testsuite/end2end/server_file.yml` set up so that the default server is the server nicknamed "Fujitsu" that had the behavior of not including the Interop namespace in the enumeration result.
* Issue:
  ```
  make end2end
  grep ToleratedServerIssueWarning end2end_*.log
  ```
* Verify that the make succeeds and that there is one occurrence of `ToleratedServerIssueWarning`.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -101,7 +101,7 @@ from .cim_constants import CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_CLASS, \
     CIM_ERR_METHOD_NOT_FOUND, CIM_ERR_METHOD_NOT_AVAILABLE, \
     CIM_ERR_NOT_SUPPORTED, CIM_ERR_NOT_FOUND, CIM_ERR_FAILED, \
     CIM_ERR_NAMESPACE_NOT_EMPTY
-from .exceptions import CIMError, ParseError
+from .exceptions import CIMError, ParseError, ModelError
 from ._warnings import ToleratedServerIssueWarning
 from ._nocasedict import NocaseDict
 from .cim_obj import CIMInstanceName, CIMInstance
@@ -237,6 +237,7 @@ class WBEMServer(object):
         Raises:
 
             Exceptions raised by :class:`~pywbem.WBEMConnection`.
+            ModelError: An issue with the model implemented by the WBEM server.
             CIMError: CIM_ERR_NOT_FOUND, Interop namespace could not be
               determined.
             CIMError: CIM_ERR_NOT_FOUND, Namespace class could not be
@@ -255,6 +256,7 @@ class WBEMServer(object):
         Raises:
 
             Exceptions raised by :class:`~pywbem.WBEMConnection`.
+            ModelError: An issue with the model implemented by the WBEM server.
             CIMError: CIM_ERR_NOT_FOUND, Interop namespace could not be
               determined.
             CIMError: CIM_ERR_NOT_FOUND, Namespace class could not be
@@ -268,11 +270,13 @@ class WBEMServer(object):
     def namespace_paths(self):
         """
         :class:`py:list` of :class:`~pywbem.CIMInstanceName`: Instance paths
-        of all namespaces of the WBEM server.
+        of the CIM instances in the Interop namespace that represent the
+        namespaces of the WBEM server.
 
         Raises:
 
             Exceptions raised by :class:`~pywbem.WBEMConnection`.
+            ModelError: An issue with the model implemented by the WBEM server.
             CIMError: CIM_ERR_NOT_FOUND, Interop namespace could not be
               determined.
             CIMError: CIM_ERR_NOT_FOUND, Namespace class could not be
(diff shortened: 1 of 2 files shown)


pywbem/_server.py
```diff
@@ -1024,6 +1029,44 @@ class variable.
...
+                            self.conn.url, self.interop_ns),
+                    ServerWarning, stacklevel=2)
+            else:
+                # Because the different classes used to represent namespaces
+                # have different keys, it seems best to copy one of them.
+                org_path = self._namespace_paths[0]
+                interop_path = org_path.copy()
+                interop_path['Name'] = self.interop_ns
+                try:
+                    self._conn.GetInstance(
```
KSchopmeyer: The namespace input parameter should not be there.  In this case, I believe that the namespace is in the instance_path.

andy-maier:
@KSchopmeyer Karl, I decided that the two situations in which the `namespaces` property could be set, but the `namespace_paths` property could not, deserve raising a `ModelError` instead of just issuing a `ServerWarning`. As a result, I'd like you to run this again against the Hitachi server, to double check that it does not run into any of these situations.

KSchopmeyer:
Here is sample of the grep

andy-maier:
@KSchopmeyer I believe you closed this PR by accident. Reopening it.

pywbem/_server.py
```diff
@@ -1024,6 +1042,49 @@ class variable.
...
+            assert interop_path.namespace is not None
+            try:
+                self._conn.GetInstance(interop_path)
+            except CIMError as exc:
+                raise ModelError(
+                    _format("Server at {0} does not return the Interop "
+                            "namespace {1!A} when enumerating class {2!A}, "
+                            "and adding it to the list of namespaces is not "
+                            "possible because verification of the instance "
+                            "it via GetInstance on path {3!A} fails: {4}",
```
KSchopmeyer: I think the "it via" should be expanded to "it recieved via".

KSchopmeyer:
I make some comments and the end2end test itself was marked failed.  Let me test some more things and I can run more detailed tests on this.  log output and mof for Fujitsu_namespace in shared files.

pywbem/_server.py
andy-maier: Fixed by removing "it".

andy-maier:
The comments have been addressed, and the PR has been rebased on latest master after merging the approved PRs.

@KSchopmeyer Please re-run the end2end test.

andy-maier:
On running the test against Fujitsu vs Hitachi: I mistakenly thought the Hitachi server was the culprit, but as you rightfully indicated, it was the server you nicknamed "Fujitsu". Corrected that in the initial comment.

andy-maier:
I have taken a look at the `test_end2end_34.tmp.log` file in Dropbox, and it raises:
```
pywbem.exceptions.ModelError: Server at https://10.2.104.20 does not return the Interop namespace 'interop' when enumerating class 'CIM_Namespace', and adding it to the list of namespaces is not possible because verification of the instance it via GetInstance on path CIMInstanceName(classname='FUJITSU_Namespace', keybindings=NocaseDict({'SystemCreationClassName': 'FUJITSU_System', 'SystemName': 'FUJITSU:ETERNUS SMI-S Agent', 'ObjectManagerName': 'FUJITSU:ETERNUS SMI-S Agent', 'Name': 'interop', 'CreationClassName': 'FUJITSU_Namespace', 'ObjectManagerCreationClassName': 'FUJITSU_ObjectManager'}), namespace='interop', host=None) fails: 6 (CIM_ERR_NOT_FOUND): CIM_ERR_NOT_FOUND, Connection id: 1-3970, Connection id: None
```

This means the server supports a namespace named `interop` but does not have a CIM instance representing it. So that is consistent with not enumerating it.

I guess this PR needs to be changed to add the interop namespace just to the `namespaces` sttribute, and to give up on trying to establish an instance path. As a result, the `namespace_paths` property would not include it, and we would describe that this is possible.

andy-maier:
The latest commit no longer tries to establish an instance path if the Interop namespace is missing.

andy-maier:
@KSchopmeyer Karl, my latest update and your test seem to have crossed each other. Please make sure you test with the latest update of this PR, as of 10min ago.

andy-maier:
**Discussion point:** Now that we ended with adding the Interop namespace only to the `namespaces` property but not to the `namespace_paths` property, I'd like us to briefly sit back and think about whether we really want to do it that way. Personally, I have no problem with it, because it is now documented that way. Let's quickly discuss in our next call (or here via comments).

andy-maier:
I am merging this PR in order to bring the various pieces together. The discussion point is still open and should be resolved, if needed with subsequent PRs.

andy-maier:
I have created issue https://github.com/pywbem/pywbem/issues/1553 to track resolution of the remaining discussion point.

PR #1543, opened 2018-12-07
[WIP] Fixes #1540: Extend WBEMConnection_mock  class resolution

### LINKED ISSUE #1540: fix issue in mocker createclass
My TODO in mocker createclass bit me.  Code is incomplete and does not propagate key from association superclass to subclass.

KSchopmeyer:
WIP

Status
Currently code finished and passes most of the tests. 

Also, in process of rework code to simplify since it duplicates the same code 3 times to resolve properties, methods, parameters.

### DIFF
--- a/PKG-INFO
+++ b/PKG-INFO
@@ -1,6 +1,6 @@
 Metadata-Version: 1.2
 Name: pywbem
-Version: 0.12.1.dev221
+Version: 0.12.1.dev229
 Summary: pywbem - A WBEM client
 Home-page: https://pywbem.github.io/pywbem/
 Author: Tim Potter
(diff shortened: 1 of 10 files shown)


andy-maier:
As discussed in our call today, I think we should look at the overall picture in pywbem of:
* mock support and how does it correspond to repositories
* MOF compiler and how does it correspond to repositories
* in-memory repositories
* plug interfaces for user-provided repositories
* use of a real WBEM server as a repository

andy-maier:
After discussion in our call today, we decided that the design of this PR should go in, and that we would clean up the design later.

Given that, my comments are reduced to these:
* The commit message should be condensed a lot. It should have a title line that brings it to the point, and a description of what is done and why it is done, not how it is done and what its design is. 
* There are some documentation fixes that are unrelated to the main problem and that should be done separately.
* The prints should be removed.
* The unit test that fails should be fixed (see the error in test_recorder.py).

PR #1544, opened 2018-12-09
Moved new warning classes into separate module and added to docs

andy-maier:
This is a small fixup after merging PR #1539.

Ready for review and merge.

Also, there is the following **discussion point**:

* The warning class has been named `ServerWarning`, but it is defined as: "indicates an issue with the WBEM server that has been tolerated by pywbem". With that definition, a class name of `ToleratedServerIssueWarning` would be more appropriate. Discuss which name to use finally.


andy-maier:
We agreed to rename the `ServerWarning` class to `ToleratedServerIssueWarning`.

PR #1550, opened 2018-12-25
Added profile definitions and more end2end testcases

andy-maier:
This is an attempt to get some of the functionality in Karl's scripts into end2end testcases.

Ready for review and test against a real WBEM server.

How to test:
* Have a `testsuite/end2end/server_file.yml` prepared with the servers to test against
* Run `make end2end` and verify that the tests succeed.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -470,6 +470,11 @@ This version contains all fixes up to pywbem 0.12.4.
   There is an example file `server_file_example.yml`.
   There are some initial tests, and users can define their own tests.
 
+* For the end2end tests, added a file `profiles.yaml` that defines the
+  discovery-related characteristics of a number of DMTF and SNIA management
+  profiles, and that is used to drive profile discovery related tests against
+  WBEM servers.
+
 **Known issues:**
 
 * See `list of open issues`_.
--- a/testsuite/end2end/_utils.py
+++ b/testsuite/end2end/_utils.py
@@ -1,0 +1,22 @@
+"""
+Utility functions for end2end tests.
+"""
+
+from __future__ import absolute_import, print_function
+
+
+def _latest_profile_inst(profile_insts):
+    """
+    Determine the instance with the latest version in a list of registered
+    profile instances and return that instance.
+    """
+    profile_versions = [inst['RegisteredVersion'] for inst in profile_insts]
+    latest_i = None
+    latest_version = [0, 0, 0]
+    for i, profile_version in enumerate(profile_versions):
+        profile_version = [int(v) for v in profile_version.split('.')]
+        if profile_version > latest_version:
+            latest_version = profile_version
+            latest_i = i
+    latest_profile_inst = profile_insts[latest_i]
+    return latest_profile_inst
--- a/testsuite/end2end/profiles.yml
+++ b/testsuite/end2end/profiles.yml
@@ -1,0 +1,860 @@
+# This file defines discovery-related characteristics of management
+# profiles. It is used by pywbem end2end testcases.
+#
+# The format is a list of profile definition items, where each item describes
+# one profile. The order of profile definition items determines the order in
+# which the profiles are tested.
+#
+# Each profile definition item is a dictionary with the following items:
+# * registered_org: Registered organization of the profile (as a string, e.g.
+#   'SNIA').
+# * registered_name: Registered name of the profile.
+# * autonomous: Boolean indicating whether the profile is autonomous.
+# * central_class: Class name of the central class of the profile.
+# * scoping_class: Class name of the scoping class of the profile. May be
+#   null or non-null for autonomous profiles.
+# * scoping_path: Iterable of class names of the scoping path (from central to
+#   scoping, excluding central and scoping classes). Must be null for
+#   autonomous profiles.
+# * doc: Informal reference to the standards document describing the profile.
+#
+# TODO #1554: Add support for representing profile versions.
+# TODO #1552: Update profile definitions from additional input.
+
+# -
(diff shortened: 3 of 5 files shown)


KSchopmeyer:
Ran test against single server Fujitsu (it seems to respond well to central_insts, etc.   It returned

```

[1m============================= test session starts ==============================[0m
platform linux -- Python 3.4.3, pytest-3.8.2, py-1.5.3, pluggy-0.7.1
rootdir: /home/kschopmeyer/pywbem/githubpywbem/pywbem, inifile:
plugins: cov-2.5.1
collected 113 items

test_general_profile.py FFsssssFFFFFsFFFFFFFFFFFFsFsFsssssssFssssssssssssssssssssssssssssssssssssssssssssssssssss
test_server.py F.F...F.F...F.F...F.F...

The actual  log file is about 100k so put it in pywbem.

Here are some early notes:

1. issue of unorderable types:

[1m    def _latest_profile_inst(profile_insts):[0m
[1m        """[0m
[1m        Determine the instance with the latest version in a list of registered[0m
[1m        profile instances and return that instance.[0m
[1m        """[0m
[1m        profile_versions = [inst['RegisteredVersion'].split('.')[0m
[1m                            for inst in profile_insts][0m
[1m        latest_i = None[0m
[1m        latest_version = (0, 0, 0)[0m
[1m        for i, profile_version in enumerate(profile_versions):[0m
[1m>           if profile_version > latest_version:[0m
[31m[1mE           TypeError: unorderable types: list() > tuple()[0m

**Comment by andy-maier** [comment]
Karl, I have fixed that error, and made some more changes:
- Commented out the profiles that have insufficient definition data (e.g. component profiles without central class, scoping class or scoping path), and added an assertion verifying that.
- Fixed the error that central class, scoping class and scoping path of the profile definition had not been passed to get_central_instances().

Please run again.

**Comment by KSchopmeyer** [comment]
Reran test this morning.  This is just  for Fujitsu server since that one generally does fairly well in the other tests.

Overall:

======= 27 failed, 16 passed, 59 skipped, 384 warnings in 134.92 seconds =======

The detailed results are in the Shared directory as a 

test_end2end_34.tmp.log

With the console output as:

more-end2end.txt

I am spending the rest next hours working may way through the get_central_instances for Fujitsu for the ones that fail to determine exactly what the objects in the server look like.

**Comment by andy-maier** [comment]
I have finally got an OP running in a virtual CentOS 7 on my macbook, and with this branch checked out, `make end2end` only fails the test_general_profile.py tests on 3 profiles that are not properly set up in this default installation of OP. Bottom line, the branch should now be good enough to find the actual issues in the SMI testlab servers, instead of falling over its own issues in the pywbem code or test code.

**Comment by KSchopmeyer** [comment]
Since this is our test code, it would be fine with me if we checked it in and fixed issues afterwards.  My only question is about syncing the definitions I have for the central_instances sript in the try directory and your definitions here since I would guess that is going to change as we learn more and it would be better if we changed one place.  Even that we could do post this PR.

**Comment by andy-maier** [comment]
Karl, I have added a TODO to apply changes to the profile definitions as per Karl's and Mike's input, and will merge the PR now.

**Comment by andy-maier** [comment]
Karl, since you said the profiles are now defined in a JSON file, I looked that up in your PR #1533, and I think that is the approach we should use to define the profile characteristics. So I'll change this PR to do that.

**Comment by andy-maier** [comment]
I have updated this PR to use a file `profiles.yaml` in the `testsuite/end2end` directory as the definition of the discovery-related profile characteristics. The content of that file uses the content of the earlier implementation of this PR (i.e. your earlier script in Dropbox, plus my updates from SMI-S), and I verified that it contains everything your `profiledictionary.json` file in PR #1533 has (in fact, it has several more scoping paths, and several corrections of invalid profile names, along with TODOs to clarify that). I used YAML format because that allows for having comments. So that `profiles.yaml` file now is the best version of profile definitions we have, pending any additional information Mike may have.

**Comment by andy-maier** [comment]
@KSchopmeyer Karl, may I ask for your review of the approach with the `profiles.yaml` file, and of course any code review you may want to do.
Then we can merge this PR and rebase your PR #1533 on top of it, including changing your try/central_instances.py script to use the profiles.yaml file.

File: testsuite/end2end/profiles.yaml [hunk 1]
```diff
@@ -0,0 +1,856 @@
...
+-
+  registered_org: SNIA
+  registered_name: Array
+  autonomous: true
+  central_class: CIM_ComputerSystem
+  scoping_class: CIM_ComputerSystem
+  scoping_path: null
+  doc: SMI-S Block
+-
+  # Note: This profile has been deprecated in SMI-S 1.7, and the DMTF
```
KSchopmeyer: Actually that statement is incorrect since SNIA is bypassing 1.7 because is was so full of errors.  The  conclusion for 1.8 will be that either the  WBEM Server or Object manager profiles will be acceptable. This  was required since NOBODY implemented the WBEM Server profile and probably never will

andy-maier: Ok. Needs to be corrected.

andy-maier: Karl: SMI-S 1.8 is out and its Revision 3 still states that the SNIA Server profile is deprecated.

andy-maier: DONE. Changed the respective note to:
```
  # Note: The SNIA Server profile has been deprecated in SMI-S 1.7, listing
  #       the DMTF WBEM Server profile as a replacement. There is an intention
  #       in the SMI-S community to change this again, allowing the SNIA
  #       Server profile and the DMTF WBEM Server profile to be implemented.
  #       However, as of the current SMI-S 1.8 Revision 3, the SNIA Server
  #       profile is still listed as deprecated.
```

PR #1558, opened 2018-12-30
Adds a test script to the try directory.

KSchopmeyer:
Adds a test script currently being used  to test get_central_classes and
the reference difference algorithms to its own branch.

Revamped to use the profile file and server_file from end2end.  I suggest we commit this as it is only a temporary test tool. 

NOTE: This depends on the implementation of get_centralinstances and the difference method defined in pr #1411.

NOTE: This script has been update for the new tests diretory structure and uses the ServerDefinitionFile and ServerDefinition classes from that work.

### DIFF
--- a/try/run_central_instances.py
+++ b/try/run_central_instances.py
@@ -1,0 +1,1987 @@
+#!/usr/bin/env python
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+#
+"""
+This is a script that executes a number of tests on servers defined in
+a file of servers to test the characteristics of the servers themselves
+and the pybem algorithms for the get_central_instances() method. It
+gets the servers for which tests are to be executed from a file, allows the
+user to select particular servers, and then executes all of the defined tests
+and display methods or just those selected on the command line.
+
+Output is sent to stdout.
+
+It is not publically supported but to be used as apywbem project internal test
+tool.
+
+NOTE: Since this is a development support tool, in includes an import that
+is not part of the pywbem install imports (tabulate).  This package must
+be manually  imported for this script to execute correctly
+"""
+from __future__ import absolute_import, print_function
+
+import sys
+import os
+import argparse as _argparse
+from collections import namedtuple
+from collections import defaultdict
+import textwrap
+import threading
+import datetime
+import traceback
+import yaml
+from operator import itemgetter
+try:
+    import tabulate
+except ImportError as ie:
+    raise ImportError('tabulate is not part of pywbem install and must be '
+                      'manually installed: %s' % ie)
+import six
+
+import pywbem
+from pywbem._nocasedict import NocaseDict
+from pywbem._cliutils import SmartFormatter as _SmartFormatter
+from pywbem import __version__
+
+from tests.end2endtest.utils.server_definition_file import \
+    ServerDefinitionFile
+
+VERBOSE = False
+TABLE_FORMAT = 'simple'
+
+SERVER_FILE = 'server_file.yml'
+SERVER_EXAMPLE_FILE = 'server_file_example.yml'
+
+SCRIPT_DIR = os.path.dirname(__file__)
+
+# Named tuple to define Central Class, Scoping Class, ScopingPath.
+ProfileDef = namedtuple('ProfileDef', ['central_class', 'scoping_class',
+                                       'scoping_path', 'type',
+                                       'comments'])
+
+# TODO: Most of these varaiables should be within the main and test servers
+# methods
+# if not empty this is a list of profiles found in the test but not in the
+# all_profiles_dict or where the definitions are incomplete in the
+# all_profiles_dict
+PROFILES_WITH_NO_DEFINITIONS = defaultdict(list)
+
+SERVERS_FOR_PROFILE = defaultdict(list)
+
+# summary info on profiles, etc. for each server. A row is added for each
+# server tested.
+GET_CENTRAL_INST_ROWS = []
+
+# results dictionary for Central instance error.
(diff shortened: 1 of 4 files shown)


try/smi_registered_profile_dictionary.json
```diff
@@ -0,0 +1,694 @@
+{
```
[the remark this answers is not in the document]
KSchopmeyer: Already  done and will check in.  However, I also suggest that we move those definitions from end2end into testsuite or somewhere to make them appear more general for tests.  That would  be another pr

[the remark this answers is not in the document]
KSchopmeyer: Modified to use the yaml file profiles.py.  Should probably also use the servers file but that may be a bit more complex so will  do separately.

andy-maier:
Karl, PR #1606 moves the profiles.yml file up, and adds a manualtest directory. Please consider whether the new run_central_instances.py script should be located in the testsuite/manualtest directory.

andy-maier:
Hi Karl. Not sure what you intended to do with this PR, but my understanding was that you wanted to migrate it to use the profiles.yml file.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+        possible_profiles = [inst.path for inst in possible_profiles]
+    elif output == 'name':
+        possible_profiles = [profile_name(org_vm, inst)
+                             for inst in possible_profiles]
+
+    return possible_profiles
+
+
+def test_reference_direction_fast(nickname, server, all_profiles_dict,
+                                  org_vm, st):
```
andy-maier: `profiles_path` is used as the default. I suggest to name the variable accordingly, consistent with how it is done for the server definition file.

andy-maier: `profile_definitions_filename` does not seem to be used. Please run `make check` on the script.

andy-maier: `servers_path` is used as the default. I suggest to name the variable accordingly.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+        path = obj.copy()
+    else:
+        assert False
+
+    path.host = None
+    path.namespace = None
+    return path
+
+
+class ServerDefinitionFileError(Exception):
```
andy-maier: I don't think the `ServerDefinitionFileError` class should be copied, but instead imported.

Because the script is invoked as a script, its directory does not count as a package even if it contains an `__init__.py` file. So you have to specify the path starting from the main directory:
```
form tests.servers.server_file import ServerDefinitionFileError
```

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+    """
+    An error in the WBEM server definition file.
+    """
+    pass
+
+
+class ServerDefinitionFile(object):
```
andy-maier: I don't think the `ServerDefinitionFile` class should be copied, but instead imported.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+                            sd_list.append(sd)
+            else:
+                raise ValueError(
+                    "Server group or server nickname {0!r} not found in WBEM "
+                    "server definition file {1!r}".
+                    format(nickname, self._filename))
+        return sd_list
+
+
+class ServerDefinition(object):
```
andy-maier: I don't think the `ServerDefinition` class should be copied, but instead imported.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+                self._check_sg(nick, server_groups, servers, visited_sg_nicks)
+            elif nick not in servers:
+                raise ServerDefinitionFileError(
+                    "Item {0!r} in server group {1!r} in WBEM server "
+                    "definition file {2!r} is not a known server or server "
+                    "group".
+                    format(nick, sg_nick, self._filename))
+
+    @property
+    def filename(self):
```
andy-maier: I think the `filename()` method should be added to the version of the class in server_file.py.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+        try:
+            server_dict = self._servers[nickname]
+        except KeyError:
+            raise ValueError(
+                "Server nickname {0!r} not found in WBEM server definition "
+                "file {1!r}".
+                format(nickname, self._filename))
+        return ServerDefinition(nickname, server_dict)
+
+    def list_all_servers(self):
```
andy-maier: I think I should add the `list_all_servers()` method to the version of the class in server_file.py.

try/run_central_instances.py
```diff
@@ -0,0 +1,2341 @@
...
+        Boolean controlling whether the returned server certificate is to be
+        verified.
+        """
+        return self._no_verification
+
+    @property
+    def default_namespace(self):
+        """
+        Boolean controlling whether the returned server certificate is to be
+        verified.
```
andy-maier: The description does not fit the attribute name.

try/run_central_instances.py
KSchopmeyer: Done

try/run_central_instances.py
KSchopmeyer: Fixed.  however, In the lastest code I changed to use the same class in the test code and added this property to that class.

try/run_central_instances.py
KSchopmeyer: Done.  Note that I added the default_namespace to this class.  We had discussed this and agreed that there was really no need for it in the tests.  However, if this class is to be more general, there may be a need for this as an optional property simply because it it a part of WBEMConnection and useful in writing other scripts using the ServerDefinition object.

try/run_central_instances.py
KSchopmeyer: Done.  In fact, I used all of that class except the list_servers since the functionality and interface is slightly different.

try/run_central_instances.py
KSchopmeyer: Dropped completely since we use the ServerDefinitionFile from the tests now

try/run_central_instances.py
KSchopmeyer: This one confusing since we are not using the servers path at all but an instance of ServerDefinition.

DISCUSISON:  What did you mean here?

KSchopmeyer: The server here is an instance of ServerDefinition class not path to profile, etc.

DISCUSSION: What did you mean here.

andy-maier:
On `default_namespace` in the server definition: We agreed that it makes sense to have it for testcases that work directly with the implementation without going through profile advertisement and discovery of central oinstances. Not having that would effectively not allow writing testcases that could be used for more than a specific server.

However, the name "default_namespace" is misleading. The PRP calls this "implementation nmamespace", so the attribute in the server definition file, and the attributes in the Python classes representing it, should be called `implementation_namespace`.

PR #1559, opened 2018-12-30
[2] Added reference direction to profile definitions; made some items optional

andy-maier:
**Note: This PR is on top of PR #1561. Review only the delta!**
For details, see the commit message.
Ready for review and merge.

PR #1560, opened 2018-12-30
End2end testcases for SNIA Server profile; some end2end base support

andy-maier:
**Note: This PR is on top of PR #1559. Review only the delta!**
Ready for review and merge.


KSchopmeyer:
I ran this against Fujitsu, a server that we know and that seems fairly consistent.  Put results in the shared directory.  

 ======= 5 failed, 39 passed, 59 skipped, 477 warnings in 158.03 seconds ========

KSchopmeyer:
Picture not nearly as pretty when we run against all servers.

Suggestion, make initial test that effectively  skips any server that does not respond to an initial standard test (connect and something simple like get known class) so that it is not tested any further and not considered part of failures.   In the SNIA test center case it is because there are always a number of servers down for one reason another and running them through any tests after this initial test is fruitless.

andy-maier:
The last commit fixes some of the issues. See its commit message for details.

PR #1569, opened 2019-01-02
Fixes #1568: Setting last_raw_request/response independent of debug

### LINKED ISSUE #1568: Use of conn.last_raw_reply when not set
The recently introduced handling of ill-formed XML in error responses (PR #1538) uses the `WBEMConnection._last_raw_reply` attribute to look at the ERROR element.

However, the code in `WBEMConnection._methodcall()` only sets that attribute when `debug` is enabled.

It seems that we should set the `_last_raw_request` and _last_raw_reply` attributes unconditionally, and that is possible without any extra conversions.

Details are in Twelveservers_test_end2end_34.tmp.log, line 1178, for the first occurrence of the `AttributeError` caused by that. There are 88 occurrences of that in that file.

andy-maier:
Ready for review, test, and merge.

Testing: `make end2end` on at least the Dell/Compellent, EMC3, EMC4 servers.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -454,6 +454,12 @@ This version contains all fixes up to pywbem 0.12.4.
   `ToleratedServerIssueWarning` and to continue with the next approach for
   determining the central instances. See issue #1438.
 
+* The `last_raw_request` and `last_raw_reply` properties of `WBEMConnection`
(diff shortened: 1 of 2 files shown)


andy-maier:
Tests against the SMI testlab have been run and have shown that this change works.

PR #1573, opened 2019-01-02
Add CIM-XML request and response data to CIM/XMLParseError

andy-maier:
This is the second part of improvements on ParseError.

Ready for review and merge.

Particular review items:
* The `str()`of ParseError now includes the CIM-XML response, which may get quite lengthy.
* It does not contain the CIM-XML request, because that is not normally needed to understand what is going on.

Testing: It would be good to run this PR with `make end2end` against the smigood group.

### DIFF
--- a/pywbem/exceptions.py
+++ b/pywbem/exceptions.py
@@ -222,6 +222,15 @@ class ParseError(Error):
 
     Derived from :exc:`~pywbem.Error`.
 
+    This class supports attaching the CIM-XML request and response data in
+    properties :attr:`~pywbem.ParseError.request_data` and
+    :attr:`~pywbem.ParseError.response_data`, respectively. When exceptions
+    of this class are raised by pywbem, these properties will always be
+    set.
+
+    The CIM-XML response data is part of the `str()` representation of the
+    exception.
+
     This exception is a base class for two more specific exceptions:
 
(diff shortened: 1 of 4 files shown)


pywbem/exceptions.py
```diff
@@ -222,6 +222,9 @@ class ParseError(Error):
 
     Derived from :exc:`~pywbem.Error`.
 
+    Optionally, the CIM-XML request and response data are attached to the
```
KSchopmeyer: The work optionally bothers me because it does not define the condition that is optional.

KSchopmeyer:
I believe that we have tested this as part of the other tests

pywbem/exceptions.py
andy-maier: Actually, the optionally refers to the ability of the exception class. In actual usage, I think we can guarantee that both are attached.

andy-maier: So, since this is the class description, so we want to describe the abilities of the class, or how we use it?

KSchopmeyer: I think we just want to state what the characteristics of the optionality are. i.e. generally wney attached and when not attached, even if it just points to behavior in the exception class.  It is just that the word optionally only means you cannot depend on something unless the condition for the optionality is understood.

andy-maier: Ok, I will add something to clarify that.

PR #1574, opened 2019-01-02
Shortened end2end test reports.

andy-maier:
Ready for review and merge.

This change only affects the `make end2end` test and gets rid of the many DeprecationWarning and ResourceWarning occurrences. It also shortens the lengthy tracebacks. If they are needed, they can be re-enabled by setting `TESTOPTS='--tb=auto'`. The review is more or less the question whether we want to have this shortening.


KSchopmeyer:
I believe that the tests this morning covered this.

PR #1578, opened 2019-01-04
Updated profile definitions from review comments

andy-maier:
This reflects Mike's comments from 2019-01-04.


testsuite/end2end/profiles.yml
```diff
@@ -338,13 +402,18 @@
...
 -
-  # TODO 2018-12 AM: Clarify the scoping path of this profile.
+  # Note: Special case: This profile has two scoping paths, see below.
   registered_org: SNIA
   registered_name: Enhanced Zoning and Enhanced Zoning Control
   autonomous: false
   central_class: CIM_ZoneService
   scoping_class: CIM_AdminDomain
-  scoping_path: null
+  scoping_path:
```
KSchopmeyer: If we are to generalize this file in the future would we not be better off to keep the whole scoping_path together and make it a list of lists.

KSchopmeyer:
I also added a test_needed flag.

testsuite/end2end/profiles.yml
andy-maier: What do you mean with keeping it together? In today's DMTF profile approach, the scoping path is an attribute of the scoped profile. I could imagine moving it to the scoping profile, for each profile it references. That would allow having different scoping paths for the same scoped profile, depending on which ptofile references it.

KSchopmeyer: you have

scoping_path:
    - CIM_HostedService
  scoping_path2:
    - CIM_HostedService

if there were three alternatives it would be:
scoping_path:
  -  ...
scoping_path1:
  -  ...
scoping_path3:
  - ...
n separate named lists.  Tonight not sure how to express it but in python it would be
[ ['CIM_HostedService'], [] ]

andy-maier: Karl, Ok now I get it. I don't want to have alternatives, because these are not really alternatives but instead from a perspective of a referencing profile, there is only one correct way. It just becomes more than one from a perspective of the referenced profile because the two profiles referencing it assume a different scoping path.

That's why I brought up the idea should we not attach the scoping path to the referencing profile. For now, just for our profile definition here in pywbem.

andy-maier:
I have updated the PR with the results of the discussion form the call with Mike on friday. The only remaining TODO (besides adding versioning) is about the case Karl commented above, where the "Enhanced Zoning and Enhanced Zoning Control" profile has two scoping paths.

**Discussion point:**
* Should we attach the definition of the scoping path to the profile references in the scoping profiles? That would support the fact that the "Enhanced Zoning and Enhanced Zoning Control" profile has a different scoping path, depending on which profile references it. However, that would require representing these references in profiles.yml, which we don't, currently.

PR #1585, opened 2019-01-06
Fixes #1582: End2end: Added testcases for SMI-S top level specification

### LINKED ISSUE #1582: end2end test for reference direction and profile tree
One idea in the recent discussions about scoping class methodology and different reference directions between SNIA and DMTF was that the top level registered specifications would determine the reference direction for all profiles in the tree below them.

That is something we can put in an end2end test.

Plus that the profile structure below them is actually a tree and that the trees of multiple top level specifications have no shared nodes.

andy-maier:
This PR adds an end2end test that asserts that the tree of referenced profiles under a top level specification has no shared or circular references.

Testing: This test is likely to fail on most implementations, because they tend to have RP instances per profile document, and not per profile implementation.

**Discussion point:** Need to discuss how central instances are supposed to be discovered from an RP instance if the same RP instance represents multiple profile implementations.

Update: I created issue #1588 for this discussion point.

### DIFF
--- a/testsuite/end2end/assertions.py
+++ b/testsuite/end2end/assertions.py
@@ -899,3 +899,92 @@ def _assert_association_consistency(
                 far_paths_msg,
                 first_far_paths,
                 far_paths_msg + " of first executed association approach")
+
+
+def std_uri(instance):
+    """
+    Return canonical WBEM URI of the path of the instance, nullifying the
+    host, or the empty string.
+    """
+    if instance is None or instance.path is None:
+        return ''
+    else:
+        path = instance.path.copy()
+        path.host = None
+        return path.to_wbem_uri(format='canonical')
+
+
+def assert_profile_tree(conn, profile_inst, profile_ancestry,
+                        reference_direction, tls_org, tls_name):
+    """
+    Assert that a profile tree is a tree without circular references,
+    when navigating to referenced profiles in the specified reference
+    direction.
+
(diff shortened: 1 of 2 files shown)


testsuite/end2end/test_profile_generic.py
```diff
@@ -97,3 +98,117 @@ def test_warn_undefined_profiles(  # noqa: F811
...
+        # incorrectly swapped ends
+        top_profiles_swapped = server.conn.Associators(
+            tls_inst.path,
+            AssocClass='CIM_ElementConformsToProfile',
+            ResultRole='ConformantStandard')
+
+        if top_profiles_swapped:
+            raise AssertionError(
+                "The instance representing the {0} {1!r} specification "
+                "references profiles via ECTP using swapped roles on "
```
KSchopmeyer: Lets spell out ECTP

andy-maier: DONE

andy-maier:
I am reworking the PR to no longer attempt to do the test generically, but to do it specifically for  SMI-S. Reason is that DMTF wrapper specs use CIM_ReferencedProfile from RegSpec to RegProf, while SMI-S uses ECTP.

andy-maier:
I have updated the PR to relax the profile tree checking as described earlier, plus it has checks on the reference from the top level spec to the top level profiles.

PR #1590, opened 2019-01-09
End2end: Updated profile.yml to reflect comments by Mike

andy-maier:
**Note: This PR is on top of PR #1585, review only the last commit**
No separate test needed, this can be tested along with PR #1591 which is based on this PR.


andy-maier:
The descriptions already reference the erratas and state that they reflect the errata content. We can update the profile.yml file as the status of these erratas changes, but I don't think we need to say that we will do that.

PR #1591, opened 2019-01-09
End2end: Raising AssertionError for all failures; Improved exc msg

andy-maier:

Particular review items may be:
* Approach for asserted version of WBEM operations by adding class `WBEMConnectionAsserted`. Alternatively, this could be done similarly to how it is done for `WBEMServer`, by generalizing `server_func_asserted()`and `server_prop_asserted()`.

Testing: end2end test on Brocade or Brocade1 server. pytest should no longer show the re-raise traceback of the ValueError exceptions for invalid integers, but a single AssertionError.

### DIFF
--- a/testsuite/end2end/_utils.py
+++ b/testsuite/end2end/_utils.py
@@ -7,7 +7,7 @@ from __future__ import absolute_import, print_function
 from copy import deepcopy
 
 from pywbem import CIMInstance, CIMInstanceName, \
-    CIMError, CIM_ERR_INVALID_CLASS
+    Error, CIMError, CIM_ERR_INVALID_CLASS, WBEMConnection
 
 
 def latest_profile_inst(profile_insts):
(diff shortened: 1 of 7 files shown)


andy-maier:
An end2end test has been run, the log is in `end2endcleanupoutput_smigoodsmall_test_end2end_34.tmp.log` in Karl's Dropbox folder.
All error output in that log is as desired by this PR.

However, the Brocade server was down, and Brocade and Brocade1 are the only ones that had the `CIMXMLParseError` on invalid integer due to `TYPE=''`. A re-run of the end2end test with this PR on Brocade or Brocade1 would be good to have.

KSchopmeyer:
This is tne one I ran the test against last night

andy-maier:
Karl and I agreed that we would run end2end test with this PR against the Brocade or Brocade1 server in the testlab, to verify it.

KSchopmeyer:
Did test on brocade alone and now doing general smigoodsmall test.

andy-maier:
You did this with the wrong PR. The exveption text says it.

KSchopmeyer:
test with smigoodsmall worked.  Removed test_needed.

KSchopmeyer:
Marked as discussion because I do not see with branch.

PR #1596, opened 2019-01-10
Fixes #1594: End2end: Support for groups in groups in server definition file

### LINKED ISSUE #1594: End2end: Support for groups in groups in server definition file
Add support for groups in groups in the server definition file.
Requires checking of loops (circular references).

One uase case is to switch the "default" server group quickly to select another existing group as the default. Today, one has to copy the group content into the "default" group, or to rename the desired group to "default" and rename the current "default" group to something else.

andy-maier:

Testing: Create nested groups in server definition file and check whether they run as expected.

### DIFF
--- a/testsuite/end2end/pytest_end2end.py
+++ b/testsuite/end2end/pytest_end2end.py
@@ -19,7 +19,7 @@ from assertions import assert_association_a1, assert_association_a2, \
 
 # Server nickname or server group nickname in WBEM server definition file
 TESTSERVER = os.getenv('TESTSERVER', 'default')
-SD_LIST = list(ServerDefinitionFile().iter_servers(TESTSERVER))
+SD_LIST = ServerDefinitionFile().list_servers(TESTSERVER)
 
 # Profile definition file (in end2end directory).
 PROFILES_YAML_FILE = 'profiles.yml'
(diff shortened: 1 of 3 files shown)


andy-maier:
I re-read the exception message and must have misread it yesterday. It says there is a loop in the definitions. That must be a bug.

andy-maier:
I fixed the bug in the loop detection in the last update to this PR.

The reason that I did not find the bug during my manual testing was that the bug only showed up when the group that had the other group is defined after that other group. In my test file, it was defined before. 

The PR is ready to re-run with your nested group server definition file.

KSchopmeyer:
I actually hit that where it was order dependent but thought it was me.  Ran simple tests OK.  Am going to do some more complex tests but looks good so far.

KSchopmeyer:
Only marked update needed because of suggestion to make set of final list to avoid duplicate testing.. The duplicate was:

  test1:
    - Cisco
    - test2
    - EMC
  test2:
    - EMC

make end2end TESTSERVER=test1 TESTCASES=test_server

Full result at : sd-groups-test1-server_test_test_end2end_34.tmp.log

Their is one weird thing.  It outputs a lot of 

/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py:1740: ToleratedServerIssueWarning: Ignoring socket error ECONNRESET (connection reset), continuing with reading the response., Connection id: None
  conn_id=self.conn_id)

messages in warnings_summary and those appear to be exactly the same but not folded.

andy-maier:
On the weird thing you mentioned: The connection ID causes the occurrences of this message not to be folded into a single one.

andy-maier:
The latest update no longer has duplicates in the server list.

PR #1598, opened 2019-01-11
Fixes #1597: End2end: Remembering unreachable servers to save time

### LINKED ISSUE #1597: End2end: Optimize non-reachable servers
If a server is not reachable, every testcase currently tries again to reach it, which consumes a lot of time.

In the wbem_connection fixture, the information whether a server is usable, could be remembered.

There are two options:
* on the server_definition object
* using the pytest cache fixture

andy-maier:
**Note: This PR is on top of PR #1596. Review only the last commit.**

Testing: end2end test with servers that are unreachable. There should only be one warning at the end about each unreachable server.


KSchopmeyer:
DISCUSSION: I ran against a single failed server and it showed skipped everything so passed.  What is result if all servers down?? Should we somehow show that some server were not avialable?

andy-maier:
For each server that is down, its test steps are shown as skipped and a single warning is shown in the warnings summary. That is per server that is down, and also applies also true if all servers are down.

KSchopmeyer:
Karl  approves as it stands.

PR #1606, opened 2019-01-13
Test: Reorganized testsuite

andy-maier:

**Note:** The `server_file.yml` file needs to be moved to fit the new directory structure:
```
mv testsuite/end2end/server_file.yml tests/servers/
```

**Testing:** 
* Run the tests in the `tests/manualtests` directory.
* Run the end2end tests. For the end2end tests, the `make end2end` command has remained the same, but the makefile no longer switches directories, so if you run py.test manually, run it from the main directory.


.gitignore
```diff
@@ -31,7 +31,7 @@ moflextab.py
 testsuite/runtests.log
 /testsuite/testtmp/
 /testsuite/schema/mof*/
-/testsuite/end2end/server_file.yml
+server_file.yml
```
KSchopmeyer: Change name to something more complete like end2endtest_server_definitions.yml

By the way I can see many more uses fo this file so maybe something even more general

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
 Testing
 -------
 
-To run unit tests in the currently active Python environment, issue one of
-these example commands:
+All of the following `make` commands run the tests in the currently active
+Python environment. Depending on how the `pywbem` package is installed in that
+Python environment, either the directories in the main repository directory
+are used, or the installed package.
```
KSchopmeyer: Not sure what or the installed package means

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+the main repository directory, and are located in the `testsuite` directory
+structure:
 
 ::
 
-    $ make test                                              # Run all unit tests
-    $ PYTHONPATH=. py.test testsuite/test_cim_obj.py -s      # Run only this test source file
-    $ PYTHONPATH=. py.test InitCIMInstanceName -s            # Run only this test class
-    $ PYTHONPATH=. py.test -k InitCIMInstanceName or Bla -s  # py.test -k expressions are possible
+    testsuite
```
KSchopmeyer: Why not go whole hog here and change testsuite to tests also

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+     +-- unittest            Unit tests for single modules of the package
+     |    +-- pywbem              Unit tests for the pywbem package
+     |    +-- pywbem_mock         Unit tests for the pywbem_mock package
```
KSchopmeyer: Onth comment here and the items right below, do we really mean Unit tests or is it the individual dest defintions? I am still trying to sort it out since this was a big change.

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+     |    +-- functiontest        Unit tests for testsuite/functiontest
+     |    +-- end2endtest_utils   Unit tests for testsuite/end2endtest/utils
+     |    +-- servers             Unit tests for testsuite/servers
+     |    +-- utils               Utility functions for use by unit tests
+     +-- functiontest        Function tests for the entire pywbem client
+     +-- end2endtest         End2end tests against WBEM servers
+     |    +-- utils               Utility functions for use by end2end tests
+     +-- manualtest          Manual tests
+     +-- servers             WBEM server definition file used by some tests and module for accessing it
+     +-- profiles            Simple definitions of management profiles used by some tests
```
KSchopmeyer: profiles is such a generic word. We want it to mean something like registered_profiles but that phrase  poor also. If you think of something better than profiles, fine, otherwise ignore this comment.

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+     +-- schema              The CIM schema MOF files used by some MOF tests
+     +-- dtd                 The CIM DTD file used by some CIM-XML validation tests
+
+There are multiple types of tests in pywbem:
+
+1. Unit tests and function tests
```
KSchopmeyer: how about separate unit tests and function tests. i.e. what are the difference so that there are two of these things.

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+   validate their results automatically.
+
+   They are run by executing:
+
+   ::
+
+       $ make test
+
+   Test execution can be modified by a number of environment variables, as
+   documented in the make help (issue `make help`).
```
KSchopmeyer: change word issue to execute or something.

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+       testsuite/end2endtest/server_file.yaml
+
+   from the provided example, and by executing:
+
+   ::
+
+       $ make end2end
+
+   Again, test execution can be modified by a number of environment variables,
+   as documented in the make help (issue `make help`).
```
KSchopmeyer: change word issue to something that means execute this  cli cmd

docs/development.rst
```diff
@@ -152,26 +152,107 @@ The top-level document to open with a web browser will be
...
+   There are several Python scripts and shell scripts that can be run manually.
+   The results need to be validated manually.
+
+   These scripts are in the directory:
+
+   ::
+
+       testsuite/manualtest
+
+   and are executed by changing into that directory and executing the
```
KSchopmeyer: Do you really have to change into that directory?

docs/development.rst
KSchopmeyer: That would bring pywbem into line with a lot of other python projects.

docs/development.rst
KSchopmeyer: Remember that I am still trying to just get the whole restructure in my head.

KSchopmeyer:
Ran through the standard test on two servers and have executed make end2end Fujitsu, make end2end, and make end2end TESTSERVER=smidown and make end2end TESTSERVER=smidown TESTCASES=test_server.

Considering the size of this change, remarkable.

.gitignore
andy-maier: The file can also be used by the manual scripts or by pywbemtools tests or by wbemcli tests, so I agree with the broader use. That's why I would not like to have end2end in its name. However, changing server_file to server_definitions sounds good to me. Can we defer that tro a separate update?

docs/development.rst
andy-maier: It refers back to the phrase "Depending on how the `pywbem` package is installed...".

docs/development.rst
andy-maier: Sounds good to me. DONE.

docs/development.rst
andy-maier: The subdirectories under `testsuite/unittest` (e.g. `pywbem`) contain the unit tests for the various areas.

docs/development.rst
andy-maier: Whenever we talk about it we say either "management profiles" or "registered profiles". IMO, that is explicit enough so that we can leave the directory name at `profiles` and the file name at `profiles.yml`.

docs/development.rst
andy-maier: I have looked at some of the scripts and it seems at least one has a dependency on running with the main directory being the current directory. So I'll change that accordingly.

andy-maier: DONE

docs/development.rst
andy-maier: DONE - changed "the installed package" to "the installed `pywbem` package".

andy-maier:
Karl, the renaming of testsuite to tests caused all schema MOF files to be listed as changed files.

I have reflected all those of your comments where I added DONE. Those are marked as discussion resolved. The remaining ones are up for your comments or for discussion.

.gitignore
KSchopmeyer: OK.

KSchopmeyer:
Passed end2end test with smigoodsmall, same number of failures as other executions of this test.

the run_cim_operations.py fails with:

  File "tests/manualtest/run_cim_operations.py", line 53, in <module>
    from unittest_extensions import RegexpMixin
ImportError: No module named 'unittest_extensions'

PR #1613, opened 2019-01-15
Add notebooks to demo mock capabilities

KSchopmeyer:
Adds Jupyter notebooks to demo the mock functions.

In the process of this we found errors in the code examples in the document for the mocker and corrected them.


KSchopmeyer:
I have left a TODO for more work but this set of notebooks actually executes.  Would like to commit if possible and then extend

docs/mocksupport.rst
```diff
@@ -83,17 +83,18 @@ executing WBEM operations on the faked connection:
     import pywbem_mock
 
     # MOF string defining qualifiers, class, and instance
-    mof = """
-        # CIM qualifier types (declarations)
+    mof = ''''
```
andy-maier: One single quote too much.

docs/notebooks/datamodel.ipynb
```diff
@@ -97,23 +95,23 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
```
andy-maier: This notebook is changed just because you used a different Python version (Python 3) than whoever committed the file last time.

We need to decide whether it is ok for the Python versions stated in the notebooks to be different. For example, if a notebook says 2.7, does it work to run it on 3.4? Or vice versa?

KSchopmeyer: Good call and I do not have an answer. 

1. we want to move to python 3 so that tends to be what we should work with I think.

2. I guess that this notebook changed simply because I opened it once to look as something. Note that none of the other notebooks I did not open changed.

3. I am not sure what the impact is if we do change python versions.

4. From the point of view of the user, I don't think the python version is really clearly defined for the user. Further, they get to chose python 2/3 version on startup if they want to work with notebooks and not just view them.

In other words I don't have a clean answer here.

docs/mocksupport.rst
KSchopmeyer: Fixed

PR #1633, opened 2019-01-24
Try: No longer use percent formatting strings in makefile

andy-maier:
This may help with issue #1631, but it is just a very rough guess.

I tested this under native Windows.


KSchopmeyer:
Not sure we are going to get this tested by Raamesh.  Lets commit and ask him to try again.

PR #1658, opened 2019-02-26
Fixes #1630: Tools to explore the issues of operation response timing

KSchopmeyer:
This modifies a test and develops a second test to allow us to
explore the issues of response time documented in issue #1630

1. It adds more detailed tests to the manual test  (tests/manual/run_enum_performance.py erformance and a better output display. It also adds command line argument processing so test variations can be defined.  This tool still works only with OpenPegasus because it uses a special OpenPegasus provider to generate the responses.

2. Adds a new test script response_performance.py (right now located in tests/performance) that
creates an xml enumerateinstances response of varying size instance and varying number of nstances returned and then executes the TupleParser.

This is a command line script with input parameters to control  the test.  It allows the user to select a variety of:

1. number of instances in the response

2. Size of the response in bytes.

3. Choice of running a test with no profiler running or with either the pyinstrument (call stack profiler) or the cProfile.

A table showing the response time for each combination of number of instances and size of the instances is displayed and if one of the profilers is defined by the input parameter, the
results of that profiler is displayed.

In addition, if the cProfiler is called it writes a file of the profile statistics to a file cprofiler.profile that can be explored with the stats module by executing:

   python -m python -m pstats <Name of file with profile data>

DISCUSSION:   

1. Is this useful  in finding the issues in response time?

2. Is there more tooling that would help?

PR #1659, opened 2019-02-27
Added response_performance.sh script

andy-maier:
**Note: This PR is on top of PR #1658, so review only the delta (last commit)**


andy-maier:
Updated the script accordingly.

I'm against doing that in run_response_performance.py: It would require adding the ability to have multiple runs, sothere can be a consistent run-ID on each log file. Also, unless we introduce a config file as input, changing the parameters would require editing the Python script. I think separating that out in a shell script makes it clear that that is just a convenience, and the interface to run_response_performance.py is the real interface.

KSchopmeyer:
DISCUSSION: I added an option log to the performance script and now it probably does just about  what you have in the script except for one variance.  We do not isolate the pyinstrument to profile just the  test code.

KSchopmeyer:
Also, since I modified some things on the cmd line of run_response_performance they need to be changed in the script (i.e. table -> cprofile and stack -> pyinst

andy-maier:
Abandoned, because Karl's python script now does all of that.

PR #1660, opened 2019-02-28
Added perf related files to .gitignore

andy-maier:


andy-maier:
> Why not simply *.profile
Do we understand what these files contain?
Can there be ore than one such file?

andy-maier:
I changed the .gitignore file to ignore `cprofilerstats*.profile`. Ignoring all of `*.profile` seems a bit too general to me. I did verify that pyinst does not create any `*.profile` files.

andy-maier:
We agreed to go with this version.

PR #1669, opened 2019-03-02
Performance improvement 9: Avoid inference of is_array and embedded_object

andy-maier:
see last commit


andy-maier:
For non-embedded object, the change trades two "if string in dict" tests with two "try & access dict" operations. I have no experience to say which one is faster.

For embedded objects, the change eliminates two "if string in dict" tests. However, embedded objects are rare.

Bottom line: I agree to leave that section of the code as is, until we gain more knowledge about the performance behavior.

KSchopmeyer:
Changed vote to APPROVED. :-)

PR #1694, opened 2019-03-16
Improved Python version checking by using 'python_requires' keyword

andy-maier:

The test that the new pip based Python version check actually works was made with a Python version requirement that required >=2.7 and thus was expected to fail on Python 2.6. The failing installs on Python 2.6 can be seen in this Travis run:
https://travis-ci.org/pywbem/pywbem/builds/507180522

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -42,6 +42,9 @@ Released: not yet
   `embedded_object` value is not inferred from the property or parameter value
   in that case, so this saves performance.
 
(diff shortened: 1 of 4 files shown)


andy-maier:
The pythion_requires keyword is processed by the setuptools package, afaik. Given that, it should be active for any way the package is installed, e.g. "python setup.py install".

andy-maier:
I have updated the PR by leaving the runtime check in.

Issue #1679, opened 2019-03-09
Redesign of CIM-XML parsing -> save the code


andy-maier:
Right now, pywbem has a SAX parser that builds up a tupletree for the complete response (in tupletree.py). The tupletree is then parsed into another kind of tupletree that contains CIM objects at its leafs (in tupleparse.py). That second tupletree is then accessed by the operation methods to do some checking and to get at the CIM objects.

This approach consumes a lot of cycles and memory for responses with large amounts of objects.

One idea to improve that is to use incremental parsing as offered eg by cElementTree, and to build up a response object directly. When using the iter*() methods, the complete response should never needed to be completely materialized that way.

andy-maier:
I have started working on this issue, and am trying the following approach:
* Use of event-driven parsing of start and end tags using cElementTree/ElementTree.
* The list of XML elements representing the operation results (result elements) are converted into CIM objects which are yielded by the parser.
* XML elements above the result elements in the XML tree are parsed by handling the start and end events incrementally, and are used to retrieve some attributes (e.g. DTD version) and to build up a response object with them. Reason for that approach is that we want to avoid having to construct the XML tree for all result objects.
* XML elements below the result elements in the XML tree are parsed by constructing the entire Element hierarchy for the subtree. That is done for each result object separately. Reason for that approach is that it is easier to deal with a tree of Element objects, and the whole subtree for each result object is needed anyway to construct the corresponding CIM object from it.
* A side effect of this overall approach is that the division of CIM-XML parsing into rather small and simple functions that correspond to the XML elements can continue to be used.

andy-maier:
PR #1695 implements the new parser based on `ElementTree.iterparse()`.
A future PR will need to implement its usage in pywbem.

andy-maier:
The work on the new parser has been stuck for a while due to the need for a light weight pull parser (which would need to be written as well). I suggest to move this to beyond 1.0.0.

andy-maier:
We agreed to move this one beyond 1.0.0

PR #1695, opened 2019-03-16
Issue #1679: Redesigned CIM-XML parser for better performance -> moved to attic

andy-maier:
WIP, with these ToDos:
* Double check new implementation with tupleparse.py to make sure any special cases are covered and have testcases.

Left for future work, beyond this PR:
* Usage of the new CIM-XML parser in pywbem (configurable between new parser and tupleparser)
* Implementation and use of listener support in new parser (indication delivery requests)


andy-maier:
Addresses a first part of #1679.

andy-maier:
Moved to 1.1.0

PR #1696, opened 2019-03-17
Some changes to improve coverage.

KSchopmeyer:
See commit messages for details.

Adds tests to test_tupleparser in those areas where the coverage shows missing, primarily in the methods that are used only by the server.  Note that leaves a bunch of TODOs for other methods that are not tested in test_tupleparser.py but that show tested in the coverage report.

It also adds test for the mof compiler to pick up a few obvious holes in the coverage (second commit)

All this only picks up coverage 1 %.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -45,6 +45,12 @@ Released: not yet
 * Added the 'python_requires' keyword to the package definition, which makes pip
   aware of the supported Python versions.
 
+* Comment out those methods in the TupleParser that are just WBEM server parser
+  methods including iparamvalue, etc. They should never be received by the
+  pywbem client. We left the code in case the server is implemented but this
+  was better than adding the test for these methods. Add test for is_subclass
+  function in cim_operations.py
+
 **Known issues:**
 
 * See `list of open issues`_.
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -9730,8 +9730,14 @@ def is_subclass(ch, ns, super_class, sub):
       sub:
         The subclass.  This can either be a string or a
         :class:`~pywbem.CIMClass` object.
-    """
 
+    Returns:
+        :class:`py:bool`: Boolean True if the assertion is True (sub is a
+        subclass of super_class) or False if it is not a subclass.
+
+    Raises:
+        CIMError if the the sub is not a valid class in the repo
+    """
     lsuper = super_class.lower()
     if isinstance(sub, CIMClass):
         subname = sub.classname
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -1683,76 +1683,6 @@ class TupleParser(object):
 
         return name(tup_tree), attrs(tup_tree), child
 
-    def parse_simplereq(self, tup_tree):
-        """
-          ::
-
-            <!ELEMENT SIMPLEREQ (IMETHODCALL | METHODCALL)>
-        """
-
-        self.check_node(tup_tree, 'SIMPLEREQ')
-
-        child = self.one_child(tup_tree, ['IMETHODCALL', 'METHODCALL'])
-
-        return name(tup_tree), attrs(tup_tree), child
-
-    def parse_imethodcall(self, tup_tree):
-        """
-          ::
-
-            <!ELEMENT IMETHODCALL (LOCALNAMESPACEPATH, IPARAMVALUE*)>
-            <!ATTLIST IMETHODCALL
-                %CIMName;>
-        """
-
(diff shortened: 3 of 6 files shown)


KSchopmeyer:
NOTE: Sadly this did not make much difference in coverage. Still at 84%.

It takes about 110 lines of code moved from missing to covered to increase the coverage by 1%.  

By that count, the major issues are in the following modules::

```
Module                                          statements  missing    coverage  % of total
pywbem/cim_operations.py                  1779         369        79%      3.5
pywbem/mof_compiler.py                     1103         237        79%      2
pywbem/cim_http.py                              370         189         49%      1.6
pywbem/_recorder.py                            489         127         74%      1 
pywbem/_server.py                               298          115         65%      1
pywbem/cim_obj.py                             1758           88         95%      .7
pywbem/_subscription_manager.py       276          50         82%      .4
. . .
Total                                    10293     1610      84%

By this measure, the major culprits are cim_operations and the compiler.  However, the issues in each are a number of small areas, not one block of code.  A large part right now is that the parser  exceptions for the various request calls are not covered and the recorder for each operation are not covered. I think
we could get about 4% more by adding code to test_cimoperations to cover these two areas and add more tests to test_mof_compiler.py in particular for CreateClass, aliases, assocDeclarations, ,etc. and rollback.

However, one big area that we really cannot change much is cim_http.py where a lot of the missed coverage is because of the PY2 vs. PY3 code.

**Comment by andy-maier** [comment]
Here are my 2 cents:
* I think we should ignore cim_http.py. The plan is to replace that with the use of "requests" in 1.0.0.
* cim_operations.py will see some change in 1.0.0 due to the parser redesign. I think that covers some of the little areas that are not covered today. So I tend to say let's ignore that too, before 1.0.0.
* Coverage improvements in the other areas most likely will remain valid also in 1.0.0. But: We wanted 0.14.0 to be a quick performance improvement release, and the last 0.x.x release, so I don't think we should invest in that before 1.0.0.

File: docs/changes.rst [hunk 1]
```diff
@@ -45,6 +45,12 @@ Released: not yet
 * Added the 'python_requires' keyword to the package definition, which makes pip
   aware of the supported Python versions.
 
+* Comment out those methods in the TupleParser that are just WBEM server parser
+  methods including iparamvalue, etc. They should never be received by the
+  pywbem client. We left the code in case the server is implemented but this
+  was better than adding the test for these methods. Add test for is_subclass
+  function in cim_operations.py
+
```
andy-maier: I disagree with commenting out the server-related parse functions. We don't know whether somebody used pywbem to build a server. This kind if thing should be done in a new major version, i.e. in 1.0.0.

pywbem/cim_operations.py
```diff
@@ -9730,8 +9730,14 @@ def is_subclass(ch, ns, super_class, sub):
...
         :class:`~pywbem.CIMClass` object.
-    """
 
+    Returns:
+        :class:`py:bool`: Boolean True if the assertion is True (sub is a
+        subclass of super_class) or False if it is not a subclass.
+
+    Raises:
+        CIMError if the the sub is not a valid class in the repo
+    """
```
andy-maier: While it is a good idea to document the return value and exceptions, it is not in scope of what the PR says it does.

docs/changes.rst
KSchopmeyer: In retrospect I agree.  I am going to modify to add tests for these functions.
I agree also with cim_http, in particular since is is really better covered than the numbers but part is with py27 and rest with py3 tests.

The one place we could gain probably easily is the compiler I would like to try to clean some of those up this week.

pywbem/cim_operations.py
KSchopmeyer: Added to PR doc.

docs/changes.rst
KSchopmeyer: DONE: Restored then and added tests for them.

KSchopmeyer:
My conclusion after this exercise is that it is going to be difficult to get coverage much higher than what we have.

docs/changes.rst
KSchopmeyer: I forget to update the PR documentation.  Am doing that.

pywbem/cim_operations.py
KSchopmeyer: I assume you did not want me to take the change out.  I tried to only add tests for a those parse_... methods that were reporting missing in coverage and to be honest, I am not sure if RETURNVALUE was in that catagory or not.

KSchopmeyer:
I think I responded to your issues and I modified the pr documentation to reflect what it is doing now.

PR #1699, opened 2019-03-21
Modify mof_compiler.py to remove unused production defs.

KSchopmeyer:
Modify mof_compiler to remove unused productions for
associationDeclaration, IndicationDeclaration, and the association
features lists

The pywbem compiler YACC includes productions for association declaration
and indication declaration.  While these are defined in DSP0004,
in this compiler they are never used because the the order of
executing the CreateClass production was classDeclaration,
associationDeclaration, indicationDeclaration and the syntax of
association and indication declarations is really a subset of the
class declaration.

Since the association and Indication abnf definitions are the same
syntax as the class declaration except that the keywords "Indication"
/ "Association" (i.e. the qualifiers association and indication) are the
first entry in the qualifier list, the pywbem  mof compiler always uses
the class declaration because it is the first entry.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -34,6 +34,14 @@ Released: not yet
 
 * Fixed AttributeError on __offset in CIMDateTime.repr(). See issue #1681.
 
+* Removed the associationDeclaration and IndicationDeclaration mof parser
+  production rules from mof_compiler.py because: a) They were orderd in
+  p_mp_createClass so that the classDeclaration production was always used,
+  b) When reordered, they still created a YACC reduce/reduce conflict where the
+  result was that YACC used the classDeclaration production to resolve the
+  conflict, c) they represent exactly the same syntax as the classDeclaration.
+  In effect, these production rules were never really used executed.
+
 **Enhancements:**
 
 * Added the possibility to specify a value of `False` for the `embedded_object`
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -79,6 +79,7 @@ from __future__ import print_function, absolute_import
 import sys
 import os
 import re
+# import logging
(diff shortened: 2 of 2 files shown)


KSchopmeyer:
NOTE: As a test, I tried reordering the three entites in the CreateClass production

    """mp_createClass : classDeclaration
                      | assocDeclaration
                      | indicDeclaration
                      """
to

    """mp_createClass : assocDeclaration
                      | indicDeclaration
                      | classDeclaration
                      """


 to see if it would then use assocDeclaration.  It still fell through to the classDeclaration declaration and did not execute the other two so I am missing some difference here.

andy-maier:
One explanation for falling through would be that the comma is missing between ASSOCIATION/INDICATION and the rest, so it does not match at all. Can you verify that?

KSchopmeyer:
What is happening.   I tested with the mof_compiler unchanged (i.e. without this pr) except for the order of assoc, indic, class in the create_class:

```

def p_mp_createClass(p):
    """mp_createClass : classDeclaration
                      | assocDeclaration
                      | indicDeclaration
                      """

to

def p_mp_createClass(p):
    """mp_createClass : assocDeclaration
                      | indicDeclaration
                      | classDeclaration
                      """

As before, we the yacc parser finds a reduce-reduce conflict but the conflict is actually in:

```

def p_qualifierName(p):
    """qualifierName : identifier
                     | ASSOCIATION
                     | INDICATION
                     """
    p[0] = p[1]

The conflict is between rules for identifier and ASSOCIATION and the ASSOCIATION rule is first in the rule table so it wins

Here is the wierd part, at this point the parser which had started parsing based on the assocDeclaration, ends up back the the classDeclaraction rule. for the next state.

That part I really do not understand. 

I commited the whole thing while I try to sort out exactly why the yacc processor did what it did.

PR #1704, opened 2019-04-03
Start 1.0.0

andy-maier:
**Discussion point:** Do we want to go to 1.0.0 as the next version.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -13,6 +13,30 @@ Change log
       :revisions: 1
 
 
+pywbem 1.0.0
+------------
+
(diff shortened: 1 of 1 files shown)


andy-maier:
Discussion point is: Do we want to go to 1.0.0 as the next version.

andy-maier:
We agreed to defer the actual start of 1.0.0 until we have understood whether there is additional performance improvements possible.

andy-maier:
In today's call, we agreed to start 1.0.0, and to do anything else as fixes in 0.14.x (unless something really serious comes up). So I'm merging this PR.

PR #1712, opened 2019-04-05
Add simple enum performance manual test

KSchopmeyer:
Add a file that executes a simple test of the response time for
EnumerateInstances and Open/Pull enumerate instances. This test does not
use any of the functionality added since pywbem 0.10.0 and only measures
the overall request response time.  This was done to allow it to be used
to measure response time changes between versions.

It executes only against the OpenPegasus server because it uses a
special provider in that server that generates response sizes and
response counts (number of instances returned) based on a method sent to
the server.

The primary goal of this test to be able to conduct a test over a range
of versions of pywbem and compare the results.

### DIFF
--- a/tests/manualtest/run_enum_performancesimple.py
+++ b/tests/manualtest/run_enum_performancesimple.py
@@ -1,0 +1,516 @@
+#!/usr/bin/env python
+
+"""
+Test CIM enumerate performance with a defined class that can be controlled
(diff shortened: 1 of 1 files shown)


andy-maier:
It fails due to a flake8 issue:
```
tests/manualtest/run_enum_performancesimple.py:266:81: E501 line too long (84 > 80 characters)
```

KSchopmeyer:
flake8 issue corrected.

Issue #1714, opened 2019-04-09
How to build a WBEM server environment for IBM SVC?


tanshaolong:
Hi everybody,

I have a IBM SVC V7000 device. I want to manage the storage by pywbem, but I don’t know how to build a WBEM server environment for IBM SVC. Who would tell me the steps? Or give some docs to refer? Thank you.

Best regards
Ray

tanshaolong:
According by my test, I think the WBEM server of IBM SVC is embedded in SVC controller. I have connected it, but I get exception when I run end2end tests. Who would give some hint for the error messages? Thank you.

The below is WBEM server definition.
servers:
  srv1:
    description: My WBEM server
    url: https://192.168.**.**:5989
    user: superuser
    password: ******
    no_verification: true
implementation_namespace: root/ibm

The below is the error message when run end2end test.
# make end2end
makefile: Running end2end tests
py.test --color=yes --pythonwarnings=default,ignore::DeprecationWarning,ignore::PendingDeprecationWarning -v --tb=short  tests/end2endtest -s
============================================================================ test session starts =============================================================================
platform linux2 -- Python 2.7.5, pytest-4.4.0, py-1.8.0, pluggy-0.9.0 -- /usr/bin/python2
cachedir: .pytest_cache
rootdir: /root/pywbem
plugins: cov-2.5.1
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/main.py", line 209, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/_pytest/main.py", line 248, in _main
INTERNALERROR>     config.hook.pytest_collection(session=session)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/pluggy/hooks.py", line 289, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/pluggy/manager.py", line 68, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/usr/lib/python2.7/site-packages/pluggy/manager.py", line 62, in <lambda>
INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
[...]

PR #1724, opened 2019-04-11
0.14: Added and fixed profile definitions for end2end tests

andy-maier:
This PR resolves some of the issues reported in issue #1714.

**Discussion point:** Complete the details for the Cascading profile. Using CIM_ComputerSystem for both central and scoping class, which consistent with what I found in a particular storage array. Asked Mike for clarification.

Update: Mike explained the situation with the Cascading profile, i.e. that in the version it appeared the central and scoping classes were not specified in the profile yet. Updated the PR with his suggestions. It is now ready to go. The end2end tests pass with the storage array I tested against.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,6 +31,8 @@ Released: not yet
 
 * Test: Fixed AttributeError in end2end assertion functions (Issue #1714)
 
+* Test: Added and fixed profile definitions for end2end tests. (Issue #1714)
+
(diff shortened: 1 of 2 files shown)


andy-maier:
Mike explained the situation with the Cascading profile, i.e. that in the version it appeared the central and scoping classes were not specified in the profile yet. Updated the PR with his suggestions.

PR #1727, opened 2019-04-11
Add changes to gain more performance:

KSchopmeyer:
1. Remove six  usage in tupleparser.py

1. Change from using isinstance() to using type() in tupleparser.
   This change seemed to actually gain  significant performance at least
  for python 3.

3. Change function that created set to create list because
   faster.  This actually seemed to gain some performance.

NOTE: This involves moving from using isinstance() to type() for this performance sensitive code and so we should really be sure that there are no issues.  The reason this was considered at all is that the isinstance had become one of the largest time consumers in the response_performance test as of version 0.14.0

I did not mark this for review since I am still doing testing, ex. against OpenPegasus. but am marking it now..

### DIFF
--- a/pywbem/tupleparse.py
+++ b/pywbem/tupleparse.py
@@ -120,13 +120,14 @@ def kids(tup_tree):
 
     The child elements are represented as tupletree nodes.
 
-    Child nodes that are not XML elements (e.g. text nodes) in tup_tree have
-    been filtered out.
+    Child nodes that are not XML elements (e.g. text nodes) in tup_tree are
+    filtered out.
     """
     k = tup_tree[2]
     if k is None:
         return []
-    return [x for x in k if isinstance(x, tuple)]
+    # pylint: disable=unidiomatic-typecheck
+    return [x for x in k if type(x) == tuple]
 
 
 class TupleParser(object):
@@ -148,17 +149,16 @@ class TupleParser(object):
 
         The child nodes must be text nodes (no element nodes).
         """
-        for inst in tup_tree[2]:
-            try:
-                assert isinstance(inst, six.string_types)
-            except AssertionError:
-                raise CIMXMLParseError(
-                    _format("Element {0!A} has unexpected child elements: "
-                            "{1!A} (allowed is only text content)",
-                            name(tup_tree), inst),
-                    conn_id=self.conn_id)
-        data = u''.join(tup_tree[2])
-        assert isinstance(data, six.text_type)
+        try:
+            data = u''.join(tup_tree[2])
+
+        except TypeError:
+            raise CIMXMLParseError(
+                _format("Element {0!A} has unexpected child elements: "
+                        "{1!A} (allowed is only text content)",
+                        name(tup_tree), tup_tree[2]),
+                conn_id=self.conn_id)
+
         return data
 
     # pylint: disable=too-many-arguments
@@ -222,10 +222,10 @@ class TupleParser(object):
                 conn_id=self.conn_id)
 
         if allowed_children is not None:
-            invalid_children = set()  # Eliminate duplicates
+            invalid_children = []
             for child in kids(tup_tree):
                 if name(child) not in allowed_children:
-                    invalid_children.add(name(child))
+                    invalid_children.append(name(child))
             if invalid_children:
                 if not allowed_children:
                     allow_txt = "no child elements are allowed"
@@ -235,7 +235,7 @@ class TupleParser(object):
                 raise CIMXMLParseError(
                     _format("Element {0!A} has invalid child element(s) "
                             "{1!A} ({2})",
-                            name(tup_tree), list(invalid_children), allow_txt),
+                            name(tup_tree), set(invalid_children), allow_txt),
                     conn_id=self.conn_id)
 
         if not allow_pcdata:
@@ -2137,7 +2137,7 @@ class TupleParser(object):
           CIMXMLParseError: There is an error in the XML.
         """
 
-        if isinstance(val, list):
+        if type(val) == list:  # pylint: disable=unidiomatic-typecheck
             return [self.parse_embeddedObject(obj) for obj in val]
         if val is None:
             return None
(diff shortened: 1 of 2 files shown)


pywbem/tupleparse.py
```diff
@@ -92,6 +91,21 @@
...
+    STRING_TYPES = basestring
+except NameError:
+    STRING_TYPES = str
+
+try:
+    # attempt to evaluate unicode
+    unicode  # pylint: disable=pointless-statement
+    TEXT_TYPE = unicode
+except NameError:
+    TEXT_TYPE = str
```
andy-maier: I don't understand the reason for this change. It boils down to exactly the same types six provides. Please explain.

andy-maier:
The next release has not been started yet, so before merging this we need to start it and that requires agreeing which one it should be.

pywbem/tupleparse.py
KSchopmeyer: Because at the time I had not put it up for review but just to test against CI.  The isinstance was the most expensive item in cprofile that we had control off so it seemed logical to see what could be done with it.  Actually you are just about correct the difference between using six and directly defining the types is alsmost nothing. The gain is in removing the number of isinstance calls since it was running at about 650,000 for our std test. and was #2 on the cprofile list

KSchopmeyer:
Performance changes with python 2 and python 3::

```

Starting profiler=none, response-count=[100, 1000] response-size=[100, 1000] runid=default log=False

Results: profile=none, response_counts=[100, 1000],
   response-sizes=[100, 1000], runid=default 20190412-100506
  Exp inst    Act inst    Response      XML    Parse time    Instances
      size        size       Count     size          sec.      per sec
----------  ----------  ----------  -------  ------------  -----------
       100         583         100    58619          0.05         1997
       100         586        1000   586919          0.43         2299
      1000        1033         100   103619          0.04         2238
      1000        1036        1000  1036919          0.5          1985


ks/more_performance branch

(perftest3) kschopmeyer@leonard:~/pywbem/perftest/pywbem(ks/more_performance)$ pp tests/manualtest/run_response_performance.py 
Starting profiler=none, response-count=[100, 1000] response-size=[100, 1000] runid=default log=False

Results: profile=none, response_counts=[100, 1000],
   response-sizes=[100, 1000], runid=default 20190412-100430
  Exp inst    Act inst    Response      XML    Parse time    Instances
      size        size       Count     size          sec.      per sec
----------  ----------  ----------  -------  ------------  -----------
       100         583         100    58619          0.04         2368
       100         586        1000   586919          0.38         2602
      1000        1033         100   103619          0.04         2608
      1000        1036        1000  1036919          0.42         2356


PYTHON 2

Master

(perftest2) kschopmeyer@leonard:~/pywbem/perftest/pywbem(master)$ pp tests/manualtest/run_response_performance.py 
Starting profiler=none, response-count=[100, 1000] response-size=[100, 1000] runid=default log=False

Results: profile=none, response_counts=[100, 1000],
   response-sizes=[100, 1000], runid=default 20190412-100212
  Exp inst    Act inst    Response      XML    Parse time    Instances
      size        size       Count     size          sec.      per sec
----------  ----------  ----------  -------  ------------  -----------
       100         583         100    58619          0.06         1715
       100         586        1000   586919          0.56         1775
      1000        1033         100   103619          0.06         1804
      1000        1036        1000  1036919          0.69         1458

ks/more_performance

(perftest2) kschopmeyer@leonard:~/pywbem/perftest/pywbem(ks/more_performance)$ pp tests/manualtest/run_response_performance.py 
Starting profiler=none, response-count=[100, 1000] response-size=[100, 1000] runid=default log=False

Results: profile=none, response_counts=[100, 1000],
   response-sizes=[100, 1000], runid=default 20190412-100326
  Exp inst    Act inst    Response      XML    Parse time    Instances
      size        size       Count     size          sec.      per sec
----------  ----------  ----------  -------  ------------  -----------
       100         583         100    58619          0.06         1795
       100         586        1000   586919          0.55         1831
      1000        1033         100   103619          0.05         1837
      1000        1036        1000  1036919          0.63         1586

File: pywbem/tupleparse.py (hunk 1 above)
Comment (KSchopmeyer) [reply pywbem/tupleparse.py]: Note that I completed the code and now there is no use of isinstance in tupleparse.py

Comment (andy-maier) [reply pywbem/tupleparse.py]: After discussing it in today's call and after understanding it better, the reason the `type(...) == six.string_types` did not work is that `six.string_types` is a tuple (with one item). So it should be `type(...) in six.string_types` instead. Could you please try with that, instead of using STRING_TYPE?

Comment (andy-maier) [reply pywbem/tupleparse.py]: Also, one of the uses of STRING_TYPE can be eliminated competely: In pcdata(), the entire for loop can be removed, and the join can be protected by handling TypeError and then raising the CIMXMLParseError.

**Comment by andy-maier** [comment]
We agreed to roll this one back to 0.14.2 (i.e. once done in master).

**Comment by andy-maier** [comment]
One more comment: Please rebase to master and add a change log entry.

**Comment by KSchopmeyer** [comment]
As a check I ran a test with python 3 replacing the 2 isinstance in tupleparse.py with type(...)  == str.  The differences were minimal and the isinstance does not show up in the top 20 entries in the cprofile. Conclusion is that this is not a significant candidate for change

**Comment by KSchopmeyer** [comment]
At this point cprofile reports that the major user is TuplePars.check_node followed by tupletree.startElementthe following:

```

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       32    0.123    0.004    0.432    0.013 {method 'Parse' of 'pyexpat.xmlparser' objects}
    35220    0.122    0.000    0.218    0.000 tupleparse.py:165(check_node)
    35220    0.074    0.000    0.079    0.000 xmlreader.py:332(items)
    35220    0.068    0.000    0.154    0.000 tupletree.py:96(startElement)
  30816/4    0.055    0.000    0.974    0.243 tupleparse.py:2069(parse_any)
    74812    0.048    0.000    0.064    0.000 tupletree.py:115(characters)

My system is report in the range of 2350 inst/sec for the test up from about 2200 in current master (in tests run one right after the other).

I have not figured out any significant changes to either.

andy-maier:
Approved. I will roll it back into 0.14.2.

andy-maier:
PR #1737 rolls this PR back into 0.14.2.

PR #1732, opened 2019-04-16
Fixes #1729: Fixed Appveyor CI setup for UNIX environments under Windows

### LINKED ISSUE #1729: Fix Appveyor CI for native Windows and Windows with UNIX-env
The manual-ci-run branch currently fails on Appveyor, plus the runtime environment used on Appveyor is incorrect.

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -50,6 +50,9 @@ Released: not yet
   property name in the returned instance. Works in the conn_lite=True mode but
   fails in conn_lite=False mode because the test was case insensitive.
 
+* Test: Fixed Appveyor CI setup for UNIX-like environments under Windows
+  (Issue #1729)
+
 **Enhancements:**
 
 * Changed GetCentralInstances methodology in WBEMServer.get_central_instances()
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -304,10 +304,10 @@ present in the PATH.
   * ``chmod``
   * ``tar``
 
-  These commands can be made available in the PATH via a UNIX-like environment
-  such as `CygWin`_, `MSYS2`_, Babun, or Gow. If they are not all available in
-  the PATH, the `Chocolatey package manager`_ must be installed and its
-  ``choco`` command must be available in the PATH.
+  These commands can either be made available via a UNIX-like environment
+  such as `CygWin`_, MSYS2, Babun, or Gow, or if they are not available
+  in the PATH, they will be installed using the
+  `Chocolatey package manager`_ by invoking its ``choco`` command.
 
   This script will also install the ``M2Crypto`` Python package into the active
   Python environment, so it must be run with the desired Python environment
@@ -328,8 +328,6 @@ present in the PATH.
 
 .. _`CygWin`: https://cygwin.org/
 
-.. _`MSYS2`: https://www.msys2.org/
-
 
 .. _`Installing to a UNIX-like environment under Windows`:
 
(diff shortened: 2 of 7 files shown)


andy-maier:
I found that there are some issues when using it on a local MSYS2 and CygWin. This definitely needs to be tested locally. One issue is a pip issue I have described in pypa/pip#6423.

docs/changes.rst
```diff
@@ -39,6 +39,9 @@ Released: not yet
 
 * Test: Added and fixed profile definitions for end2end tests. (Issue #1714)
 
+* Test: Fixed Appveyor CI setup for UNIX environments under Windows
```
KSchopmeyer: I think the the common acronym for this is WSL (Windows Subsystem for Linux). Probably more people will understand if we use that terminology.

andy-maier: WSL is something different than what CygWin and MSYS2 provide: https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux. Neither CygWin nor MSYS2 register itself as a WSL, nor do they intend to serve as a runtime compatibility layer for running Linux executables in WLF format. Instead, they serve mostly as a development environment for executables to be run on native Windows (in addition to be run in their respective shells).

andy-maier: Karl agreed in the slack channel. DONE.

andy-maier:
I have updated the PR to remove MSYS2 from the set of environments to be tested.
I'm not quite sure how to go about it in the documentation.

KSchopmeyer:
My first question is whether MSYS2 it a goal we need to achieve. 

NOTE: I made a mistake in first testing this and tested it against Microsoft WSL (https://docs.microsoft.com/en-us/windows/wsl/install-win10) which builds a linux subsystem into windows using one of the std linux flavors  (ex ubuntu,  within the windows file system but with a separate terminal for WSL and with the WSL filesystem hitten within the \user\<name>. I am not sure if this is worth adding to our tests or not at this point.

andy-maier:
On testing WSL: I don't have Windows 10 at home, nor does Appveyor.

andy-maier:
In a call last week, we agreed to remove the support for MSYS2 again.

andy-maier:
The latest update removes MSYS2 support.
The PR is now ready for final review.

docs/changes.rst
```diff
@@ -50,6 +50,9 @@ Released: not yet
   property name in the returned instance. Works in the conn_lite=True mode but
   fails in conn_lite=False mode because the test was case insensitive.
 
+* Test: Fixed Appveyor CI setup for UNIX environments under Windows
```
KSchopmeyer: Shouldn't this be unix-like to be consistent with other wording

docs/intro.rst
```diff
@@ -305,7 +305,7 @@ present in the PATH.
   * ``tar``
 
   These commands can be made available in the PATH via a UNIX-like environment
-  such as `CygWin`_, `MSYS2`_, Babun, or Gow. If they are not all available in
+  such as `CygWin`_, MSYS2, Babun, or Gow. If they are not all available in
```
KSchopmeyer: Chocolatey is mentioned just above here so shouldn't the note about this being a preq be mentioned there and maybe it should be in the list above this at about 305

docs/changes.rst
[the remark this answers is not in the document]

docs/intro.rst
andy-maier: DONE, but differently:

There are two difficulties with your proposal:
* The `choco` comamnd is only needed if these other commands are not in the PATH
* While these other commands can be made available by the user via a UNIX-like environment, that is not the case for the `choco` command.

Therefore, I don't think we should simply add the `choco` command to the list above.

Instead, I have tried addressing this by changing the paragraph to:

> These commands can either be made available via a UNIX-like environment such as `CygWin`_, MSYS2, Babun, or Gow, or if they are not available in the PATH, they will be installed using the `Chocolatey package manager`_ by invoking its ``choco`` command.

andy-maier:
Addressed both comments, and assuming your agreement to the one that I solved differently.

PR #1735, opened 2019-04-16
Fix issue with calling tupleparser in this test

KSchopmeyer:
The test directly calls the tuple parser but this call channged from a
direct method call to a Class.method call with the release of pywbem
0.13 so we test the version and create an appropriate method.

This change allows the test to be executed with earlier versions of pywbem so that we can compare results between pywbem versions.

Added pywbem version to runid since we are using this tool to test across multiple pywbem versions

I assume that this is for 1.0.0 so is marked to be rolled back to 14.2

### DIFF
--- a/tests/manualtest/run_response_performance.py
+++ b/tests/manualtest/run_response_performance.py
@@ -18,7 +18,7 @@ from pyinstrument import Profiler
 import six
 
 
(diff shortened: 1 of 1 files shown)


andy-maier:
PR #1738 rolls this PR back into 0.14.2.

PR #1739, opened 2019-04-17
Fixes #1736: Improved quality of TypeError exceptions raised for WBEMConnection ops args

### LINKED ISSUE #1736: Improve exception message when passing int objects to method params
Issue #1733 revealed that the TypeError that is raised when passing int objects is not easy to understand and needs to be improved:

```
Params = [
    ('ElementType', 2),
    ('InPool', sp_paths[0]),
    ('Size', 1024000),
    ('TheElement', None)
]
result, outparams = conn.InvokeMethod(
    'CreateOrModifyElementFromStoragePool',
    os_inst_path,
    Params)
```

This raises the following `TypeError`, due to the fact that the 'ElementType' parameter is specified as a Python `int` object in the input parameters of the method, when it should have been a `pywbem.Uint32` object:

```
File "", line 1, in
File "pywbem/cim_operations.py", line 4081, in InvokeMethod
result = self._methodcall(MethodName, ObjectName, Params, **params)
File "pywbem/cim_operations.py", line 1995, in _methodcall
ptuple = (p[0], p[1], infer_type(p[1]),
File "pywbem/cim_operations.py", line 1948, in infer_type
_format("Unsupported parameter type {0}", type(obj)))
TypeError: Unsupported parameter type <type 'int'>
```

The exception message does not really allow a user to understand what it even complains about, that they can correct it themselves, or how that would be done.

This should be improved.

andy-maier:

### DIFF
--- a/pywbem/cim_operations.py
+++ b/pywbem/cim_operations.py
@@ -1902,7 +1902,10 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
                                        namespace=self.default_namespace)
         else:
             raise TypeError(
-                _format("Invalid type for ObjectName: {0}", type(objectname)))
+                _format("The 'ObjectName' argument of the WBEMConnection "
+                        "operation has invalid type {0} (must be a string, "
+                        "a CIMClassName, or a CIMInstanceName)",
+                        type(objectname)))
 
         # Create HTTP extension headers for CIM-XML.
         # Note: The two-step encoding required by DSP0200 will be performed in
@@ -1924,7 +1927,7 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
 
         # Create parameter list
 
-        def infer_type(obj):
+        def infer_type(obj, param_name):
             """
             Infer the CIM data type name of a parameter value.
             """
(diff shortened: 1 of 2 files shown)


andy-maier:
Karl, your comment seems to have gotten lost...

pywbem/cim_operations.py
```diff
@@ -1941,11 +1944,16 @@ def infer_type(obj):
...
+                return infer_type(obj[0], param_name) if obj else None
             elif obj is None:
                 return None
+            if isinstance(obj, int):
+                hint = " (use a CIM integer type such as pywbem.Uint32)"
+            else:
+                hint = ""
             raise TypeError(
-                _format("Unsupported parameter type {0}", type(obj)))
+                _format("Method parameter {0!A} has unsupported type {1}{2}",
```
KSchopmeyer: Suggest rewording this to something like "Method Parameter {0!A} type {1} cannot be used to infer a valid CIM data type{2}.

andy-maier: DONE - changed text to:
"Method parameter {0!A} has type {1} which cannot be used to infer a valid CIM data type{2}"

andy-maier:
Resolved the comment.

PR #1747, opened 2019-05-02
Fix issue in mocker with property list on EnumerateInstance

KSchopmeyer:
Found issue with case sensitivity of test for property list on EnumerateInstances with conn_lite=False.
Fixed the issue and also cleaned up some other tests when I figured out how we could do both the
conn_lite=True and conn_lite=False tests in a single test. Did not work for this particular test however.

This is clearly a bug

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -45,6 +45,11 @@ Released: not yet
   been iterating the generator property in that object.  Documentation of
   that example and the example were corrected. (see issue #1741)
 
+* Fix issue in pywbem_mock/_wbemconnection_mock.py with EnumerateInstances that
+  includes a property list with a property name that differs in case from the
+  property name in the returned instance. Works in the conn_lite=True mode but
+  fails in conn_lite=False mode because the test was case insensitive.
+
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
The following test failed:

Chocolatey installed 1/1 packages. 
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
where swig
C:\ProgramData\chocolatey\bin\swig.exe
Done installing Swig
Curl is already installed ... skipping
Grep is already installed ... skipping
Installing Win32OpenSSL-1_1_0j ...
curl -o Win32OpenSSL-1_1_0j.exe -sSL https://slproweb.com/download/Win32OpenSSL-1_1_0j.exe --retry 3 --retry-connrefused --retry-delay 10
curl: (7) Failed to connect to slproweb.com port 443: Connection refused
Error: Command returned rc=7
make: *** [makefile:417: install_os_py2.7.done] Error 7
Command exited with code 2

andy-maier:
Rolled back into 0.14.2 via PR #1748

PR #1760, opened 2019-05-30
Fixes #330: Removed Python 2.6 support

### LINKED ISSUE #330: Remove support for Python 2.6
More and more things are falling apart for PyWBEM on Python 2.6:
- `virtualenvwrapper` still works but issues a failure about `stevedore` package using `logging.NullHandler` (see [this commit](http://git.openstack.org/cgit/openstack/stevedore/commit/?id=72e3e1177dc23b4f3d358294d68b58c01d7c5931) which removed the `NullHandler` workaround as part of a general removal of Python 2.6 support).
- Our long standing issue #26 on the `pbr` package is specific to Python 2.6.
- The Sphinx build problem discussed in PR #328 is caused by Sphinx using the latest (2.0.3) version of the GitPython Pypi package that imports `OrderedDict` from `collections` without a fallback to the `ordereddict` package. The GitPython team has added a workaround for this in their master branch, but that fix is not yet released. So this one should be resolved with the next release of GitPython.

In addition, there are some "soft factors" that may cause problems at some point down the road:
- PyWBEM is using some packages that seem to work on Python 2.6 but do not specifically claim support for Python 2.6 (However, I was also surprised how many of the packages we use, still specifically claim Python 2.6 support):
  - The `stevedore` package states a minimum of 2.7 (it is the only one of those we use that does that).
  - There are several packages we use that claim Python 2 support without being more specific and that possibly did not intend to imply Python 2.6 support with that claim. One of those is Pylint (and its dependents) - I seem to recall that they have dropped Python 2.6 support (but did not double check that).
- The Python core team has stopped updating Python 2.6 with Python 2.6.9 in 10/2013.
- In addition, all Python 2.6.x releases after 2.6.6 were source-only releases, which probably is the reason that RHEL 6.x (and possibly other Linux distros) stayed on the last binary release 2.6.6.
- The OpenStack team has decided in 11/2015 to drop support for Python 2.6 (the stevedore issue mentioned above is a direct result of that decision since the `stevedore` package is maintained by an OpenStack sub-team).

**Discussion point:**

The discussion to have is whether we still need support for Python 2.6 in PyWBEM.

It would be very helpful is PyWBEM users would state their requirements for Python 2.6 support as comments to this issue.

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -24,6 +24,20 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Removed Python 2.6 support.
+
+  The Python Software Foundation stopped supporting Python 2.6 with the
+  2.6.9 release in October 2013. Since then, many Python package projects have
+  continued releasing versions for Python 2.6, including pywbem. In 2017 and
+  2018, a number of Python package projects have actively removed support for
+  Python 2.6 and it has become an increasingly difficult task for pywbem to
+  keep supporting Python 2.6. For this reason, Python 2.6 support has been
+  removed from pywbem in its 1.0.0 version.
+
+  This allowed eliminating a lot of Python version dependent code,
+  eliminating the dependency to the unittest2 package, and lifting a number
+  of restrictions in test code.
+
 **Deprecations:**
 
 **Bug fixes:**
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,26 +30,16 @@ sudo: required
 # to install Python.
 matrix:
   include:
-#    - os: linux
-#      language: python
-#      python: "2.6"
-#      env:
-#        - PACKAGE_LEVEL=minimum
-    - os: linux
-      language: python
-      python: "2.6"
-      env:
-        - PACKAGE_LEVEL=latest
     - os: linux
       language: python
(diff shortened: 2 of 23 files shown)


docs/changes.rst
```diff
@@ -24,6 +24,20 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Removed Python 2.6 support.
+
+  The Python Software Foundation has stopped supporting Python 2.6 with the
```
KSchopmeyer: Drop word has.   Stronger with just "Foundation stopped"

docs/changes.rst
```diff
@@ -24,6 +24,20 @@ Released: not yet
...
+  2.6.9 release in october 2013. Since then, many Python package projects have
+  continued releasing versions for Python 2.6, and so has pywbem. In 2017 and
```
KSchopmeyer: Change: "2.6 and so has pywbem" to "2.6 including pywbem"

andy-maier: done

docs/changes.rst
andy-maier: done

PR #1766, opened 2019-06-06
Ks/add mock tutorial to table of tutorials

KSchopmeyer:
See commit message for more information.

I propose that we roll this one back since it is a bug but have one more pr to do so we can roll back both at same time.


KSchopmeyer:
Accidently added an unwanted change to this pr. Closing as invalid

PR #1767, opened 2019-06-06
Add ref in doc to pywbem_mock Jupyter notebook

KSchopmeyer:
Added a reference in the table of notebooks to the mock notebook.  I thought I had done it for v 0.13.0 but found in a demo that it was missing.

I also marked this for rollback but I have another but to submit and since this is low priority, I propose we wait for my other bug before any minor revision release.


andy-maier:
PR #1770 rolls this change back into 0.14.4.

PR #1781, opened 2019-07-19
Fixes #1780: Fixed incorrectly supported IncludeQualifiers on Open...() operations

### LINKED ISSUE #1780: IncludeQualifiers incorrectly supported by pywbem's OpenEnumerateInstances
As per DSP0200 1.4, IncludeQualifiers is not supported by the OpenEnumerateInstances operation (and similar Open... operations).

However, pywbem implemented support for IncludeQualifiers in `WBEMConnection.OpenEnumerateInstances()` (and similar Open... methods). When that parameter is passed as true or False, a WBEM server that correctly implements DSP0200 rightfully rejects the operation (OpenPegasus does this, for example).

Fix:

1. Change the implementation and documentation of the IncludeQualifiers parameter of `WBEMConnection.OpenEnumerateInstances()` (and similar Open... methods) to be ignored. It cannot be removed because that would break users that specify it. A user that specified it with True cannot rely on qualifiers to be included anyway, as per DSP0200, so that is no change for the client.

2. Change the implementation and documentation of the IncludeQualifiers parameter of `WBEMConnection.IterEnumerateInstances()` (and similar Iter... methods) to not pass it on if pull operations are used.

andy-maier:
Ready for review.

Needs to be rolled back to 0.14.4.


pywbem/cim_operations.py
```diff
@@ -5557,22 +5563,25 @@ class (or subclasses).
...
               causes the server-implemented default to be used. :term:`DSP0200`
               defines that the server-implemented default is `False`.
 
-            This parameter has been deprecated in :term:`DSP0200`. Clients
-            cannot rely on qualifiers to be returned in this operation.
+            If this method uses pull operations, this parameter will be ignored
+            and the returned instances will not contain any qualifiers.
+            If this method uses traditional operations, clients cannot rely
+            on returned instances containing qualifiers, as described in
+            :term:`DSP0200`.
```
KSchopmeyer: I notice that you removed the statement about deprecation. "This parameter has been deprecated in :term:`DSP0200`. Shouldn't we just leave that in

PR #1810, opened 2019-08-18
Partly fixes #1803: 0.14: Added support for testing with installed version

### LINKED ISSUE #1803: tests/README clarifications/questions
When looking at the `tests/README` file it reads: 

> Depending on how the `pywbem` package is installed in that
Python environment, either the `pywbem` and `pywbem_mock` directories in the
main repository directory are used, or the installed `pywbem` package.

How does a user run tests against a system wide install?  When you run `make test` in the root of the source tree, pywbem directory is in the current working directory, thus it will get picked up by the interpreter.  Furthermore, if you move pywbem and pywbem_mock directories, then you run into an issue where make encounters the following:

``` bash
make: *** No rule to make target 'pywbem/mof_compiler.py', needed by 'pywbem/mofparsetab.py'.  Stop.
```
as the files: `pywbem/mofparsetab.py` and `pywbem/moflextab.py` get created during the test.

How does a user run the unit tests from the source tree against installed system wide libraries?  This is something I would like to have the ability to do, thanks!

andy-maier:
This PR is ready for review and merge.

For trying it out, see the new section "Testing installed versions of pywbem" in docs/development.rst.


tasleson:
@andy-maier I followed your newly added documentation except substituting pip3 when interacting with system python.  I ran the tests against a test packaged version of `0.14.4` on fedora 30, that might be an issue if some changes are required between latest and last release.

When I run the `make test` I get a number of failures, I've attached the full output.
[pywbem_test_result.txt.gz](https://github.com/pywbem/pywbem/files/3516643/pywbem_test_result.txt.gz)

I see a bunch of
```
WARNING: Couldn't create 'pywbem.mofparsetab'. [Errno 13] Permission denied: '/usr/lib/python3.7/site-packages/pywbem/mofparsetab.py'
WARNING: Couldn't write lextab module 'pywbem.moflextab'. [Errno 13] Permission denied: '/usr/lib/python3.7/site-packages/pywbem/moflextab.py'
```
and
```
tests/unittest/utils/cmd_line_test_utils.py:150: in wbemcli_test
    assert_rc(exp_rc, rc, stdout, stderr)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

exp_rc = 0, rc = 127, stdout = ''
stderr = '/bin/sh: wbemcli: command not found\n'
```

andy-maier:
Thanks, that's good input.

* On the mofparsetab and moflextab files, we need to become more tolerant when creating them as a prereq for the test. They are most likely part of the distribution, so we need to use them instead of trying to update them. I can have a look at that.

* On wbemcli: I can only speculate that the distro package did not package wbemcli. Looking at the RPM in Fedora 30 testing (list is here: https://rpms.remirepo.net/rpmphp/zoom.php?rpm=pywbem), it installes "wbemcli" under the name "pywbemcli". See these lines in their .spec file:
  ```
  # wbemcli are conflicting with sblim-wbemcli
  mv -v %{buildroot}/%{_bindir}/wbemcli %{buildroot}/%{_bindir}/pywbemcli
  mv -v %{buildroot}/%{_bindir}/wbemcli.py %{buildroot}/%{_bindir}/pywbemcli.p
  ```
  I'm not sure how far we should go to accomodate such distro renames.

tasleson:
> it installs the "wbemcli" command under the name "pywbemcli". See these lines in their .spec file:

Yes, that was done as we had a conflict with another package with the same command name.  Adding an env variable would be good.

andy-maier:
I have updated the PR with:

* support for specifying the name of the wbemcli command via an env var TEST_WBEMCLI_NAME

* support for tolerating ply version mismatches (between installed ply package version and the ply version that was used to generate the MOF parser table files of pywbem) and skipping of testcases that compile MOF if the installed pywbem version is below 0.14.5. This has successfully addressed such mismatches when I tested it locally.

This should address both issues.

Also, I have added some words of caution about version mismatches between the installed pywbem version and the checked out repo version, to the Testing section.

docs/development.rst
```diff
@@ -268,6 +268,120 @@ active, as long as the Python `tox` package is installed in it:
...
+
+2. Create and activate a virtual Python environment of the intended Python
+   version, that is based on the system Python:
+
+   ::
+
+       $ virtualenv --system-site-packages --no-setuptools --no-pip --no-wheel .virtualenv/test
+       $ source .virtualenv/test/bin/activate
+
+   The pywbem project is set up so that Git ignores the ``.virtualenv``
```
KSchopmeyer: Question: The use of the .virtualenv as a dir name. Is that not user/env specific.  Thus, when I setup virtualenv and virtualenvwrapper it sets .virtualenvs as the the name and virtualenvwrapper uses that

WORKON_HOME="/home/kschopmeyer/virtualenvs"

tests/utils.py
```diff
@@ -0,0 +1,110 @@
...
+
+def import_installed(module_name):
+    """
+    Import a Python module/package, controlling whether it is loaded from the
+    normal Python module search path, or from an installed version (excluding
+    the module in the current directory).
+
+    The TEST_INSTALLED environment variable controls this as follows:
+
+      * If not set (or set to an empty string), the normal Python module search
```
KSchopmeyer: Why include the option of setting to an empty string with the not set.  Isn't it easier to simply use
existences or non-existence of the env var??

docs/development.rst
andy-maier: NOISSUE - WORKON_HOME is used only by virtualenvwrapper which is on top of virtualenv. virtualenvwrapper is not really useable in scripts, so scripts need to use virtualenv. virtualenv gets the directory specified directly, which is the case in the example code shown.

tests/utils.py
andy-maier: DONE - After first thinking it is strange that it should be enabled when set to an empty string, I looked at other variables in the makefile and they all enable their function when just being defined, regardless of the value. That was also the case for TEST_INSTALLED in the makefile, so it was even inconsistently implemented. So I agree with the suggestion and have changed the code accordingly.

PR #1815, opened 2019-08-21
Additional changes to support testing against installed version

tasleson:
The wbemcli handling is pretty straight forward.  The additional check for mof files is possibly incorrect, but even with these two commits I'm still getting:

`=== 149 failed, 10326 passed, 964 skipped, 417 warnings in 124.80s (0:02:04) ===`

without I get:

`=== 745 failed, 10311 passed, 381 skipped, 417 warnings, 2 error in 251.41s (0:04:11) ===`


tests/utils.py
```diff
@@ -39,7 +39,13 @@ def skip_if_moftab_regenerated():
...
 
     pywbem = import_installed('pywbem')  # noqa: E402
-    from pywbem import mofparsetab, moflextab
+
+    try:
+        # If you cannot write to the directory you can't make these files, thus they don't exist
+        # and you cannot import them.
+        from pywbem import mofparsetab, moflextab
+    except ImportError:
+        pytest.skip("We have no mofparsetab or moflextab to import!")
```
andy-maier: They should always exist: when testing against the local directory, the makefile ensures they get created. when testing against an installed version, the installed version needs to contain them. with a pypi package that is the case. with a distro provided os level package that ought to be the case as well.

andy-maier:
What are some of the 149 failures with these two commits?

tests/utils.py
andy-maier: Nevertheless, we can leave this in as an additional safeguard. Accepted.

andy-maier:
I have created PR #1833 and #1835 to pursue the two commits from this PR.

andy-maier:
Closing this PR, as its content is covered with the other two PRs. We follow up on the remaining issues via issue #1803.

Issue #1802, opened 2019-07-28
pywbem_mock issues to be fixed


KSchopmeyer:
EnumerateInstances  with properties returns instance with no properties unless DeepInheritance is set.

The cause is a code error in the mock code for enumerate instances

andy-maier:
Fixed by PR #1827 for master and PR #1830 for 0.14.5.

PR #1827, opened 2019-09-18
Ks/#1803 mock enum inst propertylist issue

KSchopmeyer:
The mock enumerateinstances returns no properties if there is a property list and deep inheritance is False.  This pr fixes the issue and makes the test more complete

See commit for details


andy-maier:
I corrected the issue fixed from #1803 to #1802.

PR #1830, opened 2019-09-18
Fixes #1802 in 0.14:  mock enum inst propertylist issue

### LINKED ISSUE #1802: pywbem_mock issues to be fixed

### LINKED ISSUE #1803: tests/README clarifications/questions

andy-maier:
Roll back PR #1827 to 0.14.

### DIFF
--- a/PKG-INFO
+++ b/PKG-INFO
@@ -1,6 +1,6 @@
 Metadata-Version: 1.2
 Name: pywbem
-Version: 0.14.5.dev13
+Version: 0.14.5.dev17
 Summary: pywbem - A WBEM client
 Home-page: https://pywbem.github.io/pywbem/
 Author: Tim Potter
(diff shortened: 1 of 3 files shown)


andy-maier:
I corrected the issue fixed from #1803 to #1802.

PR #1853, opened 2019-09-28
Fixes #1851: Improved wbemcli and pywbem mock error handling and messages

andy-maier:
For details, see the commit message.
Ready for review and merge.

PR #1854, opened 2019-09-29
0.14: Improved pywbem mock and wbemcli error handling and messages

andy-maier:
This PR rolls back PR #1853 into 0.14.5.

### DIFF
--- a/wbemcli.py
+++ b/wbemcli.py
@@ -82,7 +82,7 @@ as its current Python namespace, so the functions shown below can directly be
 invoked (e.g. ``ei(...)``).
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 # We make any global symbols private, in order to keep the namespace of the
 # interactive sheel as clean as possible.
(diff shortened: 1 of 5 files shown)


wbemcli.py
```diff
@@ -191,29 +203,30 @@ def _remote_connection(server, opts, argparser_):
             stats_enabled=opts.statistics)
 
         try:
-            build_mock_repository(CONN, opts.mock_server, opts.verbose)
+            _build_mock_repository(CONN, opts.mock_server, opts.verbose)
         except ValueError as ve:
-            argparser_.error('Build Repository failed: %s' % ve)
+            _error("Building the mock repository failed: %s" % ve)
```
KSchopmeyer: Why not use argparser_.error here also. I must be missing something.

andy-maier: `argparser.error()` displays a usage line because it reports command line errors. The ones I used `_error()` on are not argument errors, so I did not want to have the usage line.

PR #1865, opened 2019-10-02
Fixes #1415: Removed the **extra keyword args from WBEMOperation methods

### LINKED ISSUE #1415: Remove the `**extra` argument from the `WBEMConnection` operation methods
Remove the `**extra` argument from the `WBEMConnection` operation methods. It has no use except if someone implemented private extension in a server, and is error-prone in case a parameter name is mistyped (happened a few times!).

andy-maier:
Ready


andy-maier:
This PR needs PR #1866 to succeed. Abandoning this one, since 1866 contains this fix.

PR #1876, opened 2019-10-05
Fixes #1875: Removed the use of the 'pbr' package

### LINKED ISSUE #1875: pbr has too many undesirable side effects
Known issues with using pbr:
* Having to have a PKG-FILE to make the package installable from the source tarball. That installation is used by some Linux packagers (see issue #1245). The PKG-FILE required extra effort to be kept current, and was constantly out of date during development of a new version, so that the benefit of pbr showing the exact development version was basically nullified.
* Having to check out a large depth in Travis to make sure pbr gets the last version tag (See comment in .travis.yml).
* Having to set the git tag twice in the release procedure to make pbr use the right version (see tagging described in DEVELOP.md).
* pbr claims to be able to increment the major or minor version of a new development version based upon markers in the commit message text, but that never worked (see issue #1408). Example: That resulted in the master branch development version to be shown as 0.14.1.devNN because the last tag as seen from master was 0.14.0, while the stable_0.14 branch had already advanced to 0.14.5.
* It is suspected that the difficulties with PyInstaller (see issue #1410) also don't get better due to pbr.

Proposal is to remove the use of pbr and to remove the PKG-INFO file from the repo.

andy-maier:
Ready, see commit message.


.travis.yml
```diff
@@ -1,15 +1,10 @@
...
 
-# Pywbemn uses pbr to determine its version, which uses the newest tag
-# it finds back in the commit history. We need to make sure that tag
-# gets cloned, so we increase the depth of the shallow repo clone Travis
-# creates, beyond its default of 50.
-# Note that it is not clear whether the depth flag really counts number
-# of commits. For pywbem, it started failing with the default depth 50
-# when the number of commits since the last tag exceeded about 150.
+# Pywbem no longer uses pbr to determine its version, so the depth
+# does not need to be increased beyond its default.
```
KSchopmeyer: Probably these statement not necessary since after one more version nobody will know that we ever set depth to something else.

pywbem/__init__.py
```diff
@@ -64,8 +64,7 @@
 _python_m = sys.version_info[0]  # pylint: disable=invalid-name
 _python_n = sys.version_info[1]  # pylint: disable=invalid-name
 
-# !!! Make sure to keep the supported Python versions in sync
-# between setup.py, setup.cfg and pywbem/__init__.py !!!
+# Keep the checked Python versions in sync with setup.py
```
KSchopmeyer: Not shure what the word checked means in this context.

andy-maier: The code below verifies the supported Python versions. Any suggestio s for a better wording?

.travis.yml
andy-maier: Agreed

andy-maier: DONE

pywbem/__init__.py
andy-maier: DONE - changed to "Keep these Python versions in sync with setup.py"

Issue #1414, opened 2018-10-10
Conversion of test cases from unittest to pytest for 1.0.0


andy-maier:
This issue is about converting existing test cases from unittest to pytest, thereby completing its test cases as much as possible.

Remaining test scripts to be converted:
* test_mof_compiler.py - Created issue #2178
* test_indicationlistener.py - Created issue #2179
* run_cim_operations.py - Created issue #2180
* run_operationtimeouts.py - Created issue #2181

Completed:
* test_cim_obj.py - DONE in PR #1878/#1910 and #2042
* test_tupletree.py - DONE in PR ?
* test_cim_xml.py - DONE in PR #2043
* test_cim_http.py - DONE in PR #2061 
* test_statistics.py - DONE in PR #2117
* test_recorder.py - DONE in PR #2119

Note: Test scripts that don't exist yet:
* test_mock_dmtf_cim_schema.py

andy-maier:
Agreed on this:
* Create new issues for the remaining cases:
  - the two test_* files targeting 1.0.0
  - the two run_* files targeting 1.1.0

andy-maier:
Created new issues as documented in original description, so this issue can be closed as partially fixed.

PR #1878, opened 2019-10-06
Issue #1414: Converted test_cim_obj.py to pytest

andy-maier:
Ready for review and merge. For details, see the commit messages.

This PR intentionally contains multiple commits, to separate e.g. the fixes in the pywbem modules, the test infrastructure enhancements, and the conversion to pytest in test_cim_obj.py.

There are 2 enhancements/cleanup and 3 bug fixes in cim_obj.py.
Test coverage of cim_obj.py increased from 94.88%  to 96.94%.

PR #1910 rolls this PR back into 0.15.0.


pywbem/cim_obj.py
```diff
@@ -2733,28 +2759,40 @@ def update(self, *args, **kwargs):
...
     def update_existing(self, *args, **kwargs):
         """
-        Update already existing properties of this CIM instance.
+        Update the values of already existing properties of this CIM instance.
 
-        Existing properties will be updated, and new properties will be
-        ignored without further notice.
+        From the specified new properties, only those properties will be
+        updated that already exist in the instance. No new properties will
+        be added to the instance, ignoring their specified new properties
```
KSchopmeyer: Change "ignoring their specified new properties without further notice" to something like ; "they will be ignored without further notice".

Change "only those properties will be only those properties will be updated that already exist in the instance to"
TO
"only properties that already exist in in the instance will be updated"

andy-maier:
I think you don't need to review the changes in the test file.

pywbem/cim_obj.py
andy-maier: DONE - changed the sentences to:
"From the specified new properties, only properties that already exist in the instance will be updated. New properties will not be added to the instance; they will be ignored without further notice."

PR #1899, opened 2019-10-16
Fixes #1898: Test: Fixed Python version on Appveyor CI

### LINKED ISSUE #1898: Fix usage of Python 3.x on Appveyor
It turns out that the Appveyor tests on native Windows always use Python 2.7 (the system Python). The tox.ini file that is used to generate the virtual Python env does specify a PATH, but that is not sufficient.

See the builds in this recent Appveyor run: https://ci.appveyor.com/project/andy-maier/pywbem/builds/28007379

andy-maier:
Ready for review and merge.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,10 +10,15 @@ environment:
 #      UNIX_PATH: none
 #      PYTHON_CMD: python
 
+# Note: Python 3.4 is disabled on Windows because lxml does not install anymore,
+#       it fails with missing include files from libxml2 and libxslt, although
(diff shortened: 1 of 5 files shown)


andy-maier:
In our meeting today, Karl and I decided to update the documentation about how to deal with this when someone wants to develop on py34/win

andy-maier:
I found that a number of things need to be improved in the development chapter of the docs, before we can add information about py34 on Windows and lxml. I will do that in a separate PR.

Issue #1890, opened 2019-10-12
MOF Compiler issue with CreateInstance


KSchopmeyer:
The pywbem mof compilermay have a logic issue with its CreateInstance processing (see p_mp_createInstnace(p) line 618. The logic is such that it can overwrite existing instances in its working repository of instances.

The compiler  calls p.parser.handle.CreateInstance()  to create the new instance.  The default for this is in MOFWBEMConnection or a user can subsitute another repository with a CreateInstance method..

If the CreateInstance fails with CIM_ERR_ALREADY_EXISTS, it falls back to first try ModifyInstance and if this fails it next tries DeleteInstance the existing instance and recreate it with the new instance with CreateInstance.

NOTE: None of this actually happens when the compiler uses MOFWBEMConnection because the CreateInstance method in that code simply appends any new instance to the existing list of instances without checking for duplicates which is what currently causes duplicates in the mocker. See Issue #1852 

However, underlying this is the question of whether the compiler should be attempting to modify or replace existing instances in a repository based on a new instance with the same path or should simply generate an error if the instance already exists.

**DISCUSSION** Should the mof compiler ever try to replace existing instances in a repository with a new instance.   What I  found was in DSP0004, section 6.3 Instance Declarations is:

6.3   Instance Declarations
Any instance declaration is treated as an instruction to create a new instance where the key values of the object do not already exist or an instruction to modify an existing instance where an object with identical key values already exists.

The implication here is that if ModifyInstance exists, the rule is to modify the existing instance. However, this does not seem to extend to the idea of delete and replace which is being done by the current compiler

andy-maier:
I think the delete and replace approach was implemented as an additional safety net in case the modification does not work for some reason. I think we need to understand under which circumstances that would be needed.

andy-maier:
PR #1895 fixes this for master, and PR #1897 for 0.15.0.

KSchopmeyer:
**DISCUSSION** I think the real question here might be whether concept of retry with modify, etc. should be an issue for the client or the server.  In the case of the pywbem mof_compiler it actually works only if the server returns ALREADY_EXISTS for the create and NOT_SUPPORTED for the modify.

andy-maier:
We agreed on the following approach for this issue:
* Drop the third step of delete/create once modify fails
* Make sure that the error code that is now returned from the mock support causes modify to be used when create fails.

andy-maier:
[...]

PR #1909, opened 2019-10-24
Fixes issue #1890 Modify mof compiler and mocker to clean up CreateInstance

KSchopmeyer:
Modify mof compiler and mocker to clean up CreateInstance

Fix the mof_compiler to remove the logic in the createinstance function
to only attempt ModifyInstance if a CreateInstance fails with
ALLREADY_EXISTS and not attempt to DeleteInstance and CreateInstance.

Fix both the mof_compiler and the corresponding adaption of
MOFWBEMConnection in the mocker to correctly handle create instance from
compiler so that instances with the same path will modify instance with
the same path in the repository.

1. Removes the logic in compiler createinstance to delete the instance
in the repository and recreate it if the ModifyInstance fails and forces
an exception if the ModifyInstance fails

2. Adds test for Create duplicate instance in test_mof_compiler. This
test creates a new adaptation of MOFWBEMConnection
MOFWBEMConnectionInstDups that correctly handles the duplicate instance
creation by using the ModifyInstance which is part of this adaption

3. Duplicate this logic in the mocker _mocmofwbemconnection.py by adding
a corresponding ModifyInstance method.

4. Modify the test test_compile_instances_fail_dup in
test_wbemconnection_mock.py which originally tested for dups assuming
that a dup instance creation from the compiler was a failure from
earlier tests to confirm that the dup instance causes the original
instance to be modified.

5. Add to test for existing instance in mocker CreateInstance to
consider an instance from the compiler with either None for
CIMInstanceName or no keys as invalid.  This forces users of the pywbem
mocker to use the instance alias logic to set the instance path in
instances created in the mock repository.  This avoids the issue where
instances are created with a CIMInstanceName that consistes only of the
classname but no keybindings.


pywbem_mock/_mockmofwbemconnection.py
```diff
@@ -194,6 +195,34 @@ def CreateInstance(self, *args, **kwargs):
...
+        if self.default_namespace not in self.instances:
+            raise CIMError(
+                CIM_ERR_FAILED,
+                _format('ModifyInstance failed. No instance repo exists. '
+                        'Use compiler instance alias to set path on '
+                        'instance declaration. inst: {0!A}', mod_inst))
+
+        if mod_inst.path not in self.instances[self.default_namespace]:
+            raise CIMError(
+                CIM_ERR_FAILED,
```
andy-maier: Should that not result in CIM_ERR_NOT_FOUND?

pywbem/mof_compiler.py
```diff
@@ -1915,6 +1910,9 @@ class MOFWBEMConnection(BaseRepositoryConnection):
     This class implements the
     :class:`~pywbem.BaseRepositoryConnection` interface.
 
+    This implementation does not force paths on instances and just appends
+    each new createinstance to the instance repository
```
andy-maier: On "each new createinstance": Change to "each new instance".

docs/mocksupport.rst
```diff
@@ -600,6 +600,11 @@ There are two ways to build a mock repository:
   * Build from MOF definitions of the objects which are compiled into the
     repository. See :meth:`~pywbem_mock.FakedWBEMConnection.compile_mof_string`
     and See :meth:`~pywbem_mock.FakedWBEMConnection.compile_mof_file`.
+    If instances are to be compiled "instance of ..." they must be compiled
+    with the instance alias (ex:"instance of CIM_xxx as $xxx {") which adds
+    the CIMInstanceName with keybindings from the key properties to the
+    instance path. If compiled instances have duplicate paths, each new
```
andy-maier: 1. Suggest to change this sentence to:
"The MOF definition of any instances must specify an instance alias (ex: "instance of CIM_xxx as $xxx") which causes the instance path of that instance in the repository to be set up according to its key properties."

2. Why is the presence of the alias a precondition for building an instance path from the key properties? I suppose the alias name itself is not used in that process at all, so why can that not be done also for instances that do not have an alias? If you want to use the alias to safely identify the instances, it must be part of something in the repo. Where does the alias go in the repo?

docs/mocksupport.rst
```diff
@@ -600,6 +600,11 @@ There are two ways to build a mock repository:
...
+    instance will modify the already created instance.
```
andy-maier: change "already created" to "already existing".

pywbem/mof_compiler.py
```diff
@@ -636,17 +636,12 @@ def p_mp_createInstance(p):
                     if p.parser.verbose:
                         p.parser.log(
                             _format("ModifyInstance not supported. "
-                                    "Deleting instance of {0!A}: {1}",
+                                    "Create instance of {0!A}: {1} failed",
                                     inst.classname, inst.path))
```
andy-maier: I find this log message slightly confusing.
Plus, in the old code, this log message introduced that we are now deleting the instance. If we just want to log that ModifyInstance failed, we should do that for all status codes, and I guess that will be done by whoever handles the exception that is raised.

Bottom line, can we not remove the logging at this point? That would remove lines 635 to 640.

docs/mocksupport.rst
```diff
@@ -242,13 +242,13 @@ if there are classes in the repository.
   setting key properties as a dynamic provider for the class in a real WBEM
   server might have. This operation requires that the mock repository is in
   full mode and that the class of the new instance exists in the mock
-  repository. It fails with not supported if the mock repository is in lite
+  repository. It fails with NOT_SUPPORTED if the mock repository is in lite
```
andy-maier: Change "NOT_SUPPORTED" to the full name of the CIM status code "CIM_ERR_NOT_SUPPORTED".

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: Agreed and modified and the one above it should be CIM_ERR_INVALID_NAMESPACE. DONE.

pywbem/mof_compiler.py
KSchopmeyer: Agreed. DONE

docs/mocksupport.rst
KSchopmeyer: 1. Made change.

2. Because I was uncomfortable building the instance path within the CreateInstance and there must be a path to act as the key in the repository.  It is the existance of the alias name that causes the path to be set in the instance and I did not want to mess with that.  No the only reason for setting the alias is that is what causes the path to be set into the instance. The alternative is to set the path in all createed instances and I was uncomfortable proposing that change as part of this particular issue. I created issue #1911 to marked discussion over issue of setting keybindings in new instances independent of the alias flag.

docs/mocksupport.rst
KSchopmeyer: Agreed. DONE

pywbem/mof_compiler.py
KSchopmeyer: Sounds OK to me.  DONE

docs/mocksupport.rst
KSchopmeyer: DONE

docs/mocksupport.rst
andy-maier: OK, since issue #1911 was created, this comment can be considered done.

PR #1918, opened 2019-10-31
Fixes #1917: Issues with mock assocs

### LINKED ISSUE #1917: Issue in pywbem_mock associators algorthim
1. The algorithm for class referencenames does not account for the fact that we are really trying to process for all target classes from the name provided with the operation request and its superclasses. 

 2. Altough it is not documented clearly in the spec, at least two other implementations return error if the ResultClass or AssocClass do not exist. Note that the errors returned are BAD_Paramenter if the reference Result class or associator AssocClass does not exist and NOT_Found if the associator ResultClass does not exist.

KSchopmeyer:
See commit text for details.

NOTE: I added one piece of code to the tests that is disabled unless a flag is set.  This saves the request for each test as a pywbemcli command and puts the results in the file also.  I am using that to test against openpegasus with this complex association and compare results.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -35,8 +35,10 @@ import locale
 import traceback
 import re
 from xml.dom import minidom
+from collections import Counter
 from mock import Mock
 import six
+
 try:
     # time.perf_counter() was added in Python 3.3
     from time import perf_counter as delta_time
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -224,6 +224,10 @@ Released: not yet
 * Added dependency to pywin32 package for Windows, and pinned it to version 225
   to work around an issue in its version 226. (See issue ##1946)
 
+* pywbem_mock display_repository() comment defintion that surrounds comments
+  in the output was defined as # but mof comments are // so changed. (see
+  issue #1951)
```
andy-maier: The title says this PR fixes issue #1917, but this comment says it fixes #1951. Please clarify and make it consistent.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -35,8 +35,10 @@
 import traceback
 import re
 from xml.dom import minidom
+from collections import Counter
```
andy-maier: `collections.Counter` was introduced in Pyhton 2.7, and thus prevents that this PR can be rolled back into 0.15 (which still needs to support Python 2.6).

andy-maier: DONE - I have updated PR #1955 to use `backport_collections.Clounter` for Python 2.6.

PR #1929, opened 2019-11-02
Improved coverage of function tests via last_* connection attrs

andy-maier:
For coverage improvements, see my comment further down.
Ready for review.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -243,6 +243,9 @@ Released: not yet
   `AssocClass` parameters of association operations are specified using a
   `CIMClassName` object. (See issue #1907)
 
+* Improved test coverage of function tests by verifying the last_request,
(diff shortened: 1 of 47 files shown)


andy-maier:
Not sure the coveralls report linked in the coveralls comment above shows the difference in coverage correctly. What mattered to me with this PR is that the `_to_pretty_xml()` function in pywbem/cim_operations.py is now tested much better.

It does that in fact, and that can be seen by searching for "_to_pretty_xml" in pywbem/cim_operations.py in these coveralls reports:

* https://coveralls.io/builds/26729873 - master as of commit 3d6570b2
* https://coveralls.io/builds/26740439 - this PR as of commit 418e2de

It shows that the function itself is now fully tested, and that the two functions calling it are now also fully tested.

PR #1939, opened 2019-11-08
Fixes #1879: Enabled py38 for native Windows in Appveyor

### LINKED ISSUE #1879: Add Python 3.8 to supported releases and tests
According to https://www.python.org/dev/peps/pep-0569/, Python 3.8 final will be released on 2019-10-14.

andy-maier:
This PR enabled Python 3.8 for native Windows to checkout whether it is now supported on Appveyor.

Python 3.7 and 3.8 for CygWin on Appveyor is still commented out.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -47,15 +47,13 @@ environment:
 #      UNIX_PATH: none
 #      PYTHON_CMD: python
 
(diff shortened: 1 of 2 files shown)


andy-maier:
Assuming approval since it only affects the test environments.

Issue #1158, opened 2018-04-01
Removing from a repo with mof_compiler script


KSchopmeyer:
Removing items from the repo with mof_compiler appears to be incomplete.  After a correct install,

We executed the following:

```

(pywbem27) kschopmeyer@sheldon:~/pywbem/githubpywbem/pywbem(master)$ mof_compiler -r -s http://sheldon testsuite/test.mof
Error deleting instance root/cimv2:PyWBEM_AllTypes
     6 CIM_ERR_NOT_FOUND: PyWBEM_AllTypes
Error deleting instance root/cimv2:PyWBEM_AllTypes
     6 CIM_ERR_NOT_FOUND: PyWBEM_AllTypes
Error deleting instance root/cimv2:PyWBEM_MemberOfPersonCollection
     6 CIM_ERR_NOT_FOUND: PyWBEM_MemberOfPersonCollection
Error deleting instance root/cimv2:PyWBEM_MemberOfPersonCollection
     6 CIM_ERR_NOT_FOUND: PyWBEM_MemberOfPersonCollection
Error deleting instance root/cimv2:PyWBEM_MemberOfPersonCollection
     6 CIM_ERR_NOT_FOUND: PyWBEM_MemberOfPersonCollection
Error deleting instance root/cimv2:PyWBEM_PersonCollection
     6 CIM_ERR_NOT_FOUND: PyWBEM_PersonCollection
Error deleting class root/cimv2:PyWBEM_AllTypes
     9 CIM_ERR_CLASS_HAS_INSTANCES: PyWBEM_AllTypes
Error deleting class root/cimv2:PyWBEM_MemberOfPersonCollection
     9 CIM_ERR_CLASS_HAS_INSTANCES: PyWBEM_MemberOfPersonCollection
Error deleting class root/cimv2:PyWBEM_PersonCollection
     9 CIM_ERR_CLASS_HAS_INSTANCES: PyWBEM_PersonCollection

```

The results are inconsistent.

1. The PyWBEM_AllTypes class is definitily in the repo.

```

    cimcli gc PyWBEM_AllTypes

    // ===================================================
    // PyWBEM_AllTypes
   // ===================================================

       [Version ( "1.0.0" ), 

```

2. Instances still exist for PyWBEM_MemberOfPersonCollection

```

kschopmeyer@sheldon:~/pgogtest$ cimcli ni PyWBEM_MemberOfPersonCollection
[...]

PR #1943, opened 2019-11-10
Ks/#add log to mof compiler options

KSchopmeyer:
See commit message for description

NOTE: I marked this rollback needed only because I was using it to sort out the issues with the issue #1158 problem and that is marked rollback needed.

The only help this provides to to be able to better sort out what a server is really replying when the script tries to delete and the delete fails.


andy-maier:
Please rebase since PR #1946 which fixes the Appveyor issue has been merged.

PR #1957, opened 2019-11-21
Fixes #1956: mof_compiler does not send CreateInstance, CreateClass to conn

### LINKED ISSUE #1956: pywbem MOF compiler cannot compile to remote connection
The pywbem MOF compiler does not create classes, instances, or qualifer declarations in the defined target connection.

While it uses the defined target connection if defined to get information from the server if the target is specified (conn=...) The MOFWBEMConnection definitions for CreateClass, CreateInstance and SetQualifier only write to the local repository in MOFWBEMConnection, not to the remote repository.

This, for example, The set qualifier consists only of the code:

```
    def SetQualifier(self, *args, **kwargs):
        """Create or modify a qualifier type in the local repository of this
        class.

        For a description of the parameters, see
        :meth:`pywbem.WBEMConnection.SetQualifier`.
        """

        qual = args[0] if args else kwargs['QualifierDeclaration']
        try:
            self.qualifiers[self.default_namespace][qual.name] = qual
        except KeyError:
            self.qualifiers[self.default_namespace] = \
                NocaseDict({qual.name: qual})

That is why the remove works.  The remove recompiles the mof into the local repository the MOFWBEMConnection) and uses the results of that to try to remove the mof from the remote repository.

### PR DESCRIPTION
**Pull request by KSchopmeyer**
Fixed issue where mof_compiler does not write new instances to the implementation defined by the conn attribute.  It wasgetting from that repo but writing new class and instances to the cache defined by MOFWBEMConnection.

ANDY: This changes a basic perception of MOFCompiler class, the use of MOFWBEMConnection as the handle when the conn constructor parameter exists and is a WBEMConnection class or subclass.

This PR bypasses the the MOFWBEMConnection completely when conn is specified on MOF_Compiler by removing the line in MOFCompiler.__init__ that defines the handle as MOFWBEMConnection. Note that there is an alternative where we use this intermediate and then transfer all of the newly created instances to the conn repository if the compile is successful, i.e. make a single compile atomic. Note that even this creates questions, i.e. should it be part of the MOFCompiler or the mof_compile script if we were to make this atomic.  However, as it was implemented, it simply wrote all CreateClass, CreateInstance, and SetQualifier requests to the MOFWBEMConnection and then threw them away.

The adaption of MOFWBEMConnection to MyMOFWBEMConnection was dropped.

NOTE: mof_compiler script still uses the MOFWBEMConnection as the local interface when --dry-run or --remove options are defined.

**DISCUSSION** Should we implement this as atomic or leave with this change.

Example of use with this PR:

mof_compiler -s http://localhost tests/unittest/pywbem/test.mof  -v --log all=file -I tests/schema/mofFinal2.51.0/
MOF repository: Namespace root/cimv2 in WBEM server http://localhost
Compiling file './tests/unittest/pywbem/test.mof'
Creating class 'root/cimv2':'PyWBEM_Person'
Created class 'root/cimv2':'PyWBEM_Person'
Creating class 'root/cimv2':'PyWBEM_PersonCollection'
Created class 'root/cimv2':'PyWBEM_PersonCollection'
Creating class 'root/cimv2':'PyWBEM_MemberOfPersonCollection'
Created class 'root/cimv2':'PyWBEM_MemberOfPersonCollection'
Creating instance of 'PyWBEM_Person'.
Creating instance of 'PyWBEM_Person'.
Creating instance of 'PyWBEM_Person'.
Creating instance of 'PyWBEM_PersonCollection'.
Creating instance of 'PyWBEM_MemberOfPersonCollection'.
Creating instance of 'PyWBEM_MemberOfPersonCollection'.
Creating instance of 'PyWBEM_MemberOfPersonCollection'.
Creating class 'root/cimv2':'PyWBEM_AllTypes'
Created class 'root/cimv2':'PyWBEM_AllTypes'
Creating instance of 'PyWBEM_AllTypes'.
Creating instance of 'PyWBEM_AllTypes'.

pywbemcli confirms the new objects existL

pywbemcli> class find pywbem*
  root/cimv2:PyWBEM_AllTypes
  root/cimv2:PyWBEM_MemberOfPersonCollection
  root/cimv2:PyWBEM_Person
  root/cimv2:PyWBEM_PersonCollection

### DIFF
--- a/mof_compiler
+++ b/mof_compiler
@@ -43,7 +43,7 @@ from pywbem._logging import LOG_DESTINATIONS, \
     LOG_DETAIL_LEVELS, LOGGER_SIMPLE_NAMES, DEFAULT_LOG_DETAIL_LEVEL, \
     DEFAULT_LOG_DESTINATION, configure_loggers_from_string
 
-MOFCOMP_LOG_FILENAME = 'mofcompServer.log'
+MOFCOMP_LOG_FILENAME = 'mofcompserver.log'
 
 
 def main():
@@ -252,6 +252,9 @@ Example:
     if args.user and not passwd:
         passwd = getpass('Enter password for %s: ' % args.user)
 
+    if args.dry_run and args.remove:
+        argparser.error('--dry-run not allowed with --remove.')
+
     # if client cert and key provided, create dictionary for
     # wbem connection
     x509_dict = None
@@ -272,10 +275,15 @@ Example:
     if args.log:
         configure_loggers_from_string(args.log, MOFCOMP_LOG_FILENAME, conn)
 
+    # Define the connection to be use as the compile handle
+    # If remove or dry_run set new objects are not written to the
+    # repository defined by conn. However, if conn is defined, CIM
+    # objects needed for the compile are retrieved from the implementation
+    # defined by conn
     if args.remove or args.dry_run:
-
+        # Write classes/instances to local repo but get them from
+        # conn if it is defined.
         conn_mof = MOFWBEMConnection(conn=conn)
-
         # Make it simple for this script to use these attributes:
         conn_mof.default_namespace = conn.default_namespace
         conn_mof.url = conn.url
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -1897,8 +1897,15 @@ BaseRepositoryConnection.register(WBEMConnection)  # pylint: disable=no-member
 class MOFWBEMConnection(BaseRepositoryConnection):
     """
     A CIM repository connection to an in-memory repository on top of an
-    underlying repository, that is used by the MOF compiler to provide rollback
-    support.
+    underlying repository API, that is used by the MOF compiler to provide
+    rollback support.  It only implementats the BaseRepositoryConnection
+    methods required to implement the mof compiler rollback functionality and a
+    minimal repository for tests within the pywbem tests suite..
+
+    MOFWBEMConnection gets CIM objects from a repository defined
+    as the in-memory repository if the constructor parameter conn is None
+    or, if conn is defined, from the repository defined by conn.  It always
+    writes qualifiers, instances, and classes to the in-memory repository.
 
     This class implements the
     :class:`~pywbem.BaseRepositoryConnection` interface.
@@ -2089,14 +2096,21 @@ class MOFWBEMConnection(BaseRepositoryConnection):
         cc = args[0] if args else kwargs['NewClass']
         if cc.superclass:
             try:
-                _ = self.GetClass(cc.superclass, LocalOnly=True,  # noqa: F841
-                                  IncludeQualifiers=False)
+                # Since this may cause recursive GetClass calls
+                # IncludeQualifiers = True to insure reference properties on
+                # instances with aliases get built correctly.
+                self.GetClass(cc.superclass, LocalOnly=True,
+                              IncludeQualifiers=True)
             except CIMError as ce:
                 if ce.status_code == CIM_ERR_NOT_FOUND:
                     ce.args = (CIM_ERR_INVALID_SUPERCLASS, cc.superclass)
                     raise
                 raise
 
+        # TODO: Not sure why this code puts cc in local repo here and then
+        # again at the end. Note that it executes tests on the class
+        # before again inserting it into the repo and these can cause
+        # exceptions.
         try:
             self.compile_ordered_classnames.append(cc.classname)
 
@@ -2295,10 +2309,8 @@ class MOFCompiler(object):
 
             If the provided object is a WBEM connection (i.e.
             :class:`~pywbem.WBEMConnection` or
-            :class:`~pywbem_mock.FakedWBEMConnection`), a
-            :class:`~pywbem.MOFWBEMConnection` object is created from the
-            provided object, and used by the MOF compiler to interface with
-            the repository.
+            :class:`~pywbem_mock.FakedWBEMConnection`),  the MOF compiler
+            connects directly to interface defined by handle.
 
             `None` means that no CIM repository will be associated. In this
             case, the MOF compiler can only process standalone MOF that does
@@ -2340,7 +2352,6 @@ class MOFCompiler(object):
 
         if isinstance(handle, WBEMConnection):
             conn = handle
-            handle = MOFWBEMConnection(conn)
         elif handle is None:
             conn = None
         elif isinstance(handle, BaseRepositoryConnection):
@@ -2452,6 +2463,8 @@ class MOFCompiler(object):
 
         except CIMError as ce:
             if hasattr(ce, 'file_line'):
+                # pylint: disable=no-member
+                # file_line, attribute dynamically added by error code
                 self.parser.log(
                     _format("Fatal Error: {0}:{1}",
                             ce.file_line[0], ce.file_line[1]))
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -1897,21 +1897,27 @@ BaseRepositoryConnection.register(WBEMConnection)  # pylint: disable=no-member
 class MOFWBEMConnection(BaseRepositoryConnection):
     """
     A CIM repository connection to an in-memory repository on top of an
-    underlying repository API, that is used by the MOF compiler to provide
-    rollback support.  It only implementats the BaseRepositoryConnection
-    methods required to implement the mof compiler rollback functionality and a
-    minimal repository for tests within the pywbem tests suite..
+    optional underlying WBEM connection.
 
-    MOFWBEMConnection gets CIM objects from a repository defined
-    as the in-memory repository if the constructor parameter conn is None
-    or, if conn is defined, from the repository defined by conn.  It always
-    writes qualifiers, instances, and classes to the in-memory repository.
+    If a WBEM connection is provided with the conn parameter, that connection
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: mof_compiler [hunk 1]
```diff
@@ -34,16 +34,144 @@ import argparse
...
 
 from pywbem._logging import LOG_DESTINATIONS, \
     LOG_DETAIL_LEVELS, LOGGER_SIMPLE_NAMES, DEFAULT_LOG_DETAIL_LEVEL, \
     DEFAULT_LOG_DESTINATION, configure_loggers_from_string
 
-MOFCOMP_LOG_FILENAME = 'mofcompServer.log'
+MOFCOMP_LOG_FILENAME = 'mofcompserver.log'
+
+
+class MyMOFWBEMConnection(MOFWBEMConnection):
```
andy-maier: I think these changes should be in MOFWBEMConnection. That class is supposed to do use the connection provided, for all operations.

KSchopmeyer: We have already deviated from MOFWBEMConnection and its definition of the repository so that this would mean a number of other changes to MOFWBEMConnection which would change the behavior of this particular class which is 'sort of' the default class for an absolutely minimal repo and what is used in test_mofcompiler.py 

In reality, the only reason to put these in 1.0.0 is to prepare for the rollback to 0.15.0 and to achieve the same functional characteristic as 0.15.0.

I agree as part of releasing 1.0.0 we want to redo this but for now I propose that what is there is the simplest way to get the CreateInstance functionality into 1.15.0.

i.e. I propose to leave this and make it the last changes to 0.15.0, assuming that I will redo much for 1.0.0 **DISCUSSION**

KSchopmeyer: The code changes for MyMOFWBEMConnection in mof_compiler simply implements the use of the conn attribute so CreateClass, CreateInstance, setQualifier send to the server defined by conn rather than put directly into the local server.



Putting the code from MyMOFWBEMConnection back into MOFWBEMConnection will cause refactoring in the mocker since the mocker makes use of the fact that MOFWBEMConnection CreateClass, CreateInstance, and SetQualifier do not call the interface defined by the conn attribute but are implemented directly in the FakedWBEMConnection that is the mocker adaption of MOFWBEMConnection. In the mocker it would mean that a CreateClass, etc. creates a loop because it calls the server defined by conn which is the Mock server which goes back through WBEMConnection.CreateClass to get back to the FakedWBEMConnection in an endless loop.

This is because _MockMOFWBEMConnection uses the conn attribute to attach to FakedWBEMConnection assuming that the CreateClass, etc. are short-circuited directly the the repository.

My mistake was not figuring that this caused mof_compiler not to work early in the mock project and fixing it all.

NOTE: This creates another issue we should consider, the dry_run  and remove concepts in mof_compiler.  When we use that mode, we still write Classes, etc. but write them to the local repository. To actually write to to the remote repository we replace MOFWBEMConnection with the local MyMOFWBEMConnection from mof_compiler.  Thus, we will have to redo that logic also or drop the dry-run and create another version of MOFWBEMConnection in mof_compiler that is local only.

In effect.  MOFWBEMConnection only writes to local repo but reads from remote repo if thee conn repo attribute is set..
 MYMOFWBEMConnection reads from **and writes to** remote repo if the conn attribute is set.

KSchopmeyer:
Alternate proposal.  Add an init parameter to MOFWBEMConnection (ex. write_remote) that when True would send CreateClass, CreateInstance, SetQualifier to remote host if conn was set (The behavior of the MyMOFWBEMConnection adaption.  When False it would have the current behavior where even when conn is True it writes CreateClass, CreateInstance, SetQualifier local (The current MOFWBEMConnection behavior).

That way we encapuslate both behaviors in a single class.

andy-maier:
Karl and I agreed to go with the alternate proposal stated above.

KSchopmeyer:
Implemented the alternate proposal and that works.  This leaves only a single question for discussion, and that is marked in the lead comment.

pywbem/mof_compiler.py
```diff
@@ -1897,8 +1897,15 @@ def DeleteQualifier(self, *args, **kwargs):
 class MOFWBEMConnection(BaseRepositoryConnection):
     """
     A CIM repository connection to an in-memory repository on top of an
-    underlying repository, that is used by the MOF compiler to provide rollback
-    support.
+    underlying repository API, that is used by the MOF compiler to provide
```
andy-maier: I think with gaining more clarity about the nature of this construct, I suggest describing it as:
"A CIM repository connection to an underlying WBEM connection with an in-memory roll-back log on top of it, that provides the ability to roll back changes."

pywbem/mof_compiler.py
```diff
@@ -1897,8 +1897,15 @@ def DeleteQualifier(self, *args, **kwargs):
...
+    rollback support.  It only implementats the BaseRepositoryConnection
```
andy-maier: Typo "implementats"

andy-maier: I think with gaining more clarity about the nature of this construct, I suggest changing the first sentence to:

"A CIM repository connection to an in-memory store on top of an optional underlying WBEM connection.  If the underlying WBEM connection is provided, the connection is the target for any operations and the in-memory store acts as a rollback log to provide support for rolling back the operations. If the underlying WBEM connection is not provided, the in-memory store acts as a CIM repository that is targeted by the operations."

andy-maier: Typo "implementats"

pywbem/mof_compiler.py
```diff
@@ -1897,8 +1897,15 @@ def DeleteQualifier(self, *args, **kwargs):
...
+    methods required to implement the mof compiler rollback functionality and a
+    minimal repository for tests within the pywbem tests suite..
```
andy-maier: I think with gaining more clarity about the nature of this construct, I suggest changing the first paragraph to:

"A CIM repository connection to an in-memory store on top of an optional underlying WBEM connection.  If the underlying WBEM connection is provided, the connection is the target for any operations and the in-memory store acts as a rollback log in support of rolling back the operations. If the underlying WBEM connection is not provided, the in-memory store acts as a CIM repository that is targeted by the operations."

pywbem/mof_compiler.py
```diff
@@ -1897,8 +1897,15 @@ def DeleteQualifier(self, *args, **kwargs):
...
+
+    MOFWBEMConnection gets CIM objects from a repository defined
+    as the in-memory repository if the constructor parameter conn is None
+    or, if conn is defined, from the repository defined by conn.  It always
+    writes qualifiers, instances, and classes to the in-memory repository.
```
andy-maier: I think this is mostly stated by the proposed new text.

pywbem/mof_compiler.py
```diff
@@ -2089,14 +2096,21 @@ def CreateClass(self, *args, **kwargs):
         cc = args[0] if args else kwargs['NewClass']
         if cc.superclass:
             try:
-                _ = self.GetClass(cc.superclass, LocalOnly=True,  # noqa: F841
-                                  IncludeQualifiers=False)
+                # Since this may cause recursive GetClass calls
+                # IncludeQualifiers = True to insure reference properties on
+                # instances with aliases get built correctly.
```
andy-maier: We are in CreateClass. This does not drive recursive GetClass calls. There is one check for the direct superclass, and if that fails, the CreateClass fails. It is up to the MOF input to define the classes in the right order, unless I missed something.

pywbem/mof_compiler.py
```diff
@@ -2089,14 +2096,21 @@ def CreateClass(self, *args, **kwargs):
...
             except CIMError as ce:
                 if ce.status_code == CIM_ERR_NOT_FOUND:
                     ce.args = (CIM_ERR_INVALID_SUPERCLASS, cc.superclass)
                     raise
                 raise
 
+        # TODO: Not sure why this code puts cc in local repo here and then
+        # again at the end. Note that it executes tests on the class
+        # before again inserting it into the repo and these can cause
+        # exceptions.
```
andy-maier: The one at the end updates `self.class_names`, while this here updates `self.classes`.

I think that both should be at the end, since I don't think we want any updates to the repo in case of raising exceptions.

pywbem/mof_compiler.py
KSchopmeyer: Agreed. Done

pywbem/mof_compiler.py
KSchopmeyer: DONE. Paragraph gone

pywbem/mof_compiler.py
KSchopmeyer: Change wording to additional GetClass. The purpose of the note is that we need LocalOnly and IncludeQualifiers to be set True for the additional operations to work correctly.

pywbem/mof_compiler.py
KSchopmeyer: I also questioned this one which is why I added the comment but it has been there for a long time and this code is convoluted enough that I am uncomfortable because I am not sure if the subsequent  code in mp_createclass that could repeatedly do GetClass, find_mof and compile new mof to the cache might be caught in something I don't understand.

On the other hand a failure here causes the process to stop so nothing is really lost. Further, rollback starts with the class_names dictionary to get classes.

The only thing I can suggest here is to try moving that code  and see what happens.

andy-maier: You have added code to clean up the added objects in case an exception is raised. That is good enough from my perspective, for now.

Issue #1966, opened 2019-11-26
SSL Error


kautzy:
hi,

i upgraded a system to debian buster. following error occurs when running the check_esxi_hardware.py script. i think a similar ticked was closed a couple of minutes ago... please let me know if i can provide more useful information ;-)

```
root@monitor:/usr/local/src# /usr/local/nagios/libexec/check_esxi_hardware.py -H host2 -U XXXX -P XXXX -v
20191126 17:32:07 LCD Status: True
20191126 17:32:07 Connection to https://host2.vm.xiotron.com
20191126 17:32:07 Found pywbem version 0.14.6
20191126 17:32:07 Check classe OMC_SMASHFirmwareIdentity
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/pywbem/cim_http.py", line 646, in connect
    return self.sock.connect((self.host, self.port))
  File "/usr/lib/python3.7/ssl.py", line 1150, in connect
    self._real_connect(addr, False)
  File "/usr/lib/python3.7/ssl.py", line 1141, in _real_connect
    self.do_handshake()
  File "/usr/lib/python3.7/ssl.py", line 1117, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: UNSUPPORTED_PROTOCOL] unsupported protocol (_ssl.c:1056)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/nagios/libexec/check_esxi_hardware.py", line 720, in <module>
    instance_list = wbemclient.EnumerateInstances(classe)
  File "/usr/local/lib/python3.7/dist-packages/pywbem/cim_operations.py", line 2467, in EnumerateInstances
    **extra)
  File "/usr/local/lib/python3.7/dist-packages/pywbem/cim_operations.py", line 1737, in _imethodcall
    conn_id=self.conn_id)
  File "/usr/local/lib/python3.7/dist-packages/pywbem/cim_http.py", line 813, in wbem_request
    client.endheaders()
  File "/usr/lib/python3.7/http/client.py", line 1224, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.7/http/client.py", line 1016, in _send_output
    self.send(msg)
  File "/usr/local/lib/python3.7/dist-packages/pywbem/cim_http.py", line 480, in send
    self.connect()  # pylint: disable=no-member
[...]

Issue #1950, opened 2019-11-13
SSL error


allen0817:
hi andy-maier
env 
Python 3.5.7 (default, Aug 16 2019, 10:17:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux
pywbem 0.14.0

equipment  : ibm ds8000

code  and error
>>>import pywbem 
>>> conn = pywbem.WBEMConnection('https://192.168.1.10:6989', ('customer', 'cust0mer'), 'root/ibm',None,None,None,True)
>>> res = conn.ExecQuery('WQL', 'select * from IBMTSSVC_StorageVolume')
Traceback (most recent call last):
  File "/usr/local/python3/lib/python3.5/site-packages/pywbem/cim_http.py", line 646, in connect
    return self.sock.connect((self.host, self.port))
  File "/usr/local/python3/lib/python3.5/ssl.py", line 1027, in connect
    self._real_connect(addr, False)
  File "/usr/local/python3/lib/python3.5/ssl.py", line 1018, in _real_connect
    self.do_handshake()
  File "/usr/local/python3/lib/python3.5/ssl.py", line 996, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/local/python3/lib/python3.5/ssl.py", line 641, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL] EC lib (_ssl.c:728)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/python3/lib/python3.5/site-packages/pywbem/cim_operations.py", line 4196, in ExecQuery
    **extra)
  File "/usr/local/python3/lib/python3.5/site-packages/pywbem/cim_operations.py", line 1737, in _imethodcall
    conn_id=self.conn_id)
  File "/usr/local/python3/lib/python3.5/site-packages/pywbem/cim_http.py", line 813, in wbem_request
    client.endheaders()
  File "/usr/local/python3/lib/python3.5/http/client.py", line 1103, in endheaders
    self._send_output(message_body)
  File "/usr/local/python3/lib/python3.5/http/client.py", line 934, in _send_output
    self.send(msg)
  File "/usr/local/python3/lib/python3.5/site-packages/pywbem/cim_http.py", line 480, in send
    self.connect()  # pylint: disable=no-member
[...]

PR #1967, opened 2019-11-27
Added troubleshooting for SSL issues; added OpenSSL version to ConnectionError raised for SSL issues

andy-maier:
See commit message.
The reported SSL errors were in issues #1950 and #1966.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -547,6 +547,59 @@ Python versions). If this issue shows up, try installing the
 
 .. _Binary lxml package for Windows: https://www.lfd.uci.edu/~gohlke/pythonlibs/#lxml
 
+ConnectionError raised with [SSL: UNSUPPORTED_PROTOCOL]
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On newer versions of the operating system running the pywbem client,
+communication with the WBEM server may fail with::
+
+    pywbem.exceptions.ConnectionError: SSL error <class 'ssl.SSLError'>:
+      [SSL: UNSUPPORTED_PROTOCOL] unsupported protocol (_ssl.c:1056)
+
+For example, this happened after an upgrade of the client OS to Debian buster
+using Python 3.7, with OpenSSL 1.1.1d.
+
+This is an error that is created by the OpenSSL library and handed back up to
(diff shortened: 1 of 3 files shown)


docs/appendix.rst
```diff
@@ -547,6 +547,55 @@ Python versions). If this issue shows up, try installing the
...
+This is an error that is created by the OpenSSL library and handed back up to
+the SSL module of Python which hands it up to pywbem. The error indicates that
+OpenSSL and the WBEM server do not agree about which SSL/TLS protocol level to
+use.
+
+Pywbem specifies SSL parameters such that the highest SSL/TLS protocol version
+is used that both the client and server support. Thus, pywbem does not put any
+additional restrictions on top of OpenSSL.
+
+On Debian buster, OpenSSL 1.1.1d increased the security settings to require at
```
KSchopmeyer: Remove first word " On "

docs/appendix.rst
```diff
@@ -547,6 +547,55 @@ Python versions). If this issue shows up, try installing the
...
+least TLS 1.2 (see https://stackoverflow.com/a/53065682/1424462). This means
+most likely that the WBEM server side does not yet support TLS 1.2 or higher.
+
+This can be fixed for example by adding TLS 1.2 support to the server side
+(preferred) or by lowering the minimum TLS level OpenSSL requires on the client
+side (which lowers security). The latter can be done by changing the
+``MinProtocol`` parameter in the ``/etc/ssl/openssl.cnf`` config file on
```
KSchopmeyer: Note that this is Linux.  On windows it is probably in C:\OpenSSL-Win32\ ( or >C:\OpenSSL-Win64\) . Note that has been required only since OpenSSL 1.0 and before that was an option

docs/appendix.rst
andy-maier: DONE.
Changed to: "Debian buster includes OpenSSL 1.1.1d and increased its security settings to require at least TLS 1.2".

docs/appendix.rst
andy-maier: DONE.
CHanged to: "The latter can be done by changing the ``MinProtocol`` parameter in the OpenSSL config file on the client OS (typically ``/etc/ssl/openssl.cnf`` on Linux and OS-X,
and ``C:\OpenSSL-Win64\openssl.cnf`` on Windows)."

andy-maier:
Addressed the comments as proposed, considering this PR approved.

PR #1993, opened 2019-12-02
Fixed several PyLint issues

andy-maier:
We are again down to:

```
+-----------------------+------------+
|message id             |occurrences |
+=======================+============+
|fixme                  |204         |
+-----------------------+------------+
|using-constant-test    |8           |
+-----------------------+------------+
|no-value-for-parameter |2           |
+-----------------------+------------+
```


andy-maier:
Yes, apparently PyLint no longer supports these; that's how it surfaced to me.

PR #1995, opened 2019-12-03
Fixes #1235,#1997: Redesigned error handling in MOF compiler

### LINKED ISSUE #1235: mof_compiler adds file_line attribute to CIMError exceptions
The mof_compiler overloads the CIMError exception when it catches it by adding an extra attribute for line number and filename.

While this is not a major issue is is just one more thing that makes the compiler code hard to read and also an opening for an error if a new except CIMError is added and the developer forgets to add that extra line of code.

andy-maier:
See commit message.

When reviewing, have a closer look at:
* Exception class structure (MOFCompileError and its derived classes)
* Disabling of error logging in compile methods of FakedWBEMConnection (see also issue #1997).

No rollback into 0.16 planned.

### DIFF
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -99,11 +99,12 @@ from ._server import WBEMServer
 from .cim_constants import CIM_ERR_NOT_FOUND, CIM_ERR_FAILED, \
     CIM_ERR_ALREADY_EXISTS, CIM_ERR_INVALID_NAMESPACE, \
     CIM_ERR_INVALID_SUPERCLASS, CIM_ERR_INVALID_PARAMETER, \
-    CIM_ERR_NOT_SUPPORTED, CIM_ERR_INVALID_CLASS, _statuscode2string
+    CIM_ERR_NOT_SUPPORTED
 from .exceptions import Error, CIMError
 from ._utils import _format
 
-__all__ = ['MOFParseError', 'MOFWBEMConnection', 'MOFCompiler',
+__all__ = ['MOFCompileError', 'MOFParseError', 'MOFDependencyError',
+           'MOFRepositoryError', 'MOFWBEMConnection', 'MOFCompiler',
            'BaseRepositoryConnection']
 
 # The following pylint is applied for the complete file because invalid
@@ -341,36 +342,45 @@ def t_error(t):
     return t  # Return the error token for the YACC parser to handle
 
 
-class MOFParseError(Error):
+class MOFCompileError(Error):
     """
-        Exception raised when MOF cannot be parsed correctly, e.g. for
-        syntax errors. Derived from :exc:`~pywbem.Error`.
+    Base class for exceptions indicating issues with compiling MOF.
+
+    Derived from :exc:`~pywbem.Error`.
     """
 
+    error_kind = "MOF compile error"
+
     # pylint: disable=super-init-not-called
-    def __init__(self, parser_token=None, msg=None):
+    def __init__(self, msg, parser_token=None):
         """
         Parameters:
 
-          parser_token:
-            PLY parser token for the error (that is, the ``p`` argument
-            of a PLY parser function). This token contains information on
-            the location of the error in the MOF file, which is copied
-            into this object, and is accessible via properties.
-
           msg (:term:`string`):
-            Message text supplied by the creator of the error
-        """
+            Message text describing the error.
 
+          parser_token (lex.LexToken or yacc.YaccProduction):
+            PLY lex or yacc parser token (that is, the ``p`` argument of a yacc
+            parser function or the ``t`` argument of a lex parser function).
+            This token is used to obtain the MOF source text and location
+            information.
+
+            `None` will result in no MOF source text and location information
+            to be obtained.
+        """
+        assert msg is not None
+        self._msg = msg
         if parser_token is None:
             self.args = (None, None, None, None)
         else:
+            assert isinstance(
+                parser_token,
+                (lex.LexToken, yacc.YaccProduction))
             mof_ = parser_token.lexer.parser.mof
-            self.args = (parser_token.lineno,
+            self.args = (parser_token.lexer.lineno,
                          _find_column(mof_, parser_token),
                          parser_token.lexer.parser.file,
                          _get_error_context(mof_, parser_token))
-        self._msg = msg
 
     @property
     def lineno(self):
@@ -409,19 +419,12 @@ class MOFParseError(Error):
     @property
     def msg(self):
         """
-        :term:`string`: Brief description of the problem.
+        :term:`string`: Message text describing the error.
         """
         return self._msg
 
     def __str__(self):
-        ret_str = 'MOFParseError:\n'
-        if self.lineno is not None:
-            ret_str += _format('{0}:{1}:{2} msg={3}\n{4}',
-                               self.file, self.lineno, self.column,
-                               self.msg, self.context)
-        else:
-            ret_str += _format("{0}", self.msg)
-        return ret_str
+        return self.get_err_msg()
 
     def get_err_msg(self):
         """
@@ -431,7 +434,7 @@ class MOFParseError(Error):
 
         ::
 
-            Syntax error:<file>:<lineno>: <msg>
+            <error kind>:<file>:<lineno>:<colno>: <msg>
             <context - MOF line>
             <context - position indicator line>
 
@@ -439,8 +442,8 @@ class MOFParseError(Error):
 
           :term:`string`: Multi-line error message.
         """
-        ret_str = 'Syntax error:'
-        disp_file = 'NoFile' if self.file is None else self.file
+        ret_str = '{0}:'.format(self.error_kind)
+        disp_file = 'String' if not self.file else self.file
         if self.lineno is not None:
             ret_str += _format("{0}:{1}:{2}",
                                disp_file, self.lineno, self.column)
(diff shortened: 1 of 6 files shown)


andy-maier:
**Discussion needed:**
* Review exception class structure
* Discuss failure of test_remove()

andy-maier:
I was able to reproduce the error message part of what you said.

With pywbem at master:
```
$ pywbemcli -m tests/unit/mof_with_error.mof class enumerate
-Syntax error:tests/unit/mof_with_error.mof:2:19
class CIM_Foo {
        badtypedef InstanceID;
                   ^^^^^^^^^^
Mock MOF file 'tests/unit/mof_with_error.mof' failed compiling (see above)
Aborted!
```

With pywbem at this PR:
```
$ pywbemcli -m tests/unit/mof_with_error.mof class enumerate
-Mock MOF file 'tests/unit/mof_with_error.mof' failed compiling (see above)
Aborted!
```

The "Generating LALR tables" message has always been printed when YACC re-generates them:
```
ply//yacc.py:3436:        errorlog.debug('Generating %s tables', method)
```
We invoke yacc with errorlog at debug level in order to get its error messages when generating the tables. Not sure exactly what your scenario is.

My installation of pywbem was `make install` with the intended branch checked out. I don't think it works well to combine the git repo as a source and the `-e` flag, because the former makes a *temporary* copy of the repo it then installs from, and the latter relies on a *permanent* repo clone. So I recommend either installing a local repo clone with `-e` or installing from a git source without `-e`.

andy-maier:
On following up the missing error message: The MOFCompiler class still produces the error message with pywbem at master:
```
$ python -c "import pywbem; mc=pywbem.MOFCompiler(handle=None); mc.compile_file('../pywbemtools/tests/unit/mof_with_error.mof', 'root')"
MOF parsing error:../pywbemtools/tests/unit/mof_with_error.mof:2:19 MOF grammar error
class CIM_Foo {
        badtypedef InstanceID;
                   ^^^^^^^^^^
  File "<string>", line 1, in <module>
  File "/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/mof_compiler.py", line 2726, in compile_file
    return self.compile_string(mof, ns, filename=filename)
  File "/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/mof_compiler.py", line 2682, in compile_string
    rv = self.parser.parse(mof, lexer=lexer)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 1201, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 192, in call_errorfunc
    r = errorfunc(token)
  File "/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/mof_compiler.py", line 557, in p_error
    raise MOFParseError(msg=msg, parser_token=p)
pywbem.mof_compiler.MOFParseError: MOF parsing error:../pywbemtools/tests/unit/mof_with_error.mof:2:19 MOF grammar error
class CIM_Foo {
        badtypedef InstanceID;
                   ^^^^^^^^^^
```
BTW, this requires fixing pywbem.mof_compiler.py to modify `self.parser.handle` only when not None. Otherwise, with the code at master, one gets this:
```
$ python -c "import pywbem; mc=pywbem.MOFCompiler(handle=None); mc.compile_file('../pywbemtools/tests/unit/mof_with_error.mof', 'root')"
  File "<string>", line 1, in <module>
  File "/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/mof_compiler.py", line 2725, in compile_file
    return self.compile_string(mof, ns, filename=filename)
  File "/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/mof_compiler.py", line 2669, in compile_string
    self.parser.handle.default_namespace = ns
AttributeError: 'NoneType' object has no attribute 'default_namespace'
```

andy-maier:
The FakedWBEMConnection class still raises the exception correctly, but does not longer print it:
```
$ python -c "import pywbem_mock; conn=pywbem_mock.FakedWBEMConnection(); conn.compile_mof_file('../pywbemtools/tests/unit/mof_with_error.mof')"
  File "<string>", line 1, in <module>
  File "pywbem_mock/_wbemconnection_mock.py", line 583, in compile_mof_file
    mofcomp.compile_file(mof_file, namespace)
  File "pywbem/mof_compiler.py", line 2726, in compile_file
    return self.compile_string(mof, ns, filename=filename)
  File "pywbem/mof_compiler.py", line 2682, in compile_string
    rv = self.parser.parse(mof, lexer=lexer)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 333, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 1201, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/ply/yacc.py", line 192, in call_errorfunc
    r = errorfunc(token)
  File "pywbem/mof_compiler.py", line 557, in p_error
    raise MOFParseError(msg=msg, parser_token=p)
pywbem.mof_compiler.MOFParseError: MOF parsing error:../pywbemtools/tests/unit/mof_with_error.mof:2:19 MOF grammar error
class CIM_Foo {
        badtypedef InstanceID;
                   ^^^^^^^^^^
```
It no longer prints it because it is now invoked with `log_func=None`. The reason for dong that is that a user may want to react to the error by catching the exception, and may not be forced to have it printed. This was one of the changes in this PR (see commit message).

What I recommend to do in pywbwemtools, is this (in `_pywbemcli_operations.py`):
1. Catch the new base class MOFCompileError instead of MOFParseError.
2. Print the exception instead of relying on it having been printed.
This should both be done only when using pywbem 1.0.0 or higher.

andy-maier:
The last commit updates the change history to be more clear about what a user should change.

andy-maier:
I updated the change history to better clarify what a user should do in response to these incompatible changes.

andy-maier:
I did set this PR back to WIP in order to move out the fix for issue #1996 into a separate PR. It is more complex than anticipated, and may need to be rolled back.

andy-maier:
Updated the PR to remove the change that fixes PR #1996. This PR here is now ready for review.

pywbem/mof_compiler.py
```diff
@@ -2119,8 +2298,7 @@ def CreateClass(self, *args, **kwargs):
                               IncludeQualifiers=True)
             except CIMError as ce:
                 if ce.status_code == CIM_ERR_NOT_FOUND:
-                    ce.args = (CIM_ERR_INVALID_SUPERCLASS, cc.superclass)
-                    raise
+                    raise CIMError(CIM_ERR_INVALID_SUPERCLASS, cc.superclass)
```
KSchopmeyer: As long as we are modifying messages why not add target classname rather than just the superclassname

pywbem/mof_compiler.py
```diff
@@ -563,9 +665,7 @@ def p_mp_createClass(p):
                             if embedded_inst.value not in dep_classes:
                                 dep_classes[embedded_inst.value] = ce
 
-                            continue
-
-                    for cln, err in dep_classes.items():
+                    for cln, _ in dep_classes.items():
```
KSchopmeyer: Am I missing something.  If the value rtn is _ why are we processing as dep_classes.items() rather than just dep_classes to retrieve each cln???

pywbem/mof_compiler.py
```diff
@@ -1290,11 +1428,11 @@ def p_referenceInitializer(p):
...
         except KeyError:
-            ce = CIMError(
-                CIM_ERR_FAILED,
-                _format("Unknown alias: {0!A}", p[1]))
-            ce.file_line = (p.parser.file, p.lexer.lineno)
-            raise ce
+            raise MOFParseError(
+                msg=_format(
+                    "Cannot compile reference initialization because "
+                    "it specifies an unknown alias: {0!A}", p[1]),
```
KSchopmeyer: How about change to ' instance alias not previously defined in the MOF file'.  That covers our issues about definition before use and also insures the user can find reference to instance alias vs. class alias in doc.

pywbem/mof_compiler.py
```diff
@@ -1515,24 +1656,33 @@ def p_instanceDeclaration(p):
...
                 if p.parser.verbose:
                     p.parser.log("Can't find file to satisfy class")
-                ce = CIMError(CIM_ERR_INVALID_CLASS, cname)
-                ce.file_line = (p.parser.file, p.lexer.lineno)
-                raise ce
+                raise MOFDependencyError(
+                    msg=_format(
+                        "Cannot compile instance of {0!A} because its class "
+                        "does not exist in the CIM repository and a MOF file "
+                        "for it cannot be found on the search path",
```
KSchopmeyer: Change 'it' to 'that class'

KSchopmeyer:
I reviewed this and made a number of comments to the code/messages.

I also stuffed the following comment in one of the comment boxes above:

COMMENT/KS: Agree with your conclusion about pywbemtools. Will add this after we commit this code but long before we release 1.0.0.

I also agree that we will probably need more in change history defining user todos for incompatibilities but also probably need a section i the documentation on user moving from Vesion 0,x to 1.0.0 as we get closer to release.

pywbem/mof_compiler.py
andy-maier: Because I was sloppy... If you agree not to show the original exception text, we can even change it to a list.

pywbem/mof_compiler.py
andy-maier: I agree. This boils down to specifying a reasonable human readable status description. There are several other cases in this class that have similar issues (status description contains only a very short string without being really understandable unless you know how everything works). I suggest to improve that in a separate PR. I have created issue #2014 for that.

pywbem/mof_compiler.py
andy-maier: DONE - changed dict to a list.

pywbem/mof_compiler.py
andy-maier: DONE - changed to:
"Cannot compile reference initialization because the instance alias {0!A} was not previously defined in in the MOF"

pywbem/mof_compiler.py
andy-maier: DONE

andy-maier:
Addressed all comments as suggested, except for one where I created a new issue #2014.

Issue #1996, opened 2019-12-03
MOF compiler: Does not raise errors from ModifyClass when compiling an existing class


andy-maier:
When the MOF compiler compiles a class, it invokes `p_mp_createClass()` which invokes `handle.CreateClass()` and falls back to `handle.ModifyClass()` if the `handle.CreateClass()` method returns that the class already exists.

The issue is that if the `handle.ModifyClass()` call raises a CIMError, that error is logged but not re-raised.

I verified the code back to branch stable_0.8 and it never raised such an error.

The situation is complicated by the fact that `MOFWBEMConnection.CreateClass()` already replaces the class if it already exists, so that method never returns that the class already exists. Issue #991 describes that behavior. So if the MOFCompiler handle is a `MOFWBEMConnection` object, the issue described above does not have any bad consequences. However, the MOFCompiler handle does not have to be a `MOFWBEMConnection` object, so in the general case (e.g. with a real WBEM server), the issue does have bad consequences.

Proposal:
* Change `MOFWBEMConnection.CreateClass()` to no longer replace an existing class, but to return that it exists. This fixes issue #991.
* Change `p_mp_createClass()` to raise an error if `handle.ModifyClass()` returns an error.
* Add according testcases.

andy-maier:
Addresses in PR #1995 in master. Will not be rolled back because it never has been there.

KSchopmeyer:
We agree that we want to do part 2 and 3 of the above proposal but not part 1. Thus, we will raise an error if handle.ModifyClass() returns and error form p_mp_CreateClass and create test cases for this.

This also means that we abandon issue #991

andy-maier:
Issue #991 was abandoned.
PR #2011 fixes this issue here in master, and PR #2012 in 0.16.0.

PR #1999, opened 2019-12-05
[WIP] Fixes #991,#1996: MOF Compiler: Fixed ignoring error returned by ModifyClass

### LINKED ISSUE #991: MOF compiler: CreateClass updates existing class
This issue resulted from the following TODO in mof_compiler.py:

W:2117, 0: TODO: should we see if it exists first with (fixme)

The conclusion in a discussion about this was that we should check for existence and reject if it exists, because at this level, the functions should behave like WBEM operations.

andy-maier:
See commit message.

TODO:
* Fix #991 
* In the test_class_replace() testcase, provoke error in ModifyClass


andy-maier:
Abandoned because we decided not to enhance MOFWBEMConnection that way, but to keep it as is.

PR #2011, opened 2019-12-07
Fixes #1996: Fixed that MOF compiler did raise error when ModifyClass failed

### LINKED ISSUE #1996: MOF compiler: Does not raise errors from ModifyClass when compiling an existing class

andy-maier:
See commit message.

### DIFF
--- a/pywbem/mof_compiler.py
+++ b/pywbem/mof_compiler.py
@@ -613,6 +613,8 @@ def p_mp_createClass(p):
             p.parser.log(
                 _format("Error modifying class {0!A}: {1}, {2}",
                         cc.classname, ce.status_code, ce.status_description))
+            ce.file_line = (p.parser.file, p.lexer.lineno)
+            raise
 
 
 def p_mp_createInstance(p):
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -259,6 +259,11 @@ Released: not yet
   a local repository. It was documented that way, but failed with
   AttributeError. (See issue #1998)
 
+* Fixed the error that the MOF compilation of a class could fail but the
+  error was not surfaced. This only happened when the MOF compiler was invoked
+  against a WBEM server, when the class already existed, and when the
+  ModifyClass operation that was attempted in this case, failed.
+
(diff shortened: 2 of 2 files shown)


pywbem/mof_compiler.py
```diff
@@ -726,6 +726,13 @@ def p_mp_createClass(p):
             p.parser.log(
                 _format("Error modifying class {0!A}: {1}, {2}",
                         cc.classname, ce.status_code, ce.status_description))
+            raise MOFRepositoryError(
+                msg=_format(
+                    "Cannot compile class {0!A} because the CIM repository "
```
KSchopmeyer: Since we are fixing, this msg bothers me slightly because the first user question is why is the error on ModifyClass. Maybe slight extension to message that compiled class previously exists in in repository.

andy-maier: DONE - changed to:
"Cannot compile class {0!A} because the class already existed in the CIM repository and ModifyClass failed"

Issue #893, opened 2017-12-21
Use of ssl.create_default_context() to auto-adjust to better security


andy-maier:
In `cim_http.py`, around line 590:
```
                # Note: PROTOCOL_SSLv23 allows talking to servers with TLS but
                # not with SSL. For details, see the table in
                # https://docs.python.org/3/library/ssl.html#ssl.wrap_socket
                # TODO ks 4/16: confirm that we cannot use the default_context()
                # Selects the highest protocol version that both the
                # client and server support (SSLV23)
                ctx = SSL.SSLContext(SSL.PROTOCOL_SSLv23)
```
Specifying the protocol this way does not dynamically adjust to improved security in the future. The `ssl.create_default_context()` function exists for that purpose. That function should instead be used.

In addition, there is the function `create_pywbem_ssl_context()` in `cim_http.py` which is not used currently (that function incorrectly uses Purpose.CLIENT_AUTH, see https://security.stackexchange.com/a/84822 and https://github.com/mjs/imapclient/issues/317).

andy-maier:
**Discussion point:** Do we consider this mandatory for 0.13?

andy-maier:
We decided in the call today that this issue is optional for 0.13.

andy-maier:
This issue has become moot due to the use of the "requests" package. See issue #280.

Issue #889, opened 2017-12-21
Hard coded default directories for CA certificates


andy-maier:
From `cim_http.py`, line 126:
```
# TODO 5/16 ks This is a linux based set of defaults:

#: Default directory paths for looking up CA certificates for linux.
DEFAULT_CA_CERT_PATHS = \
    ['/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt',
'/etc/ssl/certs', '/etc/ssl/certificates']
```

The directories are hard coded and don't even exist on every Linux distribution. They probably don't exist on MacOS and definitely don't exist on Windows.

One possibility to improve the situation might be the use of the Python `SSLContext.load_default_certs()` function (to be investigated): https://docs.python.org/2/library/ssl.html#ssl.SSLContext.load_default_certs

Note that `SSLContext` has been introduced in Python 2.7 and has not been rolled back to Python 2.6, so it could be used once we have dropped support for Python 2.6 (see issue #330).

andy-maier:
The hard coded directories will go away in master due to migrating to the use of the requests package. See issue #280 and its PR #2023. This will only be done for the master branch, and will not be rolled back into 0.16.0. In 0.16, the hard coded directories will remain, for compatibility.

Closing this issue as pursued by another PR.

Issue #496, opened 2016-09-07
Get back to a single SSL environment across Python 2 and 3


andy-maier:

KSchopmeyer:
We have concluded that we may not be able get back to a single ssl.  

Karl agreed to generate an overview of the alternatives.

andy-maier:
I think the direction should be to switch to using the "requests" package.

In addition to solving this issue here, it would speed up connections because the "request" package maintains a pool of SSL sockets, and we so far go through an entire SSL handshake for each single operation.

**Discussion point:** Do we agree on this directions, or do we want to look at more options?

KSchopmeyer:
The two alternatives today appear to be:

1. Go to requests
2. Go completely to python ssl

Some of the open questions include:
1.. This means that anybody using the cert callback would not be getting a call back so we would have to tell them somehow. We propose that we would keep the callback but put a deprecation warning on it. that is better than removing it and causing in all probability an exception.  We need to think this one out since if someone was using the callback for real security, just not calling it means a security hole.

2. TOCTOU issues since this was the real reason that that Redhat went to m2crypto 

3. Issues in what we do with the callback in the future. It would appear that we could not remove it until pywbem 2.0

4. Is there a way we could simulate the callback with either the python ssl or requests code so that the user did have a way to inspect the server cert???

andy-maier:
Will be addressed when migrating to the "requests" package. See issue #280. Has been addressed in PR #2023.

Issue #86, opened 2016-03-12
ResourceWarnings in py3 during test


KSchopmeyer:
While probably not critical, there is no way I could see to disconnect a connection when finished.

Andy: Generalized the issue from ResourceWarnings about open sockets / WBEMConnection has no close() to cover all ResourceWarnings during test.

andy-maier:
Because the connection is only conceptual, it does not need to be closed. The WBEMConnection object will be garbage collected once unused, but that does not matter for that.

KSchopmeyer:
Note that with python 3, the run_cim_operations.py test repeatedly reports:

  timeout=self.timeout)
/home/kschopmeyer/pywbem/githubpywbem/pywbem/pywbem/cim_operations.py:596: ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('192.168.3.138', 59717), raddr=('192.168.3.138', 5988)>
  timeout=self.timeout)

Which is where I figured out that there was no timeout.

KSchopmeyer:
In retrospect this may be a more serious problem than I originally thought but not sure yet.

Note that we are running tests with timeout = 0 at this point

andy-maier:
Ok. The discussion in this issue of the `requests` package: https://github.com/kennethreitz/requests/issues/1882 seems to say that the warning is not an issue, but it remains unclear why it is being raised.

I don't think we want to explicitly destroy the socket after each CIM operation; that would be an efficiency problem.

Maybe you are right and an explicit `close()` on the `WBEMConnection` object would be a solution to avoid the warning. But we need to investigate what we would do in such a `close()` method.

KSchopmeyer:
Most other platforms have a close and that seems logical.  I agree. This is really a 0.9.0 issue and should not be 0.8.0

andy-maier:
Moving to 0.9.0.

andy-maier:
**Discussion point:** The `requests` package is able to maintain a socket pool and handle socket reuse automatically. Because we want to use that package, I don't think that we need a `close()` method on `WBEMConnection`.

KSchopmeyer:
What we do here depends on the package we use in the future.  If we go to requests, Andy's point above is correct.  Left open until we pick next ssl/ http package to work with

andy-maier:
This issue waits for the decision to move or not to move to using the requests package (see issue #280).

[...]

PR #2023, opened 2019-12-11
Fixes #280: Migrated HTTP communication to use requests

### LINKED ISSUE #280: Replace "httplib/m2crypto" with "requests
We are having trouble install pywbem on a fresh installed ubuntu.
The root cause is pywbem has dependency to m2crypto which turns out to have a bunch of dependencies that cannot be handled by "pip".
Could we just replace the httplib with the more popular "requests" lib?  As far as I could see, it greatly reduces the complexity of `cim_http.wbem_request` and it removes the dependency of m2crypto which make the installation on multiple platforms much easier than before.

Here are some details of the installation error.

```
building 'M2Crypto.__m2crypto' extension

swigging SWIG/_m2crypto.i to SWIG/_m2crypto_wrap.c

swig -python -D__x86_64__ -I/usr/include/python2.7 -I/usr/include -I/usr/include/x86_64-linux-gnu -includeall -modern -builtin -outdir build/lib.linux-x86_64-2.7/M2Crypto -o SWIG/_m2crypto_wrap.c SWIG/_m2crypto.i

unable to execute 'swig': No such file or directory

error: command 'swig' failed with exit status 1
```

This error could be resolved by `apt-get install swig`
Then I received many warnings from SWIG.  Here is just a part of it.

```
    SWIG/_bio.i:64: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_rand.i:21: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_evp.i:169: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_dh.i:36: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_rsa.i:43: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_dsa.i:31: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_ssl.i:241: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_ssl.i:242: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_x509.i:323: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_pkcs7.i:44: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_pkcs7.i:44: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_util.i:11: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_ec.i:111: Warning 454: Setting a pointer/reference variable may leak memory.
    SWIG/_engine.i:168: Warning 454: Setting a pointer/reference variable may leak memory.
    x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/include/python2.7 -I/usr/include -I/usr/include/x86_64-linux-gnu -I/tmp/pip_build_root/M2Crypto/SWIG -c SWIG/_m2crypto_wrap.c -o build/temp.linux-x86_64-2.7/SWIG/_m2crypto_wrap.o -DTHREADING
    SWIG/_m2crypto_wrap.c:7121:2: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
     *x509v3_lhash() {
      ^
    SWIG/_m2crypto_wrap.c: In function ‘x509v3_lhash’:
    SWIG/_m2crypto_wrap.c:7122:5: warning: return from incompatible pointer type [enabled by default]
         return lh_new(NULL, NULL); /* Should probably be lh_CONF_VALUE_new but won't compile. */
         ^
    SWIG/_m2crypto_wrap.c: In function ‘_wrap_new__STACK’:
    SWIG/_m2crypto_wrap.c:8379:1: warning: label ‘fail’ defined but not used [-Wunused-label]
     fail:
     ^
    SWIG/_m2crypto_wrap.c: In function ‘_wrap_sk_new_null’:
    SWIG/_m2crypto_wrap.c:8522:1: warning: label ‘fail’ defined but not used [-Wunused-label]
     fail:
```

And someone else in our team met a openssl header issue later.

andy-maier:
Ready for review and manual tests (see below for hints).

In addition to issue #280, it also addresses these issues:
* Return to a single SSL environment for both py2 and py3 (issue #496)
* Use of ssl.create_default_context() to auto-adjust to better security (issue #893)
* Hard coded default directories for CA certificates (issue #889)
* cim_operations appears to have no disconnect (issue #86)

**Note: If you want to test this PR with pywbemtools, it requires using PR https://github.com/pywbem/pywbemtools/pull/467.**

* Particular review points:
  - Changes to ca_cert parameter of WBEMConnection
  - Removal of communication via UNIX sockets
  - Removal of local authentication schemes
  - The new pywbem.PYWBEM_USES_REQUESTS attribute for indicating requests support to users

* Manual tests against real WBEM servers, including error behavior:
  - Different client envs:
     - Python 2 on Linux
     - Python 3 on Linux
     - Python 2 on native Windows
     - Python 3 on native Windows
  - tests for ca_certs:
    - No server certificate verification
    - Verification of self-signed server certificate with specified path to PEM file
    - Verification of self-signed server certificate with specified path to directory
  - tests for 2-way auth (x509 parm):
    - client certificate in PEM file, client private key in PEM file
    - client certificate and client private key in single PEM file (specified as 'cert_file')

* Tests that have been moved to new issue #2038:
    - Verification of CA-chain server certificate with specified path to PEM file
  - Verification of CA-chain server certificate with specified path to directory
  - Verification of CA-chain server certificate with certifi-provided root certificates


andy-maier:
On pypy3, make develop failed because typed-ast needs pgenheader.h and is being compiled .

andy-maier:
After reading through https://github.com/psf/requests/issues/2966 it seems that using system-provided certificates is a bigger undertaking than I first thought. We could move that out of this PR into a separate one. Using the certs from certifi would now be the default and probably also if and when we introduce support for system-provided certs.
@KSchopmeyer Karl, what do you think about this?

andy-maier:
The latest commit to this PR has removed the initial non-working code for system-provided certificates. That does not mean a decision is made. Let's discuss and decide then.

andy-maier:
The PR now has removed any support for system-provided certificates. For that, issue #2037 has been opened.

andy-ypsilon:
Is it possible configure proxies as an option, because in our envinronment direct communication is not possible very often. I use proxies argument within requests for REST API calls.

andy-maier:
We think that proxies should work without any changes in pywbem. Please try it out. You simply point pywbem to the proxy IP address. However, there are many different kinds of proxies, and we have not worked out which capabilities exactly are needed.

docs/client/security.rst
```diff
@@ -144,23 +142,18 @@ When using HTTPS, the TLS/SSL handshake protocol requires that the server always
...
+:exc:`pywbem.ConnectionError` with an according error message.
 
-If any of these validations fails, the WBEM operation methods of the
-:class:`~pywbem.WBEMConnection` object raise a :exc:`pywbem.AuthError`.
-
-If verification was disabled via the `no_verification` parameter, none of these
-validations of the server certificate happens.
+Validation of the server certificate can be disabled via the `no_verification`
+parameter of :class:`~pywbem.WBEMConnection`. Note that disabling certificate
+validation makes the communication of pywbem with the WBEM server vulnerable
```
KSchopmeyer: No need for the phrase "Noe that"

docs/client/security.rst
```diff
@@ -188,22 +181,13 @@ Authentication errors
...
-  - user is not authorized to access resource
-
-.. _`Default CA certificate paths`:
-
-Default CA certificate paths
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. autodata:: pywbem.cim_http.DEFAULT_CA_CERT_PATHS
+:exc:`pywbem.AuthError` when the WBEM server returns HTTP status 401
+"Unauthorized" and the retries in the client are exhausted. The server
```
KSchopmeyer: Should we mention retries in the client.  We removed that code and delegated any retries to requests and its subpackages.

docs/mof_compiler.help.txt
```diff
@@ -47,15 +47,12 @@ Connection security related options:
...
-                        Default: Searches for matching certificates in the
-                        following system directories:
-                        /etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt
-                        /etc/ssl/certs
-                        /etc/ssl/certificates
+  --cacerts cacerts     CA certificates to be used for verifying the server
+                        certificate presented by the WBEM server during TLS/SSL
+                        handshake:
+                        FILE: Use the certs in the specified cert file;
+                        DIR: Use the certs in the specified cert directory.
```
KSchopmeyer: Somewhere we should mention the c_rehash requirement for the directory.  Maybe the documentation.

pywbem/_features.py
```diff
@@ -0,0 +1,39 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+#
+
+"""
+Pywbem indicates support for version dependent features to users of its API.
+
+These are features that have been introduced at some point, and users may
+have a need to depent on the feature being supported (i.e. active).
```
KSchopmeyer: depent to depend

pywbem/_features.py
```diff
@@ -0,0 +1,39 @@
...
+If a feature is supported, the corresponding attribute is `True`. Note that
+before a feature was introduced, the corresponding attribute is undefined.
+
+The attributes listed here do not allow *controlling* whether a feature is
+active or not; they only *indicate* whether a feature is supported and active.
```
KSchopmeyer: change a feature to the feature

pywbem/cim_http.py
```diff
@@ -445,551 +233,161 @@ def wbem_request(url, data, creds, cimxml_headers=None, debug=False, x509=None,
...
-    key_file = None
-    cert_file = None
-
-    if use_ssl and x509 is not None:
-        cert_file = x509.get('cert_file')
-        key_file = x509.get('key_file')
-
-    local_auth_header = None
+    target = '/cimom'
+    target_url = '{}{}'.format(conn.url, target)
```
KSchopmeyer: Here is where I made temp change to use:  See separate comment since this was just a workaround to continue tests.


    target = '/cimom'
    host, port, ssl = parse_url(conn.url)
    target_url = '{}:{}{}'.format(host, port, target)

pywbem_os_setup.bat
```diff
@@ -36,6 +36,18 @@ goto end
 
 :do_install
 
+echo %myname%:
+echo %myname%: ================================================================
+echo %myname%:
+echo %myname%: Starting with pywbem 1.0.0, the installation of pywbem does not
```
KSchopmeyer: There may be some confusion over the word of installation since we really have two different installations:

1. user installation through pip

2. Development installation via clone.

We should clarify that.

KSchopmeyer:
NOTE: I also found the following code to log actual requests at the HTTPConnection level and inserted it for these tests with conn.debug around it.  (Not the whole context but the debug_requests_on.  Just a question but would it be logical to get it into our code with the debug,  or some other flag.  

def debug_requests_on():
    '''Switches on logging of the requests module.'''
    HTTPConnection.debuglevel = 1

    logging.basicConfig()
    logging.getLogger().setLevel(logging.DEBUG)
    requests_log = logging.getLogger("requests.packages.urllib3")
    requests_log.setLevel(logging.DEBUG)
    requests_log.propagate = True


def debug_requests_off():
    '''Switches off logging of the requests module, might be some side-effects'''
    HTTPConnection.debuglevel = 0

    root_logger = logging.getLogger()
    root_logger.setLevel(logging.WARNING)
    root_logger.handlers = []
    requests_log = logging.getLogger("requests.packages.urllib3")
    requests_log.setLevel(logging.WARNING)
    requests_log.propagate = False


@contextlib.contextmanager
def debug_requests():
    '''Use with 'with'!'''
    debug_requests_on()
    yield
    debug_requests_off()

KSchopmeyer:
Finally for Christmas day.  I also ran with Python 2 locally and passed the tests with these fixes. The remaining test is with windows.

KSchopmeyer:
We are getting OpenSSL information on errors with requests that are http based and should not involve SSL.

(pywbem3) C:\Users\kscho\dev\pywbem\examples>enuminstances.py
Usage: C:\Users\kscho\dev\pywbem\examples\enuminstances.py server_url username password namespace'               ' classname
Using internal defaults
Requesting url=http://localhost, ns=root/cimv2, class=CIM_ComputerSystem
TARGET_URL http://localhost:5988/cimom
Operation failed: Failed to establish a new connection: [WinError 10061] No connection could be made because the target machine actively refused it; OpenSSL version used: OpenSSL 1.0.2o  27 Mar 2018

KSchopmeyer:
**Testing status:** Wed 26 Dec
                 http     https/no-verify    https/ca-certfile   https/ca-certdir  mutual    certify
python 3   OK           OK                  OK                        OK                   OK

python 2   OK           OK                 OK                         OK                 OK

windows   OK           OK                OK

docs/client/security.rst
andy-maier: It is true that the retries are done in the "requests" package or its sub-packages,  but that is still in the WBEM client. I think the user ought to understand that the WBEM client performs retries.

docs/mof_compiler.help.txt
andy-maier: In the docs, we are referencing the corresponding OpenSSL docs section. Using c_rehash is not a requirement, it is just a convenient way to establish the required structure.

docs/client/security.rst
andy-maier: DONE

pywbem/_features.py
andy-maier: DONE

pywbem/_features.py
andy-maier: DONE

pywbem_os_setup.bat
andy-maier: With item 2, you mean "make install"? That also uses pip, just with the work directory as a source.
In any case, the statement applies to both ways of installing pywbem.

pywbem/cim_http.py
andy-maier: Has been done in WBEMConnection ctor.

andy-maier:
The latest update to this PR along with PR #2048 and the to be created one for the ssl messages for http connections should address all comments by Karl so far.

docs/mof_compiler.help.txt
andy-maier: We discussed this and agreed to add a reference to the docs, since it is not obvious that just putting them in a directory is not sufficient.

andy-maier: DONE: The last update does this.

KSchopmeyer:
Testing done using url-normalize branch

PR #2031, opened 2019-12-15
Fixes #2030: Clarified x509 parm of WBEMConnection; added checks and tests

### LINKED ISSUE #2030: x509 parm easy to get wrong
The x509 parameter of class WBEMConnection is rather complex and easy to get wrong. This is shown for example in [this Travis run of pywbemtools](andy/use-requests).

Some checking would help.

andy-maier:
See commit message.


KSchopmeyer:
Approved assuming it passes tests.

PR #2043, opened 2019-12-24
Fixes #2051: Test: Converted test_cim_xml.py to pytest; Fixed errors in cim_xml response elements

### LINKED ISSUE #2051: cim_xml.py: Add PARAMVALUE child to IMETHODRESPONSE
DSP0203 2.3 added PARAMVALUE* as a child to IMETHODRESPONSE, in support of pull operations. This needs to be added in cim_xml.py.

Note this is only relevant if pywbem code is used to build a WBEM server, which is not the case today. So this counts as cleanup rather than a fix.

Does not need to be rolled back.

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -333,6 +333,9 @@ Released: not yet
 * Test: Fixed an error in test_format_random() for the backslash character.
   (See issue #2027)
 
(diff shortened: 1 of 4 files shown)


andy-maier:
I've set this to priority because the fixes for issues #2052 and #2053 depend on it.

PR #2059, opened 2019-12-30
Partly fixes #2052: Partial support for DTD 2.4 req to set TYPE on KEYVALUE

### LINKED ISSUE #2052: CIM-XML: TYPE attribute of KEYVALUE is required in DTD 2.4
DSP0203 2.4.0 changed the TYPE attribute of KEYVALUE to be required. This is an incompatible change.

When creating CIM-XML, pywbem needs to ensure that the TYPE attribute is present to satisfy DTD 2.4. When parsing CIM-XML, pywbem should tolerate an absent TYPE attribute for compatibiity to DTD 2.3.

andy-maier:
See commit message.

No rollback to 0.16 planned.

See the **discussion point** in issue #2052.


KSchopmeyer:
I would propose that we commit what we have and make the discussion of adding typing info to ciminstance name to a separate issue.  Therefore, I marked this as approved by me but left the discussion flag on.

andy-maier:
I have set issue #2052 back to "partly fixed" with the discussion needed flag on, so we can discuss it there.

PR #2061, opened 2019-12-30
Converted unit tests of cim_http.py to pytest and added tests

andy-maier:


andy-maier:
DONE - I have added an entry to the change log.

PR #2068, opened 2020-01-08
Partly fixes #2062: Part 1 of refactor of pywbem_mock

### LINKED ISSUE #2062: Refactor pywbem mock to separate components

KSchopmeyer:
This is one part of a multipart set of PRs that will refactor pywbem mock.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -411,9 +411,33 @@ class FakedWBEMConnection(WBEMConnection, ResolverMixin):
 
     ################################################################
     #
-    #   Methods to insert data into mock repository
+    #   Methods to manage namespaces
     #
     ################################################################
+
+    def _validate_namespace(self, namespace):
+        """
+        Validate whether a namespace exists in the mock repository.
+
+        Parameters:
+
+          namespace (:term:`string`):
+            The name of the CIM namespace in the mock repository. Must not be
+            `None`.
+
+        Raises:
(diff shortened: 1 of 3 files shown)


andy-maier:
I have fixed the PyYAML install issue on Python 3.4. Please rebase this PR on master.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -270,7 +270,7 @@ class FakedWBEMConnection(WBEMConnection, ResolverMixin):
     """
     def __init__(self, default_namespace=DEFAULT_NAMESPACE,
                  use_pull_operations=False, stats_enabled=False,
-                 timeout=None, response_delay=None):
+                 timeout=None, response_delay=None, verbose=False):
```
andy-maier: Adding a "verbose" argument at an external interface goes beyond refactoring. Besides that, it is not used. I suggest to remove it.

KSchopmeyer: Removed.  Since I concluded that my previous approach to separating the repository data store from the CIMOM functions was too complex, I concluded that the API for the data store will be based on the current interface of dictionaries and once a CIMOM function (ex. get_class) validates that the namespace exists it can use dictionary methods on that repo.  There are still some function that do not do this but will clean them up.  Since most databases do support table to dictionary, this should work if the repository ever wants to use a database as the store.

Documented. that.  Next step is to move the repository to its own class.

PR #2074, opened 2020-01-09
Release 0.16.0

andy-maier:
The fix for the removal of Python 3.4 support in PyYAML 5.3 should be released, otherwise pywbem users have a show stopper issue on Python 3.4.
This motivates the release of pywbem 0.16.0 which contains that fix.


andy-maier:
wrong target branch - closing.

Issue #2041, opened 2019-12-24
Performance improvement for equality operator of CIM objects


andy-maier:
The current implementation of the `__eq__()` and `__ne__()` methods on CIM objects uses an internal method `_cmp()` on the CIM object, which is a traditional comparator function that determines less-than and larger-than in addition to equality. Since less-than and larger-than comparisons have been removed from the CIM object classes, the equality comparison could be made more efficient by determining only equality but no longer less-than or larger-than.

Note that this issue here proposes something different than #1062 which was abandoned.

andy-maier:
Addressed in PR #2091 and #2092, both for master.

PR #2091, opened 2020-01-26
Added performance tests for CIM object equality test

andy-maier:
This is a first stage of addressing issue #2041, in which a unittest module for performance measurements of the equality test for CIM objects is added, so that the original performance can be determined.


andy-maier:
Changed the implementation of the measurement function to use the Python "statistics" module instead of "numpy", since only mean and std dev are needed.

PR #2092, opened 2020-01-26
Fixes #2041: Improved performance of and cleaned up equality test for CIM objects

### LINKED ISSUE #2041: Performance improvement for equality operator of CIM objects

andy-maier:
This is the second stage of addressing issue #2041, in which the actual performance improvement gets implemented.

Original equality performance:
* CIMInstanceName with two keybindings, last one different: 14.077 us (stddev 5.2% in 100 runs with 14 outliers)
* CIMClass with 10 properties, last one different: 197.023 us (stddev 3.0% in 100 runs with 11 outliers)

Improved equality performance:
* CIMInstanceName with two keybindings, last one different: 12.952 us (stddev 4.6% in 100 runs with 11 outliers)
* CIMClass with 10 properties, last one different: 173.040 us (stddev 3.5% in 100 runs with 7 outliers)

As a result, this change is not a significant performance improvement, but it did clean up the code and eliminated the old comparator-based logic.

**Update** related to Karl's suggestion in a review comment to merge 4 lines of conditional into an expression:

This makes the improvement slightly better. Note that there are still variations in the measured result:

With the original code (4 lines of conditional):
* CIMInstanceName with two keybindings, last one different: 13.448 us (std.dev. 4.3% in 100 runs with 12 outliers)
* CIMClass with 10 properties, last one different: 176.183 us (std.dev. 2.8% in 100 runs with 14 outliers)

Latest code in this PR (4 lines condensed into one expression):
* CIMInstanceName with two keybindings, last one different: 13.703 us (std.dev. 6.1% in 100 runs with 12 outliers)
* CIMClass with 10 properties, last one different: 182.927 us (std.dev. 3.2% in 100 runs with 17 outliers)

### DIFF
--- a/pywbem/_utils.py
+++ b/pywbem/_utils.py
@@ -97,6 +97,52 @@ def _to_bytes(obj):
     return obj.encode("utf-8")
 
 
+def _eq_name(name1, name2):
+    """
+    Test two CIM names for equality.
+
+    The comparison is performed case-insensitively.
+
+    One or both of the names may be `None`; Two names that are `None` are
+    considered equal.
+    """
+    if name1 is None:
+        return name2 is None
+    else:
+        if name2 is None:
+            return False
+        else:
+            return name1.lower() == name2.lower()
+
+
+def _eq_item(item1, item2):
+    """
+    Test two items (CIM values, CIM objects) for equality.
+
+    One or both of the items may be `None`; Two items that are `None` are
+    considered equal.
+    """
+    if item1 is None:
+        return item2 is None
+    else:
+        if item2 is None:
+            return False
+        else:
+            return item1 == item2
+
+
+def _eq_dict(dict1, dict2):
+    """
+    Test two dictionary objects for equality.
+
+    The dictionaries must not be `None`; This is ensured by the respective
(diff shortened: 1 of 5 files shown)


pywbem/_utils.py
```diff
@@ -97,6 +97,65 @@ def _to__bytes(obj):
...
+    """
+    Test two CIM names for equality.
+
+    The comparison is performed case-insensitively.
+
+    One or both of the names may be `None`; Two names that are `None` are
+    considered equal.
+    """
+    if name1 is None:
+        if name2 is None:
```
KSchopmeyer: Would this be any faster if these 4 lines were one return false if None else True.   .??

andy-maier: It ended up being a little faster. See updated initial comment.

PR #2103, opened 2020-01-30
Fixes #2002: Switched to use SSLContext in WBEMListener

### LINKED ISSUE #2002: Listener uses deprecated SSL construct
Listener.py uses the deprecated ssl socket methd

server.socket = ssl.wrap_socket(server.socket,
                                                certfile=self._certfile,
                                                keyfile=self._keyfile,
                                                server_side=True)

This has been deprecated in favor of a similar function that uses SSLcontext.

The current python documentation for ssl.wrap_socket(...) includes the following:

Deprecated since version 3.7: Since Python 3.2 and 2.7.9, it is recommended to use the SSLContext.wrap_socket() instead of wrap_socket(). The top-level function is limited and creates an insecure client socket without server name indication or hostname matching.

SSLContext was added in python 2.7.9

See https://bugs.python.org/issue28022

https://bugs.python.org/issue24372

Note that this may cause problems in support of Python 2.7.9 and earlier.

andy-maier:
See commit message.
It would be good to test the listener using HTTPS.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -469,6 +469,12 @@ Released: not yet
   init parameter and attribute, utilizing the proxy support of the requests
   package. (see issue #2040)
 
+* Changed the HTTPS support of `pywbem.WBEMListener` from using the deprecated
+  `ssl.wrap_socket()` function to using the `ssl.SSLContext` class that was
+  introduced in Python 2.7.9. This causes more secure SSL settings to be used.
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Ran manual tests with both a pywbem listener and simulated indication generation and very limited with open pegasus.  I have found no errors and think we should commit this change.

APPROVED

PR #2105, opened 2020-01-31
Fixes #1925: Made all pywbem sub-namespaces private, except config

### LINKED ISSUE #1925: Remove the old Python sub-namespaces
Remove the old Python sub-namespaces (e.g. `pywbem.cim_obj`) and go with only `pywbem`. 

This includes the `pywbem.mof_compiler` namespace that was made available in the `pywbem`namespace as per issue #634.

It also includes renaming all sub-modules to have a leading underscore, in order to document they are not for customer use.

Because this is a potentially incompatible change, it cannot be rolled back before 1.0.0.

This issue was created from issue #635.

andy-maier:
See commit message.

**Discussion point: I think we can roll the renaming of the module files back to 0.17.0 if we add symbols for the original namespace names. The advantage would be that we can more easily roll back future changes, if the file names are the same**


andy-maier:
We agreed to roll this back to 0.17.0 (compatibly, by adding symbols for the old namespace names)

.gitignore
```diff
@@ -29,8 +29,8 @@ pylint_done
 cprofilerstats*.profile
 /build/
 /build_doc/
-mofparsetab.py
-moflextab.py
+*mofparsetab.py
```
KSchopmeyer: I don’t understand the asterisk on this and next line

andy-maier: The intention is to exclude both the new name `_mofparsetab.py` and the old name `mofparsetab.py` which can also still exist because an old branch had been checked out and generated the file.

If you are not happy with the asterisk approach to this, we can also specify both file names explicitly. Let me know.

andy-maier: Update: I changed this to use two entries for each these two files in `.gitignore`, instead of using wildcards, for better clarity. Also, I added removal of the old files in make clean (since I stumbled across this).
DONE

andy-maier:
@KSchopmeyer Pls approve if you agree with the resolution of the comment.

andy-maier:
Fixed docs references to `py:exceptions...` which by mistake had been renamed to `py:_exceptions...`

PR #2106, opened 2020-02-01
Fixes #1423: Fixes and context for HTTPError and CIMError exceptions and change to use ModelError

### LINKED ISSUE #1423: Context data for pywbem exceptions
Recent runs in the SNIA center showed that the information we have in exceptions that happen, is not optimal, particularly when the server is returning generic error messages.

The idea would be to add context and/or detail data to the pywbem exceptions that are raised.

For example:
- ParseError should have the CIM-XML reply
- CIMError should have the operation that was issued.

This list needs to be completed.

andy-maier:
See commit message.

See the discussion point in issue #1423. Will be done in a separate PR.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -145,6 +145,11 @@ Released: not yet
     for installing pywbem. If you have automated the pywbem installation,
     this step should be removed from your automation.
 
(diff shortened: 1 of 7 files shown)


andy-maier:
Update: See comment in issue #1423.

PR #2114, opened 2020-02-02
Fixes #805: Added support for garbage checking

### LINKED ISSUE #805: Need memory leak tests
All of our tests are short term and do not give any basis for determining if there are any memory leaks.

andy-maier:
See commit messages for details.

Invoke `make help` to understand how to run the memory leak tests, and look at the description of the new `leaktest` target and of the new `YAGOT*` env vars.

A complete `YAGOT=1 make test` runs for about 4h on my Mac, where without leak checking it takes about 4min. This seems to be caused by the many runs of the Python garbage collector that happen. Note that the `make leaktest` which runs just the specific leak tests is very fast.

Some tests that fail due to leaks have been excluded as XFAIL, and they can be run with: `TESTOPTS='--runxfail' make leaktest`.

**Discussion points (1): DONE**
* The implemented approach can be enabled for the pytest testcases, but the problem are the garbage objects pytest leaves in several situations (e.g. skipped testcase, failed testcase, use of pytest.raises()). -> Put on back burner for now
* Are independent memory leak testcases needed? -> Yes - DONE, with `make leaktest`.
* Should it be useable outside of the pytest context, e.g. in the manual test scripts? -> No
* Should a user of the pywbem library be able to enable it for their use of pywbem? -> No

**Discussion points (2): DONE**
* Find a non-leaking alternative to `collections.OrderedDict` we can use for `NocaseDict` on Python 2.7 (See https://stackoverflow.com/q/60134482/1424462). Starting with Python 3.2, `collections.OrderedDict` has been fixed to be leak-free. Note that the standard `dict` does not have leaks but is order-preserving only starting with CPython 3.6 and on other Python implementations starting with 3.7.
  -> DONE: decided to document this for python 2.7 and to run the leak tests only on python 3.
* Find a non-leaking alternative to `xml.dom.minidom` for use in cim_xml.py. It creates reference cycles even for simple usages. `xml.etree.ElementTree` did not have leaks at least in the few testcases I have added. If we stay with minidom: Change cim_xml.py and its users to no longer create standalone minidom objects but have them anchored under a Document object. 
See https://bugs.python.org/issue15290, https://bugs.python.org/issue4851. I was able to reproduce this issue: AttributeError "ownerDocument" when using minidom `Element.setAttribute()` on Python 3.7. Can be seen when running `test_leaks_standalone_Element_one_attribute()` with `--runxfail` on Python 3.7.
  -> DONE: decided to migrate from minidom to ElementTree, in a separate PR. Created issue #2121 to capture that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -215,6 +215,7 @@ script:
   # - if [[ -n $_MANUAL_CI_RUN ]]; then make pylint; fi
   - make pylint
   - make test
+  - make leaktest
 
 after_success:
 # Note: In case of error 422 (Couldn't find a repository matching this job),
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -15,6 +15,7 @@
 pytest>=4.3.1,<5.0.0; python_version < '3.5'
 pytest>=4.3.1; python_version >= '3.5'
 testfixtures>=6.9.0
+yagot>=0.5.0
 requests-mock>=1.6.0
 # lxml 4.4.0 removed Python 3.4 support.
 # lxml 4.4.1 added Python 3.8 support. However, while providing binary wheel
--- a/makefile
+++ b/makefile
@@ -313,9 +313,10 @@ help:
 	@echo "  builddoc   - Build documentation in: $(doc_build_dir)"
 	@echo "  check      - Run Flake8 on sources"
 	@echo "  pylint     - Run PyLint on sources"
-	@echo "  test       - Run unit and function tests"
+	@echo "  test       - Run unit and function tests (in tests/unittest and tests/functiontest)"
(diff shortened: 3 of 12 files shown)


andy-maier:
We agreed to invest a little on the independent memory tests, specifically for the cim objects, and only then come back to solving the larger problems with the entire test suite. Also, we don't think a public pywbem API should be provided for the leak checks.

andy-maier:
The last update adds specific leak tests for underlying classes that have leaks.

andy-maier:
We dicussed the options:
* For the leaks in `collections.OrderedDict`, we will document the issue but not solve it.

andy-maier:
The latest update adjusts to the latest changes in Yagot.

andy-maier:
Now using yagot from Pypi. Ready for review.

PR #2125, opened 2020-03-19
Fixes #2126 Add option to disable pull operations to mocker

### LINKED ISSUE #2126: Extend pywbem_mock to allow user to disable pull operations
The pull operations are an optional feature of wbem and as such we should reflect that optionality in the mocker.

This proposes adding a new property to WBEMConnection that would allow the user to disable the Open... , Pull..., operations.

The name would be forbid_pull and the property would be settable either from the constructor or the property itself.  When set True, it would cause any request to the Open... operations, and Pull... operations to return CIM_ERR_NOT_SUPPORTED when this attribute was set True.

This is considered an enhancement but it could be used now to test pywbemcli.

KSchopmeyer:
Add a property to Fake_WBEMConnection to forbid_pull which when set to
False (either in the __init__ or a property setter) forces all call to
the Open... request operations to return CIM_ERR_NOT_SUPPORTED.

Adds a test to validate property is set correctly and the operations
react correctly.

I set rollback because I would like to make that test part of pywbemcli.

ANDY: Forget discussion of the forbid word.  Changed the __init__ option to disallow-pull-operations. In case the question comes up, I do not want to merge the use_pull and disable_pull_operations (one is the client, the other is the server) and I specifically want to have a client that uses pull with server mock that disallows it to test client ability to handle error response if server mock cannot handle pulls.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -487,6 +487,10 @@ Released: not yet
   On Python versions before 2.7.9, pywbem will continue to use the deprecated
   `ssl.wrap_socket()` function. (See issue #2002)
 
+* Add property to pywbem_mock Fake_WBEMConnection to allow the user to modify
+  the mocker behavior to forbid the use of the pull operations.
+  (See issue #2126)
+
 **Cleanup:**
 
 * Test: Removed pinning of distro version on Travis to Ubuntu xenial (16.04)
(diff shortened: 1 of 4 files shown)


andy-maier:
Please rebase on master once PR #2127 has been merged, which fixes the issue on Travis with Python 3.4 and Pygments.

andy-maier:
**DIscussion:** The rollback to 0.17 has 12 conflicts in `pywbem_mock/_wbemconnection_mock.py`. If we do the rollback, I think you need to do that, Karl.

PR #2127, opened 2020-03-20
Updated minimum package versions for py34/py38

andy-maier:
No review needed.


andy-maier:
Rolled back into 0.5.1 via PR #2128.

PR #2132, opened 2020-03-24
Separate CIM repository from Fake_WBEMConnection

KSchopmeyer:
Creates a separate file that is the CIMRepository and contains all of
the Server operation request method responders.  _Fake_Wbemconnection will
continue to contain the direct call methods that are a part of setting
up a mock.

In addition, the InvokeMethod server request responder remains in
Fake_WBEMConnection since it is not part of any repository.

The mock code now calls the Server CIM repository request handlers in
the CIMRepository class.

NOTE: The server InvokeMethod responder remains in Fake_WBEMConnection since there is no method handler in most CIM repositories.

**NEXT PR part3b**
This pr separates the CIMRepository methods from the fake connection.  However, it still leaves those methods with an interface tied to the implementation of imethodcall. They are defined with a single input argument **params and return a tuple that is required by the mocker.  

The next PR is to separate the CIM repository methods from the mocker completely by expanding the work done in the mocker to make the APIs and returns from CIM repository be the CIM objects  and options themselves.  Since that is another significant code change (lots of small changes) I am separating it from this first PR..


andy-maier:
Please rebase on master to pick up PR #2127 which fixes the issue with Pygments on Python 3.4.

andy-maier:
Please rebase and resolve the conflicts.

PR #2133, opened 2020-03-26
Fixed incorrect coverage reported after pytest

andy-maier:
This change was rolled over from pywbemtools PR https://github.com/pywbem/pywbemtools/pull/550.
No review needed.


andy-maier:
I verified that in the latest master commit before this PR, on Python 3.8 with minimum package levels, the coverage was reported as 53% (see https://travis-ci.org/github/pywbem/pywbem/jobs/666222317), and in this PR, the coverage is now correctly reported as 87% (see https://travis-ci.org/github/pywbem/pywbem/jobs/667130601)

PR #2144, opened 2020-04-04
Fixed pylint issues; fixed issue in exc message

andy-maier:
See commit message.


andy-maier:
I assume removing the review needed label means approved :-)

PR #2151, opened 2020-04-09
Pinned mock version to support setup.py install from source tar

andy-maier:
See commit message.


andy-maier:
Closing this PR because its commit is included in PR #2152 and these commits really belong together.

PR #2152, opened 2020-04-09
Fixes #2150: Enabled setup.py install from source dist archive

### LINKED ISSUE #2150: requirements.txt is missing in Pypi's sdist archive
Hi,
is there a reason for not having requirements.txt in the sdist tar.gz at Pypi?

I made a [Gentoo ebuild](https://github.com/onkelbeh/HomeAssistantRepository/tree/master/dev-python/pywbem), which compiles without problems with the tar.gz here from github, which includes the requirements.txt. If I take the archive from Pypi (where the requirements.txt is missing), compilation fails:

```
>>> Compiling source in /var/tmp/portage/dev-python/pywbem-0.16.0/work/pywbem-0.16.0 ...
 * python3_7: running distutils-r1_run_phase distutils-r1_python_compile
python3.7 setup.py build -j 10
  File "setup.py", line 72, in <module>
    requirements = get_requirements('requirements.txt')
  File "setup.py", line 52, in get_requirements
    with open(requirements_file, 'r') as fp:
FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
 * ERROR: dev-python/pywbem-0.16.0::HomeAssistantRepository failed (compile phase):
```
These 2 archives should be identical or at least contain the same, I would prefer using an archiv from pypi in general, because this has certain advantages in Gentoo's build system.

andy-maier:
See commits.


minimum-constraints.txt
```diff
@@ -59,7 +59,9 @@ mock==2.0.0
 ordereddict==1.1
 ply==3.10
 PyYAML==5.1
-six==1.10.0
+six==1.10.0; python_version <= '3.7'
```
KSchopmeyer, line 62: Boy am I getting picky.  These kinds of comparisons are easier to understand and review of version number is same for both tests and use something like < rather than <= for compare.

Two occurances

minimum-constraints.txt
```diff
@@ -82,6 +84,10 @@ requests==2.20.1
 yagot==0.5.0
 requests-mock==1.6.0
 
+# Virtualenv
+virtualenv==14.0.0; python_version <= '3.7'
```
KSchopmeyer, line 88: same as above comment

minimum-constraints.txt
andy-maier: DONE

minimum-constraints.txt
andy-maier: DONE

andy-maier:
Did the requested change in all of:
* minimum-constraints.txt
* dev-requirements.txt
* requirements.txt

PR #2154, opened 2020-04-10
0.17: Pinned mock version to support setup.py install from source tar

andy-maier:
Rolls back PR #2151 into 0.17.1.


andy-maier:
Closing this PR because its commit is included in PR #2155 and these commits really belong together.

PR #2186, opened 2020-04-17
[WIP] Issue #2062: Create the DefaultInstanceProvider Class

KSchopmeyer:
WIP - separation complete and the provider registration defined.  Tests very limited.

1. Create a new class DefaultInstanceProvider and move CreateInstance
and DeleteInstance from MainProvider to this class.  The methods in this
class will be allowed to have user defined providers.  Also creats a class
ProviderDispatcher which provides request argument validation and routes the
request to the default provider unless a user provider is defined

2. Create the ProviderDispatcher class in DefaultInstanceProvider that
does the routing to the default provider for the method or to a user
defined provider and does standard parameter validation so that the default
or user provider do not have to do any input parameter validation.

3. Create BaseProvider which is the top level of the provider class hierarchy and contains those methods that are common to all of the subclasses.  

4. Create the DefaultInstanceProvider that contains the code to implement CreateInstance and
DeleteInstance if no use provider is defined.  Note that these methods have a significant amount of code removed and placed into ProviderDispatcher so that the method input parameters are validated in ProviderDispatcher before the request method in DefalutInstanceProvider or the user provider subclass is called.

5. Created the methods to 
    a. Register a user (FakedWBEMConnection.register_provider(...)
   b. get a registered provider from the registry and call that provider.

Created tests for the provider registration and the use of user instance providers for CreateInstance and ModifyInstance.

NOTE int this PR

1. Change DefaultInstanceProvider to InstanceWriteProvider

2. Change name of file _defaultinstanceprovider.py to something else.  Note that for now we are putting multiple classes into that file.

3. Extend register_provider to cover method providers

.4. Rewrite the Invokemethod code and create the MethodProviderClass

### DIFF
--- a/pywbem_mock/_baseprovider.py
+++ b/pywbem_mock/_baseprovider.py
@@ -1,0 +1,531 @@
+#
+# (C) Copyright 2018 InovaDevelopment.com
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+#
+# Author: Karl  Schopmeyer <inovadevelopment.com>
+#
+
+"""
+A CIM provider creates WBEM server responses to operations defined in DSP0200.
+
+The BaseProvider WBEM server provider handles provider required data that
+is common to both the MainProvider, DefaultInstanceProvider, and any
+registered instance providers including:
+
+  * Access to the CIM repository
+  * Access to the Method repository
+  * Local provider methods that are common to both the main provider and other
+    providers.
+  * CIMInstance request operations except for selected operations that are
+    handled by the BaseInstanceProvider to allow for specialized
+    instance providers to be implemented.
+  * Associator request operations
+
+This provider uses a separate CIM repository for the objects maintained by
+the CIM repository and defined with the interfaces in `BaseRepository`.
+"""
+
+from __future__ import absolute_import, print_function
+
+from copy import deepcopy
+
+from pywbem import CIMError, \
+    CIM_ERR_NOT_FOUND, \
+    CIM_ERR_ALREADY_EXISTS, CIM_ERR_INVALID_NAMESPACE, \
+    CIM_ERR_NAMESPACE_NOT_EMPTY
+
+from pywbem._utils import _format
+
+# pywbem_mock implementation configuration variables that are used in
+# request responsders.
+
+# Default Max_Object_Count for Fake Server if not specified by request
+_DEFAULT_MAX_OBJECT_COUNT = 100
+
+# Maximum Open... timeout if not set by request
+OPEN_MAX_TIMEOUT = 40
+
+# Per DSP0200, the default behavior for EnumerateInstance DeepInheritance
+# if not set by server.  Default is True.
+DEFAULT_DEEP_INHERITANCE = True
+
+# Value of LocalOnly parameter for instance retrevial operations.  This is
+# set to False in this implementation because of issues between different
+# versions of the DSP0200 specification that defined incompatible behavior
+# for this parameter that were resolved in Version 1.4 by
+# stating that False was the recommended Server setting for all instance
+# retrevial requests and that clients should use False also to avoid the
+# incompatibility.
+INSTANCE_RETRIEVE_LOCAL_ONLY = False
+
+
+class BaseProvider(object):
+    """
+    BaseProvider is the top level class in the provider hiearchy and includes
+    methods required by the subclasses such as MainProvider and
+    DefaultInstanceProvider.  This class is not intended to be executed
+    directly.
+    """
+
+    def __init__(self):
+        """
+        Set up dummy instance variables.
+        """
+        self.cimrepository = None
+
+    @property
+    def disable_pull_operations(self):
+        """
+        Boolean Flag to set option to disable the execution of the open and
+        pull operation request handlers in the mock CIM repository. This
+        emulates the characteristic in some CIM servers that did not
+        implement pull operations. The default is to allow pull operations.
+        All pull operations requests may be forbidden from executing by
+        setting disable_pull_operations to True.
+
+        This attribute is settable. For details, see the description of the
+        same-named init parameter of
+        :class:`this class <pywbem.FakedWBEMConnection>`.
+        """
+        return self._disable_pull_operations
+
+    @disable_pull_operations.setter
+    def disable_pull_operations(self, disable):
+        """Setter method; for a description see the getter method."""
+        # Attribute will always be boolean
+        if disable is None:
+            disable = False
+        if isinstance(disable, bool):
+            # pylint: disable=attribute-defined-outside-init
+            self._disable_pull_operations = disable
+        else:
+            raise ValueError(
+                _format('Invalid type for disable_pull_operations: {0!A}, '
+                        'must be a boolean', disable))
+
+    def __repr__(self):
+        return _format(
+            "MainProvider("
+            "cimrepository={s.cimrepository}, "
+            "host={s.host}, "
+            "disable_pull_operations={s.disable_pull_operations})",
+            s=self)
+
+    @property
+    def namespaces(self):
+        """
+        Return the namespaces that exist in the CIM repository.
+        """
+        return self.cimrepository.namespaces
+
+    ###############################################################
+    #
+    #   Access  to data store
+    #
+    #   The following methods provide access to the data store for
+    #   a single namespace. There are separate methods to provide
+    #   access to each of the 3 object stores within each namespace of
+    #   the CIM Repository CIMClass, CIMInstance, and CIMQualifierDeclaration.
+    #
+    ##################################################################
+
+    def get_class_store(self, namespace):
+        """
+        Returns the class object store for the specified CIM namespace within
+        the CIM repository.  This method validates that the namespace exists in
+        the data store.
+
+        This method does not validate that the namespace exists. Be certain
+        namespace is validated against CIM repository before calling this
+        method
+
+        Parameters:
+
+          namespace(:term:`string`):
+            The name of the CIM namespace in the CIM repository (case
+            insensitive). Must not be `None`. Leading or trailing
+            slash charactes are ignored.
+
+        Returns:
+
+          Instance of class derived from :class:`~pywbem_mock.`BaseObjectStore`
+          which is the object store for classes in the CIM repository.
+
+        """
+        return self.cimrepository.get_class_store(namespace)
+
+    def get_instance_store(self, namespace):
+        """
+        Returns the instance object store for the specified CIM namespace
+        within the CIM repository.  This method validates that the namespace
+        exists.
+
+        This method does not validate that the namespace exists. Be certain
+        namespace is validated against CIM repository before calling this
+        method
+
+        Parameters:
+
+          namespace(:term:`string`):
+            The name of the CIM namespace in the CIM repository (case
+            insensitive). Must not be `None`. Leading or trailing
+            slash characters are ignored.
+
+        Returns:
+
+           Instance of CIM Repository class
+           derived:class:`~pywbem_mock.`BaseObjectStore` which is the object
+           store for instances in the CIM repository.
+
+        Raises:
+
+          :exc:`~pywbem.CIMError`: CIM_ERR_INVALID_NAMESPACE: Namespace does
+            not exist.
+        """
+
+        return self.cimrepository.get_instance_store(namespace)
+
+    def get_qualifier_store(self, namespace):
+        """
+        Returns the qualifier declaration object store for the specified CIM
+        namespace within the CIM repository.  This method validates that the
+        namespace exists.
+
+        This method does not validate that the namespace exists. Be certain
+        namespace is validated against CIM repository before calling this
+        method
+
+        Parameters:
+
+          namespace(:term:`string`):
+            The name of the CIM namespace in the CIM repository (case
+            insensitive). Must not be `None`. Leading or trailing
+            slash characters are ignored.
+
+        Returns:
+
+          Instance of CIM Repository class derived
+          from:class:`~pywbem_mock.`BaseObjectStore` which is the object store
+          for qualifier declarations in the CIM repository.
+
+        Raises:
+
+          :exc:`~pywbem.CIMError`: CIM_ERR_INVALID_NAMESPACE: Namespace does
+            not exist.
+        """
+
+        return self.cimrepository.get_qualifier_store(namespace)
+
+    ################################################################
+    #
+    #   Methods to manage namespaces
+    #
+    ################################################################
+
+    def validate_namespace(self, namespace):
+        """
+        Validates that a namespace is defined in the CIM repository.
+        Returns only if namespace is valid. Otherwise it generates an
+        exception.
+
+          Parameters:
+
+            namespace (:term:`string`):
+              The name of the CIM namespace in the CIM repository (case
+              insensitive). Must not be `None`. Any leading or trailing
+              slash characters are ignored.
+
+          Raises:
+
+            CIMError: CIM_ERR_INVALID_NAMESPACE: Namespace does not exist.
+        """
+        try:
+            self.cimrepository.validate_namespace(namespace)
+        except KeyError:
+            raise CIMError(
+                CIM_ERR_INVALID_NAMESPACE,
+                _format("Namespace does not exist in CIM repository: {0!A}",
+                        namespace))
+
+    def add_namespace(self, namespace):
+        """
+        Add a CIM namespace to the CIM repository.
+
+        The namespace must not yet exist in the CIM repository.
+
+        Parameters:
+
+          namespace (:term:`string`):
+            The name of the CIM namespace in the CIM repository. Must not be
+            `None`. Any leading or trailing slash characters are removed before
+            the string is used to define the namespace name.
+
+        Raises:
+
+          ValueError: Namespace argument must not be None.
+          :exc:`~pywbem.CIMError`: CIM_ERR_ALREADY_EXISTS if the namespace
+            already exists in the CIM repository.
+        """
+
+        if namespace is None:
+            raise ValueError("Namespace argument must not be None")
+
+        try:
+            self.cimrepository.add_namespace(namespace)
+        except ValueError:
+            raise CIMError(
+                CIM_ERR_ALREADY_EXISTS,
+                _format("Namespace {0!A} already exists in the CIM repository ",
+                        namespace))
+
+    def remove_namespace(self, namespace):
+        """
+        Remove a CIM namespace from the CIM repository.
+
+        The namespace must exist in the CIM repository and must be empty.
+
+        Parameters:
+
+          namespace (:term:`string`):
+            The name of the CIM namespace in the CIM repository (case
+            insensitive). Must not be `None`. Leading or trailing
+            slash characters are ignored.
+
+        Raises:
+
+          ValueError: Namespace argument must not be None
+          :exc:`~pywbem.CIMError`:  CIM_ERR_NOT_FOUND if the namespace does
+            not exist in the CIM repository.
+          :exc:`~pywbem.CIMError`:  CIM_ERR_NAMESPACE_NOT_EMPTY if the
+            namespace is not empty.
+          :exc:`~pywbem.CIMError`:  CIM_ERR_NAMESPACE_NOT_EMPTY if attempting
+            to delete the default connection namespace.  This namespace cannot
+            be deleted from the CIM repository
+        """
+        if namespace is None:
+            raise ValueError("Namespace argument must not be None")
+
+        try:
+            self.cimrepository.remove_namespace(namespace)
+        except KeyError:
+            raise CIMError(
+                CIM_ERR_NOT_FOUND,
+                _format("Namespace {0!A} does not exist in the CIM repository ",
+                        namespace))
+        except ValueError:
+            raise CIMError(
+                CIM_ERR_NAMESPACE_NOT_EMPTY,
+                _format("Namespace {0!A} contains objects.", namespace))
+
+    ################################################################
+    #
+    #   Common Repository access methods used by MainProvider and
+    #   InstanceProviders
+    #
+    ################################################################
+
+    def get_class(self, namespace, classname, local_only=None,
+                  include_qualifiers=None, include_classorigin=None,
+                  property_list=None):
+        # pylint: disable=invalid-name
+        """
+        Get class from CIM repository.  Gets the class defined by classname
+        from the CIM repository, creates a copy, expands the copied class to
+        include superclass properties if not localonly, and filters the
+        class based on propertylist and includeClassOrigin.
+
+        It also sets the propagated attribute.
+
+        Parameters:
+
+          classname (:term:`string`):
+            Name of class to retrieve
+
+          namespace (:term:`string`):
+            The name of the CIM namespace in the CIM repository (case
+            insensitive). Must not be `None`. Leading or trailing slash
+            charactes are ignored.
+
+          local_only (:class:`py:bool`):
+            If `True`, or `None`only properties and methods in this specific
+            class are returned. `None` means not supplied by client and the
+            normal server default is `True`
+            If `False` properties and methods from the superclasses
+            are included.
+
+          include_qualifiers (:class:`py:bool`):
+            If `True` or `None`, include qualifiers.  `None` is the server
+            default if the parameter is not provided by the client.
+            If `False` do not include qualifiers.
+
+          include_classorigin (:class:`py:bool`):
+            If `True` return the class_origin attributes of properties and
+            methods.
+            If `False` or `None` (use server default), class_origin attributes
+            of properties and methods are not returned
+
+          property_list (list of :term:`string`):
+            Properties to be included in returned class.  If None, all
+            properties are returned.  If empty list, no properties are returned
+
+        Returns:
+
+          Copy of the CIM class if found with superclass properties
+          installed and filtered per the method arguments.
+
+        Raises:
+          CIMError: (CIM_ERR_NOT_FOUND) if class Not found in CIM repository or
+          CIMError: (CIM_ERR_INVALID_NAMESPACE) if namespace does not exist
+        """
+
+        class_store = self.get_class_store(namespace)
+
+        # Try to get the target class and create a copy for response
+        try:
+            cls = class_store.get(classname)
+        except KeyError:
+            raise CIMError(
+                CIM_ERR_NOT_FOUND,
+                _format("Class {0!A} not found in namespace {1!A}.",
+                        classname, namespace))
+
+        # Use deepcopy to assure copying all elements of the class.
+        # cls.copy() does not copy all elements.
+        cc = deepcopy(cls)
+
+        # local_only server default is True so True or None remove properties
+        if local_only is True or local_only is None:
+            for prop, pvalue in cc.properties.items():
+                if pvalue.propagated:
+                    del cc.properties[prop]
+            for method, mvalue in cc.methods.items():
+                if mvalue.propagated:
+                    del cc.methods[method]
+
+        self.filter_properties(cc, property_list)
+
+        # Remove qualifiers if specified.  Note that the server default
+        # is to include_qualifiers if include_qualifiers is None
+        if include_qualifiers is False:
+            self._remove_qualifiers(cc)
+
+        # class_origin default is False so None or False cause removal
+        if not include_classorigin:
+            self._remove_classorigin(cc)
+        return cc
+
+    def class_exists(self, namespace, classname):
+        """
+        Test if class defined by classname parameter exists in
+        CIM repository defined by namespace parameter.
+
+        Returns `True` if class exists and `False` if it does not exist.
+
+        Exception if the namespace does not exist
+        """
+        class_store = self.get_class_store(namespace)
+        return class_store.exists(classname)
+
+    @staticmethod
+    def filter_properties(obj, property_list):
+        """
+        Remove properties from an instance or class that aren't in the
+        property_list parameter
+
+        obj(:class:`~pywbem.CIMClass` or :class:`~pywbem.CIMInstance):
+            The class or instance from which properties are to be filtered
+
+        property_list(list of :term:`string`):
+            List of properties which are to be included in the result. If
+            None, remove nothing.  If empty list, remove everything. else
+            remove properties that are not in property_list. Duplicated names
+            are allowed in the list and ignored.
+        """
+        if property_list is not None:
+            property_list = [p.lower() for p in property_list]
+            for pname in obj.properties.keys():
+                if pname.lower() not in property_list:
+                    del obj.properties[pname]
+
+    @staticmethod
+    def find_instance(instance_name, instance_store, copy_inst=None):
+        """
+        Find an instance in the CIM repository by iname and return
+        that instance. the copy_inst controls whether the original
+        instance in the CIM repository is returned or a copy.  The
+        only time the original should be returned is when the user is
+        certain that the returned object WILL NOT be modified.
+
+        Parameters:
+
+          instance_name: CIMInstancename to find in the instance_store
+
+          instance_store (:class:`~pywbem_mock.BaseObjectStore):
+            The CIM repository to search for the instance
+
+          copy_inst: boolean.
+            If True do copy of the instance and return the copy. Otherwise
+            return the instance in the CIM repository
+
+        Returns:
+            None if the instance defined by iname is not found.
+            If it is found, the complete instance or copy is returned.
+
+        """
+
+        if instance_store.exists(instance_name):
+            if copy_inst:
+                return instance_store.get(instance_name).copy()
+            return instance_store.get(instance_name)
+
+        return None
+
+    def get_registered_provider(self, namespace, provider_type, classname):
+        """
+        If there is a provider registered for this namespace, provider_type,
+        and classname return that object (the instance of the
+        DefaultInstanceProvider subclass).
+
+        If no provider is registered, return None.
+
+        Parameters:
+
+          namespace (:term:`string`):
+            The namespace in which the request will be executed.
+
+          provider_type (:term:`string`):
+            String containing keyword ('instance' or 'method') defining the
+            type of provider.
+
+          classname (:term:`string`):
+            Name of the class defined for the operation
+        """
+
+        if not self.provider_registry:
+            return None
+
+        if classname not in self.provider_registry:
+            return None
+
+        if namespace not in self.provider_registry[classname]:
+            return None
+
+        tp = self.provider_registry[classname][namespace]
+
+        if provider_type not in tp:
+            return None
+
+        return tp[provider_type]
--- a/pywbem_mock/_defaultinstanceprovider.py
+++ b/pywbem_mock/_defaultinstanceprovider.py
@@ -1,0 +1,763 @@
+#
+# (C) Copyright 2018 InovaDevelopment.com
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+#
+# Author: Karl  Schopmeyer <inovadevelopment.com>
+#
+
+"""
+This module adds support for user-defined providers.  User defined
+providers may be created by the use of pywbem_mock to extend the capability
+of the mocker for special processing for certain classes to be mocked.
+
+This module contains two classes:
+
+1. ProviderDispatcher - Routes the calls for request methods that allow
+user-defined providers either to the default method in DefaultInstanceProvider
+or to a user defined method in the provider registry.
+
+2. DefaultInstanceProvider - The default implementation for the request
+responders that may include user defined providers.  This is the method that
+will be called if there is no provider registered for a namespace and class
+name.
+
+User defined providers may be created for specific CIM classes and specific
+namespaces to override one or more of the operation request methods defined in
+:class:`~pywbem_mock:DefaultInstanceProvider` with user methods defined in a
+subclass of :class:`~pywbem_mock:DefaultInstanceProvider`.
+
+A user defined provider is created as follows:
+
+1. Define the subclass of :class:`~pywbem_mock:DefaultInstanceProvider` with an
+__init__ method and the methods that will override any of the request methods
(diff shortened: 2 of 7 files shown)


pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+              The class defined in NewInstance  does not exist in the
+              namespace.
+        """
+
+        new_instance = NewInstance
+        if not isinstance(new_instance, CIMInstance):
+            raise CIMError(
+                CIM_ERR_INVALID_PARAMETER,
+                _format("NewInstance parameter is not a valid CIMInstance. "
+                        "Rcvd type={0}", type(new_instance)))
```
andy-maier, line 201: A check like this should not have to be handled by the provider. The provider is not the first stage where client calls are handled. It is the last stage just before the repository.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        self.validate_namespace(namespace)
```
andy-maier, line 203: see previous comment.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+                                           include_classorigin=True)
+        except CIMError as ce:
+            if ce.status_code != CIM_ERR_NOT_FOUND:
+                raise
+
+            raise CIMError(
+                CIM_ERR_INVALID_CLASS,
+                _format("Cannot create instance because its creation "
+                        "class {0!A} does not exist in namespace {1!A}.",
+                        new_instance.classname, namespace))
```
andy-maier, line 221: see previous comment.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        # Exception if property in instance but not class or types do not
+        # match
+        for iprop_name in new_instance:
+            if iprop_name not in target_class.properties:
+                raise CIMError(
+                    CIM_ERR_INVALID_PARAMETER,
+                    _format("Property {0!A} specified in NewInstance is not "
+                            "exposed by class {1!A} in namespace {2!A}",
+                            iprop_name, target_class.classname, namespace))
```
andy-maier, line 278: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+            cprop = target_class.properties[iprop_name]
+            iprop = new_instance.properties[iprop_name]
+            if iprop.is_array != cprop.is_array or \
+                    iprop.type != cprop.type:
+                raise CIMError(
+                    CIM_ERR_INVALID_PARAMETER,
+                    _format("Instance and class property {0!A} types "
+                            "do not match: instance={1!A}, class={2!A}",
+                            iprop_name, iprop, cprop))
```
andy-maier, line 288: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
andy-maier, line 203: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+            target_class,
+            new_instance,
+            namespace=namespace)
+
+        # Check for duplicate instances
+        if instance_store.exists(new_instance.path):
+            raise CIMError(
+                CIM_ERR_ALREADY_EXISTS,
+                _format("NewInstance {0!A} already exists in namespace "
+                        "{1!A}.", new_instance.path, namespace))
```
andy-maier, line 315: This check is already done in `instance_store.create()`, raising `ValueError`, so instead of doing this check it could be an exception handling around `instance_store.create()` that transforms the ValueError into a CIMError.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        Raises:
+
+            CIMError:  CIM_ERR_INVALID_NAMESPACE, CIM_ERR_INVALID_PARAMETER
+              CIM_ERR_NOT_FOUND
+        """  # noqa: E501
+        # pylint: enable=line-too-long
+
+        namespace = InstanceName.namespace
+        self.validate_namespace(namespace)
```
andy-maier, line 425: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+        # Set LocalOnly to False as fixed value.
+        LocalOnly = INSTANCE_RETRIEVE_LOCAL_ONLY
+
+        instance_store = self.get_instance_store(namespace)
+
+        if not self._class_exists(namespace, iname.classname):
+            raise CIMError(
+                CIM_ERR_INVALID_CLASS,
+                _format("Class {0!A} for GetInstance of instance {1!A} "
+                        "does not exist.", iname.classname, iname))
```
andy-maier, line 437: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        return self._get_instance(iname, instance_store,
+                                  LocalOnly,
+                                  IncludeClassOrigin,
+                                  IncludeQualifiers, PropertyList)
```
andy-maier, line 442: This method is missing in the class hierarchy in the current code.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        namespace = InstanceName.namespace
+        self.validate_namespace(namespace)
+
+        iname = InstanceName
+
+        # Validate namespace and get instance CIM repository
+        instance_store = self.get_instance_store(namespace)
+
+        if not self._class_exists(namespace, iname.classname):
```
andy-maier, line 496: This method is missing in the class hierarchy in the current code.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+        namespace = InstanceName.namespace
+        self.validate_namespace(namespace)
+        iname = InstanceName
+
+        # Set LocalOnly to False as fixed value.
+        LocalOnly = INSTANCE_RETRIEVE_LOCAL_ONLY
+
+        instance_store = self.get_instance_store(namespace)
+
+        if not self._class_exists(namespace, iname.classname):
```
andy-maier, line 433: This method is missing in the class hierarchy in the current code.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+                CIM_ERR_INVALID_PARAMETER,
+                _format("NewInstance parameter is not a valid CIMInstance. "
+                        "Rcvd type={0}", type(new_instance)))
+
+        self.validate_namespace(namespace)
+        instance_store = self.get_instance_store(namespace)
+
+        # Requires corresponding class to build path to be returned
+        try:
+            target_class = self._get_class(namespace,
```
andy-maier, line 208: This method is missing in the class hierarchy in the current code.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+            * `host`: value ignored.
+
+        Raises:
+            CIMError: CIM_ERR_INVALID_NAMESPACE
+            CIMError: CIM_ERR_INVALID_CLASS
+            CIMError: CIM_ERR_NOT_FOUND
+        """
+
+        namespace = InstanceName.namespace
+        self.validate_namespace(namespace)
```
andy-maier, line 489: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        # Validate namespace and get instance CIM repository
+        instance_store = self.get_instance_store(namespace)
+
+        if not self._class_exists(namespace, iname.classname):
+            raise CIMError(
+                CIM_ERR_INVALID_CLASS,
+                _format("Class {0!A} in namespace {1!A} not found. "
+                        "Cannot delete instance {2!A}",
+                        iname.classname, namespace, iname))
```
andy-maier, line 501: This check could be done before calling the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+        if not instance_store.exists(iname):
+            raise CIMError(
+                CIM_ERR_NOT_FOUND,
+                _format("Instance {0!A} not found in CIM repository namespace "
+                        "{1!A}", iname, namespace))
```
andy-maier, line 507: This check could be done before calling the provider.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -305,6 +314,11 @@ def __init__(self, default_namespace=DEFAULT_NAMESPACE,
         self._mofwbemconnection = _MockMOFWBEMConnection(self,
                                                          self.mainprovider)
 
+        # TODO: This is just prototype. The registry would be
+        # namespace, classname, providersubclassinstance.
+        # Should be in BaseRepository or DefaultInstanceProvider.
```
andy-maier, line 319: I don't think the provider registry should be in BaseRepository or DefaultInstanceProvider. It belongs where the dispatching of provider calls is done, and that should be here in this class.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1009,32 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+
+        assert classname is not None
+
+        # TODO validate that classname is in repository
+
+        if namespace not in self.provider_registry:
+            self.provider_registry[namespace] = namespace
+        if classname not in self.provider_registry[namespace][classname]:
+            self.provider_registry[namespace][classname] = provider
+        print("PROVIDERREG %r" % self.provider_registry)
```
andy-maier, line 1036: I think what you want are two levels of dicts: key namespace -> key classname -> value provider object.
If so, the logic in the lines above seems wrong. Did you mean this:
```
       if namespace not in self.provider_registry:
            self.provider_registry[namespace] = NocaseDict()
        if classname not in self.provider_registry[namespace]:
            self.provider_registry[namespace][classname] = provider
```

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1009,32 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
                     else:
                         _uprint(dest, obj.tomof())
 
+    def register_provider(self, namespace, classname, provider):
+        """
+        Register the provider object for namespace and class.
+        TODO: The registration data must be in the providers somewhere since
+        there is no link back to this class from the providers.
```
andy-maier, line 1016: I don't think so. The providers should be light-weight because there are many user-written providers. The providers do not need to know that or where they are registered. For example, if a particular instance provider is registered for two classes, it will be called for these two classes and sees which class a call is for at the time it is called. I don't think it needs to know when handling class A that it is also registered for class B.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1009,32 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
                     else:
                         _uprint(dest, obj.tomof())
 
+    def register_provider(self, namespace, classname, provider):
+        """
+        Register the provider object for namespace and class.
```
andy-maier, line 1014: Right now, we are only supporting user-provided instance providers. We have two options:
* Do not yet introduce a notion of provider types. In that case, this function should document that it registers instance providers (only).
* Introduce a notion of provider types already now, but support only instance providers at this point.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,527 @@
...
+
+from ._baseprovider import BaseProvider
+
+from ._baseprovider import INSTANCE_RETRIEVE_LOCAL_ONLY
+
+# None of the request method names conform since they are camel case
+# pylint: disable=invalid-name
+
+
+class ProviderDispatcher(BaseProvider):
```
andy-maier, line 63: It seems strange that a provider dispatcher should itself be a provider.

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Moved to ProviderDispatcher. DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Moved to ProviderDispatcher. DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Already commented above

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Fixed.  DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Fixed DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Fixed DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Fixed DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE.  NOTE: The GetInstance should not have been in the DefaultInstanceProvider but in MainProvider.  It was in both. Removed from DefaultInstanceProvider

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: I know.   However, with your earlier comments we have 2 things to do before we hand over to the provider.

1. Decide which provider (default or user)

2. Handle the validation processing (your comments).  To do this, the container has to include access to the methods that actually validate namespace, validate class exists, etc. That code, in turn requires access to the cimrepository. 

One Alternative I was thinking about was to create new methods in the DefaultInstanceProvider that would be something like CreateInstanceDispatcher and DeleteInstanceDispatcher,  That would eliminate the ProviderDispatcher completely. Each of these methods would 1. Validate request parameters, 2. Determine if the default is to be called or the user provider.  

**DISCUSSION**

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: These are all Instance providers, The remaining type is really the Method provider which is a significantly different bird I am afraid.  I will attack that tomorrow.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: If you want to register a provider for multiple classes we can do either:

1. Chang class to list of classes.

2, The user files a registration for each class.

DISCUSSION

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: You are right.   FIX

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Cannot do that because that means maintaining reference to WBEMConnection class in MainProvider/BaseProvider/DefaultInstanceProvider.  That causes python circular reference.
**DISCUSSION**  The choices are live with circular reference or create attribute with cimrepository in the provider classes.   There is even another issue here in that we have to do this for  MainProvider and DefaultInstanceProvider  and ProviderDispatcher.  I am beginning to think the circular reference is not so bad.

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE

pywbem_mock/_defaultinstanceprovider.py
andy-maier: That's actually right, I agree.

pywbem_mock/_wbemconnection_mock.py
andy-maier: The circular reference is not an actual problem at runtime, it is more a design flaw if one needs that.

pywbem_mock/_wbemconnection_mock.py
andy-maier: right, all user provided providers are instance providers of some sort. Still, the C-U-D type, the method type, the association traversal type, the read-enum type are all different types. So when I said "instance provider" in my comment, I meant to say C-U-D instance provider. Anyway, it is just a comment about wording, so let's wait how it looks like when you have the method provider.

pywbem_mock/_defaultinstanceprovider.py
andy-maier: I think the provider dispatcher should also have a reference to the repository.

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE in ProviderDispatcher

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Moved to ProviderDispatcher

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Good catch.  Modified. DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: It is now done in the ProviderDispatcher method.  DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: I just concluded late last night the the design of the provider registery should be:

dict of classes
       [namespaces] : provider

and the register_provider parameters are list of namespaces and list of classes.

The class must be there. The namespaces is optional

Further the provider  can be expanded to "instance provider", "method "provider" 

WILL fix this morning and upload along with first examples

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: I think I sorted that out in my mind late last night.  The provider registry should hiearchy of dictionaries where the top level is class name and within each classname are dict of namespaces, and dicts to represent each provider type if that provider exists for the defined class and namespaces.  Will try to create that this morning

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: It does now.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+        """
+        provider_types = ('instance', 'method')
+
+        if provider_type not in provider_types:
+            raise ValueError("{0!A} is not a valid provider type. "
+                             "Valid provider types are {1!A}.", provider_type,
+                             provider_types)
+
+        if not isinstance(provider, DefaultInstanceProvider):
+            raise ValueError(
```
andy-maier, line 1055: This should be a TypeError, since we are reporting an invalid type.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+
+        if classnames is None:
+            raise ValueError(
+                _format('register_provider classnames argument '
+                        'must be string or list '
+                        'of strings. None not allowed.'))
+
+        if namespaces is not None:
+            if not isinstance(namespaces, (list, tuple, six.string_types)):
+                raise ValueError(
```
andy-maier, line 1067: This should be a TypeError, since we are reporting an invalid type.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+        provided with this call is to be executed as the request response
+        method for the namespace and class defined in lieu of the default
+        provider.
+
+        Providers can only be registered for request response methods
+        CreateInstance and DeleteInstance and InvokeMethod
+
+        Parameters:
+
+          namespace (:term:`string or :class:`py:list` of :term:`string):
```
andy-maier, line 1029: Change `namespace` to `namespaces` to match the actual parameter name.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+            Namespace or namespaces for which the provider is being registered.
+            If empty or None, the classname is registered for all namespaces
+
+          classname (:term:`string or :class:`py:list` of :term:`string):
```
andy-maier, line 1033: ditto: `classnames`

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+            Classname or classnames for which the provider is being
+            registered. At least one classname is required
+
+          # TODO Requirement is that it is a class that is a subclass of
+          # DefaultInstanceProvider
+          provider (:class:):
```
andy-maier, line 1039: I think in order to make the provider types and which operations they can implement more clear to users, we should have a separate class for each provider type. These classes have the default implementation and only those operation methods that can be overridden, plus they identify the provider type, so that a separate provider_type parameter is not needed when registering the provider. These classes can have a provider_type attribute that is set in the base class init.

Finally, I think we should name our C-U-D instance provider not just instance provider but something that expresses this subset of instance operations.

Proposal for the class names of the default provider classes user providers would inherit from:
* `InstanceWriteProvider` - C-U-D
* `MethodProvider` - InvokeMethod (both static and non-static)

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+        if not isinstance(provider, DefaultInstanceProvider):
+            raise ValueError(
+                _format("{0!A} is not a valid subclass of "
+                        "DefaultInstanceProvider. ", provider))
+
+        if classnames is None:
+            raise ValueError(
+                _format('register_provider classnames argument '
+                        'must be string or list '
+                        'of strings. None not allowed.'))
```
andy-maier, line 1063: Instead of just excluding None, I think this should test for the allowed types, and raise TypeError.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+
+            if isinstance(namespaces, six.string_types):
+                namespaces = [namespaces]
+
+            for namespace in namespaces:
+                if namespace not in self.namespaces:
+                    raise ValueError(
+                        _format("register_provider namespaces argument "
+                                '"{0!A}" does not exist in {1!A} '
+                                'namespaces.', namespace, namespaces))
```
andy-maier, line 1080: I suggest to change this exception message to: 
```
                        _format('Namespace "{0!A}" does not exist in the '
                                'CIM repository', namespace))
```

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+        classname = classnames
+        assert isinstance(classname, six.string_types)
+
+        if namespaces:
+            for namespace in namespaces:
+                if not self.mainprovider.class_exists(namespace, classname):
+                    raise ValueError(
+                        _format("register_provider classname argument "
+                                'class "{0!A}" class does not exist in '
+                                'namespace {1!A}', classname, namespace))
```
andy-maier, line 1098: Proposal for exception message:
```
                        _format('Class "{0!A}" does not exist in namespace '
                                '"{1!A}" of the CIM repository',
                                classname, namespace))
	
```

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+
+        if classname in self.provider_registry:
+            class_reg = self.provider_registry[classname]
+            if provider_type in class_reg:
+                raise KeyError(
+                    _format("Classname {0!A} cannot register multiple times"
+                            'for provider type {1!A}.',
+                            classname, provider_type))
```
andy-maier, line 1106: I think this restriction should only apply for a combination of namespace and class name. It is valid to register one provider for class A in namespace X and another provider for class A in namespace Y.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+                _format('register_provider classnames argument '
+                        'must be string or list '
+                        'of strings. None not allowed.'))
+
+        if namespaces is not None:
+            if not isinstance(namespaces, (list, tuple, six.string_types)):
+                raise ValueError(
+                    _format('register_provider namespaces argument '
+                            '"{0!A}" not list or string'
+                            'of strings.', namespaces))
```
andy-maier, line 1070: Proposal for exception message. Since this exception will likely not be handled, its traceback already shows the function that raised it.
```
                    _format('namespaces argument must be a string '
                            'or list/tuple of strings, but is {0}',
                            type(namespaces)))
```

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,439 @@
...
+        """
+        """
+        # TODO: This not correct way to make cimrepository available to
+        # multiple subclasses
+        super(ProviderDispatcher, self).__init__(
+            conn=None, cimrepository=None)
+        self.default_instance_provider = default_instance_provider
+
+        self.cimrepository = conn.cimrepository
+        self.provider_registry = conn.provider_registry
```
andy-maier, line 118: Instead of passing in `conn` and then using exactly two attributes from it, I suggest to pass in these two attributes. That avoids the back reference and makes it more clear what information is actually passed.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,439 @@
...
+
+        If no provider is registered, return None
+        """
+        if self.provider_registry:
+            if classname not in self.provider_registry:
+                return None
+            if self.provider_registry[classname]['providertype'] != provider_type:
+                return None
+            if not self.provider_registry[classname]['namespaces']:
+                return None
```
andy-maier, line 134: Not sure about this. Are you allowing namespaces = None to indicate that it is registered for all namespaces? In that case, it should return the provider.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,439 @@
...
+    def __init__(self, conn):
+        """
+        """
+
+        # TODO: This not correct way to make cimrepository available to
+        # multiple subclasses
+        super(DefaultInstanceProvider, self).__init__(
+            conn=conn, cimrepository=None)
+
+        self.cimrepository = conn.cimrepository
```
andy-maier, line 236: Instead of passing in `conn` and then using exactly one attribute from it, I suggest to pass in that one attribute. That avoids the back reference and makes it more clear what information is actually passed.

pywbem_mock/_baseprovider.py
```diff
@@ -0,0 +1,479 @@
...
+            self.host = conn.host
+            self.disable_pull_operations = None
+
+        # Implementation of the CIM object repository that is the data store
+        # for CIM classes, CIM instances, CIM qualifier declarations and
+        # CIM methods.
+        # See :py:module:`pywbem_mock/baserepository` for a description of
+        # the repository interface.
+        if cimrepository:
+            self.cimrepository = cimrepository
```
andy-maier, line 105: I don't like this logic. The derived classes also get a `cimrepository`, set the `cimrepository` argument to `None` so that this base class does not set it, and then set `self.cimrepository` in the derived class. I think this is equivalent to just passing `cimrepository` on to this base class where `self.cimrepository` is then set unconditionally.

pywbem_mock/_baseprovider.py
```diff
@@ -0,0 +1,479 @@
...
+              The current instance of the connection from which some
+              variables are extracted for the repository including
+              the host and disable_pull_operations attributes.
+
+            cimrepository(derived from `~pywbem_mock.BaseRepository`):
+             the CIM repository that stores CIM objects.
+        """
+
+        #   Save required attributes from the connection object
+        if conn:
```
andy-maier, line 95: Same comment as below on the logic of conditionally setting the attributes.

pywbem_mock/_baseprovider.py
```diff
@@ -0,0 +1,479 @@
...
+            self.host = conn.host
```
andy-maier, line 96: Instead of passing in `conn` and then using exactly one attribute from it, I suggest to pass in that one attribute. That avoids the back reference and makes it more clear what information is actually passed.

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,439 @@
...
+        """
+        # Validate input parameters
+
+        self.validate_namespace(namespace)
+
+        if not isinstance(NewInstance, CIMInstance):
+            raise CIMError(
+                CIM_ERR_INVALID_PARAMETER,
+                _format("NewInstance parameter is not a valid CIMInstance. "
+                        "Rcvd type={0}", type(NewInstance)))
```
andy-maier, line 156: In CIM-XML, this would be detected before the request is sent to the server because the invalid type cannot be represented in CIM-XML. I therefore believe this should result in a TypeError (but I did not check what we do for CIM-XML).

pywbem_mock/_defaultinstanceprovider.py
```diff
@@ -0,0 +1,439 @@
...
+            ns_classname = None
+        if ns_classname:
+            namespace = InstanceName.keybindings['Name']
+
+            # Reflect the namespace deletion in the CIM repository
+            # This will check the namespace for being empty.
+            self.remove_namespace(namespace)
+
+        # Delete the instance from the CIM repository
+        instance_store.delete(InstanceName)
```
andy-maier, line 439: This provider should also handle ModifyInstance.

pywbem_mock/_wbemconnection_mock.py
andy-maier: Covered the relevant parts of this comment with new comments. CLosing this one.

pywbem_mock/_wbemconnection_mock.py
andy-maier: The implemented provider registration allows for lists of classes and lists of namespaces. That is fine with me, and I think this discussion is resolved. Let me know if you still want to discuss this.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -995,6 +1012,104 @@ def _display_objects(obj_type, object_repo, namespace, cmt_begin, cmt_end,
...
+            if provider_type in class_reg:
+                raise KeyError(
+                    _format("Classname {0!A} cannot register multiple times"
+                            'for provider type {1!A}.',
+                            classname, provider_type))
+
+        self.provider_registry[classname] = NocaseDict()
+        self.provider_registry[classname]['namespaces'] = namespaces
+        self.provider_registry[classname]['providertype'] = provider_type
+        self.provider_registry[classname]['provider'] = provider
```
andy-maier, line 1111: The code above allows for namespaces = None, but it is not entirely clear to me what this means. Please add a comment that describes what it means (all namespaces?)
See also my other comment about using it.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: We do that in second test in the recursion with each classname.  However, This initial test is needed since we do not want to do the recursion with classname None.

**DISCUSSION**

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: DONE.  Note that I have been putting the method name as the beginning of each message and I kept that for this message

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: DONE.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Done but added the actual argument value.

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: This is getting messy.  First, we have to pass cim_repository to each of the subclasses but some form of self.cim_repository must exist in BaseProvider just to avoid lint errors since it is referenced.

I am passing conn to the main provider only so it can be passed to the BaseProvider just to extract host and disable_pull_operations which are used in some of the methods.

Proposed solution.

1. MainProvider pass host, disable_pull_operation, cim_repository to MainProvider. 

2. DefaultInstanceProvider Pass cim_repository

3. ProviderDispatcher .CIMRepository, provider_registry, instance of DefaultInstanceProvider

**DONE:** For next commit

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: After your proposal to allow different providers for different namespaces, I think that the combination of that and the idea of all namespaces are largely incompatible since we how have to allow the namespace to be specifically defined as a dictionary entry in the provider_registry. Therefore, I propose to drop that as a requirement and force all desired namespaces to be defined.

TODO: Not done yet

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: **DONE**

pywbem_mock/_baseprovider.py
KSchopmeyer: **DONE**

pywbem_mock/_baseprovider.py
KSchopmeyer: **DONE**

pywbem_mock/_baseprovider.py
KSchopmeyer: **DONE**

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Agree that in CIM-XML this would be detected. So the dispatcher really does not even need to test for it because the client/imethodcall/us also checks for it.  However, maybe it should be in the DefaultInstanceProvider since the user provider could make a mess and hand it back to the default provider,

**DISCUSSION:**

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: My Original goal was namespaces None means allow all namespaces.   However, with the requirement to have separate providers in separate namespaces, that gets really messy.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: That one tore up the provider repository structure.  I think I will have to a) drop the any namespace unliess I can find some special flag for it.  b) modify he repository so that it has an additional level for namespace to allow 

TODO

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Ahha. The part you did not catch is  that I had not put the ModifyInstance into the DefaultInstanceProvider.  That done for next update.

1. Agreed that MethodProvider is a completely separate beast.

2. RE: InstanceWriteProvider.. If that is name, should it not be DefaultInstanceWriteProvider?

**DISCUSSION**

I think that the idea of provider_type in register_provider makes sense. Otherwise we need to sort that out in some other way. , Create a separate registry for each provider type. Even then, the provider_type on register_provider makes it easier to define which provider_registry would be used.

On the dispatcher side, I want to be careful until I do the method provider this afternoon.

**TODO**

KSchopmeyer:
Created new set of commits in response to some of the comments.  Note that we did not resolve comments.  see TODOs and DISCUSSION items in the above comments. 

This update consists of a number of smaller commits each attacking one or more of the issues.

pywbem_mock/_defaultinstanceprovider.py
andy-maier: Ok

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Happy to close DONE

pywbem_mock/_wbemconnection_mock.py

pywbem_mock/_wbemconnection_mock.py

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: DONE. In commit Tuesday night I rewrote register_provider and get_registered_provider. Now we handle multiple providers for single class in multiple namespaces.  As part of that I removed the capability to register for all namespaces.

**DISCUSSION**

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: I agree. Marked this one resolved

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: **DONE**

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Invalid type. string or list of strings allowed as message.

Check test should be string or list.

TODO

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: We agree with this conclusiong.  DONE

pywbem_mock/_defaultinstanceprovider.py
KSchopmeyer: Agree change to TypeError

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Completely rewritten.

andy-maier:
Please squash into one commit, and on the unresolved comments mark when they are done.

pywbem_mock/_wbemconnection_mock.py

pywbem_mock/_wbemconnection_mock.py
andy-maier: Refactoring the method_callback_interface to a method provider is left to be done in a subsequent PR.

pywbem_mock/_wbemconnection_mock.py
andy-maier: Was not changed. Marking the comment as resolved - this will be reviewed further.

PR #2209, opened 2020-04-30
Regularize use of deepcopy for objects xfrd to and from repository

KSchopmeyer:
Since the .copy method for CIMClass and CIMInstance does not do a deep
copy and we want to insure that we the objects in the dictionary are
truly different than the ones in the input and maintained in the
repository we changed to use deepcopy to copy all objects before they
are placed into the dictionary and for all objects that are extracted
from the dictionary with the copy property.

Note that we were doing extraneous copies in _baserepository and so
removed them.

Adds tests for cim repository to assure that the objects are copied
on create, modify, get with copy and iter_value with copy.

### DIFF
--- a/pywbem_mock/_baseprovider.py
+++ b/pywbem_mock/_baseprovider.py
@@ -40,8 +40,6 @@ the CIM repository and defined with the interfaces in `BaseRepository`.
 
 from __future__ import absolute_import, print_function
 
-from copy import deepcopy
-
 from pywbem import CIMError, \
     CIM_ERR_NOT_FOUND, \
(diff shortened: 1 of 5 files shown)


andy-maier:
The copy() methods of pywbem CIM objects do a (manual) deep copy on mutable attributes, and no copy on unmutable attributes. IMO, that separates origial and copy sufficiently in that it prevents that changes on the original can be made through the copy. Am I missing anything here?

andy-maier:
This now fails due to issue #2214. Please rebase on master after PR #2215 is merged.

KSchopmeyer:
Marked waiting. Waiting commit of pr #2215

andy-maier:
Can continue.

PR #2211, opened 2020-05-02
Fixes #2062: part4 compiler - refactor compiler interface in mocker

### LINKED ISSUE #2062: Refactor pywbem mock to separate components

KSchopmeyer:
Refactored the code in _MockMOFWBEMConnection to add things and get them from the cim repository through the WBEMConnection client interfaces (before we were communicating directly through the cim repository interface).

Waiting commit of pr #2215

### DIFF
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -38,6 +38,7 @@ from __future__ import absolute_import, print_function
 
 import uuid
 from collections import Counter
+from copy import deepcopy
 import six
 
 # pylint: disable=ungrouped-imports
@@ -101,6 +102,7 @@ class MainProvider(ResolverMixin, BaseProvider):
     """
 
     def __init__(self, host, disable_pull_operations, cimrepository):
+        # pylint: disable=super-init-not-called
         """
         Parameters:
             conn(:class:`~pywbem_mock.FakedWBEMConnection`):
@@ -1018,7 +1020,8 @@ class MainProvider(ResolverMixin, BaseProvider):
                 _format("Class {0!A} already exists in namespace {1!A}.",
                         NewClass.classname, namespace))
 
-        new_class = NewClass.copy()
+        # Create copy because resolve_class modifies elements of class
+        new_class = deepcopy(NewClass)
 
         qualifier_store = self.get_qualifier_store(namespace)
         self._resolve_class(new_class, namespace, qualifier_store,
@@ -1068,9 +1071,6 @@ class MainProvider(ResolverMixin, BaseProvider):
                 _format("ModifyClass not valid CIMClass. Rcvd type={0}",
                         type(ModifiedClass)))
 
-        # Validate namespace in class store and get the class CIM repository
-        # for this namespace.
-        self.validate_namespace(namespace)
         class_store = self.get_class_store(namespace)
 
         if not class_store.object_exists(ModifiedClass.classname):
@@ -1079,13 +1079,14 @@ class MainProvider(ResolverMixin, BaseProvider):
                 _format("Class {0!A} does not exist in namespace {1!A}.",
                         ModifiedClass.classname, namespace))
 
-        modified_class = ModifiedClass.copy()
+        # create copy because resolve_class can modify elements of class
+        modified_class = deepcopy(ModifiedClass)
 
         qualifier_store = self.get_qualifier_store(namespace)
         self._resolve_class(modified_class, namespace, qualifier_store,
                             verbose=False)
 
-        # Add new class to CIM repository
+        # Update class in CIM repository
         class_store.update(modified_class.classname, modified_class)
 
     ####################################################################
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -448,6 +448,69 @@ class MainProvider(ResolverMixin, BaseProvider):
         clnsdict[classname] = classname
         return clnsdict
 
+def validate_modified_class(self, namespace, modified_class, original_class,
+                            super_class=None):
+    """
+    Validate and possibly modify the modified class against the original_class
+    and possibly against superclasses.
+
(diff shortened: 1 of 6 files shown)


pywbem_mock/_mainprovider.py
```diff
@@ -1060,11 +1060,33 @@ def ModifyClass(self, namespace, ModifiedClass):
...
 
+        if not isinstance(ModifiedClass, CIMClass):
+            raise CIMError(
+                CIM_ERR_INVALID_PARAMETER,
+                _format("ModifyClass not valid CIMClass. Rcvd type={0}",
+                        type(ModifiedClass)))
+
+        # Validate namespace in class store and get the class CIM repository
+        # for this namespace.
         self.validate_namespace(namespace)
```
andy-maier, line 1073: validate_namespace() is called for the second time.

pywbem_mock/_mockmofwbemconnection.py
```diff
@@ -28,30 +28,25 @@
...
+class _MockMOFWBEMConnection(BaseRepositoryConnection, ResolverMixin):
     """
-    Create an adaption of the MOF compiler MOFWBEMConnection class so that to
-    directly use the attributes that represent the repository and implement a
-    modified CreateClass that resolves the new class. This directs the compiler
-    output directly  to the dictionaries for qualifiers, and instances in the
-    FakedWBEMConnection object and replaces the CreateClass with a local method
-    that allows resolving the created class before it is inserted into the
-    repository.
+    Create an adaption of the MOF compiler MOFWBEMConnection class to interface
```
andy-maier, line 43: It is no longer an adaption of the MOFWBEMConnection class.

pywbem_mock/_mockmofwbemconnection.py
```diff
@@ -28,30 +28,25 @@
...
+from pywbem._mof_compiler import BaseRepositoryConnection
 from pywbem._utils import _format
 from ._resolvermixin import ResolverMixin
 
 # Issue #2062 TODO/ks Remove this code and use the methods in _WBEMConnection
 # in their place
 
 
-class _MockMOFWBEMConnection(MOFWBEMConnection, ResolverMixin):
+class _MockMOFWBEMConnection(BaseRepositoryConnection, ResolverMixin):
```
andy-maier, line 41: Why is this class needed at all?

pywbem_mock/_mockmofwbemconnection.py
```diff
@@ -112,38 +102,34 @@ def _setns(self, value):
...
-        that is needed and never return that error, this method is never
-        called, and is therefore not implemented.
+        """
+        Not Implemented because not used with the MOF compiler.
         """
 
         raise CIMError(
-            CIM_ERR_FAILED, 'This should not happen!',
+            CIM_ERR_FAILED, 'EnumerateInstanceNames not implemented!',
             conn_id=self.conn_id)
```
andy-maier, line 111: If this class is only used by the mof compiler, and if the mof compiler does not use this method, this is a situation where an assertion error would be appropriate, instead.

Applies to multiple such methods.

andy-maier:
Please rebase on latest master, since I merged PR #2212, to make sure they work together.

pywbem_mock/_mainprovider.py
KSchopmeyer: WHOOPS done

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: DONE

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: This class differs from the MOFWBEMConnection in that:

1. It requires the connection object, the other one treats conn as an option
2. It implements ModifyClass, MofWBEMConnection does not.
3. In only uses the local cache for classes when the CreateClass is trying to resolve dependent classes

In reality, I would assume we could bring them together but that is one more step.

**DISCUSSION**

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: Agreed.

pywbem_mock/_mockmofwbemconnection.py
andy-maier: With the change of the superclass, they are going further apart rather than closer together.

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: Changed to asserts.  DONE

KSchopmeyer: Note that we could also af used the NotImplementedError but this clarifies that the error is from the child class, not from the base class

KSchopmeyer: Better Question: Why include the unimplemnted classes in BaseRepositoryConnection since it is just for compiler support?

**DISCUSSION**

pywbem_mock/_mockmofwbemconnection.py
andy-maier: We discussed this today in a call and decided that getting proper user-provider support when the MOF compiler works against a fake connection is more important than loosing the rollback support. Withdrawing the comment. DONE.

pywbem_mock/_mockmofwbemconnection.py
KSchopmeyer: Ignore this one.

andy-maier: We decided to ignore this one. DONE

andy-maier:
It currently fails tests.

andy-maier:
This now fails due to issue #2214. Please rebase on master after PR #2215 is merged.

andy-maier:
Can continue.

PR #2218, opened 2020-05-04
Demonstration of virtualenv issue with pip

andy-maier:
This PR is only used to demonstrate the issue.


andy-maier:
This PR should stay open fir a while since its Travis runs are referenced in the virtualenv issue that was created.

PR #2225, opened 2020-05-11
Partly fixes #2062: Implement new schema compile method

### LINKED ISSUE #2062: Refactor pywbem mock to separate components

KSchopmeyer:
Currently in 3 commits, 1. original, 2. fixes to comments, 3. modify we use schema_mof_pragma_file 
as the input parameter to compile_schema_classes.

The original method for compiling parts of a DMTF schema was
compile_dmtf_schema() which compiled a specific set of classes from the
DMTF schema defined by a DMTF schema version number and the directory
where a local version of that DMTF was supposed to be located.

This change separates the compile of classes of the schema from the process
of downloading and installing the schema.

Thus, originally, the FakedWBEMConnection method to create a mock
repository with some classes from the schema was:

conn.compile_dmtf_schema(DMTF_TEST_SCHEMA_VER, TESTSUITE_SCHEMA_DIR,
                         class_names=[CIM_Namespace,'CIM_ObjectManager'],
                         verbose=True)

the separation functionality yields:

schema = DMTFCIMSchema(DMTF_TEST_SCHEMA_VER, TESTSUITE_SCHEMA_DIR,
                       verbose=False)

conn.compile_dmtf_classes(schema.schema_mof_pragma_file
         schema_classnames=[CIM_Namespace,'CIM_ObjectManager'],
         namespace=True)

In addition we changed the internal name in DMTFCIMSchema class from
schema_mof_file to schema_mof_pragma_file.  Note that the property
schema_mof_file is retained but marked deprecated.  This helps to
clarify the meaning of this file.

Added tests for the new method to compile DMTF classes. This
was a very small change, just extending the test class for the original
compile_dmtf_schema to execute both the original compile and the new
compile.

### DIFF
--- a/pywbem_mock/_dmtf_cim_schema.py
+++ b/pywbem_mock/_dmtf_cim_schema.py
@@ -15,15 +15,17 @@
 #
 #
 """
-A DMTF CIM schema is the collection of CIM qualifier declarations and CIM
-classes managed by the DMTF and released as a single tested, coherent package
-with a defined major, minor, update version number.  The MOF representation of
+A CIM schema is the collection of CIM qualifier declarations and CIM
+classes and available  a single tested, coherent package
+with a defined major, minor, update version number. The DMTF regularly
+releases updates to the DMTF CIM schema that represent all of the classes
+approved by the DMTF. The MOF representation of
 a DMTF schema release is packaged as a single zip file and is available on the
 DMTF web site ( https://www.dmtf.org/standards/cim ).
 
-Because the DMTF CIM schemas are the source of most of the classes that a WBEM
-server normally implements it was important to create a simple mechanism
-to include the DMTF CIM classes and qualifier declarations in the
+Because CIM schemas are the source of most of the classes that a WBEM server
+normally implements it was important to create a simple mechanism to include
+the DMTF CIM classes and qualifier declarations from a CIM schema in the
 :class:`~pywbem_mock.FakedWBEMConnection` mock repository.
 
 The :class:`~pywbem_mock.DMTFCIMSchema` class downloads the DMTF CIM schema
@@ -36,7 +38,7 @@ classes in the CIM schema having unique class names; this is always true for
 DMTF CIM schema releases to assure that they can be installed in the same
 namespace in a WBEM server repository.
 
-Once a DMTF CIM schema is extracted into the individual MOF files it consumes
+Once aCIM schema is extracted into the individual MOF files it consumes
 considerable space since it expands to almost 3000 files.  Therefore it is
 logical to remove all of the individual MOF files when not being used and also
 if the schema information is to be committed to be used in testing in an
@@ -63,19 +65,20 @@ Because it is usually necessary to compile only a subset of the DMTF CIM
 classes into a mock repository for any test, the
 :class:`~pywbem_mock.DMTFCIMSchema` class provides a
 :meth:`~pywbem_mock.FakedWBEMConnection.build_schema_mof` method to create a
-MOF include file that includes only a subset of the classes in the downloaded
-DMTF CIM Schema. The task of listing classes to be compiled is futher
+list of MOF pragmans that includes only a subset of the classes in the
+downloaded CIM Schema. The task of listing classes to be compiled is futher
 simplified because the pywbem MOF compiler searches the DMTF schema for
 prerequisite classes (superclasses, reference classes, and EmbeddedInstance
 classes) so that generally only the leaf CIM classes required for the
 repository need to be in the class list. This speeds up the process of
 compiling DMTF CIM classes for any test significantly.
 """
+import warnings
 import os
 from zipfile import ZipFile
 import shutil
-import six
 import re
+import six
 try:
     from urllib.request import urlopen
 except ImportError:  # py2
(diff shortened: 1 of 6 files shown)


pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -226,17 +226,38 @@ def schema_mof_dir(self):
...
+        contains the `#pragama include` statements for all of the classes in
+        the schema and the qualifier declaration `#pragma include` statements
+        to compile `qualifiers.mof` and `qualifiers_optional.mof`.
+
+        This filename is of the general form::
+
+            <schema_mof_dir>/cim_schema_<schema_version_str>.mof
+
+            ex: schemas/dmtf/moffinal2490/cim_schema_2.49.0.mof
+        """
```
andy-maier, line 251: Issue a DeprecationWarning when used.

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -226,17 +226,38 @@ def schema_mof_dir(self):
         :term:`string`: Path name of the directory in which the DMTF CIM Schema
         MOF files are extracted from the downloaded zip file. This property can
         be used as the MOF compiler search path for compiling classes in the
-        DMTF CIM schema.
+        DMTF CIM schema. This directory should also contain the
+        cim_schema_.mof pragma file.
```
andy-maier, line 230: Use a consistent term for that file. "DMTF CIM schema MOF pragma file" was used elsewhere.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -626,14 +705,15 @@ def compile_dmtf_schema(self, schema_version, schema_root_dir, class_names,
           TypeError: The 'schema_version' is not a valid tuple with 3
             integer components
           :exc:`~pywbem.MOFCompileError`: Compile error in the MOF.
-        """
+        """  # noqa: E501
+        # pylint: enable:line-too-long
```
andy-maier, line 709: Issue a DeprecationWarning.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -543,20 +545,94 @@ def compile_mof_string(self, mof_str, namespace=None, search_paths=None,
 
         mofcomp.compile_string(mof_str, namespace)
 
+    def compile_dmtf_classes(self, schema_mof_dir, schema_classnames,
+                             namespace=None, verbose=False):
```
andy-maier, line 549: I would not name this method "compile_dmtf_classes()". It could be used on any CIM schema. How about "compile_schema_classes()"?

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -543,20 +545,94 @@ def compile_mof_string(self, mof_str, namespace=None, search_paths=None,
 
         mofcomp.compile_string(mof_str, namespace)
 
+    def compile_dmtf_classes(self, schema_mof_dir, schema_classnames,
```
andy-maier, line 548: Instead of passing the path name of the schema directory in "schema_mof_dir" and using an implicitly defined file name of the pragma file, it would be more straight forward to pass the path name of the pragma file in a "schema_pragma_file" parameter.

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -421,6 +443,58 @@ class mof compiles
...
+            represented by this :class:`~pywbem_mock.DMTFCIMSchema` object.
+            This parameter can be a string if there is only a single class name
+            to be included.
+
+            If the returned string is compiled, the MOF compiler will search
+            the directory defined by `schema_mof_dir` for dependent classes
+            for each class in this list.
+
+        schema_mof_pragma_file (:term:`string`):
+            File path defining a cim schema pragma file for the set of
```
andy-maier, line 489: lower case "cim" -> "CIM"

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -421,6 +443,58 @@ class mof compiles
...
+        """
+        return DMTFCIMSchema.static_build_schema_mof(
+            schema_classes, self.schema_mof_file)
+
+    @staticmethod
+    def static_build_schema_mof(schema_classes, schema_mof_pragma_file):
+        """
+        Build a string that includes the ``#include pragma`` statements for the
+        DMTF schema CIM classes defined in `schema_classes` using the DMTF CIM
+        schema defined by this object.
```
andy-maier, line 458: This method can be used on any CIM schema, not just the DMTF CIM schema. I suggest to change the docstring of this method accordingly.

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -421,6 +443,58 @@ class mof compiles
...
 
+        Raises:
+            ValueError: If any of the classnames in `schema_classes` are not in
+            the DMTF CIM schema installed
+        """
+        return DMTFCIMSchema.static_build_schema_mof(
+            schema_classes, self.schema_mof_file)
+
+    @staticmethod
+    def static_build_schema_mof(schema_classes, schema_mof_pragma_file):
```
andy-maier, line 454: How about moving this method to become a global function in the module? It can be used on any CIM schema, not just on the DMTF CIM schema.

pywbem_mock/_dmtf_cim_schema.py
KSchopmeyer: DONE

pywbem_mock/_dmtf_cim_schema.py
KSchopmeyer: The correct name should be "schema_mof_pragma_file and the text should use the same words without the _.  Fixed some of them.  we may find some others.

DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Actually, no it would not.  The schema file includes the schema version so that somewhere the version would have to be passed or the specific file for a schema version name defined.  That is why we did it this way.

**DISCUSSION**

andy-maier: The schema file would be passed all the way through all layers. And yes, at the top its file name needs to be known. A CIM schema that is not the DMTF CIM schema might not even have the version in the file name. Or it may name the file differently from the current glob pattern.

pywbem_mock/_dmtf_cim_schema.py
KSchopmeyer: Tried to remove the DMTF everywhere but to be honest there are only the DMTF schema

pywbem_mock/_dmtf_cim_schema.py
KSchopmeyer: The advantage is that it would get rid of that stupid name.

DONE

pywbem_mock/_dmtf_cim_schema.py
KSchopmeyer: DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Either way you are making assumptions.  1. Your assume that the schema file in in the schema_mof_dir or 2. you assume that the schema_mof_pragma_file directory can be used as the search path.

I would argue that the that the schema_mof_dir is more general that the schema_mof_pragma_file but in either case you end up with a name that may include a schema version.

DISCUSSION

andy-maier: On assumptions:
> 1. Your assume that the schema file in in the schema_mof_dir or 2. you assume that the schema_mof_pragma_file directory can be used as the search path.

On 1.: The path name of the schema file contains its directory, so a separate schema_mof_dir parameter is not needed to find it.
On 2.: The pragma include statements in the schema file reference the other files, and if needed, with a relative path name, so the schema file is self-contained w.r.t. the location of other files.

Where the assumption really is, is by assuming the class mof files have a file name that can be determined from the CIM class name. That is true for the DMTF schema since schema version 2.7 or so (but not before), and it may or may not be true for other CIM schemas. That assumption does not take advantage of the fact that by just knowing the schema file everything is determined, and it requires knowing a directory under which the class mof files are located so they can be found by constructing their file name from the desired class name.

I think it boils down to the question what our requirements on the file structure of a CIM schema should be, and how general we can/want to get. At the one extreme, we would only support the exact structure of today's DMTF CIM schema. At the other extreme, we would support any MOF-legal way of including files into one top level schema file.

andy-maier:
I created issue #2226 for the flake8 issue.

pywbem_mock/_dmtf_cim_schema.py
andy-maier: ok. good enough.

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -87,15 +90,105 @@
 __all__ = ['DMTFCIMSchema']
 
 
+def build_schema_mof(schema_classes, schema_mof_pragma_file):
```
andy-maier, line 93: We currently have three ways to name the parameter for the class names in the various functions: `schema_classes`, `schema_classnames`, and `class_names`. For consistency, I suggest to name them all `class_names`. I have according comments on each case.
So please rename the `schema_classes` parameter to `class_names`, for consistency.

pywbem_mock/_dmtf_cim_schema.py
```diff
@@ -386,11 +504,12 @@ def print_verbose(msg):
     def build_schema_mof(self, schema_classes):
```
andy-maier, line 504: Please rename the `schema_classes` parameter to `class_names`, for consistency.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -543,20 +546,102 @@ def compile_mof_string(self, mof_str, namespace=None, search_paths=None,
 
         mofcomp.compile_string(mof_str, namespace)
 
+    def compile_schema_classes(self, schema_mof_pragma_file, schema_classnames,
+                               namespace=None, verbose=False):
```
andy-maier, line 550: 1. I suggest to switch the order of parameters `schema_mof_pragma_file` and `schema_classnames` to be consistent with the global function `build_schema_mof()`, and because the class names are sort of the main input.
2. Please rename the `schema_classnames` parameter to `class_names`, for consistency.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -626,14 +714,20 @@ def compile_dmtf_schema(self, schema_version, schema_root_dir, class_names,
...
+        warnings.warn(
+            "Calling compile_dmtf_schema is deprecated; it will be removed "
+            "in the next pywbem release after 1.0.0. Use",
+            DeprecationWarning, 2)
 
         schema = DMTFCIMSchema(schema_version, schema_root_dir,
                                use_experimental=use_experimental,
                                verbose=verbose)
-        schema_mof = schema.build_schema_mof(class_names)
+        schema_pragma_mof = schema.build_schema_mof(class_names)
```
andy-maier, line 728: It is just a local variable, but still worth thinking about naming consistency. Originally I thought it should be renamed to `schema_mof_pragma` for consistency with the various occurrences of `schema_mof_pragma_file`, but that is somehow not logical for a MOF string.

Which lead me to think that `schema_mof_pragma_file` is probably unlogical, too, at least when comparing it with `schema_mof_pragma`.

So how about renaming the various occurrences of `schema_mof_pragma_file` to either `schema_pragma_mof_file` or somewhat shorter to `schema_pragma_file`?

pywbem_mock/_wbemconnection_mock.py
andy-maier: We discuss this in our call and between the two options of pointing to a schema directory and pointing to the pragma file, Andy has a slight preference to point to the pragma file, which is what the latest commit does.

KSchopmeyer:
Fixed issues with attribute and parameter naming.  

NOTE: I rebased the multiple commits into a single commit.

PR #2234, opened 2020-05-23
Partly fixes #2062: part5b refactor register provider

### LINKED ISSUE #2062: Refactor pywbem mock to separate components

KSchopmeyer:

Marked WIP for the moment but it passes tests.

**DISCUSSION:** In this code I ignore  attribute error in call to post_register_setup(conn).  Not sure whether we should ignore (they supply if they want to) or error(they must supply at least a dummy). Since this is server side code, I figured that they will figure it out.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -150,6 +150,8 @@ class ProviderRegistry(object):
         4. Adding the provider to the registry of user_providers so that any
            of the request methods defined for the `provider_type` are
            passed to this provider in place of the default request processors.
+        5. Execute post_register_setup() call to the provider to allow the
+           provider to perform any special setup functionality.
 
         Providers can only be registered for the following request response
         methods:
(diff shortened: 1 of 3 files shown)


andy-maier:
Rebase on master

andy-maier:
Please rebase on latest master, deleting the commits that belong to the other PR, and the conflicts will go away.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -307,6 +311,19 @@ class must contain variables `provider_type` and
...
+                        "type: {1!A} namespaces:{2!A}",
+                        provider.__class__.__name__,
+                        ", ".join(provider_classnames),
+                        provider_type, ", ".join(namespaces))
+
+            try:
+                provider.post_register_setup(conn)
+            except AttributeError:
+                if verbose:
+                    _format("Register_Provider post_register_setup not found")
```
andy-maier, line 325: I don't think we should require that a provider has this method. It should be optional: If defined, it is called.

KSchopmeyer: Good.  No warning, etc. if missing.

KSchopmeyer:
I rebased on master and cleaned up from where I had multiple commits of my own.  Also rewrote the commit message.

Sorry but I cannot figure out how to  remove the two old commits.  Rebasing on master does not do it. Suggestions welcome.

Note that we simply try the post_register_setup and ignore if there is an attribute error.  Since this is a mocker, if there are other errors, it should probably fail or at least capture those exceptions so I do not unregister the provider upon error.

andy-maier:
I squashed the 3 commits into one and will merge it when the tests complete.

PR #2235, opened 2020-05-25
Formal and light content improvements of mock docs

andy-maier:
See commit message.

### DIFF
--- a/pywbem_mock/_baseprovider.py
+++ b/pywbem_mock/_baseprovider.py
@@ -74,7 +74,8 @@ class BaseProvider(object):
     @property
     def namespaces(self):
         """
-        Return a list of namespaces that exist in the CIM repository.
+        list of :term:`string`: List of the namespaces that exist in the CIM
+        repository.
         """
         return self.cimrepository.namespaces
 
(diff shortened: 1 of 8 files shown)


pywbem_mock/_baseprovider.py
```diff
@@ -290,50 +291,54 @@ def remove_namespace(self, namespace):
...
-        Returns:
+        Only the valid Interop namespace names can be used to register Interop
+        namespace providers.
+
+        This list is defined in :attr:`pywbem.WBEMServer.INTEROP_NAMESPACES`.
 
-          List of :term:`string`: containing the valid names for
-          namespaces that can be interop namespaces. Any name not on this list
-          cannot be an interop provider
+        Note that namespace names need to be considered case insensitive.
```
KSchopmeyer, line 301: Remove "Note that"

pywbem_mock/_baseprovider.py
```diff
@@ -452,11 +457,11 @@ class are returned. `None` means not supplied by client and the
           filtered in accord with the local_only, etc. arguments.
 
         Raises:
-          :class:`~pywbem.CIMError`: (CIM_ERR_NOT_FOUND) if class not found in
-          CIM repository.
 
+          :class:`~pywbem.CIMError`: (CIM_ERR_NOT_FOUND) if class not found in
```
KSchopmeyer, line 461: We are inconsistent on whether the status_code_name is in paren. Most other places it is not. Actually I like the paren.

andy-maier: DONE - Added parenthesis all over the uses in pywbem_mock.

pywbem_mock/_baseprovider.py
andy-maier: DONE

PR #2245, opened 2020-05-28
Fixes #2229: clean up convience functions

### LINKED ISSUE #2229: Mocker, eliminate multiple convience methods that are the same thing.
If the atttribute is not public and not documented that we are reaching for with the convience method, we should probably make it public.

KSchopmeyer:
See commit for details

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -291,7 +291,7 @@ class ProviderRegistry(object):
             # pragmas exist or generate an exception if no pragmas exist
 
             for namespace in namespaces:
-                if not conn.mainprovider.class_exists(namespace, classname):
+                if not conn._mainprovider.class_exists(namespace, classname):
                     if schema_pragma_files:
                         conn.compile_schema_classes(
                             classname,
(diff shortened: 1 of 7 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -1132,22 +1125,21 @@ def add_cimobjects(self, objects, namespace=None):
...
-
-                qualifier_store = self._get_qualifier_store(namespace)
 
                 # pylint: disable=protected-access
-                cc1 = self.mainprovider._resolve_class(cc, namespace,
-                                                       qualifier_store,
-                                                       verbose=False)
+                cc1 = self._mainprovider._resolve_class(
+                    cc, namespace,
+                    self._mainprovider.get_qualifier_store(namespace),
```
andy-maier, line 1139: To go one step further, I think we should consider also removing the `get_*_store()` convenience methods and the `namespaces` property from `BaseProvider`, since any provider now has access to the repository via its own `cimrepository` property.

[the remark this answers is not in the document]
KSchopmeyer: DONE

KSchopmeyer: DONE

PR #2250, opened 2020-06-04
Correct issues in pywbem_mock with Open/Pull

KSchopmeyer:
Fixes 3 issues found in testing with pywbemcli.

1. a code error in MainProvider  Open and pull processing.  We had some parameters out of order in a local function call.  Since this was changes made in Pywbem 1.0.0, no back port necessary.  

In the process, I found that we did not have a test for what caused this issue so added two new tests for Open/Pull

2. We were inserting extra properties in instances.  My coding error. Removed that code.

3. Restored use of CIM_ERR_NOT_FOUND in place of CIM_ERR_NOT_AVAILABLE when method not found.  This is correct since NOT_FOUND is higher priority than NOT_AVAILABLE. I misread the spec when review code during the refactor.

pywbemcli now passes tests (with a number of changes to its tests) against pywbem 1.0.0. I have not submitted that PR yet.

### DIFF
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -1567,8 +1567,8 @@ class MainProvider(BaseProvider, ResolverMixin):
     def _return_assoc_class_tuples(self, rtn_classnames, namespace, iq, ico,
                                    pl):
         """
-        Creates the correct tuples of for associator and references class
-        level responses from a list of classnames.  This is special because
+        Creates the correct tuples for associator and references class
+        level responses from a list of classnames.  This is unique because
         the class level references and associators return a tuple of
         CIMClassName and CIMClass for every entry.
         """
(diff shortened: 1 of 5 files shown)


pywbem_mock/_mainprovider.py
```diff
@@ -2557,14 +2558,15 @@ def _pull_response(self, req_type, EnumerationContext,
             eos = u'TRUE'
             rtn_objs_list = objs_list
             del self.enumeration_contexts[EnumerationContext]
-            context_id = ""
+            context = ""
```
andy-maier, line 2561: In the open method, we use context_id. Stay consistent.

KSchopmeyer: But the name in the named tuple is context so we are consistently inconsistent.  Since it is built from the method create_contextid() I went back to context_id for both of these methods.

DONE

PR #2254, opened 2020-06-08
Fixes #2247: MOF compiler changes default_namespace

### LINKED ISSUE #2247: MOF Compiler changes default namespace of conn
Here is one that has been hanging around a long time and only surfaced when we started compiling into multiple namespaces now.

The compiler itself sets the namespace being compiled into as the conn.default_namepace.

See _mof_compiler line  MOFWBEMConnection Class, 2675,

        self.parser.mof = mof
        self.parser.handle.default_namespace = ns

where ns is an argument of compile_string(...) and set into the parser.handle,

NOTE: The handle is actually the MOFWBEMConnection or WBEMConnection or FakedWBEMConnection  which means that this is changing the
default_namespace in the WBEMConnection.

This new default namespace is used heavily through the compiler.

KSchopmeyer:

### DIFF
--- a/pywbem/_mof_compiler.py
+++ b/pywbem/_mof_compiler.py
@@ -2659,9 +2659,11 @@ class MOFCompiler(object):
 
           MOFCompileError: Error compiling the MOF.
         """
-
         lexer = self.lexer.clone()
         lexer.parser = self.parser
+
+        # Save any previous parser.file and parser.mof to restore when
+        # this compile successful
         try:
             oldfile = self.parser.file
         except AttributeError:
@@ -2672,7 +2674,15 @@ class MOFCompiler(object):
         except AttributeError:
             oldmof = None
         self.parser.mof = mof
+
+        # Save default_namepace so it can be restored after the parse.
+        # Note that there may be nested calls to compile_string
+        try:
+            oldns = self.parser.handle.default_namespace
+        except AttributeError:
+            oldns = None
         self.parser.handle.default_namespace = ns
+
         if ns not in self.parser.qualcache:
             self.parser.qualcache[ns] = NocaseDict()
         if ns not in self.parser.classnames:
@@ -2687,6 +2697,7 @@ class MOFCompiler(object):
             rv = self.parser.parse(mof, lexer=lexer)
             self.parser.file = oldfile
             self.parser.mof = oldmof
+            self.parser.handle.default_namespace = oldns
             return rv
         except MOFCompileError as pe:
             # Generate the error message into log and reraise error
--- a/pywbem/_mof_compiler.py
+++ b/pywbem/_mof_compiler.py
@@ -575,7 +575,7 @@ def p_mp_createClass(p):
                       """
 
     # pylint: disable=too-many-branches,too-many-statements,too-many-locals
-    ns = p.parser.handle.default_namespace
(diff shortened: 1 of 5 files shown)


pywbem/_mof_compiler.py
```diff
@@ -2672,7 +2674,15 @@ def compile_string(self, mof, ns, filename=None):
         except AttributeError:
             oldmof = None
         self.parser.mof = mof
+
+        # Save default_namepace so it can be restored after the parse.
+        # Note that there may be nested calls to compile_string
+        try:
+            oldns = self.parser.handle.default_namespace
+        except AttributeError:
+            oldns = None
```
andy-maier, line 2683: This approach of handling the AttributeError in the other cases of `self.parser.file` and `self.parser.mof` is just a very slacky way of dealing with not properly having thought about initialization. The `self.parser.handle` is in fact properly initialized, so that is not necessary there. In fact, it creates a case of non-testability for the exception path.

I suggest to remove the exception handling for this case.

Beyond that, we could also rework the other two cases to use proper initialization rather than exception handling.

pywbem/_mof_compiler.py
```diff
@@ -2672,7 +2674,15 @@ def compile_string(self, mof, ns, filename=None):
...
         self.parser.handle.default_namespace = ns
```
andy-maier, line 2684: Besides the previous comment, can we not avoid the modification of the `default_namespace` altogether? We could set the target namespace as a new attribute on `self.parser`, like so:
```
self.parser.target_namespace = ns
```
and every time the parser invokes an operation on its `self.handle`, pass the `self.target_namespace` to the operation explicitly.

pywbem/_mof_compiler.py
KSchopmeyer: DISCUSSION:  I must be missing something because I am not certain what the exception gains but we have to be careful in this method is recursively called multiple times, not directly but as a result of the calls to the parser which may end up starting another compile. In my tests we were at points 4 levels into recursive calls to compile_string.

pywbem/_mof_compiler.py
KSchopmeyer: Refactored the MOF compiler to use a target_namespace and not have any predefined use of the default namespace, and refactored MOFWBEMConnection and MOCKMOFWBEMConnection to support a namespace  parameter on the GetClass, etc calls so that compiler can pas the defined namespace back to these methods.

pywbem/_mof_compiler.py
KSchopmeyer: I dropped the additions.  I want to discuss the original use since I am not sure why needed at all **DISCUSSION**  Those variables should be private to every nested call as I understand nested method calls in python.

pywbem/_mof_compiler.py
andy-maier: Ok, you were right, it was more than just what I suggested above. Thanks for this change!

pywbem/_mof_compiler.py
andy-maier: I'm ok not removing the exception handling in the other two cases (`self.parser.file` and `self.parser.mof`) for now, since they are unrelated to this PR. So we are DONE with this comment.

PR #2259, opened 2020-06-11
Correct issues with InvokeMethod response handling

KSchopmeyer:
Corrects issue in pywbem_mock that was causing failures in tests with pywbemcli and pywbem version 1.0.0.

We were incorrectly processing the InvokeMethod response

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -1961,18 +1961,21 @@ class FakedWBEMConnection(WBEMConnection):
                 CIM_ERR_FAILED,
                 _format("Invalid Response. Must be list or tuple)",
                         type(result)))
-        for param in result[1]:
-            if not isinstance(param, CIMParameter):
+
+        # Map output params to NocaseDict to be compatible with return
+        # from _methodcall. The input list is an iterable of CIMParameters
+        output_params = NocaseDict()
+        rtn_params = result[1]
+        rtn_params = rtn_params if isinstance(rtn_params, (tuple, list)) \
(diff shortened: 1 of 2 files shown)


andy-maier:
I don't understand this change.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -1961,18 +1961,21 @@ def _meth_InvokeMethod(self, methodname, objectname, Params, **params):
                 CIM_ERR_FAILED,
                 _format("Invalid Response. Must be list or tuple)",
                         type(result)))
-        for param in result[1]:
-            if not isinstance(param, CIMParameter):
+
+        # Map output params to NocaseDict to be compatible with return
+        # from _methodcall. The input list is an iterable of CIMParameters
+        output_params = NocaseDict()
+        rtn_params = result[1]
```
andy-maier, line 1968: First time I was reading this, I was confused about return vs output parameters, and that was caused by the variable name rtn_params. After a closer look, I agree with the code, but suggest to use a different name for that variable, eg out_param_list.

KSchopmeyer: Thats better because I spent an hour trying to figure out what was wrong.

KSchopmeyer: Made changes.   Also found some doc issues in MethodProvider when reviewing yesterday and fixed those.  Committed new version with all those changes in separate commit.

andy-maier:
On Travis, the runs for this PR succeeded, they were just not be reported back. Merging.

PR #2265, opened 2020-06-15
Fixes #2264: Changed format 'standard' of CIMInstanceName.to_wbem_uri() to sort the   keys

### LINKED ISSUE #2264: Changed format 'standard' of CIMInstanceName.to_wbem_uri() to sort the keys
For testing purposes, it is useful to guarantee the sorted order of keys in the produced WBEM URIs without lower casing (like format "canonical" does).

Therefore, it is suggested to add a new format "sorted" for that purpose.

See also https://github.com/pywbem/pywbemtools/pull/625.

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -593,6 +593,9 @@ Released: not yet
   a CIM_Namespace provider to handle the CIM_Namespace class in the interop
   namespace.  See issue #2062)
 
+* Changed format 'standard' of `CIMInstanceName.to_wbem_uri()` to sort the
+  keys in the resulting WBEM URI. (See issue #2264)
+
 **Cleanup:**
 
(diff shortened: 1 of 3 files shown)


andy-maier:
I have second thoughts on the new format "sorted": I'm not sure why anyone would have a need to use "standard" now that "sorted" exists. Also, it is fairly compatible; it changes the order from unpredictable to predictable. Finally, even if someone has a testcase with "standard" that now breaks, it was likely to break anyway had it been executed on multiple Python versions.

So I suggest that instead of introducing a new format "sorted", let's change format "standard" to always be sorted.

DISCUSSION

andy-maier:
We agreed to go as proposed above.

Issue #2232, opened 2020-05-22
Write section in documentation on user-defined providers


KSchopmeyer:
Right now that section exists but it is a TODO

andy-maier:
Note that the module docstring of `_defaultinstanceprovider.py` already contains some information on this topic.

andy-maier:
Addressed in PR #2262.

PR #2266, opened 2020-06-16
Partly fixes #2261: provider type names

### LINKED ISSUE #2261: Possible issues found while writing documentation on user-defined providers
1. post_register_setup.   It might be more logical to include this in the default providers with pass than the just putting it in the user-defined provider.

   **COMMENT** Karl - This one makes sense. 1. I shows the method in the superclass, it means that something is always called so the exception AttributeError is not really used ever.  Thus any exception would be a failure

   **RESOLUTION** Andy: In the meeting today, we agreed that we would change the approach to have an empty ´post_register_setup()` in the base class. This can be done in 1.0.0 final.

2. instancewrite provider and the registered type "instance".

   a. We do not have a clear reason for creating a type like InstanceWriteProvider rather than just InstanceProvider since the only reason for creating separate types is to all more config and operation control. In our case, we could make it so any instance request (EnumerateInstances, GetInstance, ExecQuery) could be put into the InstanceProvider and the choice of whether the operation was executed with a user provider or the defaultinstance provider was whether the method existed in the user-defined provider.

   **Andy:** We wanted to make it clear that there is no need to have user-defined providers for the other instance providers.

   **COMMENT** Karl.  Actually the way it is coded, there is no need for any working methods in a user-defined provider since the code falls back to the default for any request method that does not exist in the user-defined provider. EX. can implement CreateInstance but not ModifyInstance and it will use the Default.  Thus, there is no real need for separate InstanceWrite, InstanceRead, etc. except for clarification now that there are no user-defined providers other than WRITE.

   **RESOLUTION** Andy: In the meeting today, we agreed that we would stick with the term "instance write provider" and would change the provider type from "intance" to "instancwwrite". This should be done in 1.0.0 beta.

   b. We have the type as InstanceWrite but the type name is "instance". If we every do more instance providers, this type name is taken.

3. When we look for registered providers for a class we look for that specific class, not that class or subclasses.  I am actually not sure what is correct here.  We could require that:

   a. the user supply All exact classes so that if they implemented BLAH_ComputerSystem they should list CIM_ComputerSystem and BLAH_ComputerSystem

   b. We search base on subclasses also (ouch that is expensive unless we cach)

   c. We do subclass lookup as part of registration (blah)

   d. We ask for leaf class and do superclass lookup as part of registration.

   **Andy:** All classes that exist in a namespace have the main provider as a default. That default provider is overwritten when user-defined providers are registered for a class (or rather the subset of operations they implement). That override is only for exactly the class the provider is registered for, not for any of its subclasses. The subclasses have the main provider again. A provider that is registered for class BLAH_ComputerSystem creates only instances of that class (in CreateInstance), and will only get target instances of that class (in ModifyInstance, DeleteInstance, and InvokeMethod). Registration for subclasses needs to be done explicitly by the user who registeres the provider, by supplying the desired set of class names.

   **Karl:** Response. I am happy with that.

DISCUSSION

NOTE: Actually None of these require solution before the beta release

KSchopmeyer:
Changes name of provider from 'instance" to "instance-write"

Based on pr #2232, mock documentation update because the name of the provider type changed in the documentation also

### DIFF
--- a/docs/mockwbemserver.rst
+++ b/docs/mockwbemserver.rst
@@ -869,14 +869,14 @@ can be executed for each provider type.
 
 .. table:: Pywbem_mock provider types
 
-    ====================== ============ ========================  ================================
-    Provider Type          type name    CIM Request Operations    Default provider class
-    ====================== ============ ========================  ================================
-    method                 "method"      InvokeMethod             :ref:`Method Provider`
-    instance write         "instance"    CreateInstance           :ref:`Instance Write Provider`
-    instance write         "instance"    ModifyInstance           :ref:`Instance Write Provider`
-    instance write         "instance"    DeleteInstance           :ref:`Instance Write Provider`
-    ====================== ============ ========================  ================================
+    ====================== =================== ========================  ================================
+    Provider Type          type name           CIM Request Operations    Default provider class
+    ====================== =================== ========================  ================================
+    method                 "method"            InvokeMethod             :ref:`Method Provider`
+    instance write         "instance-write"    CreateInstance           :ref:`Instance Write Provider`
+    instance write         "instance-write"    ModifyInstance           :ref:`Instance Write Provider`
+    instance write         "instance-write"    DeleteInstance           :ref:`Instance Write Provider`
+    ====================== ============ ====== ========================  ================================
 
 Each user-defined provider is a Python class which is a subclass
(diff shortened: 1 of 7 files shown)


andy-maier:
I rebased it again and squashed the two commits.
Approved.

Issue #2272, opened 2020-06-18
Changes.rst. dropped the changes frompywbem versions 0.15 - 0.17


KSchopmeyer:
Pywbem 0.17.0 dropped the changes for 0.15.0, 0.16.0 and n 0.17 in the development of changes for 1.0.0 in changes.rst

KSchopmeyer:
Went through once and found what may be a few missing items.  Want to recheck before telecon.

It would probably be  good to add note that 17,16,15 change defs are integrated into 14 because that is when 1.0.0 was started.

andy-maier:
All the time since we started 1.0.0 from 0.14.x, we rolled changes back, not forward. So I suspect that what you found are things that were only needed on the old version (e.g. because of Python 2.6). But let's double check.

andy-maier:
PR #2281 adds a statement that 1.0.0b1 includes all fixes since 0.14.0 up to 0.17.2.

KSchopmeyer:
I am going through a second time now.

KSchopmeyer:
These are the one I found that are in the 17,16,15 changes.rst but not the 1.0.0.
0.17
Cleanup:
    Renamed all sub-modules within the pywbem namespace so they are now private 

0.16.0
Bug Fixes
"Fixed handling of Unicode string"
Enhancements
"Removed the dependency on unittest2"

0.15.0
Deprecations

The wbemcli command has been deprecated.

andy-maier:
Needs to be worked into PR #2281 .

andy-maier:
As proposed in PR #2281 , change log entries for 0.15, 0.16 amd 0.17 were added.

andy-maier:
Addressed in PR #2281.

PR #2281, opened 2020-06-19
Change log: Prepared for beta version and general improvements

andy-maier:
See commit message.
See the open items in the next comment.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -13,17 +13,22 @@ Change log
       :revisions: 1
 
 
-pywbem 1.0.0
-------------
+pywbem 1.0.0b1
+--------------
 
 This version is currently in development and is shown as |version|.
 
-This version contains all fixes up to pywbem 0.17.1.
-
 Released: not yet
 
-Incompatible changes:
-^^^^^^^^^^^^^^^^^^^^^
+This is a beta version of the upcoming version 1.0.0. Pip will only install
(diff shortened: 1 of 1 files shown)


andy-maier:
Need to add the missing change log entries listed in issue #2272 .

docs/changes.rst
```diff
@@ -163,112 +172,111 @@ details of the changes themselves and the reasons for the changes.
...
+    `compile_dmtf_schema()` to be able to do no logging, by specifying `None` for
+    the `log_func` init parameter of `MOFCompiler`. This is now the default.
+
+    MOF compile errors no are longer printed to stdout by default. To continue
+    printing the MOF compile errors to stdout, print the exception in your code.
+    (See issue #1997)
+
+  * Added pywbem_mock.config.py variables to control use of the IncludeQualifiers
+    and IncludeClassOrigin arguments on the GetInstance and EnumerateInstances
+    operations.  The default is not to ignore the request variables and
```
KSchopmeyer, line 261: Propose to replace the text starting with "The default is not to" with the following:

The default is to ignore the request variables and never return either attribute in the return instances whereas, in previous vesions the request variables determined whether they were returned. This may be changed by modifying the config variables however, the behavior of these request parameters was inconsistent between versions of the DMTF specification so the user should not depend on them. (See issue # 2065)

andy-maier: DONE - see new text in my comment further up.

PR #2289, opened 2020-06-21
Improved mock documentation

andy-maier:
See commit message.

### DIFF
--- a/docs/mockwbemserver.rst
+++ b/docs/mockwbemserver.rst
@@ -10,26 +10,26 @@ Mock WBEM server
 Overview
 --------
 
-The pywbem package contains the ``pywbem_mock`` subpackage which provides mock
-WBEM server support that enables using the pywbem library without a WBEM
-server. This subpackage is useful for testing the pywbem library itself as well
+The 'pywbem_mock' module of pywbem provides a mock WBEM server that
+enables using the pywbem client library without a real WBEM server.
+This module is useful for testing the pywbem client library itself as well
 as for the development and testing of Python programs that use the pywbem
-library.
+client library.
 
-pywbem_mock  contains the :class:`pywbem_mock.FakedWBEMConnection`
+The 'pywbem_mock' module contains the :class:`pywbem_mock.FakedWBEMConnection`
 class that establishes a *faked connection*. That class is a subclass of
 :class:`pywbem.WBEMConnection` and replaces its internal methods that use
 HTTP/HTTPS to communicate with a WBEM server with methods that communicate
 with  an in-process in-memory repository of CIM objects (the *mock repository*).
 
-:class:`~pywbem_mock.FakedWBEMConnection` acts as both the client API and a
+:class:`~pywbem_mock.FakedWBEMConnection` acts as both the client API and an
 in-process fake WBEM server. It includes methods to establish, configure, and
-visualize this fake WBEM server.  As a result, the operation methods of
+visualize this fake WBEM server. As a result, the operation methods of
 :class:`~pywbem_mock.FakedWBEMConnection` are those inherited from
 :class:`~pywbem.WBEMConnection`, so they have exactly the same input
 parameters, output parameters, return values, and even most of the raised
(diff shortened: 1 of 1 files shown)


docs/mockwbemserver.rst
```diff
@@ -10,26 +10,26 @@ Mock WBEM server
 Overview
 --------
 
-The pywbem package contains the ``pywbem_mock`` subpackage which provides mock
-WBEM server support that enables using the pywbem library without a WBEM
-server. This subpackage is useful for testing the pywbem library itself as well
+The 'pywbem_mock' module of pywbem provides mock WBEM server support that
```
KSchopmeyer, line 13: Change "mock WBEM server support" to "a mock WBEM server"

docs/mockwbemserver.rst
```diff
@@ -851,50 +853,48 @@ that somewhere an instance the class must be created each time a namespace is
...
-    instance write         "instance-write"    CreateInstance           :ref:`Instance Write Provider`
-    instance write         "instance-write"    ModifyInstance           :ref:`Instance Write Provider`
-    instance write         "instance-write"    DeleteInstance           :ref:`Instance Write Provider`
-    ====================== =================== ======================== ================================
-
-Each user-defined provider is a Python class which is a subclass
-of the corresponding default provider for the provider type as defined in
-the previous table.
+    ============== ================ ============== ==============================
+    Provider type  Type name        WBEM operation Details
```
KSchopmeyer, line 879: I prefer more than "Details". However, we do have an issue in that we don't have one clear name for the class that is the defining superclass for user-defined providers.   I used "Default provider class". Works for instances but not really for methods.  Something like "API defining class" but that is awkward. How about "Defining class"?

docs/mockwbemserver.rst
```diff
@@ -851,50 +853,48 @@ that somewhere an instance the class must be created each time a namespace is
...
 Creating user-defined instance providers
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
- User-defined instance providers can override the default implementations for
- the methods defined in the above table (``CreateInstance``, ``ModifyInstance``,
- and ``DeleteInstance``). These methods are defined in a user-defined class
- that subclasses :class:`pywbem_mock.InstanceWriteProvider`. The details of
- defining this provider are in section :ref:`Instance Write Provider`.
+
+User-defined instance write providers can override the default implementations for
+the methods defined in the above table (``CreateInstance``, ``ModifyInstance``,
```
KSchopmeyer, line 894: Shouldn't the work methods really be "operations"?

docs/mockwbemserver.rst
```diff
@@ -851,50 +853,48 @@ that somewhere an instance the class must be created each time a namespace is
...
+and ``DeleteInstance``). These methods are defined in a user-defined class
```
KSchopmeyer, line 895: Change "These methods" to "The user-defined implementation of these methods"

docs/mockwbemserver.rst
andy-maier: DONE

docs/mockwbemserver.rst
andy-maier: The links in that column target the section in the documentation that describes the provider type, so I changed the column heading to "Description".

In these sections, the language also needs to be cleaned up somewhat. Will do that in the next update.

docs/mockwbemserver.rst
andy-maier: Yes. DONE

docs/mockwbemserver.rst
andy-maier: Changed the text along these lines, but slightly different: 
```
User-defined instance write providers can override the default implementations
for the corresponding WBEM operations listed in the above table (i.e.
``CreateInstance``, ``ModifyInstance``, and ``DeleteInstance``). The Python
methods implementing these WBEM operations are defined in a user-defined class
that subclasses the :class:`~pywbem_mock.InstanceWriteProvider` class. The
details of implementing this provider type are described in
:ref:`Instance Write Provider`.
```

PR #2296, opened 2020-06-23
Fixes #2291: refactor base provider

### LINKED ISSUE #2291: Cleanup of BaseProvider methods and attributes
In the `BaseProvider` class, there are several things that can be cleaned up:

* `cimrepository` attribute: This is currently simply a public attribute, which is writeable. It could be changed to become a read-only property, with documentation.
* `find_instance()` method: It could be named more consistently as `get_instance()`.
* `find_interop_namespace()` method: It could be named more consistently as `get_interop_namespace()`.
* `get_instance_store()` method: This is purely a wrapper function around `cimrepository.get_instance_store()` and can be removed, if we access the repository through the `cimrepository` property.
* `get_qualifier_store()`: same here

If we do this, it would need to be done in beta, since it changes names.

DISCUSSION

KSchopmeyer:
Refactor issues documented in issue # 2291 as follows:

* cimrepository attribute: Change to define as a property  in
BaseProvider with documentation

* find_instance() method: Changed method name to get_bare_instance()

* find_interop_namespace() method: No change.  See issue for comments.

* get_instance_store() method and get_qualifier_store method: Removed
them both and modified tests and examples to show difference.

### DIFF
--- a/pywbem_mock/_baseprovider.py
+++ b/pywbem_mock/_baseprovider.py
@@ -62,9 +62,9 @@ class BaseProvider(object):
           cimrepository (:class:`~pywbem_mock.BaseRepository` or subclass):
             Defines the repository to be used by request responders.  The
             repository is fully initialized but contains only objects defined
-            by the :class:`pywbem_mock/FakedWbemConnection` methods that
+            by the :class:`~pywbem_mock.FakedWbemConnection` methods that
             create objects in the repository
-            (ex. :meth:`pywbem_mock/FakedWbemConnection.compile_file`)
+            (ex. :meth:`~pywbem_mock.FakedWbemConnection.compile_file`)
         """
         self._cimrepository = cimrepository
 
@@ -83,13 +83,13 @@ class BaseProvider(object):
         The CIM repository is the data store for CIM classes, CIM
         instances, and CIM qualifier declarations.  All access to the mocker
         CIM data must pass through this variable to the CIM repository.
-        See :class:`pywbem_mock/BaseRepository` for a description of
+        See :class:`~pywbem_mock.BaseRepository` for a description of
         the repository interface.
 
         The mocker may use subclasses of  :class:`~pywbem_mock.BaseRepository`
         for specific functionality.  Thus, :class:`~pywbem.InMemoryRepository`
         implements an InMemory CIM repository that must be initialized for
-        each :class:`pywbem_mock/FakedWbemConnection` construction.
+        each :class:`~pywbem_mock.FakedWbemConnection` construction.
         """
         return self._cimrepository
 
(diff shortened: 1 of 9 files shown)


pywbem_mock/_baseprovider.py
```diff
@@ -61,16 +61,38 @@ def __init__(self, cimrepository):
 
           cimrepository (:class:`~pywbem_mock.BaseRepository` or subclass):
             Defines the repository to be used by request responders.  The
-            repository is fully initialized.
+            repository is fully initialized but contains only objects defined
+            by the :class:`pywbem_mock/FakedWbemConnection` methods that
+            create objects in the repository
+            (ex. :meth:`pywbem_mock/FakedWbemConnection.compile_file`)
```
andy-maier, line 67: Use dot instead of slash in the references.

pywbem_mock/_baseprovider.py
```diff
@@ -61,16 +61,38 @@ def __init__(self, cimrepository):
...
+        The CIM repository is the data store for CIM classes, CIM
+        instances, and CIM qualifier declarations.  All access to the mocker
+        CIM data must pass through this variable to the CIM repository.
+        See :class:`pywbem_mock/BaseRepository` for a description of
+        the repository interface.
+
+        The mocker may use subclasses of  :class:`~pywbem_mock.BaseRepository`
+        for specific functionality.  Thus, :class:`~pywbem.InMemoryRepository`
+        implements an InMemory CIM repository that must be initialized for
+        each :class:`pywbem_mock/FakedWbemConnection` construction.
```
andy-maier, line 92: Use dot instead of slash in the references.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -253,13 +253,13 @@ def interop_namespace_names(self):
...
-        `InMemoryRepository`  that represents the CIM repository instance.
+        :class:`~pywbem.InMemoryRepository`: Instance of class
+        `InMemoryRepository`  that represents the CIM repository.
 
-        The CIM repository instance  is the data store for CIM classes, CIM
+        The CIM repository instance is the data store for CIM classes, CIM
         instances, and CIM qualifier declarations.  All access to the mocker
         CIM data must pass through this variable to the CIM repository.
-        See :class:`pywbem_mock/InmemoryRepository` for a description of
+        See :class:`pywbem_mock/BaseRepository` for a description of
```
andy-maier, line 262: Use dot instead of slash in the references.

andy-maier: FIXED

pywbem_mock/_baseprovider.py
andy-maier: FIXED

pywbem_mock/_baseprovider.py
andy-maier: FIXED

PR #2299, opened 2020-06-24
Fixes #2298: Second thoughts on BaseProvider.get_bare_instance()

### LINKED ISSUE #2298: Second thoughts on BaseProvider.get_bare_instance()
I've got second thoughts on the recent change of `BaseProvider.find_instance()` to `get_bare_instance()`.

For one, the name is just ugly.

More importantly though, the method gets an `instance_store` parameter, even though it is clear that it always would find the instance in the instance store for the namespace of the instance to be found, in the CIM repository that is known by the BaseProvider object. The namespace could be required to be set on the instance to be found (and is in fact set on all uses in the code).

andy-maier:
See commit message.


andy-maier:
Abandoning this PR.

PR #2308, opened 2020-06-27
Fixes #2304: Improved support for array-typed elements in ValueMapping

### LINKED ISSUE #2304: Add support for array elements to ValueMapping
The current `pywbem.ValueMapping` supports only scalar integer elements. There are several array properties in the DMTF CIM schema that have the ValueMap qualifier, for example "CIM_ComputerSystem.Dedicated", or "CIM_ManagedSystemElement.OperationalStatus".

See also the discussion in https://github.com/pywbem/pywbemtools/pull/636.

Support for array elements should be added to `pywbem.ValueMapping`.

**Update:** It turned out that array-typed elements did already work in `ValueMapping`, however the `tovalues()` method did not support lists of values, or None.

andy-maier:
See commit message.

Note that support for array-typed elements was already there, but the tovalues() method needed to be improved to accept a list/tuple of values.

### DIFF
--- a/pywbem/_valuemapping.py
+++ b/pywbem/_valuemapping.py
@@ -48,6 +48,8 @@ class ValueMapping(object):
     with the `ValueMap` and `Values` qualifiers, and the corresponding values
     of the `Values` qualifier, in both directions.
 
+    The CIM element may be a scalar or an array.
+
     This is done by retrieving the CIM class definition defining the CIM
     element in question, and by inspecting its `ValueMap` and `Values`
     qualifiers.
@@ -210,6 +212,8 @@ class ValueMapping(object):
         instance that maps CIM property values to the `Values` qualifier
         defined on that property.
 
+        The CIM property may be a scalar or an array.
+
         If a `Values` qualifier is defined but no `ValueMap` qualifier, a
         default of 0-based consecutive numbers is applied (that is the default
         defined in :term:`DSP0004`).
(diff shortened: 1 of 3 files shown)


pywbem/_valuemapping.py
```diff
@@ -743,6 +796,11 @@ def tobinary(self, values_str):
         Return the integer value or values for a `Values` string, based upon
         this value mapping.
 
+        Due to the complexity of its return value, this method only supports a
+        single `Values` string at a time. It does support array-typed elements,
+        though. Thus, if multiple `Values` strings need to be translated, this
+        method simply needs to be invoked multiple times.
```
KSchopmeyer, line 802: Remove the word simply or better "method must be invoked once for each value to be translated"

tests/unittest/pywbem/test_valuemapping.py
```diff
@@ -51,6 +51,16 @@ def element_kind(request):
     return request.param
 
 
+@pytest.fixture(params=[
```
KSchopmeyer, line 54: Couldn't you simplify this fixture by naming it "is_array and make the params True, False.  Not a real issue but when I figured that out it simplified other parts of the test.

pywbem/_valuemapping.py
andy-maier: DONE

tests/unittest/pywbem/test_valuemapping.py
andy-maier: DONE

PR #2315, opened 2020-06-29
Fixes #2314: slow performance for mock enumerate classes

### LINKED ISSUE #2314: pywbem_mock enumerate classes --di Very slow
This method is hundreds of times slower with pywbem 1.0.0b1 than with pywbem 0.17.

Confirmed that the issue is that in method get_subclassnames we are using the cim repository iter_value() with the copy parameter set to its default which is True.  This means we are creating many worthless copies.  In particular since the only thing we are using from the value is the classname string.

KSchopmeyer:
Fixes issue where we were repeatedly copying from the repo rather than
just itering causing very large slowdown of enumerate classes mock
operation.

NOTE: This should have an entry in changes.rst for b2 but this commit is to master.

    fixes issue #2314, mock enumerate classes --di very slow with large sets of classes.

Marked rollback for rollback to beta2

Marked priority because it makes it impossible to use the mocker with more than about 150 classes because of the speed of EnumerateClasses.

NOTE: I did not include a specific test. I will add one before the 1.0.0 release but that will take more time and the fix works with manual tests.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -29,6 +29,9 @@ Released: not yet
   1.0.0b1. This reduces the change log entries shown for 1.0.0b1 to just the
   changes relative to 0.17.2. (See issue #2303)
 
+* Fixed slow performance for EnumerateClasses operation in mock WBEM server.
+  (See issue #2314)
+
 **Enhancements:**
 
 * Added support for array-typed elements to pywbem.ValueMapping.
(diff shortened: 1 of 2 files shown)


andy-maier:
Rebased on branch stable_1.0 to get it into 1.0.0b2, and added a commit with the change log entry.

PR #2321, opened 2020-07-04
Fixes #2313: Fixed validation in main provider and instance write provider

### LINKED ISSUE #2313: pywbem_mock OpenExecQuery tests params before valid namespace
Minor issue but generally we test namespace first and this causes fail of one of our tests in pywbemtools.

andy-maier:
For details, see commit messages.

The commits are intentionally separated to break this large PR down into some reasonable steps. Each commit can be reviewed independently.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -34,6 +34,11 @@ Released: not yet
 * Removed the deprecated `schema_mof_file` property in `DMTFCIMSchema`, in favor
   of the `schema_pragma_file` property. (See issue #2284)
 
+* Changed the handling of invalid types of input parameters to WBEMConnection
+  operation methods to raise TypeError instead of other exceptions (KeyError,
+  AttributeError, CIMError). This does not change the behavior if valid types
+  are passed. (See issue #2313)
+
 **Deprecations:**
 
 **Bug fixes:**
@@ -48,6 +53,12 @@ Released: not yet
 
 * Test: Renamed the 'end2end' target in the makefile to 'end2endtest'.
   (Part of issue #2260)
+
+* Added type checking for input parameters to WBEMConnection operation methods.
+  Previously, invalid types could cause various exceptions to be raised,
+  including KeyError, AttributeError, or CIMError. Now, all invalid types are
+  properly checked and cause TypeError to be raised. Added testcases
+  for invalid types. (See issue #2313)
 
 **Cleanup:**
 
--- a/pywbem/_cim_operations.py
+++ b/pywbem/_cim_operations.py
@@ -221,56 +221,129 @@ def _check_classname(val):
 
 
 def _iparam_propertylist(property_list):
-    """Validate property_list and if it is not iterable convert to list.
+    """
+    Validate property_list input parameter and return it as a tuple/list,
+    or None.
 
     This is a test for a particular issue where the user supplies a single
     string instead of a list for a PropertyList parameter. It prevents
     an XML error.
     """
-    return [property_list] if isinstance(property_list, six.string_types) \
-        else property_list
+    if property_list is None:
+        pass
+    elif isinstance(property_list, (list, tuple)):
+        pass
+    elif isinstance(property_list, six.string_types):
+        property_list = [property_list]
+    else:
+        raise TypeError(
+            _format("The 'PropertyList' parameter of the WBEMConnection "
+                    "operation has invalid type {0} (must be a string, or "
+                    "a list/tuple of strings",
+                    type(property_list)))
+    return property_list
 
 
-def _validateIterCommonParams(MaxObjectCount, OperationTimeout):
+def _validate_OperationTimeout(OperationTimeout):
     """
-    Validate common parameters for an iter... operation.
+    Validate the OperationTimeout input parameter for the Iter...() and
+    Open...() operations.
 
-    MaxObjectCount must be a positive non-zero integer or None.
-
-    OperationTimeout must be positive integer or zero
+    Parameters:
+      OperationTimeout: Must be of integer type >= 0, or None.
 
     Raises:
-
-      ValueError: if these parameters are invalid
+      TypeError: Invalid type
+      ValueError: Invalid value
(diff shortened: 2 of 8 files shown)


docs/changes.rst
```diff
@@ -34,12 +34,23 @@ Released: not yet
 * Removed the deprecated `schema_mof_file` property in `DMTFCIMSchema`, in favor
   of the `schema_pragma_file` property. (See issue #2284)
 
+* Changed the handling of invalid types of input parameters to WBEMConnection
```
KSchopmeyer, line 37: I think that as much as standardization, it extended the type checking so that operations that might have been accepted even with invalid parameters are now caught by.t the client code. This could cause incompatibilites with older versions that ignored certain invalid parameters.

pywbem/_cim_operations.py
```diff
@@ -2060,39 +2133,53 @@ def infer_embedded_object(obj):
...
+        If namespace is None, the default namespace of the connection object
+        is used.
+
+        Returns:
+           The so determined namespace string.
+
+        Raises:
+          TypeError - namespace has an invalid type
+        """
+        if isinstance(namespace, six.string_types):
```
KSchopmeyer, line 2153: We test for the / chars here, and I assume other places in client.  However, we also check all over the mock code. Not for this PR, but we should probably relook at that and clean it up so that there is only one check somewhere or none because the client does it.

pywbem/_cim_operations.py
```diff
@@ -2148,37 +2254,171 @@ def _iparam_classname(classname, arg_name):
...
+        Returns:
+          * CIMClass copy without path
+          * None
+
+        Raises:
+          TypeError - klass has an invalid type
+        """
+
+        if isinstance(klass, CIMClass):
+            klass = klass.copy()
```
KSchopmeyer, line 2313: Not sure but asking.  Should this be deep copy?

pywbem/_cim_operations.py
```diff
@@ -2148,37 +2254,171 @@ def _iparam_classname(classname, arg_name):
...
+          * CIMInstance copy
+
+        Raises:
+          TypeError - instance has an invalid type
+        """
+
+        assert required is True  # For now, only required instances supported
+
+        if isinstance(instance, CIMInstance):
+            return instance.copy()
```
KSchopmeyer, line 2339: Same question as above about copy.

pywbem/_cim_operations.py
```diff
@@ -221,56 +221,129 @@ def _check_classname(val):
...
-    if (not isinstance(MaxObjectCount, six.integer_types) or
-            MaxObjectCount < 0):
+    if context is None:
         raise ValueError(
-            _format("MaxObjectCount parameter must be integer >= 0 but is "
-                    "{0!A}", MaxObjectCount))
-    if context is None or len(context) < 2:
+            _format("The 'context' parameter of the WBEMConnection "
+                    "operation has invalid value None "
+                    "(enumeration may be exhausted)"))
```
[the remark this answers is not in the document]
andy-maier: DONE - not added that :-) :-)

pywbem_mock/_instancewriteprovider.py
```diff
@@ -338,18 +338,13 @@ class for the new instance.
 
         # Parameter types are already checked by WBEMConnection operation
         assert isinstance(namespace, six.string_types)
+        assert isinstance(NewInstance, CIMInstance)
 
-        new_instance = NewInstance
+        new_instance = NewInstance.copy()
```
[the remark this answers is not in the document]
andy-maier: No.
There are two things worthwhile discussing:
* Is the original object modified: In the prior code, there was no copy at all before the path of the original object was modified. The new code shallow-copies the original object before modifying the path of the new object, which leaves the original object unchanged. So that part has improved.
* Which object is put into the repo: That is the same before and after the change, i.e. the original object is offered to `instance_store.create()` which performs a deep copy.

docs/changes.rst
andy-maier: What are you suggesting in terms of a change to the current text in the change history?

pywbem/_cim_operations.py
andy-maier: NOISSUE - This normalization you are commenting on is invoked at the first level in the WBEMConnection operations that is common for both the CIM-XML part (for parsing the namespace into its parts), and for the mocker. So the mocker already gets a normalized namespace, that is good to be a key in a case-insensitive dictionary. If the mock support strips again, that is not needed, and can be verified by having appropriate testcases.

pywbem/_cim_operations.py
andy-maier: NOISSUE - No.
We are only modifying the path, so all the rest of the object remains unchanged. When used for CIM-XML, it will not be stored longer term (and will not further be modified), and when used for the mock repo, it will be deep-copied upon adding it to the repo.

pywbem/_cim_operations.py
andy-maier: Same answer.

andy-maier:
I have responded to your comments. There is one question back to you.
So far, I made no changes to the PR (other than rebasing it).
Please review again.

docs/changes.rst
KSchopmeyer: “The May cause TypeErrors in sending client requests if parameters are incorrectly typed that could have slipped through the client and XML generation code before these strict type tests on all parameters were added.

KSchopmeyer: Something like “this may cause type errors in requests because of the new strict type checking “

KSchopmeyer:
Let’s fix the one comment and commit.

docs/changes.rst
andy-maier: The text already says that TypeError is now raised for invalid types. I tried to fit your sentence in, but any way I tried, it sounded pretty much like a duplication of the existing text about TypeError. I decided not to change it.

PR #2324, opened 2020-07-07
Fixes #2290: Reworked docs on mock server, specifically user-defined providers

### LINKED ISSUE #2290: Duplications in description of user-defined providers
The description of user-defined providers is distributed across multiple places, here for the example of instance write providers:
- Section "Creating user-defined instance providers", written in `docs/mockwbemserver.rst`.
- Section "Instance Write Provider", written in the module docstring of `pywbem_mock/_instancewriteprovider.py`.
- Section "InstanceWriteProvider", written in the class docstring of `InstanceWriteProvider` in `pywbem_mock/_instancewriteprovider.py`.

This leads to duplication that needs to be cleaned up.

andy-maier:
See commit message. The provider examples are now consistent with PR #2329.

It is probably best to review this PR by building the documentation and looking at section "Mock WBEM server".

### DIFF
--- a/docs/mockwbemserver.rst
+++ b/docs/mockwbemserver.rst
@@ -10,52 +10,53 @@ Mock WBEM server
 Overview
 --------
 
-The 'pywbem_mock' module of pywbem provides a mock WBEM server that
+The 'pywbem_mock' module of pywbem provides a *mock WBEM server* that
 enables using the pywbem client library without a real WBEM server.
-This module is useful for testing the pywbem client library itself as well
+This is useful for testing the pywbem client library itself as well
 as for the development and testing of Python programs that use the pywbem
 client library.
 
-The 'pywbem_mock' module contains the :class:`pywbem_mock.FakedWBEMConnection`
-class that establishes a *faked connection*. That class is a subclass of
+The :class:`pywbem_mock.FakedWBEMConnection` class establishes an in-process
+mock WBEM server and represents a *faked connection* to that mock WBEM server.
+That class acts as both the client API and as an API for managing the mocked
+WBEM server.
+
+The :class:`pywbem_mock.FakedWBEMConnection` class is a subclass of
 :class:`pywbem.WBEMConnection` and replaces its internal methods that use
-HTTP/HTTPS to communicate with a WBEM server with methods that communicate
-with  an in-process in-memory repository of CIM objects (the *mock repository*).
-
-:class:`~pywbem_mock.FakedWBEMConnection` acts as both the client API and an
-in-process fake WBEM server. It includes methods to establish, configure, and
-visualize this fake WBEM server. As a result, the operation methods of
+HTTP/HTTPS to communicate with a WBEM server with methods that communicate with
+the mock WBEM server. As a result, the operation methods of
 :class:`~pywbem_mock.FakedWBEMConnection` are those inherited from
-:class:`~pywbem.WBEMConnection`, so they have exactly the same input
-parameters, output parameters, return values, and even most of the raised
-exceptions, as when invoked on a :class:`~pywbem.WBEMConnection` object against
-a real WBEM server.
+:class:`~pywbem.WBEMConnection`, so they have exactly the same input parameters,
+output parameters, return values, and even most of the raised exceptions, as
+when invoked on a :class:`~pywbem.WBEMConnection` object against a real
+WBEM server.
 
-Each :class:`~pywbem_mock.FakedWBEMConnection` object creates its own mock
-repository that contains the same kinds of CIM objects a WBEM
-server repository contains: CIM classes, CIM instances, and CIM qualifier
-declarations types contained in CIM namespaces. Because
-:class:`~pywbem_mock.FakedWBEMConnection` operates only on the mock repository,
-the class does not have any connection or security-related constructor
+The mock WBEM server has an in-memory repository of CIM objects
+(the *CIM repository*). Each :class:`~pywbem_mock.FakedWBEMConnection` object
+creates its own CIM repository that contains the same kinds of CIM objects a
+WBEM server repository contains: CIM classes, CIM instances, and CIM qualifier
+declarations types, contained in CIM namespaces. Because
+:class:`~pywbem_mock.FakedWBEMConnection` operates only on the CIM repository,
+the class does not have any connection- or security-related constructor
 parameters.
 
 Like :class:`~pywbem.WBEMConnection`, :class:`~pywbem_mock.FakedWBEMConnection`
-has a default CIM namespace that is created upon
-:class:`~pywbem_mock.FakedWBEMConnection` instance creation.
-:class:`~pywbem_mock.FakedWBEMConnection` allows defining additional namespaces
-with :meth:`~pywbem_mock.FakedWBEMConnection.add_namespace()` .
+has a default CIM namespace that is created in the CIM repository upon
+:class:`~pywbem_mock.FakedWBEMConnection` object creation.
+Additional namespaces in the CIM repository can be created with
+:meth:`~pywbem_mock.FakedWBEMConnection.add_namespace()` .
 
-The mock repository must contain the CIM classes, CIM instances and CIM qualifier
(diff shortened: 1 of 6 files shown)


docs/mockwbemserver.rst
```diff
@@ -199,31 +200,32 @@ The mock WBEM server does NOT support:
...
 
-Faked WBEM operations
----------------------
+.. _`WBEM operations of a mock WBEM server`:
+
+WBEM operations of a mock WBEM server
+-------------------------------------
 
-The :class:`pywbem_mock.FakedWBEMConnection`  class supports the same WBEM
+The :class:`pywbem_mock.FakedWBEMConnection` class supports the same WBEM
```
KSchopmeyer, line 209: Can we add the work client before WBEM?

docs/mockwbemserver.rst
```diff
@@ -199,31 +200,32 @@ The mock WBEM server does NOT support:
...
 operations that are supported by the :class:`pywbem.WBEMConnection` class and
-in addition a set of methods to execute WBEM server responders for each of these
-client methods using an internal CIM repository of CIM classes, CIM instances,
-and CIM qualifier declarations.
+in addition a set of methods for setting up its mocked CIM repository, and
+for registering user-defined WBEM operation responders ("providers") for
+some the WBEM operations.
```
KSchopmeyer, line 213: Change "some the" to something like selected

docs/mockwbemserver.rst
```diff
@@ -199,31 +200,32 @@ The mock WBEM server does NOT support:
...
 
-These faked operations generally adhere to the behavior requirements defined in
-:term:`DSP0200` for handling input parameters and returning a result.
+These *faked WBEM operations* generally adhere to the behavior requirements
+defined in :term:`DSP0200` for handling input parameters and returning a result.
 
-The faked operations get the data to be returned from the mock repository of
-the faked connection, and put the data provided in operation parameters that
+The faked WBEM operations get the data to be returned from the mock repository of
```
KSchopmeyer, line 218: I think we are mixing the wording repository and phrase cim repostiory in different parts of our documentation.  Thus, here it could be mock CIM repository.  I would not get carried away trying for absolute consistency since the meaning is clear

docs/mockwbemserver.rst
```diff
@@ -836,389 +926,310 @@ mock environment.
...
 one. If a user-defined provider does not implement all WBEM operations, the
-default implementation is used.
+default implementation will be used.
 
 The following table shows the WBEM operations for which user-defined providers
 are supported, and the corresponding provider types:
 
-.. table:: WBEM operations of user-defined providers
+==============  ==============  ===========================================  ============================================
+WBEM operation  Provider type   Superclass                                   Description
```
KSchopmeyer, line 975: It took a bit to coordinate the refs and the header Description.  Maybe Provider Description or something would be a better word.

docs/mockwbemserver.rst
```diff
@@ -836,389 +926,310 @@ mock environment.
...
+operations) in user-defined providers may:
 
+* provide parameter or CIM repository validation in addition to the normal
+  request validation,
+* modify parameters of the request,
+* abort the request with a :class:`pywbem.CIMError` exception,
+* make modifications to the CIM repository.
 
-.. _`Mock CIM repository`:
+The Python operation methods may or may not call the corresponding superclass
```
KSchopmeyer, line 1104: Change may or may not call the corresponding superclass.  to "may call  the corresponding superclass method  to complete the CIM repository modification: or may implement the code to complete the modification.", Calling the corresponding superclass method insures that any validation that the superclass method is done (ex. insure that a CreateInstance does include all key properties) is done without duplicating code.

docs/mockwbemserver.rst
```diff
@@ -836,389 +926,310 @@ mock environment.
...
-Instance Write Provider
-^^^^^^^^^^^^^^^^^^^^^^^
+.. _`Mocker base classes`:
 
-.. automodule:: pywbem_mock._instancewriteprovider
+Mocker base classes
+-------------------
 
-.. autoclass:: pywbem_mock.InstanceWriteProvider
+Some bases classes are included in this documentation in order to provide the
```
KSchopmeyer, line 1195: Andy.   I am not sure but I think that the documentation for these classes pulls in the base class documentation so we should not be including base classes specifically.

pywbem_mock/_methodprovider.py
```diff
@@ -19,129 +19,81 @@
...
-      defined for the provider type. Methods that do not exist in the
-      user-defined provider default to the default method. Each of these
-      methods may:
+            if MethodName.lower() == 'method1':
+                if not isinstance(ObjectName, CIMInstanceName):
+                    raise CIMError(
+                        CIM_ERR_INVALID_PARAMETER,
+                        "CIM method {0} must be invoked on a CIM instance".
+                        format(MethodName))
+                return self.Tick(namespace, ObjectName, Params)
```
KSchopmeyer, line 70: What is a self.Tick(.

docs/mockwbemserver.rst
andy-maier: No. The derived classes have these lists of methods and attributes right after the class itself, and that includes any inherited elements. The element descriptions themselves are not pulled in, so if the base classes are not shown, the links in the attribute and method tables will not be generated as links, but just as bold text. Try it out yourself.

pywbem_mock/_methodprovider.py
andy-maier: oops. That is a leftover from an earlier version where I had a counter property and a Tick() method that increments the counter. Until I came to the point that using an existing MOF is probably better, even though the example is more abstract.

docs/mockwbemserver.rst
andy-maier: DONE

docs/mockwbemserver.rst
andy-maier: I agree that we do, and suggest the following terms:

* "CIM repository of the mock WBEM server" - fully specific term used when that context is needed.
* "mock repository" - short term for the above, that we define in the glossary that way, and introduce it on first usage in each main chapter.

docs/mockwbemserver.rst
andy-maier: DONE - changed to "Provider description" as proposed.

pywbem_mock/_methodprovider.py
andy-maier: DONE - changed back to "Method1".

andy-maier:
The last update addresses review comments.
Let's discuss the unresolved comments.

docs/mockwbemserver.rst
andy-maier: DONE

docs/mockwbemserver.rst
andy-maier: Karl agrees that we do not add "client"

docs/mockwbemserver.rst
andy-maier: We agreed to consolidate on the following terms:
* "CIM repository of the mock WBEM server" - fully specific term used when that context is needed.
* "mock repository" - short term for the above, that we define in the glossary that way, and introduce it on first usage in each main chapter.
* "provider" (subsumes "responder"), define in glossary, also relative to SNIA providers
* "provider method"  (subsumes "responder method" and "Python operation method")

docs/mockwbemserver.rst
andy-maier: Karl agrees.

docs/mockwbemserver.rst
andy-maier: I have some second thoughts on the term "mock provider": It is currently used much less than "CIM repository", and "CIM repository" seems to better hit the nail, IMO. So I will use "CIM repository", instead.

andy-maier:
The latest update resolves all comments except for one; see above.

docs/mockwbemserver.rst
andy-maier: Karl agrees

PR #2325, opened 2020-07-07
Fixes #2310: mockserver notebook update to pywbem 1.0.0

### LINKED ISSUE #2310: Jupyter notebook mockwbemserver.ipnb incorrect
This notebook reflects the api to load a schema before pywbem 1.0.0.  It needs to be updated to show both apis.  The code in the wbemserver_mock.py should suffice.

Since the old methods are not yet removed in 1.0.0b, this need not be incorporated immediatly.

KSchopmeyer:
See commit message.  Updated the code to reflect incompatible issues beteen pywbem pre 1 and pywbem 1.  This include move to user defined method provider and the replacement of compile_dmtf_schema method.


docs/notebooks/pywbemmock.ipynb
```diff
@@ -339,27 +339,65 @@
...
+    "                            ObjectName))\n",
+    "\n",
+    "        # Execute provider specific code and define any return parameters and\n",
+    "        # a return value\n",
+    "        return_params = None\n",
+    "        return_value = None\n",
+    "\n",
+    "        return (return_value, rtn_params)\n",
+    "\n",
+    "        raise CIMError(CIM_ERR_METHOD_NOT_AVAILABLE)\n",
```
andy-maier, line 400: Before merging this, I think we should flesh out which checks are done by the provider dispatcher in order to keep the provider simple. I have created issue #2326 for that.

KSchopmeyer: Done and documented for the method provider.   Note that I documented what validations we have in the pre-execution tests for the method provider in issue #2326. Note that we also don't test whether params are valid OUT/required in the response either.  The only one I can see that is missing would be test for valid param names and whether parameters are required/IN on input. To be honest we never did that in Pegasus but left it to the server method writer.  

Dropped the line with raise CIMError... from the code.   At this point the InvokeMethod itself is about 4 lines of code including displaying the input parameters.

KSchopmeyer: NOTE: I have not writted a user-defined instancewriteprovider for the notebook.

andy-maier:
Approved. Rebased it.

PR #2329, opened 2020-07-08
Fixes #2326: Simplified providers by moving checks to caller and improved docs

### LINKED ISSUE #2326: User providers still too complicated
If i look at the default implementations of user-defined instance write providers in `_instancewriteprovider.py` and at our examples for user-defined method providers (e.g. in the jupytr notebook in PR #2325 ), I think the code there is still to complicated. A lot of that is tied to the fact that the provider does a lot of checking.

I suggest that we move any checking that can be done generically, into the provider dispatcher. The default implementations of user-defined instance write providers in `_instancewriteprovider.py` can be simplified accordingly. Also, i think the default method provider in `_methodprovider.py` can get more code to show what a provider would typically do, and then still raise the same error as before, just more in the place where the actual code would be. I have some ideas on how that could be done.

andy-maier:
See commit message.

This PR is now complete from my perspective and should get a final review.
There is still this an open discussion item, see review comments below.

### DIFF
--- a/pywbem_mock/_instancewriteprovider.py
+++ b/pywbem_mock/_instancewriteprovider.py
@@ -210,7 +210,6 @@ class InstanceWriteProvider(BaseProvider):
 
           cimrepository (:class:`~pywbem_mock.BaseRepository` or subclass):
             The CIM repository to be used by mock WBEM server responders.
-            The repository must be fully initialized.
         """
         super(InstanceWriteProvider, self).__init__(cimrepository)
 
@@ -230,7 +229,7 @@ class InstanceWriteProvider(BaseProvider):
         Validation already performed by the provider dispatcher that calls
         this provider method:
         - The provider method is called only for the registered class and
-          namespace.
+          namespace (only applies to user-defined providers).
         - The Python types of all input parameters to this provider method are
           as specified below.
         - The namespace exists in the CIM repository.
@@ -249,13 +248,14 @@ class InstanceWriteProvider(BaseProvider):
           provider, depending on the model implemented by the provider.
           The CIM repository will reject any new instance that does not have
           all key properties specified.
-        - The new instance does not exist in the CIM repository.
+        - The new instance (i.e. an instance with the new instance path) does
+          not exist in the CIM repository.
           This validation needs to be done by the provider because the
-          determination of the key properties for the instance path may depend
-          on the model implemented by the provider.
+          determination of the key properties for the new instance path may
+          depend on the model implemented by the provider.
           The CIM repository will reject the creation of instances that already
           exist, so this check can be delegated to the repository once the
-          instance path has been determined.
+          new instance path has been determined.
 
         Parameters:
 
@@ -368,7 +368,7 @@ class InstanceWriteProvider(BaseProvider):
         Validation already performed by the provider dispatcher that calls
         this provider method:
         - The provider method is called only for the registered class and
-          namespace.
+          namespace (only applies to user-defined providers).
         - The Python types of all input parameters to this provider method are
           as specified below.
         - The classnames in modified_instance are consistent between
@@ -475,7 +475,7 @@ class InstanceWriteProvider(BaseProvider):
         Validation already performed by the provider dispatcher that calls
         this provider method:
         - The provider method is called only for the registered class and
-          namespace.
+          namespace (only applies to user-defined providers).
         - The Python types of all input parameters to this provider method are
           as specified below.
         - The namespace exists in the CIM repository.
--- a/pywbem_mock/_instancewriteprovider.py
+++ b/pywbem_mock/_instancewriteprovider.py
@@ -220,7 +220,7 @@ class InstanceWriteProvider(BaseProvider):
     #
     ####################################################################
 
-    def CreateInstance(self, namespace, NewInstance):
+    def CreateInstance(self, namespace, new_instance):
         """
         Implements a mock WBEM server responder for
         :meth:`pywbem.WBEMConnection.CreateInstance`.
@@ -242,10 +242,10 @@ class InstanceWriteProvider(BaseProvider):
           attributes (i.e. type, is_array, embedded_object).
 
         Validation that should be performed by this provider method:
-        - NewInstance does not specify any properties that are not
+        - new_instance does not specify any properties that are not
           allowed to be initialized by the client, depending on the model
           implemented by the provider.
-        - NewInstance specifies all key properties needed by the
+        - new_instance specifies all key properties needed by the
           provider, depending on the model implemented by the provider.
           The CIM repository will reject any new instance that does not have
           all key properties specified.
@@ -264,12 +264,16 @@ class InstanceWriteProvider(BaseProvider):
             created, in any lexical case, and with leading and trailing slash
             characters removed.
 
-          NewInstance (:class:`~pywbem.CIMInstance`):
+          new_instance (:class:`~pywbem.CIMInstance`):
             A representation of the CIM instance to be created.
 
             This object is a deep copy of the original client parameter, and may
             be modified by the provider as needed, before storing it in the
             CIM repository.
+
+            The property names in this object have been adjusted to match the
+            lexical case of the property definitions in the creation class
+            of the instance in the CIM repository.
 
             The `classname` attribute of this object will specify the
             creation class for the new instance, in any lexical case.
@@ -301,16 +305,7 @@ class InstanceWriteProvider(BaseProvider):
         # Since the existence of the class has already been verified, this
         # will always succeed.
         class_store = self.cimrepository.get_class_store(namespace)
-        creation_class = class_store.get(NewInstance.classname, copy=False)
-
-        # This default provider adjusts the lexical case of the property names
-        # in the new instance to match the property definitions in the creation
-        # class. A user-defined provider may or may not do that.
-        for inst_pn in NewInstance.properties:
-            inst_prop = NewInstance.properties[inst_pn]
-            cls_pn = creation_class.properties[inst_pn].name
-            if inst_pn != cls_pn:
-                inst_prop.name = cls_pn  # modifies NewInstance
+        creation_class = class_store.get(new_instance.classname, copy=False)
 
         # This default provider determines the instance path from the key
         # properties in the new instance. A user-defined provider may do that
@@ -319,8 +314,8 @@ class InstanceWriteProvider(BaseProvider):
         # properties exposed by the class are specified in the new instance,
         # and raises ValueError if key properties are missing.
         try:
-            NewInstance.path = CIMInstanceName.from_instance(
-                creation_class, NewInstance, namespace=namespace, strict=True)
+            new_instance.path = CIMInstanceName.from_instance(
+                creation_class, new_instance, namespace=namespace, strict=True)
         except ValueError as exc:
             raise CIMError(CIM_ERR_INVALID_PARAMETER, str(exc))
 
@@ -331,15 +326,15 @@ class InstanceWriteProvider(BaseProvider):
         # Store the new instance in the CIM repository, verifying that it does
         # not exist yet.
         try:
-            instance_store.create(NewInstance.path, NewInstance)
+            instance_store.create(new_instance.path, new_instance)
         except ValueError:
             raise CIMError(
                 CIM_ERR_ALREADY_EXISTS,
                 _format("New instance {0!A} already exists in namespace "
-                        "{1!A}.", NewInstance.path, namespace))
+                        "{1!A}.", new_instance.path, namespace))
 
         # CreateInstance returns the instance path of the new instance
-        return NewInstance.path
+        return new_instance.path
 
     ####################################################################
     #
@@ -347,8 +342,7 @@ class InstanceWriteProvider(BaseProvider):
     #
     ####################################################################
 
-    def ModifyInstance(self, ModifiedInstance,
-                       IncludeQualifiers=None, PropertyList=None):
+    def ModifyInstance(self, modified_instance, IncludeQualifiers=None):
         # pylint: disable=invalid-name,line-too-long,unused-argument
         """
         Implements a mock WBEM server responder for
@@ -357,7 +351,7 @@ class InstanceWriteProvider(BaseProvider):
         Modify an existing CIM instance in the CIM repository of the mock WBEM
         server.
 
-        NOTE: This method specifies the namespace in ModifiedInstance.path
+        NOTE: This method specifies the namespace in modified_instance.path
         rather than as a separate input parameter.
 
         The modification of the instance in the CIM repository that is performed
@@ -365,30 +359,11 @@ class InstanceWriteProvider(BaseProvider):
         addition of properties if not yet set on the instance), because
         instance-level qualifiers have been deprecated in CIM.
 
-        The set of properties that are to be modified in the CIM instance, is:
-
-        - If PropertyList is not None: All properties specified in the
-          PropertyList parameter, to the value specified in ModifiedInstance
-          including `None` (for NULL).
-
-          If a property specified in PropertyList is not specified in
-          ModifiedInstance, it must be set to the default value of the
-          property exposed by the creation class of the instance in the
-          CIM repository, or to any other value including `None` (for NULL),
-          if the property does not define a default value.
-
-        - If PropertyList is None: All properties specified in the
-          ModifiedInstance parameter, to the values specified there, including
-          `None` (for NULL).
-
-        Key properties must not be modified. Depending on the model implemented,
-        additional properties may be considered unmodifiable.
-
-        If a property in ModifiedInstance specifies the same value as in the
-        instance to be modified, that does not count as a modification
-        (for example, ModifiedInstance may legally specify key properties as
-        long as they have the same values as the key properties in the
-        instance to be modified).
+        The set of properties that are to be modified in the CIM instance has
+        already been determined by the caller so that the modified_instance
+        parameter specifies exactly the set of properties to be modified.
+        Therefore, this provider method does not have a property list
+        parameter.
 
         Validation already performed by the provider dispatcher that calls
         this provider method:
@@ -396,32 +371,30 @@ class InstanceWriteProvider(BaseProvider):
           namespace.
         - The Python types of all input parameters to this provider method are
           as specified below.
-        - The classnames in ModifiedInstance are consistent between
+        - The classnames in modified_instance are consistent between
           instance and instance path.
         - The namespace exists in the CIM repository.
         - The creation class of the instance to be modified exists in the
           namespace of the CIM repository.
         - The instance to be modified exists in the namespace of the CIM
           repository.
-        - All properties in ModifiedInstance that are to be modified are
+        - All properties in modified_instance that are to be modified are
           exposed (i.e. defined and inherited with any overrides resolved) by
           the creation class in the CIM repository, and have the same
           type-related attributes (i.e. type, is_array, embedded_object).
-        - All properties in PropertyList are exposed by the creation class in
-          the CIM repository.
         - No key properties are requested to change their values.
 
         Validation that should be performed by this provider method:
 
-        - ModifiedInstance does not specify any changed values for
+        - modified_instance does not specify any changed values for
           properties that are not allowed to be changed by the client,
           depending on the model implemented by the provider.
 
         Parameters:
 
-          ModifiedInstance (:class:`~pywbem.CIMInstance`):
+          modified_instance (:class:`~pywbem.CIMInstance`):
             A representation of the modified CIM instance, also indicating its
-            instance path.
+            instance path, with exactly the set of properties to be modified.
 
             This object is a deep copy of the original client parameter, and may
             be modified by the provider as needed, before storing it in the
@@ -435,11 +408,13 @@ class InstanceWriteProvider(BaseProvider):
             The `classname` attribute of this object will specify the creation
             class of the instance to be modified, in any lexical case.
 
-            The `properties` attribute of this object will specify the new
-            property values for the CIM instance, including `None` (for NULL),
-            with property names in any lexical case. The actual properties
-            to be modified must be determined by the provider as described for
-            the PropertyList parameter.
+            The `properties` attribute of this object will specify exactly the
+            set of properties that are to be updated, taking into account the
+            original ModifiedInstance and PropertyList input parameters of the
+            ModifyInstance() client call.
+            The lexical case of the property names has been adjusted to match
+            the lexical cae of the property definitions in the creation class
+            in the CIM repository.
 
             The `qualifiers` attribute of this object, if non-empty, should
             be ignored by the provider, because instance-level qualifiers have
@@ -449,13 +424,6 @@ class InstanceWriteProvider(BaseProvider):
             This parameter should be ignored by the provider, because
             instance-level qualifiers have been deprecated in CIM.
 
-          PropertyList (:term:`string` or :term:`py:iterable` of :term:`string` or `None`):
-            The names of the property or properties to be modified, in any
-            lexical case, or `None`.
-
-            Duplicate property names in PropertyList must be tolerated and
-            the duplicates ignored.
-
         Raises:
             :exc:`~pywbem.CIMError`: The provider may raise CIMError with any
               status code, and typically raises:
@@ -463,54 +431,24 @@ class InstanceWriteProvider(BaseProvider):
         """  # noqa: E501
         # pylint: disable=invalid-name,line-too-long
 
-        namespace = ModifiedInstance.path.namespace
+        namespace = modified_instance.path.namespace
 
-        # Get the creation class with all exposed properties and qualifiers.
-        # Since the existence of the class has already been verified, this
-        # will always succeed.
-        class_store = self.cimrepository.get_class_store(namespace)
-        creation_class = class_store.get(ModifiedInstance.classname, copy=False)
-
-        # Get the instance to be modified from the CIM repository.
-        # This is the original object in the repo, so changing this object
-        # updates it in the repo.
+        # Get a copy of the instance to be modified from the CIM repository.
         instance_store = self.cimrepository.get_instance_store(namespace)
-        instance = instance_store.get(ModifiedInstance.path, copy=False)
+        instance = instance_store.get(modified_instance.path)
 
-        # Set ModifiedInstance to have just the properties to be modified
-        if PropertyList is not None:
+        # Modify the properties of the local instance copy. The implemented
+        # approach is intentionally careful, to ensure that only the property
+        # values get updated.
+        for pn in modified_instance.properties:
+            mod_prop = modified_instance.properties[pn]
+            inst_prop = instance.properties[pn]
+            inst_prop.value = mod_prop.value  # Update the property value
 
-            # Create property set with only unique names from PropertyList
-            # The property names are stored in lower case to support case
-            # insensitivity properly.
-            property_set = set()
-            for pn in PropertyList:
-                property_set.add(pn.lower())
+        # Note that IncludeQualifiers is completely ignored.
 
-            # Add class default values for properties not specified in
-            # ModifiedInstance.
-            for pn in property_set:
-                if pn not in ModifiedInstance:
-                    # If the property in the class does not have a default
-                    # value, it is None.
-                    ModifiedInstance[pn] = creation_class.properties[pn].value
-
-            # Remove properties from ModifiedInstance that are not in
-            # PropertyList.
-            for pn in list(ModifiedInstance):
-                if pn.lower() not in property_set:
-                    del ModifiedInstance[pn]
-
-        # Adjust the lexical case of the properties in ModifiedInstance to the
-        # lexical case they have in the creation class.
-        for pn in ModifiedInstance.properties:
-            inst_prop = ModifiedInstance.properties[pn]
-            cl_prop = creation_class.properties[pn]
-            if inst_prop.name != cl_prop.name:
-                inst_prop.name = cl_prop.name  # changes ModifiedInstance
-
-        # Modify the instance in the CIM repository
-        instance.update(ModifiedInstance.properties)
+        # Replace the instance in the CIM repository with the local copy.
+        instance_store.update(modified_instance.path, instance)
 
     ####################################################################
     #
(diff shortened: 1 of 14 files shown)


KSchopmeyer:
I agree with the general approach.  Some suggestions:

1. We are going through the get_class method twice which is fairly expensive.  But we are using the class as readonly to get parameters for testing against instance.   Lets create bare_class method that makes the class in the repo available without the extra overhead.

2. I had always wanted to get the validation-checks out of the dispatcher because nobody really sees them there.  Why not create a method (ex. validate-createnstance(), etc.) for at least CreateInstance and ModifyInstance and put these as methods in InstanceWriteProvider.py so that dispatfcher simply calls the corresponding validation.  That way we have all of the "stuff" associated with these requests in their own module.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
 
     ####################################################################
     #
-    #   CreateInstance server request
+    #   CreateInstance responder
```
KSchopmeyer, line 223: I try to use the phrase server responder

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
-        Implements a WBEM server responder for
+        Implements a mock WBEM server responder for
         :meth:`pywbem.WBEMConnection.CreateInstance`.
 
-        Create a CIM instance in the local CIM repository of this class.
+        Create a new CIM instance in the CIM repository of the mock WBEM server.
 
-        The input parameters must have been validated before this method
-        is called including:
+        Validation that already has been performed by the caller:
```
KSchopmeyer, line 234: drop "that has".  Describe where this is in code.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
+        - The Python types of all parameters will be as specified.
+        - The namespace exists in the CIM repository.
+        - The creation class of the new instance exists in the namespace
+          in the CIM repository.
+        - All properties specified in the new instance are exposed by the
```
KSchopmeyer, line 239: why use exposed rather than "in"

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
-        2. The ``NewInstance`` is a valid CIM Instance.
-        3. All key properties defined in the class  of ``NewInstance`` are
-           in ``NewInstance``. This method cannot add properties or
-           property values to ``NewInstance``.
-        4. There are no properties in ``NewInstance`` that are not in the
-           corresponding class.
-        5. NewInstance path must not exist in the CIM repository.
+        Validation that should be performed by the provider:
+        - Existence of all key properties in the new instance, if the provider
+          depends on that for determining the instance path.
```
KSchopmeyer, line 245: Add "The mock CIM respository will reject any NewInstance that does not have all key properties defined."

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
+        - Non-existence of uninitializable properties in the new instance,
+          depending on the provider.
+        - Non-existence of the new CIM instance in the CIM repository.
```
KSchopmeyer, line 248: Add something like " The CIMRepository add method will reject the NewInstance if it already exists in the repository"

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
-            property values for the new CIM instance.
+            The ``properties`` attribute of this object will specify the
+            initial property values for the new CIM instance, with property
+            names in any lexical case. Key properties may or may not be
+            included.
 
-            Instance-level qualifiers have been deprecated in CIM, so any
-            qualifier values specified using the `qualifiers` attribute
-            of this object are ignored.
+            The ``path`` attribute of this object will be `None`.
```
KSchopmeyer, line 272: I did not check yet.  Do we clear it in the validation?

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
 
-        # Parameter types are already checked by WBEMConnection operation
-        assert isinstance(namespace, six.string_types)
-        assert isinstance(NewInstance, CIMInstance)
-
-        new_instance = NewInstance.copy()
+        # Get the creation class with all exposed properties and qualifiers.
+        # Since the existence of the class has already been verified, this
+        # will always succeed.
+        creation_class = self.get_class(
```
KSchopmeyer, line 293: Again, maybe bare_class since this is readonly

pywbem_mock/_instancewriteprovider.py
```diff
@@ -277,114 +220,117 @@ def __init__(self, cimrepository=None):
...
 
-        1. Namespace is valid in CIM repository.
-        2. The ``NewInstance`` is a valid CIM Instance.
-        3. All key properties defined in the class  of ``NewInstance`` are
-           in ``NewInstance``. This method cannot add properties or
-           property values to ``NewInstance``.
-        4. There are no properties in ``NewInstance`` that are not in the
-           corresponding class.
-        5. NewInstance path must not exist in the CIM repository.
+        Validation that should be performed by the provider:
```
KSchopmeyer, line 243: Should we also mention that we don't validate some special characteristics of property values (ex. EmbeddedObject type)

pywbem_mock/_instancewriteprovider.py
andy-maier: We use "exposed" in the DMTF specs, and it means "defined + inherited" elements.

pywbem_mock/_instancewriteprovider.py
andy-maier: It is cleared in `WBEMConnection.CreateInstance()`

pywbem_mock/_instancewriteprovider.py
andy-maier: We need all inherited properties.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - Changed it consistently to "mock WBEM server responder"

pywbem_mock/_instancewriteprovider.py
andy-maier: It is not in this code, which is why it states that it has been done by the caller.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - as proposed.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - as proposed.
Also added a motivation for that, by stating that only the provider can determine the instance path of the new instance.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - It turns out I was not aware that each class in the repository is stored with all exposed (defined + inherited, with overrides resolved) properties (and methods, fwiw). So we can go to the class repository and simply get it. Changed the code in both the dispatcher and the provider accordingly.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - I added checks for embedded_instance and reference_class.

andy-maier:
Updated the PR with all comments resolved as stated (and one rejected).
Please review again.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,192 +213,231 @@ def __init__(self, cimrepository=None):
...
+
+        NOTE: This method specifies the namespace in ModifiedInstance.path
+        rather than as a separate input parameter.
+
+        The modification of the instance in the CIM repository that is performed
+        by the provider should be limited to property value changes (including
+        addition of properties if not yet set on the instance), because
+        instance-level qualifiers have been deprecated in CIM.
+
+        The set of properties that are to be modified on the CIM instance, is:
```
KSchopmeyer, line 365: Think in the CIM Instance better than on the CIM instance

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,192 +213,231 @@ def __init__(self, cimrepository=None):
...
+
+        - If PropertyList is not None: All properties specified in the
+          PropertyList parameter, to the value specified in ModifiedInstance
+          (inlcuding `None` for setting to NULL).
```
KSchopmeyer, line 369: I would remove the ( ). At most a comma

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,192 +213,231 @@ def __init__(self, cimrepository=None):
...
+
+          If such a property is not specified in ModifiedInstance, it must be
```
KSchopmeyer, line 371: Change "such a property" to "a property in the PropertyList" . That such a propery occurs somewhere else (perhaps in the DMTF doc) and it is always causes some confusion for the reader trying to sort out such a.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,192 +213,231 @@ def __init__(self, cimrepository=None):
...
+          set to the default value of the property exposed by the creation
+          class of the instance in the CIM repository, or to any other value
+          (inlcuding `None`) if the property does not define a default value.
+
+        - If PropertyList is None: All properties specified in the
+          ModifiedInstance parameter, to the values specified there (inlcuding
+          `None` for setting to NULL).
+
+        Key properties must not be modified. Depending on the provider,
```
KSchopmeyer, line 380: Change "Depending on the provider" to "A user-defined provider"

KSchopmeyer:
NOTE: I have no issue with the comments you rejected. In fact exposed is really correct, just sounds wierd like something is being hidden since the opposite of exposed is not exposed.

pywbem_mock/_providerdispatcher.py
```diff
@@ -105,35 +111,86 @@ def CreateInstance(self, namespace, NewInstance):
         either to the default provider or to a provider registered for the
         namespace and class in InstanceName
         """
-        # Validate input parameters
 
-        self.validate_namespace(namespace)
+        # Verify the input parameter types (type errors have already been
+        # raised during checks in the WBEMConnection operation).
+        assert isinstance(namespace, six.string_types)
```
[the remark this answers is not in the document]
andy-maier: The WBEMConnection method already checks all of that and raises exceptions (e.g. TypeError). If you check it here again, that code will never run and will not be testable.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - Changed to "Depending on the model implemented,"

andy-maier:
Resolved further comments (except the one about the assertions), and now also rewrote the ModifyInstance code.

andy-maier:
The last update to this PR covers all four provider methods of user-defined providers, and also adjusted the namespace provider accordingly.
The documentation changes for the RTD docs that are part of PR #2324 will be adjusted as part of that PR.

This PR here is now complete from my perspective and should get a final review.
There is still this an open discussion item, see review comments above.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,337 @@ def __init__(self, cimrepository=None):
...
+        - The namespace exists in the CIM repository.
+        - The creation class of the new instance exists in the namespace
+          in the CIM repository.
+        - All properties specified in the new instance are exposed (i.e.
+          defined and inherited with any overrides resolved) by the creation
+          class in the CIM repository, and have the same type-related
+          attributes (i.e. type, is_array, embedded_object).
+
+        Validation that should be performed by this provider method:
+        - That NewInstance does not specify any properties that are not
```
KSchopmeyer, line 243: Drop That

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,337 @@ def __init__(self, cimrepository=None):
...
+        # Adjust the lexical case of the properties in ModifiedInstance to the
+        # lexical case they have in the creation class.
+        for pn in ModifiedInstance.properties:
+            inst_prop = ModifiedInstance.properties[pn]
+            cl_prop = creation_class.properties[pn]
+            if inst_prop.name != cl_prop.name:
+                inst_prop.name = cl_prop.name  # changes ModifiedInstance
+
+        # Modify the instance in the CIM repository
+        instance.update(ModifiedInstance.properties)
```
KSchopmeyer, line 510: Ignore comment.  Found code in providerdispatcher.

pywbem_mock/_methodprovider.py
```diff
@@ -180,86 +180,104 @@ def __init__(self, cimrepository=None):
...
+          - The target instance exists in the namespace of the CIM repository.
+          - The creation class of the target instance exposes the method to be
+            invoked.
+        - For class-level use:
+          - The target class exists in the namespace of the CIM repository.
+          - The target class exposes the method to be invoked.
+        - The set of specified CIM method input parameters matches exactly the
+          set of input parameters defined by the method in the CIM repository,
+          w.r.t. parameter name and type-related attributes (type, is_array,
+          embedded_object).
```
KSchopmeyer, line 218: Should add a not validated section to include 1) static qualifier, 2)IN OUT (required/optional) qualifiers

pywbem_mock/_providerdispatcher.py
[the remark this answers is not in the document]
KSchopmeyer: I am fine with the asserts.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE, all occurrences.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,337 @@ def __init__(self, cimrepository=None):
...
+
+        Validation that should be performed by this provider method:
+        - That NewInstance does not specify any properties that are not
+          allowed to be initialized by the client, depending on the model
+          implemented by the provider.
+        - That NewInstance specifies all key properties needed by the
+          provider, depending on the model implemented by the provider.
+          The CIM repository will reject any new instance that does not have
+          all key properties specified.
+        - That the new instance does not exist in the CIM repository.
```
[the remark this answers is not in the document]
andy-maier: That is done in the docs update in PR #2324 as per one of your comments there.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,337 @@ def __init__(self, cimrepository=None):
...
-            modified. Its `keybindings` attribute is required. Its
-            `namespace` attribute must be the namespace containing the instance
-            to be modified. Its `host` attribute will be ignored.
-
-            The `classname` attribute of the instance path and the `classname`
-            attribute of the instance must specify the same class name.
+            This object is a deep copy of the original client parameter, and may
+            be modified by the provider as needed, before storing it in the
+            CIM repository.
+            TBD: Should we actually deep copy?
```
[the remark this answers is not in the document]
andy-maier: DONE - removed the TBD

pywbem_mock/_methodprovider.py
andy-maier: We could in fact check that in the dispatcher.

andy-maier: I have added corresponding code to the dispatcher.

andy-maier:
I resolved all new comments, plus added support for checking Static and In.

Then, I found that I completely missed the validations that already happen in the `_imeth_...()` methods of `FakedWBEMConnection` before calling the dispatcher, and so right now there are several duplicate validations between that and the provider methods in the dispatcher. It seems to me we should consolidate that into one set of validations per method.

**DISCUSSION**

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,342 @@ def __init__(self, cimrepository=None):
...
+        creation_class = class_store.get(NewInstance.classname, copy=False)
+
+        # This default provider adjusts the lexical case of the property names
+        # in the new instance to match the property definitions in the creation
+        # class. A user-defined provider may or may not do that.
+        for inst_pn in NewInstance.properties:
+            inst_prop = NewInstance.properties[inst_pn]
+            cls_pn = creation_class.properties[inst_pn].name
+            if inst_pn != cls_pn:
+                inst_prop.name = cls_pn  # modifies NewInstance
```
andy-maier, line 313: DISCUSSION:

The adjustment of lexical case could be moved up to the dispatcher, if we believe this is good practice. If we think this is a don't care, it can be removed from here as well.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,342 @@ def __init__(self, cimrepository=None):
...
+                if pn not in ModifiedInstance:
+                    # If the property in the class does not have a default
+                    # value, it is None.
+                    ModifiedInstance[pn] = creation_class.properties[pn].value
+
+            # Remove properties from ModifiedInstance that are not in
+            # PropertyList.
+            for pn in list(ModifiedInstance):
+                if pn.lower() not in property_set:
+                    del ModifiedInstance[pn]
```
andy-maier, line 502: This is a general rule in DSP0200, and could thus be moved up to the dispatcher.

DISCUSSION

andy-maier:
Another ToDo is for the ModifyInstance dispatcher to validate that the PropertyList specifies no properties that are not exposed by the class.

pywbem_mock/_instancewriteprovider.py
andy-maier: We agreed to go for that.

pywbem_mock/_instancewriteprovider.py
andy-maier: We agreed to go for that.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE in latest commit.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE in latest commit.

andy-maier:
Karl, I implemented all change we discussed. Note that the semantics of the parameters for the provider methods for CreateInstance, ModifyInstance and InvokeMethod now have slightly changed, based on the changes to their input parameters performed by the dispatcher methods. In case of ModifyInstance, I decided to remove the propertylist parameter entirely, because now all the information is contained in the modified_instance parameter.

The are enough changes to suggest another review. Set back to review needed.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
         Parameters:
 
           cimrepository (:class:`~pywbem_mock.BaseRepository` or subclass):
-            Defines the repository to be used by request responders.  The
-            repository is fully initialized.
+            The CIM repository to be used by mock WBEM server responders.
+            The repository must be fully initialized.
```
KSchopmeyer, line 213: Not sure what we mean by fully initialized.  Not sure that sentence adds anything.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
...
-           in ``NewInstance``. This method cannot add properties or
-           property values to ``NewInstance``.
-        4. There are no properties in ``NewInstance`` that are not in the
-           corresponding class.
-        5. NewInstance path must not exist in the CIM repository.
+        Create a new CIM instance in the CIM repository of the mock WBEM server.
+
+        Validation already performed by the provider dispatcher that calls
+        this provider method:
+        - The provider method is called only for the registered class and
```
KSchopmeyer, line 232: I think this applies to user-defined providers only.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
...
+          implemented by the provider.
+        - new_instance specifies all key properties needed by the
+          provider, depending on the model implemented by the provider.
+          The CIM repository will reject any new instance that does not have
+          all key properties specified.
+        - The new instance does not exist in the CIM repository.
+          This validation needs to be done by the provider because the
+          determination of the key properties for the instance path may depend
+          on the model implemented by the provider.
+          The CIM repository will reject the creation of instances that already
```
KSchopmeyer, line 256: Maybe not here but I wonder if I clearly stated somewhere that "already exists" means that an instance with the same path exists.  i.e the key is the path

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
...
           namespace (:term:`string`):
-            The name of the CIM namespace in the CIM repository (case
-            insensitive). Must not be `None`. Leading or trailing slash
-            characters are ignored.
+            The name of the CIM namespace in which the CIM instance is to be
+            created, in any lexical case, and with leading and trailing slash
+            characters removed.
 
-          NewInstance (:class:`~pywbem.CIMInstance`):
+          new_instance (:class:`~pywbem.CIMInstance`):
```
KSchopmeyer, line 267: Why did we switch from NewInstance to new_instance for the provider?

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
...
                         "{1!A}.", new_instance.path, namespace))
 
-        # Create instance returns model path
+        # CreateInstance returns the instance path of the new instance
         return new_instance.path
 
     ####################################################################
     #
-    #   ModifyInstance server request
+    #   ModifyInstance mock WBEM server responder
```
KSchopmeyer, line 341: I see we are still using responder.  I am fine with it.

pywbem_mock/_instancewriteprovider.py
```diff
@@ -270,367 +209,280 @@ def __init__(self, cimrepository=None):
...
     #
     ####################################################################
 
-    def ModifyInstance(self, ModifiedInstance,
-                       IncludeQualifiers=None, PropertyList=None):
+    def ModifyInstance(self, modified_instance, IncludeQualifiers=None):
```
KSchopmeyer, line 345: Doesn't removing the property list make us non-compliant.  I agree that it was not worth much but I actually used it as an alternative to removing properties when, for example, I wanted to just change the value of a single property.  (get from server, modify property value, set PropertyList) so you don't have to worry about key properites since you just send the whole thing back with a propertylist.

KSchopmeyer: Ahha.  I did not look yet but does this mean that you cleaned them in the dispatcher so the property list no longer needed.

pywbem_mock/_methodprovider.py
```diff
@@ -180,86 +180,114 @@ def __init__(self, cimrepository=None):
...
+            :func:`py:tuple` (return_value, out_params): The return value and
+            output parameters of the method invocation:
 
-        Returns:
+            * return_value (:term:`CIM data type`):
+              Return value of the method invocation.
 
-            A :func:`py:tuple` of (returnvalue, outparams), with these
-            tuple items:
+            * out_params (list/tuple or dict):
```
KSchopmeyer, line 270: Why have multiple ways to construct the output parameters,  I would thing that just allowing CIMParameter would reduce the number of alternatives for the user.

pywbem_mock/_namespaceprovider.py
```diff
@@ -85,16 +87,14 @@ def __init__(self, cimrepository):
...
         if not self.find_interop_namespace():
             raise CIMError(
                 CIM_ERR_INVALID_PARAMETER,
-                _format("No Interop namespace found. "
-                        "Construction of CIM_Namespace provider aborted."
-                        "Namespaces found: ", ", ".
-                        join(self.interop_namespace_names)))
+                _format("Cannot create namespace provider (for class {0}): "
+                        "No Interop namespace found in the CIM repository. "
+                        "Interop namespaces searched: {1}",
```
KSchopmeyer, line 95: Maybe change searched to allowed.  I could go either way.

pywbem_mock/_providerdispatcher.py
```diff
@@ -95,207 +97,580 @@ def default_method_provider(self):
...
+        # Verify the input parameter types (type errors have already been
+        # raised during checks in the WBEMConnection operation).
+        assert isinstance(ModifiedInstance, CIMInstance)
+        assert isinstance(IncludeQualifiers, (bool, type(None)))
+        assert isinstance(PropertyList,
+                          (six.string_types, list, tuple, type(None)))
+        assert isinstance(ModifiedInstance.path, CIMInstanceName)
 
-        # Classnames in instance and path must match
+        # Verify consistency of the class names in the modified instance.
```
KSchopmeyer, line 236: Maybe change consistency to equality

pywbem_mock/_providerdispatcher.py
```diff
@@ -95,207 +97,580 @@ def default_method_provider(self):
...
+            * value (CIMParameter): Parameter value.
+            Types are validated.
+
+        Returns:
+
+          A tuple of (returnvalue, outparams), with these tuple items:
+
+            * returnvalue (CIM data type): Return value.
+
+            * outparams (NocaseDict): Output parameters, with:
```
KSchopmeyer, line 480: How does a user in an environment like pywbemcli use NocaseDict since that is a pywbem internal?  I understand why we have dict on input parameters, name case insensitivity but wouldn't just list of CIMParameter be easier???

pywbem_mock/_providerdispatcher.py
```diff
@@ -95,207 +97,580 @@ def default_method_provider(self):
...
+                        "not exist in namespace {1!A} of the CIM repository.",
+                        InstanceName.classname, namespace))
 
-        instance_store = self.cimrepository.get_instance_store(namespace)
+        # Verify that the instance to be deleted exists in the CIM repository.
         if not instance_store.object_exists(InstanceName):
             raise CIMError(
                 CIM_ERR_NOT_FOUND,
-                _format("Instance {0!A} not found in CIM repository",
+                _format("Instance to be deleted does not exist in the CIM "
```
KSchopmeyer, line 436: Generally we place data within the text stream.  In this case the path is at the end instead of in the text.   Just checking since one logical reason for that is that paths are long and messy.

pywbem_mock/_instancewriteprovider.py
andy-maier: I agree. Removed.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - good catch. Corrected.

pywbem_mock/_instancewriteprovider.py
andy-maier: DONE - Added some clarification here.

pywbem_mock/_instancewriteprovider.py
andy-maier: Yes, I did that as an indication that this is no longer the original NewInstance parameter provided to the client operation (e.g. lexical case of property names has been adjusted).

pywbem_mock/_instancewriteprovider.py
andy-maier: This will be changed as part of the docs update PR, which is where we discussed terminology. I just did not want to overload this PR here even more.

pywbem_mock/_methodprovider.py
andy-maier: The client operation does not return CIMParameter, but only the value (as dict(name:value)). So requiring the detour through CIMParameter is overhead, but since it is quite convenient I did not want to prohibit it.

On input parameters, the client operation is flexible to accept a list/tuple of CIMParameter and dict(name:value), which is transformed to only list/tuple of CIMParameter as input to the provider method. So there we already have the precedent case that the user has some flexibility in what they provide, and we narrow it down to a single representation.

The output parameter apply this principle as well, just this time the user provides the output parameter in the provider method (with flexibility), and we narrow it down to a single representation when passing it out of the client operation.

So there is some logic and consistency behind that.

Having said that, I'm open for discussion.

DISCUSSION

pywbem_mock/_namespaceprovider.py
andy-maier: NOISSUE: My logic was: If we find something, we have first searched it.

pywbem_mock/_providerdispatcher.py
andy-maier: DONE

pywbem_mock/_providerdispatcher.py
andy-maier: NOISSUE
Note, this is the return value of the dispatcher method. In the user-defined provider, the user does not need to create a NocaseDict. In case the output parameters are returned as dict, that is a standard dict.

Using a NocaseDict is different, that happens all the time and we describe that in the pywbem docs. But the user never needs to import the NocaseDict name.

pywbem_mock/_providerdispatcher.py
andy-maier: ANSWER
I did it that way because the path is long and messy, and this particular message text allowed to do that easily.

andy-maier:
I resolved some of the comments as proposed and marked those comments as resolved.
I responded to all other comments and left those marked unresolved.

pywbem_mock/_methodprovider.py
andy-maier: Karl agrees to the approach.

PR #2331, opened 2020-07-10
Fixes #2277:  MOF compiler ignores EmbeddedObject instance property

### LINKED ISSUE #2277: There is no code to compile value of inst with embedded object property in it
The apparent solution to compiling instances that contain impeddedobject properties is to define the value of the property as a string where the string contains the mof for the value of the embedded instance.

The compiler appears not to cover this but drops everything down to cimvalue() to evaluate the property value.  We need to add code to the compiler to test if the property has embedded... qualifier and then get the string value and recall the compiler.   Mike found this issue when trying to compile mocks for SNIA Array profile, CIM_CIMXMLCapabilities.  Have not yet tested but cannot find anywhere in the code that the correct decisions are made, i.e. instance_declaration() where the property in the class has the embedded instance or embedded object qualifier.

NOTE: I do not propose that we ever consider the generality of embeddedobject where the object could be a class and just consider the case of compiling an instance that contains an embedded instance property.

See DSP004 Appendix F1 for the rules and a example.

```
instance of CIM_CIMXMLCapabilities as $CIMXMLCap {
         InstanceID = "CIMXMLCapabilities Instance";
         AuthenticationMechanismsSupported = {0};
         ElementName = "WBEM Server CIMXMLCapabilities";
         GenericOperationCapabilities = "instance of CIM_GenericOperationCapabilitiesStructure {FQLSupported = false; CQLSupport = {0}; Validated = false; };";
         MaxConnections = 1;
         MultipleOperationsSupported = false;
         RequestedStatesSupported = {2,3,11};
         Validated = false;
         WBEMProtocolVersionsSupported = {2.0.0}; };

### PR DESCRIPTION
**Pull request by KSchopmeyer**
WARNING: This removes code in providerdispatcher that compares cls.embedded_object with instance.embedded_object.

LIMITATIONS;

1. Does not handle embedded class although that is only a few lines of code now.

2. Does not have test for recursive embedded.

Fixes #2277: MOF compiler ignores EmbeddedObject instance property.

The DMTF definition of the value of instance property MOF is one
or more strings containing the MOF that would be compiled into
CIM instances.

Originally the compiler simply passed the string through the compiler so
what went into a repository was simply a string for the value of the property.

The MOF compiler was extended to compile properties with the
EmbeddedInstance or EmbeddedObject qualifiers into CIMInstance
object(s).

The enhanced compiler creates CIMInstance objects so that the
instance compiled has a one or an  array of CIMInstance objects
corresponding

Limitations: The defintion of EmbeddedObject either an instance
or a class.  Since there are not instances of classes as
EmbeddedObject properties in the DMTF schema the compiler only
supports compiling instances.

### DIFF
--- a/tests/unittest/pywbem/test_mof_compiler.py
+++ b/tests/unittest/pywbem/test_mof_compiler.py
@@ -78,7 +78,7 @@ class MOFTest(unittest.TestCase):
         self.mofcomp = MOFCompiler(
             MOFWBEMConnection(),
             search_paths=[TEST_DMTF_CIMSCHEMA_MOF_DIR],
-            verbose=True,
+            verbose=False,
             log_func=moflog)
 
         self.partial_schema_file = None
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tests/unittest/pywbem/test_mof_compiler.py [hunk 1]
```diff
@@ -80,7 +78,7 @@ def moflog(msg):
         self.mofcomp = MOFCompiler(
             MOFWBEMConnection(),
             search_paths=[TEST_DMTF_CIMSCHEMA_MOF_DIR],
-            verbose=False,
+            verbose=True,
```
andy-maier, line 81: This should be reset to False as it displays a lot of stuff now.

tests/unittest/pywbem/testmofs/test_instance.mof
```diff
@@ -5,7 +5,7 @@
 #pragma include("test_types.mof")
 
 
-// TODO: Uncomment this to work on issues with instance MOF
+// TODO: Uncomment this to work on issues with embedded instance MOF
```
andy-maier, line 8: Not sure I understand this TODO. It is already uncommented. ??

tests/unittest/pywbem/test_mof_compiler.py
KSchopmeyer: Done

tests/unittest/pywbem/testmofs/test_instance.mof
KSchopmeyer: Actually the only thing I did was added the word embedded.  This is from a very old set of code (Bart days) and refers to the last two lines of property  definitions in the following mof which are to compile instances of  embedded instance properties.  

In retrospect, a quick review indicates that the particular MOF line indicated are actually broken.  They are trying to use an alias to substitute the mof in embedded... propeties when, in fact,  the alias inserts a path.  I removed the references in this test case that are broken

PR #2341, opened 2020-07-13
Fixed embedded_object check in CreateInstance dispatcher method

andy-maier:
**Note:** This PR is on top of the (rebased) PR #2331 .

See commit message.

### DIFF
--- a/pywbem_mock/_baseprovider.py
+++ b/pywbem_mock/_baseprovider.py
@@ -493,3 +493,50 @@ class BaseProvider(object):
             return instance_store.get(instance_name, copy=copy)
 
         return None
+
+    def is_subclass(self, klass, superclass, class_store):
+        """
+        Return boolean indicating whether a class is a (direct or indirect)
+        subclass of a superclass, or the same class, in the specified class
+        store of the CIM repository (i.e. in the namespace of that class store).
+
+        This test is done by starting at the class and walking its
+        superclasses up to the root, so it is not as expensive as determining
+        all subclasses of a given class.
+
+        Parameters:
+
+          klass (:term:`string`): Class name of the class.
+
+          superclass (:term:`string`): Class name of the superclass.
+
+          class_store (:class:`~pywbem_mock.BaseObjectStore`):
+            Class store of the CIM repository to search for the classes.
+
+        Returns:
+
+            bool: Boolean indicating whether the class is a (direct or indirect)
+            subclass of the superclass, or the same class.
+
+        Raises:
+
+            KeyError: Class or superclass does not exist in the class store.
+        """
+
+        # This check is at the begin in order to catch the case where class and
+        # superclass are the ame class and both do not exist:
+        klass_cls = class_store.get(klass)
+
+        if klass.lower() == superclass.lower():
+            # Found it
+            return True
+
+        next_klass = klass_cls.superclass
+        if next_klass is None:
+            # Exhausted, we are at the root class
+            # Check superclass existence
+            class_store.get(superclass)
+            return False
+
+        # Continue searching upwards
+        return self.is_subclass(next_klass, superclass, class_store)
--- a/pywbem_mock/_providerdispatcher.py
+++ b/pywbem_mock/_providerdispatcher.py
@@ -103,7 +103,7 @@ class ProviderDispatcher(BaseProvider):
         return "an array" if is_array else "a scalar"
 
     def _validate_property(
-            self, prop_name, instance, creation_class, namespace):
+            self, prop_name, instance, creation_class, namespace, class_store):
         """
         Validate a property of an instance against its declaration in a class.
         """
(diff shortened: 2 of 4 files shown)


pywbem_mock/_providerdispatcher.py
```diff
@@ -102,6 +102,82 @@ def _array_ness(is_array):
...
+                        "according to its creation class {3!A} in namespace "
+                        "{4!A} of the CIM repository",
+                        prop_name, prop_inst.is_array, prop_cls.is_array,
+                        instance.classname, namespace))
+
+        if isinstance(prop_inst.value, CIMInstance):
+            emb_classname_inst = prop_inst.value.classname
+            if 'EmbeddedInstance' in prop_cls.qualifiers:
+                emb_classname_cls = prop_cls.qualifiers['EmbeddedInstance']
+                # TODO: Expand test to allow inheritance
```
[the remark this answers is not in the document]
andy-maier: This tests the new or modified instance that is an input parameter of the CreateInstance or ModifyInstance client operation, against the creation class that is in the repository.

The client call may have been made by the user or by the MOF compiler. In case it is made by the user, the MOF compiler is not even involved for the input instance.

The same is true for the creation class that is already in the repository.

Maybe the MOF compiler should check less if it is not even always involved, and leave that to the client calls of the mocker, which are always involved.

Also, the subclass test is not very expensive, IMO. If we walk the superclasses of a given class up to the root, tha is sufficient for what we need here, and that is much better compared to determining all subclasses of a given class. And embedded instances are pretty rare.

DISCUSSION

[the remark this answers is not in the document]
andy-maier: I have put up an update to the PR that introduces a BaseProvider.is_subclass() method (including testcases) and that uses it for the EI test.

Please review.

pywbem_mock/_baseprovider.py
```diff
@@ -493,3 +493,42 @@ def _get_bare_instance(instance_name, instance_store, copy=None):
...
+            bool: Boolean indicating whether subclass is a (direct or indirect)
+            subclass of superclass.
+
+        Raises:
+            KeyError: Subclass or superclass does not exist in the class store.
+        """
+        if subclass.lower() == superclass.lower():
+            # Found it
+            return True
+        cls = class_store.get(subclass)
```
andy-maier, line 526: This check should be moved to the begin, in order to also fulfil the promise of raising KeyError for non-existing classes if both classes are invalid but equal.

pywbem_mock/_baseprovider.py
```diff
@@ -493,3 +493,42 @@ def _get_bare_instance(instance_name, instance_store, copy=None):
             return instance_store.get(instance_name, copy=copy)
 
         return None
+
+    def is_subclass(self, subclass, superclass, class_store):
+        """
+        Return boolean indicating whether subclass is a (direct or indirect)
+        subclass of superclass.
```
andy-maier, line 500: Clarify that subclass and superclass can also be equal for True to be returned.

pywbem_mock/_providerdispatcher.py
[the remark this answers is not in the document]
KSchopmeyer: The compiler must check everything except the is_subclass since it is the tool that compiles the property value.  I was thinking backwards when It thought of get_subclasses.  Good catch and yes, is_subclass should have been one of our standard methods long ago.

pywbem_mock/_baseprovider.py
```diff
@@ -493,3 +493,42 @@ def _get_bare_instance(instance_name, instance_store, copy=None):
...
+
+        This test is done by starting at the subclass and walking its
+        superclasses up to the root, so it is not as expensive as determining
+        all subclasses of a given class.
+
+        Parameters:
+
+          subclass (:term:`string`): Class name of subclass.
```
KSchopmeyer, line 508: Actually why  not use class in instead of  subclass.  It is looking for the suberclass of a class.

pywbem_mock/_providerdispatcher.py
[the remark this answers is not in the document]
andy-maier: ok

pywbem_mock/_baseprovider.py
andy-maier: I know, but it is a name conflict with the 'class' keyword, so we woud need to use `class_` or `klass`.
I'll use `klass`.

pywbem_mock/_baseprovider.py
andy-maier: DONE

pywbem_mock/_baseprovider.py
andy-maier: DONE

pywbem_mock/_baseprovider.py
andy-maier: DONE

PR #791, opened 2017-10-04
Weekly CI run of master branch on full set of environments

andy-maier:
This PR is used to run the CI tests on the master branch,
on the full set of OS / Python / package level combinations.

The test is scheduled to run on a weekly basis.

**Do not merge this PR!!**

PR #2342, opened 2020-07-14
Release 1.0.0b3

andy-maier:
Requesting approval for release of 1.0.0b3.

See PR #791 for a run with this code level on all environments.
See https://pywbem.readthedocs.io/en/latest/changes.html for the change log of 1.0.0b3.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -13,10 +13,10 @@ Change log
       :revisions: 1
 
 
-pywbem 1.0.0b3.dev1
--------------------
+pywbem 1.0.0b3
+--------------
 
-Released: not yet
+Released: 2020-07-15
 
 **Incompatible changes:**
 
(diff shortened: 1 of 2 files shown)


docs/changes.rst
```diff
@@ -82,9 +80,14 @@ Released: not yet
   qualifiers.  In this case, the property value is defined in the MOF as
   a string or array of strings that compiles to a CIMInstance.  This
   change does not compile CIMClass definitions.
-  Originally these compiled objects were passed through the compiler as 
+  Originally these compiled objects were passed through the compiler as
   strings. (See issue # 2277).
 
+* Mock support: Added a method BaseProvider.is_subclass() that tests whether
```
KSchopmeyer, line 86: I would move this one to enhancements or something. It is not really a bug

andy-maier: It is already in section Enhancements.

PR #2343, opened 2020-07-15
Start 1.0.0

andy-maier:
Requesting approval that we don't do another 1.0.0 beta but the next release is 1.0.0 final.


andy-maier:
We discussed this and concluded we would not have another 1. beta, so 1.0.0 final is our next 1.0 release.

PR #2345, opened 2020-07-15
Addressed Pylint issues

andy-maier:
See commit message.

New Pylint result with this PR:
```
+--------------------+------------+
|message id          |occurrences |
+====================+============+
|fixme               |131         |
+--------------------+------------+
|using-constant-test |8           |
+--------------------+------------+
|redefined-builtin   |2           |
+--------------------+------------+
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,8 +25,9 @@ Setup script for pywbem project.
 import sys
 import os
 import re
+# setuptools needs to be imported before distutils in order to work.
 import setuptools
-from distutils import log
+from distutils import log  # pylint: disable=wrong-import-order
 
 
 def get_version(version_file):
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@
 import os
 import re
 import setuptools
-from distutils import log
+from distutils import log  # pylint: disable=wrong-import-order
```
KSchopmeyer, line 29: I assume there is a reason we don't put it into the correct order.

andy-maier: Yes. setuptools needs to be import before distutils. If imported afterwards, it issues a UserWarning that it might not work. I'll add a comment to that extent.

andy-maier: DONE

PR #2348, opened 2020-07-16
Partly fixes #2253: Added NocaseList and corresponding tests

### LINKED ISSUE #2253: Mock: provider registry list methods are case sensitive
ProviderRegistry methods for listing namespaces `provider_namespaces()` and class names `provider_classes()` return case sensitive lists. For example:
```
'Interop' in registry.provider_namespaces()
```
does not properly support a case insensitive test, and thus requires the user to ensure case insensitive comparisons and tests. This can easily be forgotten and should ideally be done automatically, e.g. by returning a case insensitively comparing list.

andy-maier:
This PR adds a `NocaseList` class, but does not use it yet for anything other than testing.

The implementation and test cases are complete now.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -42,6 +42,10 @@ Released: not yet
 
 **Enhancements:**
 
+* Added a new class `pywbem.NocaseList` that is a case-insensitive but case-preserving
+  list. Objects of this class cannot be created by users, but are returned
+  from some functions (See issue #2253).
+
 **Cleanup:**
 
 **Known issues:**
--- a/pywbem/_nocaselist.py
+++ b/pywbem/_nocaselist.py
@@ -1,0 +1,335 @@
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


pywbem/_nocaselist.py
```diff
@@ -0,0 +1,326 @@
...
+compared, that is done case-insensitively. Whenever the value of list items
+is returned, they have the lexical case that was originally specified.
+
+It is used for lists of case-insensitive items such as namespaces returned
+by the pywbem mock support.
+
+Users of pywbem will notice ``NocaseList`` objects only as a result of pywbem
+functions. Users cannot create ``NocaseList`` objects.
+
+Except for the case-insensitivity of its items, it behaves like the built-in
```
KSchopmeyer, line 31: Is there anything that can be done with a list that cannot be done with the no case list? Ex. iteration/

andy-maier: No.

docs/changes.rst
```diff
@@ -42,6 +42,10 @@ Released: not yet
 
 **Enhancements:**
 
+* Added a new class `pywbem.NocaseList` that is a case-insensitive but case-preserving
+  list. Objects of this class cannot be created by users, but are returned
```
KSchopmeyer, line 46: This means that we have to duplicate this in pywbemtools to get the same functionality or use an internal class from pywbem.

andy-maier: Should we make it public? And NocaseDict as well?

KSchopmeyer:
APPROVED.

andy-maier:
This PR got manually unapplied from master, so legally, it never went in.

PR #2370, opened 2020-07-27
Used NocaseDict from nocasedict package

andy-maier:
See commit message.
Note this introduces some incompatibilities, as documented in the change history.


docs/changes.rst
```diff
@@ -58,6 +84,10 @@ Released: not yet
 * Docs: Switched to using the sphinx_rtd_scheme for the HTML docs
   (See issue #2367).
 
+* Replaced pywbem's own NocaseDict with NocaseDict from the nocasedict package
```
KSchopmeyer, line 87: Should this not be in the enhancements section or at least not in bug fixes.

pywbem/_cim_obj.py
```diff
@@ -368,7 +368,7 @@ def _qualifiers_tomof(qualifiers, indent, maxline=MAX_MOF_LINE):
     line_pos = indent + 1
 
     mof_quals = []
-    for q in qualifiers.itervalues():
+    for q in qualifiers.values():
```
KSchopmeyer, line 371: Was there a reason for this one to move to values rather than six.itervalues??

pywbem_mock/_baseprovider.py
```diff
@@ -401,10 +401,10 @@ class are returned. `None` means not supplied by client and the
 
         # local_only server default is True so True or None remove properties
         if local_only is True or local_only is None:
-            for prop, pvalue in klass.properties.items():
+            for prop, pvalue in list(klass.properties.items()):
```
KSchopmeyer, line 404: Would we be more consistent here if we went to six iteritemss for consistency.

This is probably my personal bete noir but the whole move of items/values itertems/itervalues between python 2 and 3 was a real mess and being forced to add list(..) to get the correct answer just adds to their mess.

Note that it is all my personal issue but I spend a lot of time relooking up the correct implementation for exactly this.

KSchopmeyer:
I reviewed and only had minor comments.

pywbem_mock/_baseprovider.py
andy-maier: In this loop, we are deleting items while iterating, and six.iteritems() returns an iterator on py2 and a dictionary view on py3. I'm pretty sure that neither of them guarantees to work if items are deleted while it iterates. In fact, the dictionary view raises RuntimeError if it detects that (I saw that happen at this code here before I added the list() around it). So if we used six.iteritems(), it would still need the list() around it to stabilize the iteration. In fact, for py3, the list() is needed in any case, because both items() and six.iteritems() returns the dictionary view.

pywbem/_cim_obj.py
andy-maier: My thinking was to use the values() method that is recommended for python 3. However, in python 2 that results in a small loss of efficiency compared to using itervalues(). The one advantage of the dictionary view returned by values() in python 3 (it can iterate multiple times) is not needed here, so I think going to six.itervalues() is actually a good recommendation.

pywbem_mock/_baseprovider.py
andy-maier: If you want to know what is most future-safe, it is definitely using values(), items(), keys().

They return a dictionary view object in python 3 which is basically an iterator that can be used to iterate multiple times. I think it is backwards compatible to the iterators returned by python 2's iter...().

docs/changes.rst
andy-maier: It is currently in the Cleanup section. I did not see much of a functional enhancement that would qualify it for the enhancement section.

andy-maier:
The current inconsistency between CIMInstance.keys() returning a view/list and CIMInstance.values/items() always returning a list is also a **DISCUSSION**.

docs/changes.rst
andy-maier: Ok. done.

pywbem/_cim_obj.py
andy-maier: TODO(Andy): Add comment explaining why we use values() in this case.

pywbem_mock/_baseprovider.py
andy-maier: TODO(Andy): Add comment explaining why we use list() in this case.

pywbem/_cim_obj.py
andy-maier: DONE - I just clarified the content of the qualifiers dict, that should be sufficient.

pywbem_mock/_baseprovider.py
andy-maier: DONE - added comment to all places that require using list() because of deleting while iterating.

PR #2382, opened 2020-07-30
Deprecated the iter..() methods of CIMInstanceName

andy-maier:
See commit message.

**DISCUSSION**:

* It is quite unsatisfactory to have CIMInstance.values/items() return lists. This is inconsistent, prevents deprecation of its iter..() methods, and does not provide a good path forward.
* It is equally unsatisfactory that NocaseDict.keys/values/items() returns lists on py2, even though it has dict view classes it could return also on py2.

Both could be addressed by having these methods return dict view objects on all Python versions. For NocaseDict, the price would be incompatibility with the built-in dict on py2 e.g. in cases where a dict is modified while iterating over it.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -64,6 +64,14 @@ Released: not yet
 
 **Deprecations:**
 
+* Deprecated the iterkeys(), itervalues() and iteritems() methods of
+  CIMInstanceName on Python 3, to be more consistent with the built-in dict
+  class that does not support these methods on Python 3. Use the keys(),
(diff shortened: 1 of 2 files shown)


andy-maier:
We discussed this:
* For NocaseDict, we think it is not worhtwhile to change anything for Python 2, which means it stays as it is (100% consistent with built-in dict)
* For CIMInstance.values/items(), we introduce derived dict views that benefit from the iteration through the properties NocaseDict, and resolve the CIMProperty to its value when being accessed. That will be in a separate PR.

Issue #2365, opened 2020-07-24
Make creation of mock environment faster


andy-maier:
MOF compiles during creation of mock environment take quite long. For example, users have about 300 classes in one namespace, which takes 20 sec to create it.

andy-maier:
As a first step, I think we need to understand what the time is spent for. I'll add metrics to the mock support.

andy-maier:
The statistic support ha been merged.
What we need now, are measurements using the new statistics support. Please post them here.

andy-maier:
We discussed this today, and believe that the outcome of any measurements will be that MOF compile is what consumes the time, and that caching the resulting repo would be what we need to go for.

We noted that it is not necessary to address the question of how to package profiles (i.e. issue #2409) in order to cache the repo, but we rather can do that at the level of the entire repo, as long as it is controlled by the connection definition as the top level handle to it.

andy-maier:
Karl, to confirm our expectation about where the time is spent, could you still please post some meaurements from the larger mock environment that you have access to?

PR #2403, opened 2020-08-06
Added time statistics support to pywbem_mock

andy-maier:
This is a first step to address the "make mock faster" issue #2365. It provides time statistics for the mock support, in order to find out where the time is spent.

There are no testcases for it yet, but it is ready to be tried out.

Usage:
```
conn.statistics.enabled = True  # or stats_enabled=True in ctor

# Do stuff with the faked connection, that collects time statistics:
conn.compile_schema_classes(...)
conn.compile_mof_string(...)
conn.compile_mof_file(...)
conn.add_cimobjects(...)
conn.EnumerateInstances(...) # and all other operations

print(conn.statistics.formatted())  # print formatted time statistics

stats = conn.statistics.snapshot()  # get a snapshot of the time statistics as a list of tuple(name, stats)
```

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -42,6 +42,10 @@ Released: not yet
   This provides consistency with the way the input parameters of the method
   provider are represented. (See issue #2415)
 
+* Added time statistics support to pywbem_mock, that allows measuring which
+  parts of the setup and execution of a mock environment takes how much time.
+  (Part of issue #2365)
+
 **Cleanup:**
 
 **Known issues:**
--- a/pywbem/_statistics.py
+++ b/pywbem/_statistics.py
@@ -395,7 +395,7 @@ class OperationStatistic(object):
             if not self._stat_start_time:
                 self._stat_start_time = self._start_time
 
-    def stop_timer(self, request_len, reply_len, server_time=None,
+    def stop_timer(self, request_len=None, reply_len=None, server_time=None,
                    exception=False):
         """
         This is a low-level method is called by pywbem at the end of an
@@ -438,8 +438,10 @@ class OperationStatistic(object):
         self._start_time = None
         self._count += 1
         self._time_sum += dt
-        self._request_len_sum += request_len
-        self._reply_len_sum += reply_len
+        if request_len is not None:
+            self._request_len_sum += request_len
+        if reply_len is not None:
+            self._reply_len_sum += reply_len
 
         if exception:
             self._exception_count += 1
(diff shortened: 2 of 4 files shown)


KSchopmeyer:
Tested and hit the following with an attempt to load Mikes Array mock which takes many seconds:

NOTE: I would not completely trust Mike's code either since I had to fix the compile_dmtf_schema to compile_schema_classes but I would not think that it should fail with unitialized timer.

```
\Mock Python script '/home/kschopmeyer/pywbem/pywbemtools/walkermocks/ArrayMockLoad.py' failed:

  File "/home/kschopmeyer/pywbem/pywbemtools/pywbemtools/pywbemcli/_pywbemcli_operations.py", line 355, in build_repository
    exec(file_code, globalparams, None)

  File "/home/kschopmeyer/pywbem/pywbemtools/walkermocks/ArrayMockLoad.py", line 103, in <module>
    verbose=False)

  File "/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/pywbem_mock/_wbemconnection_mock.py", line 687, in compile_schema_classes
    verbose=verbose)

  File "/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/pywbem/_statistics.py", line 607, in __exit__
    self._cm_opstat.stop_timer()

  File "/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/pywbem/_statistics.py", line 435, in stop_timer
    raise RuntimeError('stop_timer() called without preceding '

RuntimeError: stop_timer() called without preceding start_timer()

On the other hand I ran a test with a very small mock and got the following output:

```

 ppp -n mock1 -T
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli or <CTRL-r> to search history, 
pywbemcli> class enumerate --no
-/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/nocasedict/_nocasedict.py:278: UserWarning: Before Python 3.7, initializing or updating a NocaseDict object from a dict object is not guaranteed to preserve the order of its items
  self.update(*args, **kwargs)
CIM_Foo
  Count    Exc  Time               ReqLen    ReplyLen    Operation
-------  -----  -----------------  --------  ----------  -------------------------
      1      0  0.287              0/inf/0   0/inf/0     compile_mof_file(ns=None)
      4      0  0.011/0.006/0.014  0         0           CreateClass
     12      0  0.004/0.003/0.005  0         0           CreateInstance
     10      0  0.000/0.000/0.000  0         0           SetQualifier
      1      0  0.000              0         0           EnumerateClassNames

QUESTIONS are exc time and REQ/REPL lens inverted because for the operations there are 3 elements in exc time but for the compile_mof_fil there are 3 entries in REQLen and ReplyLen.

andy-maier:
On the exception: The problem was in the new context manager handling I added to the `Statistics` class (Mike's code was innocent :-) ). The context manager did not support nesting, but `compile_schema_classes()` calls `compile_mof_string()` so it was used in a nested way. I added support for nesting the context managers by changing the single value for the op_stats to become a stack, and added testcases for the use of the context manager, including nested use. BTW, I was able to reproduce the exception with nested use before I added support for it. I updated this PR accordingly. Please re-test.

On the result columns: The `Statistics.formatted()` method now hides the server time and request/reply length columns if they do not have values, so maybe there is another bug in that logic. Can you please let me know how you obtain and print the values?

andy-maier:
We discussed all 4 items in my comment above, and agree with the general approaches.

Karl wants to do a final review on this PR.

Issue #2383, opened 2020-07-30
test_indicationlistener.py fails on 0.17 with minimum package levels on py3 on macOS


andy-maier:
In Travis, the run on stable_0.17 with minimum package levels on macOS with Python 3 fails in test_indicationlistener.py. See https://travis-ci.org/github/pywbem/pywbem/jobs/711840570#L6298

```
tests/unittest/pywbem/test_indicationlistener.py ....Exception ('Connection aborted.', BadStatusLine('POST / HTTP/1.1\r\n'))
F.
. . .
=================================== FAILURES ===================================
________________________ TestIndications.test_send_100 _________________________
self = <tests.unittest.pywbem.test_indicationlistener.TestIndications testMethod=test_send_100>
    def test_send_100(self):
        """Test sending 100 indications"""
>       self.send_indications(100, 50000)
tests/unittest/pywbem/test_indicationlistener.py:248: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/unittest/pywbem/test_indicationlistener.py:228: in send_indications
    self.fail('Error return from send. Terminating.')
E   AssertionError: Error return from send. Terminating.
```

andy-maier:
The diagnostics for the assertion failures have been improved in PR #2397.
The transition to pytest in PR #2400 seems to have fixed the issue, but it is not clear why.

andy-maier:
Created a standalone testcase in PR #2402, and opened an issue against the "requests" package: https://github.com/psf/requests/issues/5550

andy-maier:
With a high probability this is not an issue in pywbem. Since the issue shows up only on pywbem 0.17, and since 0.17 is no longer fixed now that 1.0.0 final has been released, I am closing this issue as won't fix.

andy-maier:
Reopened this issue, as it shows up on the latest pywbem version as well and on other OS platforms than just macOS.

andy-maier:
Actually, I decided to create a new issue https://github.com/pywbem/pywbem/issues/2659 since the symptom of the newly occurring issues are different.

PR #2405, opened 2020-08-07
1.1: Improved logging in WBEM listener and its test module

andy-maier:
This was an outcome of the investigations on the bad status line issue #2383.


andy-maier:
Decided that we should have this change already in 1.0.0 final in case we start seeing the issue #2383 also on 1.0.0. Therefore, the content of this PR has been rebased on 1.0.0 in PR #2407, and this PR here has been abandoned.

PR #2413, opened 2020-08-08
Multiple fixes for OpenQueryInstances() in WBEMConnection, mock, testcases

andy-maier:
See commit message.

### DIFF
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -39,6 +39,7 @@ from __future__ import absolute_import, print_function
 import uuid
 from collections import Counter
 from copy import deepcopy
+import re
 import six
 from nocaselist import NocaseList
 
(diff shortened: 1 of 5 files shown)


pywbem_mock/_mainprovider.py
```diff
@@ -2506,7 +2508,19 @@ def _open_response(self, namespace, objects, pull_type, OperationTimeout,
...
             del objects[0: max_obj_cnt]
 
-        return(rtn_objects, eos, context_id)
+        return (rtn_objects, eos, context_id)
+
+    def _openquery_response(self, namespace, objects, pull_type,
+                            OperationTimeout, MaxObjectCount, ContinueOnError,
+                            QueryResultClass):
+        """
+        Like _openquery_response(), just adds QueryResultClass to the returned
```
KSchopmeyer, line 2517: I assume you meant "Like _open_response()"

pywbem_mock/_mainprovider.py
```diff
@@ -3146,16 +3160,28 @@ def OpenQueryInstances(self, namespace, FilterQueryLanguage, FilterQuery,
...
-        # Handle ReturnQueryResultClass maybe or make it exception
+        # pylint: disable=assignment-from-no-return
         instances = self.ExecQuery(namespace, FilterQueryLanguage, FilterQuery)
 
-        return self._open_response(namespace, instances,
-                                   'PullInstancesWithPath',
-                                   OperationTimeout,
-                                   MaxObjectCount,
-                                   ContinueOnError)
+        if ReturnQueryResultClass:
```
KSchopmeyer, line 3167: I think this is correct code but not sure why it became issue now since the ExecQuery always returns not implemented.

pywbem_mock/_mainprovider.py
andy-maier: DONE.

pywbem_mock/_mainprovider.py
andy-maier: This is one of the errors you get when running the pywbemtools test against pywbem 1.0.0.

PR #2421, opened 2020-08-15
Fixed erronously raised HeaderParseError for Content-type text/xml

andy-maier:
See commit message.


KSchopmeyer:
Since this was user discovered and the user is trying to work with pywbem and pywbemcli right now, I propose that rollback to 1.0.2 is justified

andy-maier:
Agreed. It has already been rolled back to 1.0.2 as PR #2422

PR #2424, opened 2020-08-16
Added support for translating PUnit and Units qualifiers to SI units

andy-maier:
See commit message.
**Note: If this PR is merged, PR https://github.com/pywbem/pywbemtools/pull/731 needs to be updated to get pywbem from master.**

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -54,6 +54,11 @@ Released: not yet
   a class is a subclass of a class. Both classes can be specified as classnames
   or as ``CIMClass`` objects.
 
+* Added support for translating the values of ``PUnit`` and ``Units``
+  qualifiers into human readable SI conformant unit strings, via new
+  functions ``pywbem.siunit_obj()`` and ``pywbem.siunit()``. These new
+  functions are marked as experimental. (See issue #2423)
+
 **Cleanup:**
 
 **Known issues:**
--- a/pywbem/_units.py
+++ b/pywbem/_units.py
@@ -1,0 +1,758 @@
+# -*- coding: utf-8 -*-
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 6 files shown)


andy-maier:
Karl and I decided that this new function is not rolled back into 1.0.

docs/changes.rst
```diff
@@ -54,6 +54,11 @@ Released: not yet
   a class is a subclass of a class. Both classes can be specified as classnames
   or as ``CIMClass`` objects.
 
+* Added support for translating the values of ``PUnit`` and ``Units``
+  qualifiers into human readable SI conformant unit strings, via new
+  functions ``pywbem.siunit_obj()`` and ``pywbem.siunit()``.
```
KSchopmeyer, line 59: Should we not that this only works on servers that support classes?

pywbem/_units.py
```diff
@@ -0,0 +1,749 @@
...
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
+from nocasedict import NocaseDict
+import six
+
+from ._cim_obj import CIMProperty, CIMMethod, CIMParameter
+from ._utils import _ensure_unicode
+
+__all__ = ['siunit_obj', 'siunit']
```
KSchopmeyer, line 106: Should we put the public functions at the top of the module and the supporting ones further down in the code since it is primarily the public modules that people will look at.

docs/changes.rst
andy-maier: These functions do not perform class operations under the covers. However, the code in pywbemtools that uses them does. I added a note that "these functions do not perform any class operations; they take the qualifiers as input."

pywbem/_units.py
andy-maier: Moved them to the top.

andy-maier:
According to the discussion in the slack channel, I marked the new functions as experimental.

PR #2439, opened 2020-09-11
Mock: Added a registry of provider dependent files

andy-maier:
This is in support of https://github.com/pywbem/pywbemtools/issues/748.

### DIFF
--- a/docs/mockwbemserver.rst
+++ b/docs/mockwbemserver.rst
@@ -1179,6 +1179,39 @@ User-defined method providers
    .. rubric:: Details
 
 
+.. _`Registry for provider dependent files`:
+
+Registry for provider dependent files
+-------------------------------------
+
+A faked WBEM connection provides a registry for provider dependent files
+in its :attr:`~pywbem_mock.FakedWBEMConnection.provider_dependent_registry`
+property of type :class:`pywbem_mock.ProviderDependentRegistry`.
+
+This registry can be used by callers to register and look up the path names of
+additional dependent files of a mock script, in context of that mock script.
+
+The pywbemtools project makes use of this registry for validating whether its
+mock cache is up to date w.r.t. additional dependent files a mock script has
+used.
+
+.. autoclass:: pywbem_mock.ProviderDependentRegistry
+   :members:
+
+   .. rubric:: Methods
+
+   .. autoautosummary:: pywbem_mock.ProviderDependentRegistry
+      :methods:
+      :nosignatures:
+
+   .. rubric:: Attributes
+
+   .. autoautosummary:: pywbem_mock.ProviderDependentRegistry
+      :attributes:
+
+   .. rubric:: Details
+
+
 .. _`Configuration of mocked behavior`:
 
 Configuration of mocked behavior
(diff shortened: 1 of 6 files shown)


andy-maier:
We agreed to pursue this path.

pywbem_mock/_providerdependentregistry.py
```diff
@@ -0,0 +1,136 @@
...
+from pywbem._utils import _format
+
+__all__ = ['ProviderDependentRegistry']
+
+
+class ProviderDependentRegistry(object):
+    """
+    A registry for provider dependent files in context of a mock script.
+
+    This registry allows registering additional dependent files in context of
```
KSchopmeyer, line 31: Should we be more complete here to state the reason for register and lookup.  Thus to register dependent files when a mock script is cached and to look up and validate the mock_script file dependencies when the cached mock script is loaded.

docs/mockwbemserver.rst
```diff
@@ -1179,6 +1179,39 @@ User-defined method providers
...
+-------------------------------------
+
+A faked WBEM connection provides a registry for provider dependent files
+in its :attr:`~pywbem_mock.FakedWBEMConnection.provider_dependent_registry`
+property of type :class:`pywbem_mock.ProviderDependentRegistry`.
+
+This registry can be used by callers to register and look up the path names of
+additional dependent files of a mock script, in context of that mock script.
+
+The pywbemtools project makes use of this registry for validating whether its
```
KSchopmeyer, line 1194: Change "makes use of this registry" to "uses this registry"

docs/mockwbemserver.rst
```diff
@@ -1179,6 +1179,39 @@ User-defined method providers
...
+.. _`Registry for provider dependent files`:
+
+Registry for provider dependent files
+-------------------------------------
+
+A faked WBEM connection provides a registry for provider dependent files
+in its :attr:`~pywbem_mock.FakedWBEMConnection.provider_dependent_registry`
+property of type :class:`pywbem_mock.ProviderDependentRegistry`.
+
+This registry can be used by callers to register and look up the path names of
```
KSchopmeyer, line 1191: Should we explain why register and lookup are used?

KSchopmeyer (commented):
Some editorial comments. Otherwise approved

pywbem_mock/_providerdependentregistry.py
andy-maier: The possible use for validating a cache in pywbemtools is already described in the mockwbemserver.rst file just before this class is shown, as:
```
The pywbemtools project makes use of this registry for validating whether its
mock cache is up to date w.r.t. additional dependent files a mock script has
used.
```

docs/mockwbemserver.rst
andy-maier: The next sentence does that:
```
The pywbemtools project makes use of this registry for validating whether its
mock cache is up to date w.r.t. additional dependent files a mock script has
used.
```

andy-maier:
I think both comments had been already addressed.

PR #2446, opened 2020-09-22
Fixes #2327-incomplete mock tests

### LINKED ISSUE #2327: Mock tests incomplete
Several of the tests in test_wbemconnection.py are incomplete  and were marked with TODO comments.  This issue consolidates the TODOs into a collection of items that need to be expanded on these tests:

This is in response to issue #1240 and only documents the TODOs, not fix them.

1. test_associator_instances - Should test more parameter options and add tests for IncludeQualifiers, PropertyList, IncludeClassOrigin

2. def test_associator_classes - review test to assure that it has covered all the essential parameter and options and test for correct properties in responses.

3. def test_get_instance -  Reorganize the test to match the  default and configurable parameter definitions

4. def test_remove_namespace(...) Test inconsistently uses methods in FakedWbemconnection, mainprovider, datastore. Make it more consistent

5. test_display_registered_providers - Extend test to include register_provider option namespace=None

6. def test_createclass - This test does not validate the parameters of each method.

7. def test_enumerateinstancenames() add error tests to this test or another test for errors.

8. In several tests, we use a fixed numeric for things like the number of returned instances.  That should be computed from
    the input, See test_enumerateinstances

9. def test_enumerateinstances_di - Expand this test for iq, ico, pl  1.  Extend test to test instances returned against expected instances.  There is a note   about inconsistency in the test definition clarify this.  PRIORITY

10. def test_modifyinstance() - Add additional tests for cases with only some properties in modified instance, 

11. Clean up tests around add and delete instance namespace (ex. test_delete9nstance_namespace) which are marked skip because they are superceeded with pywbem 1.0.0

12. def test_openassociatorinstances - Add test for actual instances returned. Today we only test for correct paths

13. Combine tests test_reference_instances_min and  test_reference_instances_opts.. Also add test for acutal instance characteristics

14. test_enumerateinstances_di - This test contains a print statement as part of a test that needs to be cleaned up.

15. test_associator_instnames - This test needs to add more test cases for the various parameter options.

16. test_createinstance - This test should be rewritten to simplify the test, either separate the error tests from this test or build in the options in a cleaner manner.

17. test_createclass - This test needs to add code for more specific tests of the results of the resolver.  The current code is very generic and does not test all the options possible with the resolve process.

18. test_enumerateinstances - Improve the algorithm for determining the propertylist results with the di parameter.

KSchopmeyer:
Added and extended tests for pywbem_mock documented in issue #2327 and
where the coverage report showed untested code.

Remove unused method _make_tuple in _mainprovider.py

NOTE: This leaves 3 of the items from issue # 2327 in the code open and two other tests (the original namespace tests) marked skipped for the moment.  Otherwise all tests extended per iseu # 2327 and the comments removed .

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,4 +1,5 @@
 
+
 .. _`Change log`:
 
 Change log
@@ -87,6 +88,9 @@ Released: not yet
 * Mock: Cleaned up the output of repr(BaseProvider) to no longer show the
   CIM repository, and to include the other attributes that were not shown so
   far. (See issue #2432)
+
+* Complete pywbem_mock tests that were documented as missing in issue.
+  (see issue # 2327)
 
 * Removed dependency on package custom-inherit and removed package from
   pywbem.  (see issue # 2436)
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -201,14 +201,6 @@ class MainProvider(ResolverMixin, BaseProvider):
     #####################################################################
 
     @staticmethod
-    def _make_tuple(rtn_value):
-        """
-        Change the return value from the value consistent with the definition
-        in cim_operations.py into a tuple in accord with _imethodcall
-        """
-        return [("IRETURNVALUE", {}, rtn_value)]
-
-    @staticmethod
     def _get_superclass_names(classname, class_store):
         """
         Get list of superclasses names from the class datastore for the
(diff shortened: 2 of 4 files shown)


andy-maier (approved):

KSchopmeyer:
**DISCUSSION**  We did NOT rewrite the tests to use simplified_test_function. As far as I can tell, the simplified_test_function is incompatible with the fixtures we have defined to create qualifiers, etc. in repository since the fixtures must be defined as a parameter for each test and the simplified_test_function controls those parameters.

andy-maier:
We agreed that simplified_test_function just cannot be used in these cases.
This PR fixes issue #2327 partly, and the issue describe what is left.

PR #2465, opened 2020-10-02
Docs: Pinned Jinja2 (used by Sphinx) to <2.11 on Python 3.4

andy-maier:


andy-maier:
Withdrawn, since recent runs succeed.

PR #2469, opened 2020-10-02
0.17: Added pip environment info; Mitigated virtualenv creation issue

andy-maier:
Rolls back PR #2467 into 0.17.6


andy-maier:
Merging this PR even though it fails to break PR dependency cycle - the fixes are rollbacks.

PR #2476, opened 2020-10-03
Deprecated propagation of key properties to path in CIMInstance

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -25,6 +25,26 @@ Released: not yet
 
 **Deprecations:**
 
+* Deprecated the propagation of key property value changes to corresponding
+  path keybindings in `CIMInstance` objects. A DeprecationWarning is now
+  issued in that case. A future release of pywbem will remove the propagation.
+  If you change key property values of a CIMInstance object that has a path set
+  and depend on the corresponding keybinding values in the path to also change,
+  then you should now change these keybindings values in your code instead of
+  relying on the automatic propagation.
+
+  Reasons for this deprecation are:
+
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -25,6 +25,24 @@ Released: not yet
...
+  If you change key property values of a CIMInstance object that has a path set
+  and depend on the corresponding keybinding values in the path to also change,
+  then you should now change these keybindings values in your code instead of
+  relying on the automatic propagation.
+
+  Reasons for this deprecation are:
+
+  - There are valid scenarios to have the keybindings different from the key
+    properties, for example when passing an instance to the ModifyInstance
+    operation that attempts to modify the key property values of an instance
```
KSchopmeyer, line 40: I would remove the last comment about an example being test.

KSchopmeyer: Interesting that we do not have a sync path method, however. We do have a method to create the path from the instance and that is sufficient.

KSchopmeyer (commented):
Left couple of comments

docs/changes.rst
andy-maier: I removed the example in parenthesis but left the item in.
I added an item:

  - Propagating the update of a key property value to the path is a side effect
    and complicates an otherwise simple operation.

andy-maier:
I updated the PR as explained in my response to the comment above. Please re-review.

docs/changes.rst
KSchopmeyer: OK

Issue #2502, opened 2020-10-11
pywbem_mock ompiling schemas into multiple namespaces fails after the first namespace


KSchopmeyer:
Compiling schemas into the mock cim repository  (using DMTFCIMSchema pragma to compile the qualifier declaractions and a set of classes from leaf classes into two different namespaces fails on the second namespace.

It fails on trying to compile the first class in the list because it cannot find the qualifier declaractions.

This  is an issue in the compiler and would fail with either pywbem_mock or another environment.

The problem is the namespace attribute is missing on an EnumerateQualifiers statement when the compiler tries to build the local qualifier list and the second compile gets the qualifier list from the first compile

PR #2503, opened 2020-10-11
Fixes issue #2502 pywbem compiler issue building in second namespace

KSchopmeyer:
The MOF compiler cannot install a DMTF schema in a second namespace
because it tries to get the qualifiers from an incorrect remote
namespace.  This is a one line fix to the   compiler.

Adds test for pywbem_mock compiling into multiple namespaces.

Marked as priority because while this is a mof compiler issue it directly affects the last pywbem 0.8.0. This is blocking pywbemtools 0.8.0 release.

NOTE: I did not add comment to change log because it still reflects 1.1.0-dev.

### DIFF
--- a/pywbem/_mof_compiler.py
+++ b/pywbem/_mof_compiler.py
@@ -816,7 +816,8 @@ def p_mp_setQualifier(p):
     ns = p.parser.target_namespace
     if p.parser.verbose:
         p.parser.log(
-            _format("Setting qualifier {0!A}", qualdecl.name))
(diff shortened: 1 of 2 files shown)


andy-maier (approved):

andy-maier:
Needs to also go into master

andy-maier:
Will add the following change history entry in a future commit:

* MOF compiler: Fixed bug where MOF compiler did not correctly install a CIM schema
  in a non-default namespace because it tried to get the qualifiers from the
  default namespace. (see issue #2502)

PR #2504, opened 2020-10-12
1.1: Fixes issue #2502 pywbem compiler issue building in second namespace

andy-maier:
Rolls back PR #2503 into 1.1.1.


andy-maier:
Will add the following change history entry in a future commit:

* MOF compiler: Fixed bug where MOF compiler did not correctly install a CIM schema
  in a non-default namespace because it tried to get the qualifiers from the
  default namespace. (see issue #2502)

Issue #2509, opened 2020-10-19
Performance: Add usage of __slots__


andy-maier:
The resource consumption measurements show that adding `__slots__` to the CIM objects reduces their memory footprint in the 10%+ range.

**DISCUSSION**: The only downside of using `__slots__` is that users cannot add attributes to the CIM objects anymore. This could still be achieved by adding `__dict__` as a slot, but I think we don't want to have that anyway.

andy-maier:
We agreed to accept the downside of users not being able to dynamically add attributes to CIM object/type objects. They can subclass them if needed.

andy-maier:
Results of using slots:

```
tests/resourcetest/test_enuminst.py 
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=1):
Debug: cpu_times: at start: pcputimes(user=0.49475424, system=0.124179096, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.496206368, system=0.124188136, children_user=0.0, children_system=0.0)
  Elapsed time:        1.474 ms
  CPU time:            1.461 ms
  Result size:          2750 B
  Peak usage:          36388 B
.
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=10):
Debug: cpu_times: at start: pcputimes(user=0.503950784, system=0.12431972, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.511262784, system=0.124327232, children_user=0.0, children_system=0.0)
  Elapsed time:        7.309 ms
  CPU time:            7.320 ms
  Result size:         23624 B
  Peak usage:          41630 B
.
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=100):
Debug: cpu_times: at start: pcputimes(user=0.545044992, system=0.12475792, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.617914368, system=0.125215736, children_user=0.0, children_system=0.0)
  Elapsed time:       73.600 ms
  CPU time:           73.327 ms
  Result size:        231222 B
  Peak usage:         240361 B
.
Resource consumption test: test_enuminst_mock(num_props=10, num_insts=1):
Debug: cpu_times: at start: pcputimes(user=0.638165824, system=0.125434256, children_user=0.0, children_system=0.0)
[...]

PR #2501, opened 2020-10-11
Resource consumption tests

andy-maier:
Ready for review.
For details, see commit message.

Notes:
* There is no measurement of CIM-XML based resource consumption at this point. It would require adding server-side CIM-XML support to the mock environment, and that seemed too big of a task right now.

Output of `make resourcetest`:
```
tests/resourcetest/test_enuminst.py 
Resource consumption test: EnumerateInstances with num_props=1, num_insts=1:
  Elapsed time:        2.395 ms
  CPU time:            2.285 ms
  Peak usage:          38895 B
  Result size:          3960 B
.
Resource consumption test: EnumerateInstances with num_props=10, num_insts=100:
  Elapsed time:      230.473 ms
  CPU time:          229.457 ms
  Peak usage:         842896 B
  Result size:       1108848 B
.
Resource consumption test: EnumerateInstanceNames with num_props=1, num_insts=1:
  Elapsed time:        0.434 ms
  CPU time:            0.440 ms
  Peak usage:           6534 B
  Result size:          1416 B
.
Resource consumption test: EnumerateInstanceNames with num_props=10, num_insts=100:
  Elapsed time:      109.410 ms
  CPU time:          108.617 ms
  Peak usage:         131364 B
  Result size:         83252 B
.
```

PR #2510, opened 2020-10-19
Added usage of __slots__ to CIM objects and CIM types

andy-maier:
**Note: This PR is on top of PR #2501 - review only the delta.**

For seeing the resource savings by this change, compare the output of `make resourcetest` between this PR and PR #2501. Results have been posted to issue #2509.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -60,6 +60,11 @@ Released: not yet
 * Reduced memory consumption of CIM objects and CIM types by defining their
   attributes to use Python slots. (see issue #2509)
 
+* Reduced memory consumption of CIM objects by using lazy initialization of
+  dictionary-type attributes. This resulted in significant savings when the
+  attribute is typically unused, for example in ``CIMInstance.qualifiers``.
+  (see issue #2511)
+
 **Cleanup:**
 
 * Test: Fixed all remaining ResourceWarnings during test. (issue #86)
--- a/pywbem/_cim_obj.py
+++ b/pywbem/_cim_obj.py
@@ -1270,14 +1270,16 @@ class CIMInstanceName(_CIMComparisonMixin, SlottedPickleMixin):
             v2 = instpath['k2']  # Access value of "k2"
             del instpath['k2']  # Delete "k2" from the instance path
         """
+        if self._keybindings is None:  # Lazy initialization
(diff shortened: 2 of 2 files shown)


docs/changes.rst
```diff
@@ -37,6 +37,9 @@ Released: not yet
   ``pywbem.configure_logging()`` function that disables logging.
   (part of issue #86)
 
+* Reduced memory consumption of CIM objects and CIM types by using slots.
```
KSchopmeyer, line 40: how about by replacing existing attributes with python slots.

KSchopmeyer:
**QUESTION:** What does this do to mocker cache compatiblity with any existing cached names? If the user already has cached mocks and upgrades to this version of pywbem, I would assume that the existing pickles are obsolete. Does this mean that they wll get an error?  Or will it simply recreate the cache????

KSchopmeyer (commented):
Made one minor comment and added a question about this change and mock caching.

docs/changes.rst
andy-maier: They are still attributes. I changed it to "... by defining their attributes to use Python slots."

andy-maier:
Your question is a very good thought. I verified it, and alas, using a mock cache that was created before this change, with this change, results in:
```
  File "c:\u\andi\projects\pywbemtools\pywbemtools\pywbemcli\_pywbemcli_operations.py", line 642, in _load_mockenv
    mockenv = pickle.load(fp)
AttributeError: 'CIMQualifier' object has no attribute '__dict__'
```

pywbem/_cim_types.py
```diff
@@ -179,28 +179,58 @@ class SlottedPickleMixin(object):
     default, without requiring that they define methods __getstate__() and
     __setstate__().
 
-    This mixin class provides these methods when running on Python 2. The
-    methods support classes with both slotted and non-slotted attributes.
+    That lead to the idea to provide these methods only when running on
```
KSchopmeyer, line 182: Love english :-).  Change lead to led.

KSchopmeyer (commented):
One small comment.

pywbem/_cim_types.py
andy-maier: DONE :-)

Issue #2511, opened 2020-10-21
Memory usage: Lazy creation of empty dicts in CIM objects


andy-maier:
The CIM object classes always allocate NocaseDict objects for their attributes, even if the attribute remains empty. For example, CIMProperty has a NocaseDict for its qualifiers.

We could create these dicts in a lazy manner, only when needed.

andy-maier:
Results of using lazy initialization (as of PR #2513):

```
tests/resourcetest/test_enuminst.py 
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=1):
Debug: cpu_times: at start: pcputimes(user=0.492316288, system=0.10439292, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.493448032, system=0.10440892, children_user=0.0, children_system=0.0)
  Elapsed time:        1.150 ms
  CPU time:            1.148 ms
  Result size:          3179 B
  Peak usage:          33401 B
.
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=10):
Debug: cpu_times: at start: pcputimes(user=0.500425824, system=0.104567728, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.505775744, system=0.104596072, children_user=0.0, children_system=0.0)
  Elapsed time:        5.400 ms
  CPU time:            5.378 ms
  Result size:         20547 B
  Peak usage:          32903 B
.
Resource consumption test: test_enuminst_mock(num_props=1, num_insts=100):
Debug: cpu_times: at start: pcputimes(user=0.530870048, system=0.104944712, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.586304448, system=0.105526544, children_user=0.0, children_system=0.0)
  Elapsed time:       56.960 ms
  CPU time:           56.016 ms
  Result size:        192596 B
  Peak usage:         190818 B
.
Resource consumption test: test_enuminst_mock(num_props=10, num_insts=1):
Debug: cpu_times: at start: pcputimes(user=0.607094336, system=0.105988456, children_user=0.0, children_system=0.0)
Debug: cpu_times: at stop: pcputimes(user=0.611855168, system=0.106005968, children_user=0.0, children_system=0.0)
  Elapsed time:        4.775 ms
  CPU time:            4.778 ms
  Result size:          8518 B
  Peak usage:          52354 B
.
[...]

PR #2513, opened 2020-10-22
Reduced memory consumption of CIM objects by using lazy initialization

andy-maier:
**Note: This PR is on top of PR #2501 - review only the delta.**

For seeing the resource savings by this change, compare the output of `make resourcetest` between this PR and PR #2501. Results have been posted to issue #2511.


KSchopmeyer (approved):
That was a good catch.   Wonder where else  lazy init might apply

Issue #2514, opened 2020-10-24
Warnings and exceptions for instance paths without keys


andy-maier:
**Update:** This issue was originally created as "Tolerate class paths in association references returned by servers" and after discussion was repurposed to add warnings and exceptions for instance paths without keys. See the discussion points, below.

The issue reported in https://github.com/pywbem/pywbemtools/issues/806 is that a WBEM server returns a reference in an association that is a class path. Here is a fragment of the response CIM-XML:
```
              <INSTANCENAME CLASSNAME="IFA_McsRedundancySetIdentity">
                <KEYBINDING NAME="SameElement">
                  <VALUE.REFERENCE>
                    <LOCALCLASSPATH>
                      <LOCALNAMESPACEPATH>
                        <NAMESPACE NAME="Intelliflash" />
                      </LOCALNAMESPACEPATH>
                      <CLASSNAME NAME="IFA_McsControllerRedundancySet" />
                    </LOCALCLASSPATH>
                  </VALUE.REFERENCE>
                </KEYBINDING>
                <KEYBINDING NAME="SystemElement">
                  <VALUE.REFERENCE>
                    <LOCALINSTANCEPATH>
                      <LOCALNAMESPACEPATH>
                        <NAMESPACE NAME="Intelliflash" />
                      </LOCALNAMESPACEPATH>
                      <INSTANCENAME CLASSNAME="IFA_Array">
                        <KEYBINDING NAME="CreationClassName">
                          <KEYVALUE VALUETYPE="string">IFA_Array</KEYVALUE>
                        </KEYBINDING>
                        <KEYBINDING NAME="Name">
                          <KEYVALUE VALUETYPE="string">172.18.76.227</KEYVALUE>
                        </KEYBINDING>
                      </INSTANCENAME>
                    </LOCALINSTANCEPATH>
                  </VALUE.REFERENCE>
                </KEYBINDING>
              </INSTANCENAME>
```

This is allowed as per DSP0201, but disallowed as per DSP0004 section 7.7.5.

Nevertheless, pywbem should tolerate this and return a CIMClassName object in such cases. All variations of class paths need to be supported as child elements of VALUE.REFERENCE: CLASSPATH, LOCALCLASSPATH, CLASSNAME

andy-maier:
While working through this issue, a number of discussion points came up:

* Standards perspective on instances without keys: The DMTF standards seem to be contradictory about this:
 
[...]

PR #2515, opened 2020-10-24
New MissingKeybindingsWarning warning for instance paths without keys

andy-maier:
See commit message.
Ready for review.

**Note: Other than discussed in the meeting, I did not raise ValueError when creating a WBEM URI, but issued the new warning as well. This was because WBEM URIs are often created to report issues with instances, for example instances without keys, and so we should be as tolerant as before.**


andy-maier:
After agreeing on the discussion points in issue #2514, this PR was repurposed, see the title change.

KSchopmeyer (approved):

PR #2516, opened 2020-10-26
Improved exception handling for received CIM-XML in tuple parser

andy-maier:
See commit message.

I don't think this PR needs to be rolled back into pywbem 1.1.

### DIFF
--- a/pywbem/_tupleparse.py
+++ b/pywbem/_tupleparse.py
@@ -150,13 +150,14 @@ class TupleParser(object):
         """
         try:
             data = u''.join(tup_tree[2])
-
         except TypeError:
-            raise CIMXMLParseError(
+            new_exc = CIMXMLParseError(
                 _format("Element {0!A} has unexpected child elements: "
                         "{1!A} (allowed is only text content)",
                         name(tup_tree), tup_tree[2]),
                 conn_id=self.conn_id)
+            new_exc.__cause__ = None
+            raise new_exc
 
         return data
 
@@ -203,8 +204,8 @@ class TupleParser(object):
             for attr in required_attrs:
                 if attr not in tt_attrs:
                     raise CIMXMLParseError(
-                        _format("Element {0!A} missing required attribute "
-                                "{1!A} (only has attributes {2!A})",
+                        _format("Element {0!A} is missing required "
+                                "attribute {1!A} (only has attributes {2!A})",
                                 name(tup_tree), attr, attrs(tup_tree).keys()),
                         conn_id=self.conn_id)
                 del tt_attrs[attr]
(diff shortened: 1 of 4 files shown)


tests/unittest/pywbem/test_indicationlistener.py
```diff
@@ -988,7 +988,7 @@ def test_WBEMListener_incorrect_headers(desc, headers, exp_status, exp_headers):
         </CIM>""",
         400,
         {
-            'CIMErrorDetails': r'Element .CIM. missing required '
+            'CIMErrorDetails': r'Element .CIM. does not have required '
```
KSchopmeyer, line 991: Actually I would prefer the original text with the word missing to "does not have". There are several cases of this

pywbem/_tupleparse.py
```diff
@@ -2015,7 +2167,7 @@ def parse_methodcall(self, tup_tree):
                                      ('LOCALCLASSPATH', 'LOCALINSTANCEPATH'))
         if not path:
             raise CIMXMLParseError(
-                _format("Element {0!A} missing a required child element "
+                _format("Element {0!A} does not have required child elements "
```
KSchopmeyer, line 2170: Actually I prefer"missing" to "does not have"

pywbem/_tupleparse.py
```diff
@@ -1987,7 +2139,7 @@ def parse_imethodcall(self, tup_tree):
 
         if not k:
             raise CIMXMLParseError(
-                _format("Element {0!A} missing child elements "
+                _format("Element {0!A} does not have any child elements "
```
KSchopmeyer, line 2142: See comment below about replaciing word "missing"

pywbem/_tupleparse.py
```diff
@@ -150,13 +150,14 @@ def pcdata(self, tup_tree):
         """
         try:
             data = u''.join(tup_tree[2])
-
         except TypeError:
-            raise CIMXMLParseError(
+            new_exc = CIMXMLParseError(
```
KSchopmeyer, line 154: This one confuses me in that I thought that __cause__ was only involves when the
raise was of the form

raise new_exc from original_exc

tests/unittest/pywbem/test_indicationlistener.py
andy-maier: Missing is an emotion. Does the element really miss its kids?

pywbem/_tupleparse.py
andy-maier: It is the portable and more recent alternative to that.

tests/unittest/pywbem/test_indicationlistener.py
andy-maier: We agreed to say "is missing"

andy-maier: DONE

pywbem/_tupleparse.py
andy-maier: DONE - changed to "is missing"

pywbem/_tupleparse.py
andy-maier: DONE - changed to "is missing required"

andy-maier:
Ready for a final review.

KSchopmeyer (approved):

PR #2532, opened 2020-11-14
Fixed DeprecationWarning issued by urllib3

andy-maier:
See commit message.
Since we are now enforcing expected warnings in tests, the unexpected DeprecationWarning causes tests to fail. This PR fixes that.


KSchopmeyer (approved):
Try o do that at runtime with a language like C++.

PR #2534, opened 2020-11-15
Migrated from Travis and Appveyor to GitHub Actions

andy-maier:
This PR adds support for using GitHub Actions to run the same tests as in Travis/Appveyor, and it removes running Travis or Appveyor. See commit messages for details. Each commit is a self-contained logical step.


KSchopmeyer (commented):
We should probably figure out what we want to do about the  security vulnerabilites statement that shows up in the code view since I would guess  that it  will show up even when merged amd that will be the top of the main page for user that view the pywbem github.

PR #2572, opened 2021-01-14
Added make perftest for indications

andy-maier:


KSchopmeyer (approved):

andy-maier:
There must be a general issue between the test client and the pywbem listener, that surfaces every now and then. It surfaces with a small likelihood in the unit tests, and surfaces with a much larger likelihood in these perf tests. This is what needs to be investigated and fixed.

andy-maier:
PR #2666 addresses the issue we previously encountered, so this PR should now also succeed.

KSchopmeyer (approved):

Issue #2256, opened 2020-06-09
MOF_Compiler and pragma namespace 


KSchopmeyer:
While MOF and the pywbem MOF compiler support the existence of a pragma namspace statement, the implement is incomplete.

1. the compile_file and compile string REQUIRE a namespace to compile mof.

2. If the namespace pragma is used, must it be predefined?  I don't think that the wbem server behavior is clearly defined in the case what the first indication of a new namespace is a SetQualifier or CreateClass.  To behonest I need to review the mocker also for its behavior.

Actually I am not really sure what the rules on use of this pragma should be.

andy-maier:
DSP0004 says: "This pragma is used to specify a Namespace path.".

First, I think we should ignore any host on the pragma namespace value. The current code assumes the pragma namespace value is directly the namespace name, which according to this definition is not necessarily the case.

Second, regarding the namespace on the pragma namespace value, we need to decide how it relates to the namespace argument of compile_file() and compile_string(), i.e. which of the two definitions has precedence. The current code uses the namespace argument of these methods as the initial target namespace, which is changed for each pragma namespace encountered, from that point on. I think that is a reasonable approach. The only thing we might want to think if is to add a flag to these methods that causes pragma namespace to be ignored, but I'm not even sure we need that.

KSchopmeyer:
There is one question here.  Must the namespace already exist in the target environment for the namespace pragma to be valid.

Andy feels it must exist.  Karl is not sure of current behavior.

What is current behavior at least in mock but I have no problem with the proposal that it already exist.

KSchopmeyer:
I agree that we do not need a flag in the compile_mof_... methods to override the use of  namespace pragma.

KSchopmeyer:
Actions: Correct the issues we agreed to above.  For the host name in the namespace pragma, If it exists, we error out. 

Fintally we agree that there are ambiguities in the compiler specification with respect to the namespace pragma and we will document how it really behaves in pywbem.  We will not make any code changes.

PR #2578, opened 2021-01-15
Fixes and tests for use of compiler namespace pragma

KSchopmeyer:
This pr fixes issue #2256 as follows:

1. Adds tests to mof compiler for no  name on the namespace pragma
just to be sure we have that covered

2. Adds tests in pywbem_mock for use of the namespace pragma.

3. Adds fix to the mof compiler for case where pybwbem_mock
compile_mof_string and compile_mof_file fail if the MOF contains a
namespace pragma that has a different namespace name than the namespace
parameter on the methods.  Note that it is just that one of the
MOFWBEMConnection lists does not get created for the namespace defined
in the pragma.

4. Extends documentation on the namespace parameter for the
compile_mof_string and compile_mof_file() methods to clarify the
behavior with the namespace pragma in the MOF.

5. Add note on limitation of name in namespace pragma to just the name and not the path. This
becomes part of the documentation on the compiler limitations.

This pr does NOT fix the issue with the pragma including host name and
there is a new issue dealing with the errors that come back if the
namespace defined in the pragma does not exist.

	addtestmof.py

### DIFF
--- a/pywbem/_mof_compiler.py
+++ b/pywbem/_mof_compiler.py
@@ -51,6 +51,11 @@ The pywbem MOF compiler will compile MOF files whose syntax complies with
    requirement is documented.
    (See `pywbem issue #1079 <https://github.com/pywbem/pywbem/issues/1079>`_).
 
+4. The namespace pragma only accepts CIM namespace name and not full
+   CIM namespace path as defined in DSP0004. If the host component is part
+   of the namespace name in a namespace pragma, the compiler generates an
+   exception.
+
 The MOF compiler API provides for invoking the MOF compiler programmatically.
 It consists of the following parts, which are described in the remaining
 sections of this chapter:
@@ -113,6 +118,13 @@ _lextab = '_moflextab'
 
 # Directory for _tabmodule and _lextab
 _tabdir = os.path.dirname(os.path.abspath(__file__))
+
+# wbem_uri Namespace path regex used to test pragma namespace
+WBEM_URI_NAMESPACEPATH_REGEXP = re.compile(
+    r'^(?:([\w\-]+):)?'  # namespace type (URI scheme)
+    r'(?://([\w.:@\[\]]*))?'  # authority (host)
+    r'(?:/|^/?)(\w+(?:/\w+)*)$',  # namespace name (leading slash optional)
+    flags=re.UNICODE)
 
 # -----------------------------------------------------------------------------
 #
@@ -599,7 +611,10 @@ def p_mp_createClass(p):
                 if p.parser.verbose:
                     p.parser.log(
                         _format("Created class {0!A}:{1!A}", ns, cc.classname))
-                p.parser.classnames[ns].append(cc.classname.lower())
+                try:
+                    p.parser.classnames[ns].append(cc.classname.lower())
+                except KeyError:
+                    p.parser.classnames[ns] = [cc.classname.lower()]
                 break
             except CIMError as ce:
                 errcode = ce.status_code
@@ -855,6 +870,7 @@ def p_compilerDirective(p):
     """compilerDirective : '#' PRAGMA pragmaName '(' pragmaParameter ')'"""
     directive = p[3].lower()
     param = p[5]
+
     if directive == 'include':
         fname = param
         if p.parser.file:
@@ -862,7 +878,20 @@ def p_compilerDirective(p):
                 fname = os.path.join(os.path.dirname(p.parser.file),
                                      fname)
         p.parser.mofcomp.compile_file(fname, p.parser.target_namespace)
+
     elif directive == 'namespace':
+        # parse the param to separate out namespace from other wbemuri pieces
+        m = WBEM_URI_NAMESPACEPATH_REGEXP.match(param)
+        ns_type = m.group(1) or None
+        host = m.group(2) or None
+        namespace = m.group(3) or None
+        if m is None or ns_type or host or namespace is None:
+            raise MOFParseError(
+                msg=_format(
+                    "Invalid pragma namespace path: {0!A}. Only namespace name "
+                    "component allowed", param),
+                parser_token=p)
+
         p.parser.target_namespace = param
         if param not in p.parser.qualcache:
             p.parser.qualcache[param] = NocaseDict()
@@ -1703,7 +1732,10 @@ def p_instanceDeclaration(p):
     try:
         cc = p.parser.handle.GetClass(cname, namespace=ns, LocalOnly=False,
                                       IncludeQualifiers=True)
-        p.parser.classnames[ns].append(cc.classname.lower())
+        try:
+            p.parser.classnames[ns].append(cc.classname.lower())
+        except KeyError:
+            p.parser.classnames[ns] = [cc.classname.lower()]
     except CIMError as ce:
         if ce.status_code == CIM_ERR_NOT_FOUND:
             if p.parser.verbose:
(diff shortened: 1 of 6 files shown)


pywbem/_mof_compiler.py
```diff
@@ -2834,7 +2842,9 @@ def compile_string(self, mof, ns, filename=None):
           ns (:term:`string`):
             The name of the CIM namespace in the associated CIM repository
             that is used for lookup of any dependent CIM elements, and that
-            is also the target of the compilation.
+            is also the target of the compilation. The compiler uses the
+            connection default_namespace if ns is None. If the MOF contains a
+            namespace pragma, that namespace will be used.
```
andy-maier, line 2847: I don't think this is correct. The namespace defined in MOF is only used from that point on in the MOF file defining it, and the previous target namespace is used upon encountering the end of that MOF file.

pywbem/_mof_compiler.py
```diff
@@ -599,7 +599,11 @@ def p_mp_createClass(p):
                 if p.parser.verbose:
                     p.parser.log(
                         _format("Created class {0!A}:{1!A}", ns, cc.classname))
-                p.parser.classnames[ns].append(cc.classname.lower())
+                try:
+                    p.parser.classnames[ns].append(cc.classname.lower())
+                except KeyError:
+                    p.parser.classnames[ns] = []
+                    p.parser.classnames[ns].append(cc.classname.lower())
```
andy-maier, line 606: How about this:
```
                try:
                    p.parser.classnames[ns].append(cc.classname.lower())
                except KeyError:
                    p.parser.classnames[ns] = [cc.classname.lower()]
```

pywbem/_mof_compiler.py
```diff
@@ -1703,7 +1707,11 @@ def p_instanceDeclaration(p):
     try:
         cc = p.parser.handle.GetClass(cname, namespace=ns, LocalOnly=False,
                                       IncludeQualifiers=True)
-        p.parser.classnames[ns].append(cc.classname.lower())
+        try:
+            p.parser.classnames[ns].append(cc.classname.lower())
+        except KeyError:
+            p.parser.classnames[ns] = []
+            p.parser.classnames[ns].append(cc.classname.lower())
```
andy-maier, line 1714: see above

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -523,7 +523,9 @@ def compile_mof_file(self, mof_file, namespace=None, search_paths=None,
             The name of the target CIM namespace in the CIM repository. This
             namespace is also used for lookup of any existing or dependent
             CIM objects. If `None`, the default namespace of the connection is
-            used.
+            used.  A namespace defined in a namespace pragma of the MOF
+            has precedence over the namespace parameter.  The namespace
+            specified in this parameter or the MOF itself must exist.
```
andy-maier, line 528: The notion of "precedence" does not sufficiently explain it. See my previous comment.

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -588,7 +590,9 @@ def compile_mof_string(self, mof_str, namespace=None, search_paths=None,
             The name of the target CIM namespace in the CIM repository. This
             namespace is also used for lookup of any existing or dependent
             CIM objects. If `None`, the default namespace of the connection is
-            used.
+            used. A namespace defined in a namespace pragma of the MOF
+            has precedence over the namespace parameter.  The namespace
+            specified in this parameter or the MOF itself must exist.
```
andy-maier, line 595: see above on "precedence".

pywbem_mock/_wbemconnection_mock.py
```diff
@@ -665,7 +669,9 @@ def compile_schema_classes(self, class_names, schema_pragma_files,
 
           namespace (:term:`string`):
             Namespace into which the classes and qualifier declarations will
-            be installed.
+            be installed. A namespace defined in a namespace pragma of the MOF
+            has precedence over this namespace parameter. The namespace specified
+            in this parameter or the MOF itself must exist.
```
andy-maier, line 674: see above on "precedence".

andy-maier (changes requested):
see comments

pywbem/_mof_compiler.py
KSchopmeyer: DONE

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Cleaned up

pywbem/_mof_compiler.py
KSchopmeyer: Cleaned up

pywbem_mock/_wbemconnection_mock.py
andy-maier: I don't see a change?

pywbem/_mof_compiler.py
```diff
@@ -2834,7 +2861,11 @@ def compile_string(self, mof, ns, filename=None):
           ns (:term:`string`):
             The name of the CIM namespace in the associated CIM repository
             that is used for lookup of any dependent CIM elements, and that
-            is also the target of the compilation.
+            is also the target of the compilation in the MOF file/string being
+            compiled. If `None`, the default namespace of the connection is
+            used. A namespace defined in a namespace pragma of the MOF
+            has precedence over the namespace parameter.  The namespace
+            specified in this parameter or the MOF itself must exist.
```
andy-maier, line 2868: he notion of "precedence" does not sufficiently explain it. See my previous comment.

andy-maier (changes requested):
See remaining comments about precedence.
Also, please rebase to resolve the conflict with changes.rst.

pywbem/_mof_compiler.py
andy-maier: It now has the same text on precedence I commented on in my other comments. Simply saying that the pragma has precedence does not sufficiently explain it, because it only precedes from the point on where it appears in the MOF, but not before. Also, we need to describe somewhere that the pragmas have a file local scope and are nested (i.e. end of file activates the namespace that was active at begin of file).

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: removed use of word precedence and change to note that the pragma becomes the namespace from the point of declaration in the compile element (string/file).  Same statement used in all of the 4 references to namespace parameter in compiling mof.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: See above

pywbem_mock/_wbemconnection_mock.py
andy-maier: It is now much better but it does not yet cover the behavior that at the end of a MOF file that had a pragma namespace, the namespace before that MOF file gets used again.

andy-maier: We looked into the code and convinced ourselves that we were wrong with the assumption the compiler would implement a nesting of namespaces along the file boundaries. Instead, the compiler only maintains one globally set target namespace that is changed when encountering a pragma namespace directive, but is not reset when encountering the end of file. Given that, the current text is correct.

andy-maier (approved):

Issue #2575, opened 2021-01-14
pywbem_mock fails with msgs about profiles if mof compiled with invalid namespace pragma


KSchopmeyer:
pywbem_mock fails with either a ModelError or CIMError naming a profile class if the MOF includes a pragma statment that refers to a namespace that does not exit and no fleshed out interop namespace exists

This is because the compiler qualifier declaration directly calls the server.create_namespace which is expecting some means of communicating with the server to create a namespace but when there is no interop namespace, or there is just an empty namespace the compiler fails with exceptions like ModelError or CIMError 

I think we need to refine this to return an exception that is more understandable, i.e. the named namespace does not exist.

KSchopmeyer:
There are two examples of this behavior in tests/unitest/ TODO.

andy-maier:
I believe this has been fixed with PR #2581 . Please verify.

PR #2581, opened 2021-01-22
 Fixed, improved, clarified support for Interop namespace in mock+WBEMServer

andy-maier:
See commit message.

I believe this also fixes Karl's issue #2575 . Please verify.

### DIFF
--- a/pywbem/_server.py
+++ b/pywbem/_server.py
@@ -382,6 +382,11 @@ class WBEMServer(object):
         update this WBEMServer object to reflect the new namespace
         there.
 
+        This method cannot create an Interop namespace because creating the
+        an Interop namespace with client operations depends on the prior
+        existence of an Interop namespace, and creating additional Interop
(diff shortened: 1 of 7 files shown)


pywbem/_server.py
```diff
@@ -382,6 +382,11 @@ def create_namespace(self, namespace):
         update this WBEMServer object to reflect the new namespace
         there.
 
+        This method cannot create an Interop namespace because creating the
+        first Interop namespace with client operations depends on the prior
```
KSchopmeyer, line 386: Should probably remove first from "first interop namespace" since there can be only one interop namespace

KSchopmeyer (commented):
One comment. Otherwise approved.

pywbem/_server.py
andy-maier: DONE

KSchopmeyer (approved):

PR #2589, opened 2021-02-07
Adds index section to generated documentation

KSchopmeyer:
This pr adds a new section to the end of the pywbem documentation, an
index.  Note that there are probably no hand created index entries but
the index gathers a number of entries from API information, etc.

This is parallel to what we did for pywbemcli and only adds a new rst
file (genindex.rst) and a reference to that file in index.rst.


KSchopmeyer:
This is a strange failue. Dying in macos and python 2.7. in the itest_indications with an issue over a bad header.  Since the only change in this pr is to docs. That means something else going on.

andy-maier (approved):

PR #2592, opened 2021-02-11
Added close() method to WBEMConnection and context manager

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -212,6 +212,12 @@ Released: not yet
 * Improvements in the log messages of the `MOFCompiler` class.
   (related to issue #2586)
 
+* Added a `close()` method to `pywbem.WBEMConnection` that closes the underlying
+  session of the 'requests' package. This avoids the ResourceWarning
+  'unclosed socket' that the 'requests' package issued so far when the Python
+  process terminates. Added the ability for `pywbem.WBEMConnection` to be used
+  as a context manager, that closes the connection at the end. (see issue #2591)
+
 **Cleanup:**
 
 * Test: Fixed all remaining ResourceWarnings during test. (issue #86)
--- a/pywbem/_cim_operations.py
+++ b/pywbem/_cim_operations.py
@@ -155,6 +155,7 @@ from ._cim_http import get_cimobject_header, wbem_request, parse_url
 from ._tupleparse import TupleParser
 from ._tupletree import xml_to_tupletree_sax
 from ._exceptions import CIMXMLParseError, XMLParseError, CIMError
+from ._exceptions import ConnectionError  # pylint: disable=redefined-builtin
 from ._statistics import Statistics
 from ._recorder import LogOperationRecorder
 from ._logging import DEFAULT_LOG_DETAIL_LEVEL, LOG_DESTINATIONS, \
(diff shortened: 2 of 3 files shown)


KSchopmeyer (commented):
Really good first step. This is much more important than just the warning.  This means that we can close sockets to servers like OpenPegasus and not depend on their timeouts to do it.  The only reason it might be difficult to kill pegasus today is that we make the connection only within the context of a command so it depends on how long each command takes.  Since the pegasus timeout of a socket is in minutes (1 or 2) could we issue enough commands to overrun the socket pool befre pegausus timed out the first one.

Next question: Does this mean we should make CIM_operations a context class????

**DISCUSSION**

pywbem/_cim_operations.py
```diff
@@ -1599,6 +1602,39 @@ def _activate_logger(cls, logger_name, simple_name, detail_level, handler,
...
+    def _verify_open(self):
+        """
+        Verify that this connection is open and raise ConnectionError otherwise.
+
+        Raises:
+          :exc:`~pywbem.ConnectionError`: Connection must be open but is closed
+        """
+        if self.session is None:
+            raise ConnectionError(
+                "Connection must be open but is actually closed",
```
KSchopmeyer, line 1614: Maybe simpler message. We do not state that connection must be open.  Is there any case where self.session is marked none by something other than our close.  If not message could be "connection closed by pywbem"

pywbem/_cim_operations.py
```diff
@@ -1599,6 +1602,39 @@ def _activate_logger(cls, logger_name, simple_name, detail_level, handler,
...
+                conn_id=self.conn_id)
+
+    def close(self):
+        """
+        Closes this connection.
+
+        *New in pywbem 1.2.*
+
+        The connection must be open. This method closes the connection and
```
KSchopmeyer, line 1623: Propose slight reordering here:   This method closes the connection in the underlying requests package causing it to close any sockets it has pooled and marks the connection closed. Any subsequent WBEM connection operation requests will generate an exception.

KSchopmeyer (commented):
I have two questions.  1 the question of context class but that was just a thought.   The second is reword of one of the comments. The third is that I think we should also test with the verify_closed somewhere like imethodcall and let it generate the exception if closed rather than let things wonder down into requests for exceptions.

KSchopmeyer:
NOTE: I left the DISCUSSION flag on but I think we have had the discussion.

pywbem/_cim_operations.py
andy-maier: We only set self.session = None in case of closing the connection.
I agree that we could simplify the message, I have changed it to "WBEMConnection is closed", to distinguish this from the general term "connection".

pywbem/_cim_operations.py
andy-maier: DONE - Changed the paragraph to:
"This method closes the session in the underlying requests package,
causing it to close any sockets it has pooled and marks the connection
closed. Any subsequent WBEM connection operation requests will generate
an exception."

andy-maier:
Added unit tests for close().
Adjusted the docstring of close() (see comment above).
Added context manager functionality.

pywbem/_cim_operations.py
```diff
@@ -848,6 +859,27 @@ def __init__(self, url, creds=None, default_namespace=None,
                 detail_levels=self._log_detail_levels)
             self.add_operation_recorder(recorder)
 
+    def __enter__(self):
```
KSchopmeyer, line 862: One more comment.  Should we mention that this class is context aware in the general documentation about the class?

Other than that question, approved

docs/changes.rst
```diff
@@ -212,6 +212,11 @@ Released: not yet
 * Improvements in the log messages of the `MOFCompiler` class.
   (related to issue #2586)
 
+* Added a `close()` method to `pywbem.WBEMConnection` that closes the underlying
+  session of the 'requests' package. This avoids the ResourceWarning
+  'unclosed socket' that the 'requests' package issued so far when the Python
+  process terminates. (see issue #2591)
+
```
KSchopmeyer, line 219: Should we also note that the class is context-aware??

KSchopmeyer (commented):
Two questions on documenting that the class is context aware.

pywbem/_cim_operations.py
andy-maier: The docstring of class WBEMConnection already does that, even with an example, on line 383 of _cim_operations.py:
```
    The WBEMConnection class can also be used as a context manager, which
    causes the connection to be closed at context manager exit:

    .. code-block:: python

        with WBEMConnection('https://myserver') as conn:
            conn.EnumerateInstances('CIM_Foo')
```

docs/changes.rst
andy-maier: Good point. Added that to the change log. DONE

andy-maier:
The comments have been resolved.

PR #2597, opened 2021-02-13
Fixes #2585 - add test to DeleteQualifier in pywbem_mock

### LINKED ISSUE #2585: Check and reject DeleteQualifier if the qualifier is used
A well behaved WBEM server should reject the DeleteQualifier operation if the qualifier is used anywhere.

The pywbem mock support currently just deletes the qualifier regardless of whether it is being used.

KSchopmeyer:

### DIFF
--- a/pywbem_mock/_mainprovider.py
+++ b/pywbem_mock/_mainprovider.py
@@ -57,7 +57,7 @@ from pywbem import CIMClass, CIMClassName, CIMInstanceName, \
     CIM_ERR_NOT_FOUND, CIM_ERR_INVALID_PARAMETER, CIM_ERR_INVALID_CLASS, \
     CIM_ERR_ALREADY_EXISTS, CIM_ERR_INVALID_ENUMERATION_CONTEXT, \
     CIM_ERR_NOT_SUPPORTED, CIM_ERR_QUERY_LANGUAGE_NOT_SUPPORTED, \
-    CIM_ERR_INVALID_QUERY
+    CIM_ERR_INVALID_QUERY, CIM_ERR_FAILED
 from pywbem._utils import _format
 
 from pywbem_mock.config import IGNORE_INSTANCE_IQ_PARAM, \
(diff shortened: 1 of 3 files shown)


pywbem_mock/_mainprovider.py
```diff
@@ -1171,8 +1171,35 @@ def DeleteQualifier(self, namespace, QualifierName):
             :exc:`~pywbem.CIMError`: (CIM_ERR_INVALID_NAMESPACE)
             :exc:`~pywbem.CIMError`: (CIM_ERR_INVALID_PARAMETER)
             :exc:`~pywbem.CIMError`: (CIM_ERR_NOT_FOUND)
-        """
-
+            :exc:`~pywbem.CIMError`: (CIM_ERR_FAILED)
+
+        """
+        def qualifier_exists(cls, qualifier_name):
```
andy-maier, line 1177: I would prefer a terminology that the qualifier is used on the class component. Existence is something I would attribute to the qualifier declaration.

pywbem_mock/_mainprovider.py
```diff
@@ -1171,8 +1171,35 @@ def DeleteQualifier(self, namespace, QualifierName):
...
+            """
+            Test for qualifier exists in any component of the class cls.
+
+            Parameters:
+              cls ()
```
andy-maier, line 1182: Add the type of the cls parameter and a colon at the end:
`cls (CIMClass):`

andy-maier (changes requested):
Minor comments

pywbem_mock/_mainprovider.py
KSchopmeyer: qualifier_exiists_in_cls. DONE

pywbem_mock/_mainprovider.py
KSchopmeyer: DONE

KSchopmeyer:
Made fixes.  Now includes test to assure that reset does not disable statistics.  Ready for review of last changes

pywbem_mock/_mainprovider.py
andy-maier: Not what I had in mind, but I see you like the "exists" terminology. Ok with me.

PR #2601, opened 2021-02-22
Fixed new pylint issues raised by pylint 2.7.0

andy-maier:
See commit message.

**DISCUSSION:**  The unpinning causes pylint 2.7.0 to be used, which increases its run time significantly (from 2 min to 22 min). See for example the test run for py39 with latest package levels. Should we exclude pylint 2.7.0?


KSchopmeyer (commented):
Agree with the issues found.  However, I believe that the execution  time is a killer and don't see how we can live with this version.  I also assume that with multiple issues filed this will be fixed fairly rapidly.  I would propose that we exclude this version

andy-maier:
I agree. The latest update pins pylint<2.7.0 and astroid<2.5.0 and adds a TODO to remove this again in the future.

PR #2608, opened 2021-03-10
Support for building cythonized distribution archive

andy-maier:
This adds support so that users can build a cythonized version of the wheel distribution archive for pywbem, by using **either one** of these commands:
`$ make buildc`
or
`$ ./setup.py bdist_wheel --cythonized`

Note that `./setup.py bdist_wheel --help` displays help text that covers the `--cythonized` option.

The cythonized version can be successfully built on macOs and Ubuntu on Python 3.5 and higher, but not on Windows.
The failing tests should only be the occasional ConnectionError failures when sending indications.

See the commit message for details.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -238,6 +238,42 @@ can be installed.
 .. _PyPI: https://pypi.python.org/pypi
 
 
+.. _`Building a compiled archive with Cython`:
+
+Building a compiled archive with Cython
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The 'Cython' package allows building a wheel distribution archive containing
+compiled code for Python packages such as pywbem. These distribution archives
+are nicknamed "cythonized" wheel distribution archives. They have the wheel
+format, but are specific to a Python version and to the target operating system.
+
+Since the code in a cythonized wheel archive is native machine code for the
+target system, this speeds up execution in compute-intensive areas of pywbem
+such as when parsing the CIM-XML responses of a WBEM server. The downside is
+that a cythonized wheel archive is significantly larger than a universal wheel
+archive, and the memory consumption is also larger.
+
+The pywbem project does not provide cythonized wheel distribution archives, but
+allows users to build them on their own. The build process for them is part of
+the regular tests to ensure they can be built.
+
+To build the cythonized wheel distribution archive of pywbem, execute:
+
+.. code-block:: bash
+
+    $ make buildc
+
+The resulting distribution archive can be installed with Pip, for example:
+
+.. code-block:: bash
+
+    $ pip install dist/pywbem-1.2.0.dev1-cp38-cp38-macosx_10_15_x86_64.whl
+
+Pip will verify that the target system and Python version matches what the wheel
+archive was built for.
+
+
 .. _`Verifying the installation`:
 
 Verifying the installation
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,6 +126,12 @@ jobs:
       matrix: ${{ fromJson(needs.set_matrix.outputs.matrix) }}
     runs-on: ${{ matrix.os }}
     steps:
+    - name: Select whether to run steps for cythonized archive
+      id: run_cythonized
+      # TODO: Remove 'andy/cythonize' branch once tests succeed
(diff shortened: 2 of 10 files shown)


rchateauneu:
Great job.

Out of curiosity, I tried on Python 2.7 and Windows 7 . Maybe this is not pywbem preferred target, indeed, but it might be of interest.

It compiles with extremely few warnings, but there was a link error. As far as I remember, I had a similar problem in the past and this problem vanished when I excluded __init__.py files from compilation. This might not be a perfect solution but at least, it works. Log file and one-line patch to setup.py are attached.

[setup.patch.txt](https://github.com/pywbem/pywbem/files/6128507/setup.patch.txt)
[pywbem_build.log](https://github.com/pywbem/pywbem/files/6128373/pywbem_build.log)

At the end, I could not find anything in pywbem\build\bdist.win-amd64 ; Maybe this is intended as this platform is not supported but it seems perfectly doable.

FYI, the setup.py options "build_ext --inplace" , sometimes implemented, allows to compile and link Python files into pyd, yielding an immediate speedup and allowing to adjust Cython directives. I use it here: https://github.com/rchateauneu/survol/blob/master/setup.py

More tests to come.

rchateauneu:
This command:
```
python3 setup.py bdist_wheel --cythonized
```
... builds successfully on:
```
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
```

... running on Windows Subsystem for Linux (WSL).

andy-maier:
@rchateauneu Remi, I have updated the PR to add documentation, and to have the cythinized version built as part of the scheduled tests and when releasing pywbem. A test or resource measurements have not been added at this point. Unless you insist on tests, I would go forward with this PR.

andy-maier:
A few more updates to the PR were needed to address basic issues with the installation of the cythonized archive.

At this point, the three environments that are capable of running the test all fail consistently with the same error that is raised in the ply module:

        if pinfo.error:
>           raise YaccError('Unable to build parser')
E           ply.yacc.YaccError: Unable to build parser

/opt/hostedtoolcache/Python/3.9.4/x64/lib/python3.9/site-packages/ply/yacc.py:3282: YaccError

andy-maier:
I found the problem: The 'module' argument of 'yacc.yacc()' was not used, and the algorithm to determine the module automatically seems to skip cythonized call levels. Fixed that in an update to the PR, and added a second commit that improves the error handling of this situation for both yacc and lex.

andy-maier:
The last changes brought it to the point where only the occasional ConnectionError when sending indications shows up.
I think we are ready for review and merge now.

docs/intro.rst
```diff
@@ -238,6 +238,42 @@ can be installed.
...
+The 'Cython' package allows building a wheel distribution archive containing
+compiled code for Python packages such as pywbem. These distribution archives
+are nicknamed "cythonized" wheel distribution archives. They have the wheel
+format, but are specific to a Python version and to the target operating system.
+
+Since the code in a cythonized wheel archive is native machine code for the
+target system, this speeds up execution in compute-intensive areas of pywbem
+such as when parsing the CIM-XML responses of a WBEM server. The downside is
+that a cythonized wheel archive is significantly larger than a universal wheel
+archive, and the memory consumption is also larger.
```
KSchopmeyer, line 255: propose change to "both the wheel archive  and memory consumption are significantly  larger that a universal wheel archive."

KSchopmeyer (commented):
One minor comment but that is a lot of code and must have been a lot of work.  I reviewed but did not test.

docs/intro.rst
andy-maier: Well, I did not measure the memory consumption, and while I think it must be larger based on the archive size, I would not necessarily call this out to be significant without having measured it. So I would like to have two separate comparison statements. NOT CHANGED.

andy-maier:
I would like to not make your proposed change. See the comment for reasons. Let me know whether you're ok with that.

docs/intro.rst
andy-maier: In the meeting today, Karl agreed to go without resolving this comment.

PR #2657, opened 2021-04-20
Added support for end2end testing using easy-server files

andy-maier:
See commit message.

Particular review points are:
* The data structure in the es_server.yml and es_vault.yml files (see comments in those files).
* The concept of WBEM server functions in the es_server.yml files and how it is used in test functions to skip the tests.
* The specific WBEM server functions that are currently defined (see es_server.yml file).

Note: The validation of the user-defined sections of the es_server.yml and es_vault.yml files with a JSON schema needs to be in a future separate issue, because that requires an extension in the underlying pytest-easy-server package.


tests/end2endtest/es_server.yml
```diff
@@ -0,0 +1,87 @@
...
+#       url: URL                   # URL of the WBEM server; for details see url
+#                                  # parameter of pywbem.WBEMConnection().
+#                                  # Mandatory.
+#
+#       default_namespace: NAMESPACE
+#                                  # Default CIM namespace to use in the
+#                                  # created pywbem.WBEMConnection object.
+#                                  # Optional, default: null.
+#
+#       functions:                 # List of functions implemented by the
```
KSchopmeyer, line 58: The word function confused me in particular because it conflicts with programming use of the word function.  I suggest either functionality or capability or even service.  Even "server process" might be logical.

tests/end2endtest/es_server.yml
```diff
@@ -0,0 +1,87 @@
...
+# vault_file: VAULT_FILE           # Path name of easy-server vault file.
+#                                  # Mandatory.
+#
+# default: NICKNAME                # Nickname of default group or server.
+#                                  # Mandatory.
+#
+# server_groups:                   # WBEM server groups.
+#                                  # Optional, default: no groups defined.
+#
+#   NICKNAME42:                    # Nickname of WBEM server group.
```
KSchopmeyer, line 14: Not really a big issue but why do we call it nickname rather than name???

tests/end2endtest/es_server.yml
```diff
@@ -0,0 +1,87 @@
...
+#
+#     user_defined:
+#
+#       docker_image: ORG/IMAGE:M.N.U
+#                                  # Docker image to pull for setting up the
+#                                  # WBEM server, or null to not pull any
+#                                  # Docker image.
+#                                  # Optional, default: null.
+#
+#       docker_port_mapping:       # Port mapping for Docker image.
```
KSchopmeyer, line 40: The port mapping confuses me:

1. What is straight mapping? I assume it is same port on both host and docker.
2. How do you map multiple ports (ex 5988 and 5989)

tests/end2endtest/es_server.yml
```diff
@@ -0,0 +1,87 @@
...
+#                                  # server. The function names are specific
+#                                  # to pywbem end2end testing and are checked
+#                                  # by the end2end testcases. If a testcase
+#                                  # requires a particular function and the
+#                                  # server does not implement it, the testcase
+#                                  # is skipped.
+#         - interop                # Has an Interop namespace
+#         - namespaces             # Supports management of namespaces
+#         - profiles               # General support for mgmt profiles
+#         - smis                   # Support for SMI-S profiles
```
KSchopmeyer, line 68: Not sure if smis is good example.  I could see specific smis profiles however. Some other possibilities include slp,  ssl (i.e. for testing ssl capabilities), subscriptions/indications, pull operations,

KSchopmeyer (commented):
I made several comments

tests/end2endtest/es_server.yml
andy-maier: 1. correct.
2. you cannot. Each server has a single URL, so you need to decide which protocol and port you use. If you want to test both protocols and ports, define two servers.

tests/end2endtest/es_server.yml
andy-maier: These functions are not examples, but actually existing end2end tests.

For SMI-S specific tests, we have two test modules test_profile_snia_smis.py that depends on the SMI-.S way of advertising profiles and test_profile_snia_server.py that depends on the SNIA Server profile. So we could split "smis" into two functions, but I think that every SMI-S implementation will support the SNIA Server profile, so you would never find one without the other. Anyway, we could still split "smis" into something like "snia-smis" and "snia-server".

tests/end2endtest/es_server.yml
andy-maier: That would be a change in easy-server, since it is part of the stuff inherited from there. Also, I wanted to have a term that is sufficiently far away from the "hostname" of the server and makes it clear that it only exists in the realm of easy-server.

tests/end2endtest/es_server.yml
andy-maier: DONE - changed to "capabilities".

tests/end2endtest/es_server.yml
andy-maier: DONE - I have split it into "snia-smis" and "snia-server".

andy-maier:
Updated the PR to address your comments. Please review again.

docs/changes.rst
```diff
@@ -265,6 +265,14 @@ Released: not yet
   the underlying connection. Used that function in the 'mof_compiler' script
   to remove a ResourceWarning about unclosed sockets. (issue #2610)
 
+* Test: Added support for end2end testing of WBEM servers based on
+  easy-server server and vault files. The server files can specify WBEM servers
```
KSchopmeyer, line 269: suggest change easy-server to easy-server python package

KSchopmeyer (commented):
Made a couple of minor comments and a note that apparently docker allows mapping not only single ports but ranges of ports (something like 5988-5989:9988-9989 although I have not used that capability.

tests/end2endtest/es_server.yml
```diff
@@ -0,0 +1,88 @@
...
+#                                  # Optional, default: null.
+#
+#       docker_port_mapping:       # Port mapping for Docker image.
+#                                  # Optional, default: straight mapping (i.e.
+#                                  # same port on host and in image).
+#
+#         image: PORT              # Port number provided by Docker image.
+#                                  # Mandatory.
+#
+#         local: PORT              # Port number used on local system.
```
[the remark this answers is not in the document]
andy-maier: Yes, it does. Changed to "host". DONE

docs/changes.rst
andy-maier: DONE - changed to "server and vault files of the 'easy-server' Python package".

andy-maier:
Added a note to the description of the port mapping to address the comment about mapping multiple ports:
```
#                                  # Note that each server defines the single
#                                  # port that is to be used for testing, so if
#                                  # you want to use both the HTTP and HTTPS
#                                  # ports, define a server for each.
```

andy-maier:
I added a second pegasus server for the second HTTP ports. That drove some further changes:
* Fixed the incorrect description of server groups in the es_server.yml file
* Made the docker container name unique by prepending the server nickname

Also, addressed the DeprecationWarning about using yaml.load().

PR #2666, opened 2021-04-23
Workaround for BadStatusLine issue

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -144,6 +144,10 @@ Released: not yet
 * Fix pywem_mock issue with Delete class not calling providers to handle
   the DeleteInstance (see issue #2643)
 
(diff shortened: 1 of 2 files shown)


docs/changes.rst
```diff
@@ -144,6 +144,9 @@ Released: not yet
 * Fix pywem_mock issue with Delete class not calling providers to handle
   the DeleteInstance (see issue #2643)
 
+* Test: Workaround for BadStatusLine issue in test_WBEMListener_send_indications
+  test function. (issue #2659)
```
KSchopmeyer, line 148: I would clarify by stating that this is NOT a fix for the issue itself, This is just to be sure users understand.

KSchopmeyer (commented):
One minor comment

docs/changes.rst

PR #2678, opened 2021-04-26
Release 1.2.0

andy-maier:
Please approve the release of pywbem 1.2.0.

The one extra commit solves an issue that only shows up in the full set of tests, but nt in the normal set of tests. That's why it is part of the release PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,9 +9,9 @@ on:
     - # cron (in UTC): minute hour day_of_month month day_of_week
       cron: '00 03 * * SUN'
   push:
-    branches: [ master, stable_1.1 ]
+    branches: [ master, stable_1.2 ]
   pull_request:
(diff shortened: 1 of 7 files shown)


KSchopmeyer:
**DISCUSSION**: I think we should release the changes as separate from the actual release so that the release PR only contains the changes necessary for the release itself.  

Since there are failures in builddoc, in more than one test, there are other issues.  

However, I think we should release at the current level of code with the changes in now.

andy-maier:
The problem is that the errors only show up in test envs that are not part of the normal set of test envs. The release PR however is tested with the full set of test envs, just like the weekly scheduled run.
I have kept the logically separate fixes in separate commits, so that the commit history will be clean. Just the relation to PRs will not be 1:1.

andy-maier:
Ready for review again.

KSchopmeyer (approved):

PR #2681, opened 2021-04-27
Fixed installation of pywinpty

andy-maier:
See commit message.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -187,3 +187,8 @@ pyzmq>=20.0.0; python_version >= '3.9'
 bleach>=3.3.0; python_version == '2.7'
 bleach>=3.1.2,<3.1.3; python_version == '3.4'
 bleach>=3.3.0; python_version >= '3.5'
+
+# pywinpty is used by terminado <- notebook <- jupyter
+# pywinpty does not declare supported Pyrthon versions and 1.0 has removed support for py27.
+pywinpty>=0.5,<1.0; os_name == "nt" and python_version == '2.7'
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -187,3 +187,8 @@ pyzmq>=20.0.0; python_version >= '3.9'
 bleach>=3.3.0; python_version == '2.7'
 bleach>=3.1.2,<3.1.3; python_version == '3.4'
 bleach>=3.3.0; python_version >= '3.5'
+
+# pywinpty is used by ???
```
KSchopmeyer, line 191: Not sure of meaning of this commentt????  :-)

KSchopmeyer (commented):
One minor comment and approved.

dev-requirements.txt
andy-maier: It means to say who pulls in that package, but since we are not yet installing and running pipdeptree, and since I did not try this on my Windows, I don't have the answer yet.

andy-maier: Karl provided which packages use pywinpty on native Windows by using pipdeptree:
```
pywinpty==0.5.5
  - terminado==0.8.1 [requires: pywinpty>=0.5]
    - notebook==5.7.4 [requires: terminado>=0.8.1]
      - jupyter==1.0.0 [requires: notebook]
      - widgetsnbextension==3.4.2 [requires: notebook>=4.4.1]
        - ipywidgets==7.4.2 [requires: widgetsnbextension~=3.4.0]
          - jupyter==1.0.0 [requires: ipywidgets]
```
Updated the requirements file with that.

PR #2689, opened 2021-05-01
Test: Added support for validating user-defined properties in easy-server files

andy-maier:
See commit message

### DIFF
--- a/tests/end2endtest/es_schema.yml
+++ b/tests/end2endtest/es_schema.yml
@@ -1,0 +1,109 @@
+# Easy-server schema file for pywbem end2end tests
+#
+# This file is used to validate the user-defined sections in the easy-server
+# server file and vault file.
+#
+# Keep this JSON schema in sync with the code in the wbem_connection() fixture
+# that uses the items from these files.
+
+user_defined_schema:
+  $schema: http://json-schema.org/draft-07/schema#
+  type: object
+  additionalProperties: false
+  required: [url]
+  properties:
+    docker_image:
+      type: string
+      description: |
+        Docker image to pull for setting up the WBEM server, or null to not
+        pull any Docker image.
+        Format: ORG/IMAGE:M.N.U
+        Optional, default: null.
+    docker_port_mapping:
+      type: object
+      description: |
+        Port mapping for Docker image.
+        Optional, default: straight mapping (i.e. same port on host and in image).
+        Note that each server defines the single port that is to be used for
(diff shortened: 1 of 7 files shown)


tests/end2endtest/es_schema.yml
```diff
@@ -0,0 +1,109 @@
+# Easy-server schema file for pywbem end2end tests
+#
+# This file is used to validate the user-defined sections in the easy-server
+# server file and vault file.
+#
+# Keep this schema in sync with the code in the wbem_connection() fixture that
```
KSchopmeyer, line 6: Change to json-schema. The word schema is so general that it confuses the  reader.

tests/end2endtest/es_schema.yml
```diff
@@ -0,0 +1,109 @@
...
+      type: [string, "null"]
+      description: |
+        Path name of file containing X.509 client certificate to be presented
+        to server, or null for not presenting a client certificate to the server.
+        Optional, default: null.
+    key_file:
+      type: [string, "null"]
+      description: |
+        Path name of file containing X.509 private key of the client certificate,
+        or null for not presenting a client certificate to the server.
```
KSchopmeyer, line 108: change "for not presenting" to "if not presenting"

tests/end2endtest/es_schema.yml
```diff
@@ -0,0 +1,109 @@
...
+      type: boolean
+      description: |
+        Boolean controlling whether the returned server certificate is to be verified.
+        Optional, default: true.
+    ca_certs:
+      type: [string, "null"]
+      description: |
+        Path name of CA certificate file or path name of directory containing CA
+        certificate files to be used for verifying the returned server certificate,
+        or null for using the pywbem default locations.
```
KSchopmeyer, line 96: change for to if

KSchopmeyer (commented):
Minor comments and approve

tests/end2endtest/es_schema.yml
andy-maier: DONE

tests/end2endtest/es_schema.yml
andy-maier: This is the input for something to happen, that's why I used "for". Otherwise it would need to say something like "if the pywbem default locations are to be used", which is probably more ugly.
Not changed, for now.

tests/end2endtest/es_schema.yml
andy-maier: same reason as previous comment, not changed for now.

PR #2695, opened 2021-05-08
Improved the running of indication listeners when using HTTPS

andy-maier:
See commit message.
This came up when implementing listener support in pywbemtools. It started out as only an improvement with no required functionality for the listener support in pywbemtools, but it ended up with new exception classes pywbemtools now has a dependency on.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -23,6 +23,12 @@ Released: not yet
 
 **Incompatible changes:**
 
+* The `WBEMListener.start()` method may raise new exceptions
+  `pywbem.ListenerPortError`, `pywbem.ListenerPromptError` and
+  `pywbem.ListenerCertificateError`. The `OSError` and `IOError` exceptions
+  raised in earlier versions may still be raised for other, less common cases.
+  For details, see the corresponding item in the Enhancements section, below.
+
 **Deprecations:**
 
 **Bug fixes:**
@@ -34,6 +40,26 @@ Released: not yet
 * Docs: Fixed an error with the autodocsumm and Sphinx 4.0.0. (issue #2697)
 
 **Enhancements:**
+
+* Improved the running of indication listeners via `WBEMListener.start()`:
+
+  - The method will now raise a new exception `pywbem.ListenerPortError` when
+    the port is in use, instead of the previous `socket.error` on Python 2 and
+    `OSError` on Python 3 that had confusing or unspecific error messages.
+
+  - The method will now raise a new exception `pywbem.ListenerCertificateError`
+    when using HTTPS and there is an issue with the server certificate file,
+    private key file, or invalid password for the private key file, instead of
+    the previous `ssl.SSLError` or `OSError` that had confusing or unspecific
+    error messages.
+
+  - The method will now raise a new exception `pywbem.ListenerPromptError`
+    when using HTTPS and the prompt for the password of the private key file
+    was interrupted or ended, instead of the previous `IOError` or `OSError`
+    that had unspecific error messages.
+
+  - If the private key file is protected with a password, the password prompt
+    now states the path name of the private key file in the prompt message.
 
 **Cleanup:**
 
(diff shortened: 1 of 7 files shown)


docs/changes.rst
```diff
@@ -23,6 +23,11 @@ Released: not yet
 
 **Incompatible changes:**
 
+* The `WBEMListener.start()` method may raise new exceptions
+  `pywbem.ListenerPortError`, `pywbem.ListenerPromptError` and
+  `pywbem.ListenerCertificateError`. For details, see the corresponding item
```
KSchopmeyer, line 28: we might comment that the old version raised IOError and OSError to cement the idea that this is an incompatibility.

KSchopmeyer:
I ran a test with the this fix and the listen.py and send_indications.py in pywbem/examples.  When invalid password entered.  The following occurred:    (you just have to love those original SSL error exceptions :-) )

```

istener)$ ./listen.py localhost 50000 50001 cert.pem key.pem
2021-05-09 10:18:22,445 INFO: Starting threaded HTTP server on port 50000
2021-05-09 10:18:22,446 INFO: Started threaded HTTP server on port 50000
2021-05-09 10:18:22,446 INFO: Starting threaded HTTPS server on port 50001
Enter password for key file key.pem: 
  File "/home/kschopmeyer/pywbem/pywbem39/pywbem/examples/./listen.py", line 148, in <module>
    _sys.exit(_main())
  File "/home/kschopmeyer/pywbem/pywbem39/pywbem/examples/./listen.py", line 99, in _main
    listener.start()
  File "/home/kschopmeyer/pywbem/pywbem39/pywbem/pywbem/_listener.py", line 1084, in start
    raise new_exc  # ListenerCertificateError
pywbem._exceptions.ListenerCertificateError: Invalid password for key file, bad key file, or bad certificate file. Original error: [SSL] PEM lib (_ssl.c:4048)
Exception ignored in: <socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 50001)>
ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 50001)>

File: pywbem/_exceptions.py [hunk 2]
```diff
@@ -699,3 +701,50 @@ def __init__(self, message, conn_id=None):
...
+    class :class:`~pywbem.WBEMListener).
+    """
+
+    def __init__(self, message):
+        """
+        Parameters:
+
+          message (:term:`string`): Error message (will be put into `args[0]`).
+
+        :ivar args: A tuple (message, ) set from the corresponding init
```
KSchopmeyer, line 718: This definition is new to me.

KSchopmeyer:
NOTE: We should probably also update the listener code in examples with this change.

pywbem/_exceptions.py
andy-maier: It causes a section "Variables" to be created with the specified attributes. We use this already in a number of places in order to document attributes that we inherit from base classes but that are not documented there. For example, the "args" attribute of "Exception".
Look at https://pywbem.readthedocs.io/en/latest/client/exceptions.html#pywbem.ConnectionError for an example.

docs/changes.rst
andy-maier: Well, they do replace the use of the earlier exceptions for the cases they are used for, but there is probably a chance that IOError on py27 and OSError still occur for some other cases that have not been replaced.
DONE - Added text: "The `OSError` and `IOError` exceptions raised in earlier versions may still be raised for other, less common cases."

andy-maier: One more: I found that the new exceptions were not included in the docs yet. I added them to the indication llistener chapter, and improved their description to state when they were added and what the base class is, for consistency with our other exception classes.

andy-maier:
Karl, I addressed all your comments.

KSchopmeyer (approved):
Made some comments.

Issue #2701, opened 2021-05-26
Subscription manager, issues


KSchopmeyer:
1.  Issue with definition of Name property.

When the SubscriptionManager Name property creates a new destination it defines the name property as:

```

SubscriptionManager, Line 1179
        dest_inst['Name'] = _format(
            'pywbemdestination:{0}:{1}:{2}',
            ownership, self._subscription_manager_id, uuid.uuid4())

```

However, when the add_server looks for owned indications: ist uses syntax for the Name property defined by the regex

```

        this_host = getfqdn()
        dest_name_pattern = re.compile(
            _format(r'^pywbemdestination:owned:{0}:{1}:[^:]*$',
                    this_host, self._subscription_manager_id))

```

Which contains the "this_host" value.   Thus, it never finds the owned destinations.

The same issue applies to the add_filter()

We will fix this to include the fqdn as the value of this component of each Name property and add code to includ it in the pattern.

```

2. Issue with comparison of paths of new subscription instance and instance from owned list.  In this case, the instance created for the new subscription instance to be created does not include the keybindings even though they are in the instance.  See the code below from SubscriptionManager, line 1336:

```
        sub_path = CIMInstanceName(SUBSCRIPTION_CLASSNAME,
                                   namespace=server.interop_ns)

        sub_inst = CIMInstance(SUBSCRIPTION_CLASSNAME)
        sub_inst.path = sub_path
        sub_inst['Filter'] = filter_path
        sub_inst['Handler'] = dest_path

        if owned:
            for inst in self._owned_subscriptions[server_id]:
                if inst.path == sub_path:
                    # It does not have any properties besides its keys,
                    # so checking the path is sufficient.
                    return sub_inst

```

The sub_inst.path is not completed with the keybindings

```
The _createfilter method  does not create a complete path for the new filter because it does not add the keybindings to the path but only sets the classname and namespace.  
3. Issue with same test in create_filter.  Created path is incomplete with no keybindings. While this is correct for the CreateInstance that follows, the test of inst.path == ..._path fails because the keybindings are not defined.
4. Same issue with dest_path
[...]

PR #2702, opened 2021-05-27
Fixes issue #2701, SubscriptionManager code issues 

KSchopmeyer:
1. Fixes an issue where the build of owned CIM Listener destination and CIM
indication filter objects and the regex to find the objects when
restoring these objects from a server did not have a common
understanding of the exact format of the Name property. The host_name
element was part of the regex but not of the string in the Name
property. This pr simply adds the host_name to the Name property.

Resolved: We send them

2. Fixes an issue where the add_subscription() comparison of the new instance path with paths from the list of owned instances is working with an incomplete path on the new instance (the keybindings have not been added).

Resolved: We don't do the modify so that code removed.

It also adds tests to confirm that when a SubscriptionManager is created, the owned subscriptions, filters, and destinations are
recovered from the server correctly and the subscription if added twice does not actually add the second time

**DISCUSSION:** 
There are a couple of issues for discussion:

1. Should we be sending the properties for SystemCreationClassName and SystemName properties to the server at all when we create filters and destinations.  OpenPegasus allows them but tests the values against what it expects and changes the values if necessary.  NOTE: This question covers a lot more than just these classes for the client since that pattern of properties in paths is common.

   Update by Andy: We decided to always send the key properties when creating instances in order to support older implementations.

2. Should we be caching the SystemCreationClassName as we are now with the ServerName to be used to build local paths to help determine if filters and destinations already exist in the server.

   Update by Andy: Don't think so, if we change the comparison as I suggested in the comments I don't think that is necessary.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -37,6 +37,21 @@ Released: not yet
   surrounded by verbose test (if p.parse.verbose:). See pywbemcli issue
   #395,
 
+* Fixes several issues in WBEMSubscriptionManager:
+  - add_filter() and add_destinations() methods  can no longer modify
+    existing instances on the WBEM server. They can only create new instances.
+  - Modified the algorithm to determine owned filters and
+    instances so they are are correctly recovered from the WBEM server when the
+    WBEMSubscriptionManager is restarted (before this they could be returned
+    as not-owned object).
+  - Change to use WBEM server systemname as the value of the SystemName
+    property.
+  - Removed code that built instance path for new filter and destination
+    instances since that was used only to try to determine if instance existed
+    to make the create/modify decision.
+  - Added the client host as a component of the Name property for owned
+    filters and destinations. (issue #2701).
+
 * Docs: Fixed an error with the autodocsumm and Sphinx 4.0.0. (issue #2697)
 
 * Jupyter Notebook: Ignored safety issues 40380..40386 in order to continue
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -74,18 +74,25 @@ with one exception:
 The :class:`~pywbem.WBEMSubscriptionManager` object remembers owned
 subscriptions, filters, and listener destinations. If for some reason that
 object gets deleted before all servers could be removed (e.g. because the
-Python program aborts), the corresponding CIM instances in the WBEM server
-still exist, but the knowledge is lost that these instances were owned by that
-subscription manager. Therefore, the subscription manager discovers owned
-subscriptions, filters, and listener destinations when a server is added.
-For filters, this discovery is based upon the Name property. Therefore, if
-the Name property is set by the user (e.g. because a management profile
-requires a particular name), the filter must be permanent and cannot be owned.
+Python program aborts or the client is terminated), the corresponding CIM
+instances in the WBEM server still exist, but the knowledge is lost that these
+instances were owned by that subscription manager. Therefore, the subscription
+manager discovers owned subscriptions, filters, and listener destinations when
+a server is added. For this discovery, is based upon the Name property.
+Therefore, if the Name property is set by the user (e.g. because a management
+profile requires a particular name), the filter must be permanent and cannot be
+owned.
+
+Since :class:`~pywbem.WBEMSubscriptionManager` does not directly modify
+existing instances of filter or destinations or subscriptions, the user must do
+this directly through the `ModifyInstance` WBEM request method and then update
+the local owned instances list by executing get_all_filters(),
+get_all_destinations(), or get_all_subscriptions().
 
 Examples
 --------
 
-The following example code demonstrates the use of a subscription manager
+The following example code demonstrates the use of a subscri tion manager
 to subscribe for a CIM alert indication on a WBEM server. The WBEM listener
 is assumed to exist somewhere and is identified by its URL::
 
@@ -187,7 +194,8 @@ class WBEMSubscriptionManager(object):
             contain the character ':' because that is the separator between
             components within the value of the `Name` property.
 
-            The subscription manager ID must be unique on the current host.
+            The subscription manager ID must be unique on the current client
+            host.
 
             For example, the form of the `Name` property of a filter instance
             is (for details, see
@@ -207,6 +215,7 @@ class WBEMSubscriptionManager(object):
         self._owned_subscriptions = {}  # CIMInstance of owned subscriptions
         self._owned_filters = {}  # CIMInstance of owned filters
         self._owned_destinations = {}  # CIMInstance owned destinations
+        self._systemnames = {}  # Dict that will contain SystemNames
 
         if subscription_manager_id is None:
             raise ValueError("Subscription manager ID must not be None")
@@ -228,7 +237,7 @@ class WBEMSubscriptionManager(object):
         return _format(
             "WBEMSubscriptionManager("
             "_subscription_manager_id={s._subscription_manager_id!A}, "
-            "_servers={s._servers}, "
+            "_servers={s._servers}, _systemnames={s._systemnames!A}"
             "...)",
             s=self)
 
@@ -244,6 +253,7 @@ class WBEMSubscriptionManager(object):
             "_owned_subscriptions={s._owned_subscriptions!A}, "
             "_owned_filters={s._owned_filters!A}, "
             "_owned_destinations={s._owned_destinations!A})",
+            "_systemnames={s._systemnames!A}",
             s=self)
 
     def __enter__(self):
@@ -344,29 +354,35 @@ class WBEMSubscriptionManager(object):
         self._owned_filters[server_id] = []
         self._owned_destinations[server_id] = []
 
-        # Recover any owned destination, filter, and subscription instances
-        # that exist on this server
+        # Get the SystemName from the WBEMServer CIM_ObjectManager for this
+        # server.
+        self._systemnames[server_id] = server.cimom_inst['SystemName']
 
-        this_host = getfqdn()
+        # Issue #2709 Expand to account for WBEMServer profile
 
+        # Get the hostname of the client system to be part of destination Name
+        this_client = getfqdn()
+
+        # Recover owned destination, filter, and subscription instances
+        # that exist on the WBEMServer
         dest_name_pattern = re.compile(
             _format(r'^pywbemdestination:owned:{0}:{1}:[^:]*$',
-                    this_host, self._subscription_manager_id))
+                    this_client, self._subscription_manager_id))
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
+
         for inst in dest_insts:
-            if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+            if re.match(dest_name_pattern, inst.path.keybindings['Name']):
                 self._owned_destinations[server_id].append(inst)
 
         filter_name_pattern = re.compile(
             _format(r'^pywbemfilter:owned:{0}:{1}:[^:]*:[^:]*$',
-                    this_host, self._subscription_manager_id))
+                    this_client, self._subscription_manager_id))
         filter_insts = server.conn.EnumerateInstances(
             FILTER_CLASSNAME, namespace=server.interop_ns)
+
         for inst in filter_insts:
-            if re.match(filter_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+            if re.match(filter_name_pattern, inst.path.keybindings['Name']):
                 self._owned_filters[server_id].append(inst)
 
         sub_insts = server.conn.EnumerateInstances(
@@ -375,6 +391,8 @@ class WBEMSubscriptionManager(object):
                               self._owned_filters[server_id]]
         owned_destination_paths = [inst.path for inst in
                                    self._owned_destinations[server_id]]
+
+        # Subscription is owned if either filter or destination is owned.
         for inst in sub_insts:
             if inst.path.keybindings['Filter'] in \
                     owned_filter_paths \
@@ -401,7 +419,6 @@ class WBEMSubscriptionManager(object):
 
             Exceptions raised by :class:`~pywbem.WBEMConnection`.
         """
-
         # Validate server_id
         server = self._get_server(server_id)
 
@@ -463,21 +480,22 @@ class WBEMSubscriptionManager(object):
 
         The form of the `Name` property of the created destination instance is:
 
-          ``"pywbemdestination:" {ownership} ":" {subscription_manager_id} ":"
-          {guid}``
+          ``"pywbemdestination:" {ownership} ":" {client_host_name} ":"
+          {subscription_manager_id} ":" {guid}``
 
         where ``{ownership}`` is ``"owned"`` or ``"permanent"`` dependent on
-        the `owned` argument; ``{subscription_manager_id}`` is the
-        subscription manager ID; and ``{guid}`` is a globally unique
-        identifier.
+        the `owned` argument; ``{client_host_name}`` is the fqdn of the client
+        host; ``{subscription_manager_id}`` is the subscription manager ID;
+        and ``{guid}`` is a globally unique identifier.
 
         Owned listener destinations are added or updated conditionally: If the
         listener destination instance to be added is already registered with
-        this subscription manager and has the same property values, it is not
-        created or modified. If it has the same path but different property
-        values, it is modified to get the desired property values. If an
-        instance with this path does not exist yet (the normal case), it is
-        created.
+        this subscription manager, is an owned destination and has the same
+        `Destination property`, a new listener_destination instance is not
+        created. If an instance with this path does not exist yet (the normal
+        case), it is created.
+
+        This method does not modify an existing destination instance.
 
         Permanent listener destinations are created unconditionally, and it is
         up to the user to ensure that such an instance does not exist yet.
@@ -698,25 +716,26 @@ class WBEMSubscriptionManager(object):
 
           In this case, the value of the `Name` property will be:
 
-          ``"pywbemfilter:" {ownership} ":" {subscription_manager_id} ":"
-          {filter_id} ":" {guid}``
+          ``"pywbemfilter:" {ownership} ":" {client_host_name} ":"
+          {subscription_manager_id} ":" {filter_id} ":" {guid}``
 
           where
           ``{ownership}`` is ``"owned"`` or ``"permanent"`` dependent on
           whether the filter is owned or permmanent;
+          ``{client_host_name}`` is the fqdn of the client hos;
           ``{subscription_manager_id}`` is the subscription manager ID;
           ``{filter_id}`` is the filter ID; and
           ``{guid}`` is a globally unique identifier.
 
-          This can be used for both owned and permanent filters.
+        Only owned  filters can use the filter_id to create the
+        `Name` property.
 
         Owned indication filters are added or updated conditionally: If the
         indication filter instance to be added is already registered with
-        this subscription manager and has the same property values, it is not
-        created or modified. If it has the same path but different property
-        values, it is modified to get the desired property values. If an
-        instance with this path does not exist yet (the normal case), it is
+        this subscription manager , it is not
         created.
+
+        This method does not modify existing CIM_IndicationFilter instances.
 
         Permanent indication filters are created unconditionally, and it is
         up to the user to ensure that such an instance does not exist yet.
@@ -793,6 +812,10 @@ class WBEMSubscriptionManager(object):
             if ':' in filter_id:
                 raise ValueError(
                     _format("Filter ID contains ':': {0!A}", filter_id))
+
+        if name is not None and owned:
+            raise ValueError("The name parameter cannot be used to add "
+                             "owned filters.")
 
         filter_inst = self._create_filter(server_id, source_namespace, query,
                                           query_language, owned, filter_id,
@@ -1131,6 +1154,10 @@ class WBEMSubscriptionManager(object):
         In order to catch any changes the server applies, the instance is
         retrieved again using the instance path returned by instance creation.
 
+        If an instance is found in the server with the same value of the
+        `Destination` property, no new instance is created and the existing
+        is returned.
+
         Parameters:
 
           server_id (:term:`string`):
@@ -1162,41 +1189,34 @@ class WBEMSubscriptionManager(object):
         # Validate server_id
         server = self._get_server(server_id)
 
+        this_client = getfqdn()
+
         # Validate the URL by reconstructing it. Do not allow defaults
         _, _, listener_url = parse_url(dest_url, allow_defaults=False)
 
-        this_host = getfqdn()
         ownership = "owned" if owned else "permanent"
 
-        dest_path = CIMInstanceName(DESTINATION_CLASSNAME,
-                                    namespace=server.interop_ns)
-
         dest_inst = CIMInstance(DESTINATION_CLASSNAME)
-        dest_inst.path = dest_path
         dest_inst['CreationClassName'] = DESTINATION_CLASSNAME
         dest_inst['SystemCreationClassName'] = SYSTEM_CREATION_CLASSNAME
-        dest_inst['SystemName'] = this_host
+        dest_inst['SystemName'] = self._systemnames[server_id]
+
         dest_inst['Name'] = _format(
-            'pywbemdestination:{0}:{1}:{2}',
-            ownership, self._subscription_manager_id, uuid.uuid4())
+            'pywbemdestination:{0}:{1}:{2}:{3}',
+            ownership, this_client, self._subscription_manager_id, uuid.uuid4())
         dest_inst['Destination'] = listener_url
 
         if owned:
-            for i, inst in enumerate(self._owned_destinations[server_id]):
-                if inst.path == dest_path:
-                    # It already exists, now check its properties
-                    if inst != dest_inst:
-                        server.conn.ModifyInstance(dest_inst)
-                        dest_inst = server.conn.GetInstance(dest_path)
-                        self._owned_destinations[server_id][i] = dest_inst
-                    return dest_inst
-            dest_path = server.conn.CreateInstance(dest_inst)
-            dest_inst = server.conn.GetInstance(dest_path)
+            # If an instance with the same destination property exists,
+            # do not create a new instance.
+            for inst in self._owned_destinations[server_id]:
+                if inst['Destination'] == dest_inst['Destination']:
+                    return inst  # has path set
+        dest_path = server.conn.CreateInstance(dest_inst,
+                                               namespace=server.interop_ns)
+        dest_inst = server.conn.GetInstance(dest_path)
+        if owned:
             self._owned_destinations[server_id].append(dest_inst)
-        else:
-            # Responsibility to ensure it does not exist yet is with the user
-            dest_path = server.conn.CreateInstance(dest_inst)
-            dest_inst = server.conn.GetInstance(dest_path)
 
         return dest_inst
 
@@ -1253,44 +1273,32 @@ class WBEMSubscriptionManager(object):
         # Validate server_id
         server = self._get_server(server_id)
 
-        this_host = getfqdn()
         ownership = "owned" if owned else "permanent"
 
-        filter_path = CIMInstanceName(FILTER_CLASSNAME,
-                                      namespace=server.interop_ns)
+        this_client = getfqdn()
 
         filter_inst = CIMInstance(FILTER_CLASSNAME)
-        filter_inst.path = filter_path
         filter_inst['CreationClassName'] = FILTER_CLASSNAME
         filter_inst['SystemCreationClassName'] = SYSTEM_CREATION_CLASSNAME
-        filter_inst['SystemName'] = this_host
+        filter_inst['SystemName'] = self._systemnames[server_id]
         if filter_id:
             filter_inst['Name'] = _format(
-                'pywbemfilter:{0}:{1}:{2}:{3}',
-                ownership, self._subscription_manager_id, filter_id,
-                uuid.uuid4())
+                'pywbemfilter:{0}:{1}:{2}:{3}:{4}',
+                ownership, this_client, self._subscription_manager_id,
+                filter_id, uuid.uuid4())
         if name:
             filter_inst['Name'] = name
+
         filter_inst['SourceNamespace'] = source_namespace
         filter_inst['Query'] = query
         filter_inst['QueryLanguage'] = query_language
 
+        filter_path = server.conn.CreateInstance(
+            filter_inst, namespace=server.interop_ns)
+        filter_inst = server.conn.GetInstance(filter_path)
+
         if owned:
-            for i, inst in enumerate(self._owned_filters[server_id]):
-                if inst.path == filter_path:
-                    # It already exists, now check its properties
-                    if inst != filter_inst:
-                        server.conn.ModifyInstance(filter_inst)
-                        filter_inst = server.conn.GetInstance(filter_path)
-                        self._owned_filters[server_id][i] = filter_inst
-                    return filter_inst
-            filter_path = server.conn.CreateInstance(filter_inst)
-            filter_inst = server.conn.GetInstance(filter_path)
             self._owned_filters[server_id].append(filter_inst)
-        else:
-            # Responsibility to ensure it does not exist yet is with the user
-            filter_path = server.conn.CreateInstance(filter_inst)
-            filter_inst = server.conn.GetInstance(filter_path)
 
         return filter_inst
 
@@ -1334,7 +1342,9 @@ class WBEMSubscriptionManager(object):
         server = self._get_server(server_id)
 
         sub_path = CIMInstanceName(SUBSCRIPTION_CLASSNAME,
-                                   namespace=server.interop_ns)
+                                   namespace=server.interop_ns,
+                                   keybindings=[('Filter', filter_path),
+                                                ('Handler', dest_path)])
 
         sub_inst = CIMInstance(SUBSCRIPTION_CLASSNAME)
         sub_inst.path = sub_path
@@ -1347,12 +1357,14 @@ class WBEMSubscriptionManager(object):
                     # It does not have any properties besides its keys,
                     # so checking the path is sufficient.
                     return sub_inst
-            sub_path = server.conn.CreateInstance(sub_inst)
+            sub_path = server.conn.CreateInstance(
+                sub_inst, namespace=server.interop_ns)
             sub_inst = server.conn.GetInstance(sub_path)
             self._owned_subscriptions[server_id].append(sub_inst)
         else:
             # Responsibility to ensure it does not exist yet is with the user
-            sub_path = server.conn.CreateInstance(sub_inst)
+            sub_path = server.conn.CreateInstance(
+                sub_inst, namespace=server.interop_ns)
             sub_inst = server.conn.GetInstance(sub_path)
 
         return sub_inst
--- a/tests/unittest/pywbem/test_subscriptionmanager.py
+++ b/tests/unittest/pywbem/test_subscriptionmanager.py
@@ -25,7 +25,6 @@ from __future__ import absolute_import, print_function
 
 import os
 import re
-from socket import getfqdn
 import pytest
 
 from ...utils import skip_if_moftab_regenerated
@@ -55,7 +54,9 @@ VERBOSE = False
 
 # The following dictionary is the mock WBEM server configuration for the
 # WBEMServerMock class.  It defines a mock WBEM server that is build with
-# defined class schemas, classnames, etc.
+# defined class schemas, classnames, etc.  It uses the same dictionary pattern
+# as the default dictionary in the ../utils directory.
+
 SUBSCRIPTION_WBEM_SERVER_MOCK_DICT = {
     'dmtf_schema': {'version': DMTF_TEST_SCHEMA_VER,
                     'dir': TESTSUITE_SCHEMA_DIR},
@@ -100,12 +101,15 @@ SUBSCRIPTION_WBEM_SERVER_MOCK_DICT = {
     # List of CIMInstances. Each entry is a CIM instance with classname,
     # and properties. All properties required to build the path must be
     # defined. No other properties are required for this test.
-    # TODO: We may expand this for more scoping tests.
     'central-instances': [],
     'scoping-instances': []
 }
 
 
+# NOTE: Please add new test functionality to the function
+#       test_subscriptionmanager(...)) below. We will leave these functions
+#       because they do provide a broad set of test functionality in just a few
+#       tests (each tests creates the mock environment)
 class BaseMethodsForTests(object):
     """
     Base Class for tests contains all common methods and class level
@@ -120,6 +124,7 @@ class BaseMethodsForTests(object):
         """
         skip_if_moftab_regenerated()
 
+        # TODO: modify this to use the same code as in test_subscriptionmanager
         server = WbemServerMock(interop_ns='interop')
         # pylint: disable=attribute-defined-outside-init
         self.conn = server.wbem_server.conn
@@ -140,7 +145,6 @@ class BaseMethodsForTests(object):
                             include_missing_properties=True,
                             include_path=True):
         # TODO AM 8/18 The inst_from_classname() method is not used.
-        # NOTE: This method was never added to pywbem apparently
         """
         Build instance from class using class_name property to get class
         from a repository.
@@ -181,9 +185,8 @@ class BaseMethodsForTests(object):
         Using Server class, get count of Filters, Subscriptions, Destinations
         from server as confirmation outside of SubscriptionManagerCode.
         """
-        this_host = getfqdn()
-
         server = WBEMServer(self.conn)
+        system_name = server.cimom_inst['SystemName']
 
         dest_name_pattern = re.compile(r'^pywbemdestination:')
         dest_insts = server.conn.EnumerateInstances(
@@ -191,7 +194,7 @@ class BaseMethodsForTests(object):
         valid_dests = []
         for inst in dest_insts:
             if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+                    and inst.path.keybindings['SystemName'] == system_name:
                 valid_dests.append(inst)
 
         filter_instances = server.conn.EnumerateInstances(
@@ -200,7 +203,7 @@ class BaseMethodsForTests(object):
         filter_name_pattern = re.compile(r'^pywbemfilter')
         for inst in filter_instances:
             if re.match(filter_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+                    and inst.path.keybindings['SystemName'] == system_name:
                 valid_filters.append(inst)
 
         filter_paths = [inst.path for inst in valid_filters]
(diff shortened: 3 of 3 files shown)


andy-maier (changes requested):
See comments on the issue.
Also, a trivial flake8 issue.

pywbem/_subscription_manager.py
```diff
@@ -343,27 +343,35 @@ def add_server(self, server):
...
+                    self._subscription_manager_id))
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
-        for inst in dest_insts:
-            if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
-                self._owned_destinations[server_id].append(inst)
+        if dest_insts:
+            self._systemnames[server_id] = dest_insts[0]['SystemName']
+            this_host = self._systemnames[server_id]
```
andy-maier, line 359: I suggest renaming this variable to something like server_host or system_name, since this_host was meant to be the host the client runs on

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1301,60 @@ def _create_filter(self, server_id, source_namespace, query,
...
-            filter_path = server.conn.CreateInstance(filter_inst)
+            filter_path = server.conn.CreateInstance(
+                filter_inst, namespace=server.interop_ns)
             filter_inst = server.conn.GetInstance(filter_path)
 
+        # If systemname does not yet exist, create it
+        if server_id not in self._systemnames:
+            self._systemnames[server_id] = filter_inst['SystemName']
+        else:
+            assert self._systemnames[server_id] == filter_inst['SystemName']
```
andy-maier, line 1319: This assert depends on server behavior. What if the server is inconsistent across different providers? Maybe issuing a Python warning would be more robust.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1301,60 @@ def _create_filter(self, server_id, source_namespace, query,
...
+
         return filter_inst
 
+    def _get_systemname(self, server_id):
+        """
+        Get the SystemName returned for anything returned for this server_id.
+        This method gets the s
```
andy-maier, line 1326: Incomplete

docs/changes.rst
```diff
@@ -31,6 +31,20 @@ Released: not yet
   surrounded by verbose test (if p.parse.verbose:). See pywbemcli issue
   #395,
 
+* Fixes bugs in SubscriptionManager where 1. host_name  of
+  CIM_ListenerDestination and CIM_Filter objects inconsistently used
+  in building the string in the Name property and building the regex to
+  find the string. The result was that starting a subscription manager
+  against a server with owned subscriptions, etc. did not restore them to
+  the _owned... instance dictionaries, 2: _create_subscriptions did not
+  create complete path for new subscription before using it,
```
andy-maier, line 40: Not sure we should mention internal methods in the change log. Describe it from the user's view.

andy-maier (changes requested):
See comments

pywbem/_subscription_manager.py
KSchopmeyer: Added.

docs/changes.rst
KSchopmeyer: Changed to reference the external methods add_...  DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

KSchopmeyer:
Current failure is due to issue#2703, make check security issue with notebook version.

pywbem/_subscription_manager.py
```diff
@@ -1190,14 +1200,27 @@ def _create_destination(self, server_id, dest_url, owned):
...
                         self._owned_destinations[server_id][i] = dest_inst
                     return dest_inst
-            dest_path = server.conn.CreateInstance(dest_inst)
+            dest_path = server.conn.CreateInstance(dest_inst,
+                                                   namespace=server.interop_ns)
             dest_inst = server.conn.GetInstance(dest_path)
             self._owned_destinations[server_id].append(dest_inst)
         else:
             # Responsibility to ensure it does not exist yet is with the user
             dest_path = server.conn.CreateInstance(dest_inst)
```
andy-maier, line 1209: Here we should also add `namespace=server.interop_ns`.

pywbem/_subscription_manager.py
```diff
@@ -1334,7 +1430,9 @@ def _create_subscription(self, server_id, dest_path, filter_path, owned):
         server = self._get_server(server_id)
 
         sub_path = CIMInstanceName(SUBSCRIPTION_CLASSNAME,
-                                   namespace=server.interop_ns)
+                                   namespace=server.interop_ns,
+                                   keybindings=[('Filter', filter_path),
+                                                ('Handler', dest_path)])
```
andy-maier, line 1435: The path is not honored by CreateInstance, so we should also add `namespace=server.interop_ns` to the two `CreateInstance()` calls further down in this method.

docs/changes.rst
```diff
@@ -31,6 +31,20 @@ Released: not yet
...
+  subscriptions, etc. did not restore them as owned, 2. add_filter(),
+  add_destination() and add_subscriptions() did not create a complete path for
+  the instance to be added to be added. While this is correct for the
+  CreateInstance(), that path is used to determine if the instance already
+  exists a few lines further down in the code. 3. The client host fqdn was being
+  used as the instance SystemName property when the local instances were built.
+  WBEM servers consistently overwrite this property if is is incorrect or not
+  defined. The SystemName returned from the add_operations is now used to
+  create new paths for the add_...() methods.  (issue #701).
+
```
andy-maier, line 47: Another fix is that the CIM instances for filters, subscriptions and listener destinations were created in the default namespace of the connection of the server that was registered, and are now correctly created in the Interop namespace regardless of the default namespace.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
+            _format(r'^pywbemdestination:owned:{0}:[^:]*$',
+                    self._subscription_manager_id))
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
-        for inst in dest_insts:
-            if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
-                self._owned_destinations[server_id].append(inst)
+        if dest_insts:
+            self._systemnames[server_id] = dest_insts[0]['SystemName']
```
andy-maier, line 358: A few lines down, the key from the path is used. I suggest that we are consistent in whether we obtain the key property from the instance or from the path, i.e. change this line here to also get it from the path.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
+            svr_name = self._systemnames[server_id]
+            for inst in dest_insts:
+                if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
+                        and inst.path.keybindings['SystemName'] == svr_name:
```
andy-maier, line 362: The code a few lines above that gets the svr_name assumes that all instances use the same server name, regardless of whether they are static, owned or persistent. Otherwise it could not arbitrarily pick it from the first instance in the result. If that assumption holds, we no longer need to compare for the server name in this line here. If that assumption does not hold, we should change the way the svr_name is obtained.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
             FILTER_CLASSNAME, namespace=server.interop_ns)
+        if server_id not in self._systemnames:
+            if filter_insts:
+                self._systemnames[server_id] = filter_insts[0]['SystemName']
+                svr_name = self._systemnames[server_id]
+
         for inst in filter_insts:
             if re.match(filter_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+                    and inst.path.keybindings['SystemName'] == svr_name:
```
andy-maier, line 377: Same comment on discovery as above: Remove the SystemName comparison.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
         self._owned_subscriptions[server_id] = []
         self._owned_filters[server_id] = []
         self._owned_destinations[server_id] = []
+        # self._systemnames[server_id] = None
+
+        # TODO expand this to account for WBEM Server profile.
```
andy-maier, line 348: Based on the comments below, I think we no longer need to remember the SystemName properties, i.e. remove `self._systemnames`.

pywbem/_subscription_manager.py
andy-maier: Whatever you did, it does not show up in the PR update.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+
         return filter_inst
 
+    def _get_systemname(self, server_id):
+        """
+        Get the SystemName returned for anything returned for this server_id.
+        This method gets the s
+        """
+        if server_id in self._systemnames:
+            return self._systemnames[server_id]
```
andy-maier, line 1344: I think if the SystemName was not previously discovered, it should be obtained from the CIMOM instance.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+
+        return None
```
andy-maier, line 1346: I think if the SystemName was not previously discovered, it should be obtained from the CIMOM instance.

pywbem/_subscription_manager.py
```diff
@@ -207,6 +206,7 @@ def __init__(self, subscription_manager_id):
         self._owned_subscriptions = {}  # CIMInstance of owned subscriptions
         self._owned_filters = {}  # CIMInstance of owned filters
         self._owned_destinations = {}  # CIMInstance owned destinations
+        self._systemnames = {}  # Dict that will contain SystemNames
```
andy-maier, line 209: Since one WBEMSubscriptionManager instance has one ID and operates on only its own instances, the `self._systemnames` attribute does not need to be a dict by ID, but can and should be a string typed attribute `self._systemname`

andy-maier, line 209: The new attribute should be added to `__str__()` and `__repr__()`.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
         self._owned_subscriptions[server_id] = []
         self._owned_filters[server_id] = []
         self._owned_destinations[server_id] = []
+        # self._systemnames[server_id] = None
```
andy-maier, line 346: If the `self._systemnames` attribute is changed to a string, there is no longer a need to create an item in that dict.

pywbem/_subscription_manager.py
```diff
@@ -1162,26 +1170,28 @@ def _create_destination(self, server_id, dest_url, owned):
...
-        dest_path = CIMInstanceName(DESTINATION_CLASSNAME,
-                                    namespace=server.interop_ns)
-
         dest_inst = CIMInstance(DESTINATION_CLASSNAME)
-        dest_inst.path = dest_path
         dest_inst['CreationClassName'] = DESTINATION_CLASSNAME
         dest_inst['SystemCreationClassName'] = SYSTEM_CREATION_CLASSNAME
-        dest_inst['SystemName'] = this_host
+        # Issue #2704, dummy set because mock requires the property
+        dest_inst['SystemName'] = svr_name or "dummy"
```
andy-maier, line 1184: If `_get_systemname()` is changed as suggested, its return value will never be `None`.

pywbem/_subscription_manager.py
```diff
@@ -1253,29 +1276,31 @@ def _create_filter(self, server_id, source_namespace, query,
...
-                                      namespace=server.interop_ns)
+        svr_name = self._get_systemname(server_id)
 
         filter_inst = CIMInstance(FILTER_CLASSNAME)
-        filter_inst.path = filter_path
         filter_inst['CreationClassName'] = FILTER_CLASSNAME
         filter_inst['SystemCreationClassName'] = SYSTEM_CREATION_CLASSNAME
-        filter_inst['SystemName'] = this_host
+        # Issue #2704, dummy set because mock requires the property
+        filter_inst['SystemName'] = svr_name or "dummy"
```
andy-maier, line 1287: If `_get_systemname()` is changed as suggested, its return value will never be `None`, and the `or "dummy"` can be removed.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+                    _format("WBEM server may be returning different SystemName "
+                            "property for different operations: received  "
+                            "{0!A} and {0!A}. Owned filters may not be found.",
+                            self._systemnames[server_id],
+                            filter_inst['SystemName']),
+                    UserWarning)
+
         return filter_inst
 
+    def _get_systemname(self, server_id):
```
andy-maier, line 1338: if the `self._systemnames` attribute is changed to become a string, the `server_id` parameter is no longer needed.

pywbem/_subscription_manager.py
```diff
@@ -1190,14 +1200,27 @@ def _create_destination(self, server_id, dest_url, owned):
...
+            self._systemnames[server_id] = dest_inst['SystemName']
+        else:
+            if self._systemnames[server_id] != dest_inst['SystemName']:
+                warnings.warn(
+                    _format("WBEM server may be returning different SystemName "
+                            "property for different operations: received  "
+                            "{0!A} and {0!A}. Owned destinations may not be "
+                            "found.", self._systemnames[server_id],
+                            dest_inst['SystemName']),
+                    UserWarning)
```
andy-maier, line 1223: If we remove the SystemName from the detection of owned instances as proposed in another comment, the detection of owned instances is not affected by different SystemName values anymore. In fact, it does not matter that much whether they are different, so I think we can remove the entire else path here.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+            self._systemnames[server_id] = filter_inst['SystemName']
+        else:
+            if self._systemnames[server_id] != filter_inst['SystemName']:
+                warnings.warn(
+                    _format("WBEM server may be returning different SystemName "
+                            "property for different operations: received  "
+                            "{0!A} and {0!A}. Owned filters may not be found.",
+                            self._systemnames[server_id],
+                            filter_inst['SystemName']),
+                    UserWarning)
```
andy-maier, line 1334: See comment on corresponding code in `_create_destination()`.

pywbem/_subscription_manager.py
```diff
@@ -1190,14 +1200,27 @@ def _create_destination(self, server_id, dest_url, owned):
...
             dest_inst = server.conn.GetInstance(dest_path)
             self._owned_destinations[server_id].append(dest_inst)
         else:
             # Responsibility to ensure it does not exist yet is with the user
             dest_path = server.conn.CreateInstance(dest_inst)
             dest_inst = server.conn.GetInstance(dest_path)
 
+        # If systemname does not yet exist, create it
+        if server_id not in self._systemnames:
+            self._systemnames[server_id] = dest_inst['SystemName']
```
andy-maier, line 1214: For consistency, use the SystemName key from the path and not from the instance.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
         else:
             # Responsibility to ensure it does not exist yet is with the user
-            filter_path = server.conn.CreateInstance(filter_inst)
+            filter_path = server.conn.CreateInstance(
+                filter_inst, namespace=server.interop_ns)
             filter_inst = server.conn.GetInstance(filter_path)
 
+        # If systemname does not yet exist, create it
+        if server_id not in self._systemnames:
+            self._systemnames[server_id] = filter_inst['SystemName']
```
andy-maier, line 1325: For consistency, use the SystemName key from the path and not from the instance.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+        svr_name = self._get_systemname(server_id)
+
+        # if svr_name exists, build keybindings, else empty keybindings
+        if svr_name:
+            kb = [('CreationClassName', inst.classname),
+                  ('SystemCreationClassName', SYSTEM_CREATION_CLASSNAME),
+                  ('SystemName', svr_name),
+                  ('Name', inst['Name'])]
+        else:
+            kb = {}
```
andy-maier, line 1386: If `_get_systemname()` is changed as suggested, its return value will never be `None`, and the else path is not needed.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+        This method gets the s
+        """
+        if server_id in self._systemnames:
+            return self._systemnames[server_id]
+
+        return None
+
+    def _build_path(self, inst, ns, server_id):
+        """
+        Common method to build a CIM_InstanceName for either filter or
```
andy-maier, line 1350: Typo: CIM_InstanceName

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+        destination. Used to build the paths for both destinations and filters.
```
andy-maier, line 1351: The sentence starting with "Used to" just repeats what the previous sentence already says.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+        Keybindings will be included only if a system name has already
+        been retrieved from an earlier add_filter or add_destination.
```
andy-maier, line 1353: If `_get_systemname()` is changed as suggested, that statement can be removed.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+
```
andy-maier, line 1354: It would be worthwhile to point out that this common method relies on the fact that CIM_IndicationFilter and CIM_ListenerDestination happen to have the same key properties.

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+        Parameters:
+          inst :class:`~pywbem.CIMInstance`:
+            Instance of CIM_IndicationFilter or CIM_IndicationHandler
```
andy-maier, line 1357: It is not CIM_IndicationHandler, but CIM_ListenerDestination.

docs/changes.rst
```diff
@@ -31,6 +31,20 @@ Released: not yet
   surrounded by verbose test (if p.parse.verbose:). See pywbemcli issue
   #395,
 
+* Fixes bugs in SubscriptionManager where 1. host_name  of
+  CIM_ListenerDestination and CIM_Filter objects inconsistently used
```
andy-maier, line 35: It is not CIM_Filter but CIM_IndicationFilter

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
+                    self._subscription_manager_id))
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
-        for inst in dest_insts:
-            if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
-                self._owned_destinations[server_id].append(inst)
+        if dest_insts:
+            self._systemnames[server_id] = dest_insts[0]['SystemName']
+            svr_name = self._systemnames[server_id]
```
andy-maier, line 359: Can we call this variable "system_name" instead?

andy-maier (changes requested):
See comments

pywbem/_subscription_manager.py
```diff
@@ -1162,26 +1170,28 @@ def _create_destination(self, server_id, dest_url, owned):
         # Validate server_id
         server = self._get_server(server_id)
 
+        svr_name = self._get_systemname(server_id)
```
andy-maier, line 1173: Can we call this variable "system_name" instead?

pywbem/_subscription_manager.py
```diff
@@ -1253,29 +1276,31 @@ def _create_filter(self, server_id, source_namespace, query,
         # Validate server_id
         server = self._get_server(server_id)
 
-        this_host = getfqdn()
         ownership = "owned" if owned else "permanent"
 
-        filter_path = CIMInstanceName(FILTER_CLASSNAME,
-                                      namespace=server.interop_ns)
+        svr_name = self._get_systemname(server_id)
```
andy-maier, line 1281: Can we call this variable "system_name" instead?

pywbem/_subscription_manager.py
```diff
@@ -1284,16 +1309,87 @@ def _create_filter(self, server_id, source_namespace, query,
...
+            The server ID of the WBEM server, returned by
+            :meth:`~pywbem.WBEMSubscriptionManager.add_server`.
+        Returns:
+            CIMInstanceName with path including keybindings or None if
+            there is systemname defined for this server_id
+
+        Returns:
+          CIMInstanceName for inst
+        """
+        svr_name = self._get_systemname(server_id)
```
andy-maier, line 1377: Can we call this variable "system_name" instead?

pywbem/_subscription_manager.py
```diff
@@ -1162,26 +1170,28 @@ def _create_destination(self, server_id, dest_url, owned):
...
             'pywbemdestination:{0}:{1}:{2}',
             ownership, self._subscription_manager_id, uuid.uuid4())
         dest_inst['Destination'] = listener_url
 
         if owned:
+            dest_path = self._build_path(dest_inst,
+                                         server.interop_ns,
+                                         server_id)
             for i, inst in enumerate(self._owned_destinations[server_id]):
                 if inst.path == dest_path:
```
andy-maier, line 1196: What if the server side implementation uses an implementation class name for SystemCreationClassName? I think the whole idea to build a path and then compare it exactly is not flexiblke enough and should be changed to compare only the key properties we know we can compare. I.e. introduce a `_match_path()` method. Also, we have the systemname already in this method, so we do not need to create it from the server_id on each loop turn.

pywbem/_subscription_manager.py
KSchopmeyer: The dummy thing is a temp hack because we need subscription, etc. providers in the mock.

I like the idea of making the WBEMServer  SystemName and/or the entire CIMOM instance.  That eliminates the whole issue of this value without extra  processing cost.

Will do.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
         # Recover any owned destination, filter, and subscription instances
         # that exist on this server
-
-        this_host = getfqdn()
-
         dest_name_pattern = re.compile(
-            _format(r'^pywbemdestination:owned:{0}:{1}:[^:]*$',
-                    this_host, self._subscription_manager_id))
+            _format(r'^pywbemdestination:owned:{0}:[^:]*$',
+                    self._subscription_manager_id))
```
andy-maier, line 354: As discussed in the call, the subscription mgr ID is unique on the client host, so the client host should be put back in, and also be compared against.

pywbem/_subscription_manager.py
```diff
@@ -343,30 +343,38 @@ def add_server(self, server):
...
+            for inst in dest_insts:
+                if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
+                        and inst.path.keybindings['SystemName'] == svr_name:
+                    self._owned_destinations[server_id].append(inst)
 
         filter_name_pattern = re.compile(
-            _format(r'^pywbemfilter:owned:{0}:{1}:[^:]*:[^:]*$',
-                    this_host, self._subscription_manager_id))
+            _format(r'^pywbemfilter:owned:{0}:[^:]*:[^:]*$',
+                    self._subscription_manager_id))
```
andy-maier, line 367: As discussed in the call, the subscription mgr ID is unique on the client host, so the client host should be put back in, and also be compared against.

andy-maier:
We had a short meeting and realized that the subscription manager ID is already documented to be unique to the "current host" which is the client host. That was the reason the client host was made part of the Name properties for CIM_IndicationFilter an CIM_ListenerDestination. So the client host consistently needs to be put back in to the Name properties, but the SystemName properties should be set as in the current PR. I have added according comments.

pywbem/_subscription_manager.py
andy-maier: In our call today, we decided that the WBEMServer class should be extended to have the entire CIMOM instance as a public property. May be in a separate PR, though.

KSchopmeyer: Modified to get SystemName from WBEMServer.cimom_inst object.  Also the 'Dummy' eliminated

Note that the the ObjectManager is already available as the property cimom_inst

DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: Removed the method completely.  Since we put value in self._systemname when we add_server, this method no longer necessary.

DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: Whole thing removed since we are depending on the SystemName from the WBEMServer

pywbem/_subscription_manager.py
KSchopmeyer: Completely removed
DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE:

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: REMOVED the method

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

docs/changes.rst
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: Changed to svr_system_name. DONE

pywbem/_subscription_manager.py
KSchopmeyer: Since the medthod does not exist, DONE

pywbem/_subscription_manager.py
KSchopmeyer: See above  DONE.

pywbem/_subscription_manager.py
KSchopmeyer: Removed.

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: I have no problem removing the TODO since that was a leftover from a much earlier pr.

pywbem/_subscription_manager.py
KSchopmeyer: TODO:   Notet that there is also an instance comparison so the same applies to that test.

KSchopmeyer:
ANDY: The whole question of deteremining modify/create for filters and destinations is even messier than the issue above.

1. for the path, we match everything except the CreationClassName and for that we should really check for subclasses.

2. For the instance equality test we must only match those properties we control.  The returned instance can have other proerties that a) we don't care about, b) we don't even know about that we cannot match.  Thus, we probably should compare just the properties we know about and then tell the modify to only change those properties.

andy-maier:
Karl, on your last comment: This is only for owned instances, right? Those are instances that were always created by the WBEMSubscriptionManager code. So we know which properties were provided.

I think we need to work out what our assumptions about the provider behavior are, i.e. which properties it can change compared to what was provided, and which unprovided properties it can add.

docs/changes.rst
andy-maier: I don't see this as done.

pywbem/_subscription_manager.py
```diff
@@ -244,6 +245,7 @@ def __repr__(self):
             "_owned_subscriptions={s._owned_subscriptions!A}, "
             "_owned_filters={s._owned_filters!A}, "
             "_owned_destinations={s._owned_destinations!A})",
+            "_systemnames={s._systemnames}",
```
andy-maier, line 248: I suggest using the `!A` modifier for consistency. It ensures that the strings are properly escaped.

pywbem/_subscription_manager.py
```diff
@@ -344,29 +346,38 @@ def add_server(self, server):
         self._owned_filters[server_id] = []
         self._owned_destinations[server_id] = []
 
-        # Recover any owned destination, filter, and subscription instances
-        # that exist on this server
+        # Get the SystemName from the WBEMServer CIM_CimomManager for this
```
andy-maier, line 349: It is CIM_ObjectManager, not CIM_CimomManager.

pywbem/_subscription_manager.py
```diff
@@ -1162,35 +1176,37 @@ def _create_destination(self, server_id, dest_url, owned):
...
 
         if owned:
+            dest_path = self._build_path(dest_inst,
+                                         server.interop_ns,
+                                         server_id)
             for i, inst in enumerate(self._owned_destinations[server_id]):
-                if inst.path == dest_path:
+                if self._match_path(inst.path, dest_path):
                     # It already exists, now check its properties
                     if inst != dest_inst:
```
andy-maier, line 1203: This does not work anymore:
* The ModifyInstance addresses the instance to be modified by the instance path, but this instance does not have its path set.
* The GetInstance also relies on a correct path.
So we need to provide a correct path that takes into account implementation freedom such as implementation class names.

I suggest that the `_match_path()` method compares only the Name property, and that `inst.path` is used as the path for Modify and Get.

pywbem/_subscription_manager.py
andy-maier: outdated

pywbem/_subscription_manager.py
```diff
@@ -344,29 +346,38 @@ def add_server(self, server):
...
             _format(r'^pywbemdestination:owned:{0}:{1}:[^:]*$',
-                    this_host, self._subscription_manager_id))
+                    this_client, self._subscription_manager_id))
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
+
         for inst in dest_insts:
             if re.match(dest_name_pattern, inst.path.keybindings['Name']) \
-                    and inst.path.keybindings['SystemName'] == this_host:
+                    and inst.path.keybindings['SystemName'] == svr_system_name:
```
andy-maier, line 369: We should remove the SystemName comparison from this condition, since the identification can now completely be done through the Name property.

Also, for providers that set the SystemName key, we cannot be sure whether the SystemName key in CIM_ObjectManager is consistent with the SystemName key in CIM_ListenerDetination and CIM_IndicationFilter, so the SystemName comparison might even not work.

This comment also applies to `create_filter()`.

pywbem/_subscription_manager.py
andy-maier: outdated, replaced with newer comment.

pywbem/_subscription_manager.py
andy-maier: outdated, replaced with newer comment.

pywbem/_subscription_manager.py
andy-maier: outdated, see my newer comment.

andy-maier (changes requested):
See the new and unresolved comments. Note there are some "hidden" comments again.

pywbem/_subscription_manager.py
KSchopmeyer: Done

pywbem/_subscription_manager.py
KSchopmeyer: Actually our CreateInstance client code will get it from the path if it exists. but I added it to the CreateInstance in any case.

docs/changes.rst
KSchopmeyer: Since the path was attached before the CreateInstance was called, we are covered but I added namespace to the CreateInstance calls.

DONE

pywbem/_subscription_manager.py

docs/changes.rst
KSchopmeyer: DONE again

pywbem/_subscription_manager.py
KSchopmeyer: Not sure where newer comment is.   This one ties to the issue of finding owned filters, etc.  I did not do anything more with it so far.

docs/changes.rst
andy-maier: ok. You are right, it uses the namespace from the path if present.

pywbem/_subscription_manager.py
andy-maier: outdated - function was removed.

pywbem/_subscription_manager.py
andy-maier: The newer comment is https://github.com/pywbem/pywbem/pull/2702#discussion_r646192446.

docs/changes.rst
```diff
@@ -37,6 +37,20 @@ Released: not yet
   surrounded by verbose test (if p.parse.verbose:). See pywbemcli issue
   #395,
 
+* Fixes bugs in SubscriptionManager where 1. host_name  of
```
andy-maier, line 40: Do not abbreviate the class name, i.e. use WBEMSubscriptionManager.

docs/changes.rst
```diff
@@ -37,6 +37,20 @@ Released: not yet
...
+  CIM_ListenerDestination and CIM_IndicationFilter objects inconsistently used
+  in building the string in the Name property and building the regex to
+  find the string. Starting a subscription manager against a server with owned
+  subscriptions, etc. did not restore them as owned, 2. add_filter(),
+  add_destination() and add_subscriptions() did not create a complete path for
+  the instance to be added to be added. While this is correct for the
```
andy-maier, line 46: Two times "to be added".

docs/changes.rst
```diff
@@ -37,6 +37,20 @@ Released: not yet
...
+  CreateInstance(), that path is used to determine if the instance already
+  exists a few lines further down in the code. 3. The client host fqdn was being
```
andy-maier, line 48: item 2 does not describe what the impact is from a user's perspective.

docs/changes.rst
```diff
@@ -37,6 +37,20 @@ Released: not yet
...
+  used as the instance SystemName property when the local instances were built.
+  WBEM servers consistently overwrite this property if is is incorrect or not
+  defined. The SystemName returned from the add_operations is now used to
```
andy-maier, line 51: If WBEM servers consistently overwrite this property, it is not an issue, but it becomes an issue for those servers that don't. That is what we should state here.

andy-maier, line 51: There is no "add_operations". Change to "add_...() methods".

pywbem/_subscription_manager.py
```diff
@@ -463,13 +476,13 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
...
+          ``"pywbemdestination:" {ownership} ":" {client_system_name} ":"
+          {subscription_manager_id} ":" {guid}``
 
         where ``{ownership}`` is ``"owned"`` or ``"permanent"`` dependent on
-        the `owned` argument; ``{subscription_manager_id}`` is the
-        subscription manager ID; and ``{guid}`` is a globally unique
-        identifier.
+        the `owned` argument; ``{client_system_name}`` is the fqdn of the client
+        system; ``{subscription_manager_id}`` is the subscription manager ID;
+        and ``{guid}`` is a globally unique identifier.
```
andy-maier, line 485: In the description of the "subscription_manager_id" parameter, we talk about the "current host" in which the ID must be unique. I suggest that we use consistent terms, so change in the text above:

"{client_system_name}" to "{current_host}"

pywbem/_subscription_manager.py
```diff
@@ -1162,35 +1176,37 @@ def _create_destination(self, server_id, dest_url, owned):
...
                         server.conn.ModifyInstance(dest_inst)
                         dest_inst = server.conn.GetInstance(dest_path)
                         self._owned_destinations[server_id][i] = dest_inst
                     return dest_inst
-            dest_path = server.conn.CreateInstance(dest_inst)
+            dest_path = server.conn.CreateInstance(dest_inst,
+                                                   namespace=server.interop_ns)
             dest_inst = server.conn.GetInstance(dest_path)
             self._owned_destinations[server_id].append(dest_inst)
         else:
```
andy-maier, line 1212: Two lines down is one more CreateInstance without namespace. I suggest adding it there for consistency.

pywbem/_subscription_manager.py
```diff
@@ -1347,7 +1420,8 @@ def _create_subscription(self, server_id, dest_path, filter_path, owned):
                     # It does not have any properties besides its keys,
                     # so checking the path is sufficient.
                     return sub_inst
-            sub_path = server.conn.CreateInstance(sub_inst)
+            sub_path = server.conn.CreateInstance(
+                sub_inst, namespace=server.interop_ns)
             sub_inst = server.conn.GetInstance(sub_path)
             self._owned_subscriptions[server_id].append(sub_inst)
         else:
```
andy-maier, line 1427: Two lines down is one more CreateInstance without namespace. I suggest adding it there for consistency.

docs/changes.rst
```diff
@@ -37,6 +37,20 @@ Released: not yet
...
+  create new paths for the add_...() methods.  (issue #701).
```
andy-maier, line 52: It is issue #2701, not #701.

pywbem/_subscription_manager.py
```diff
@@ -1253,47 +1269,102 @@ def _create_filter(self, server_id, source_namespace, query,
...
+                                           server_id)
             for i, inst in enumerate(self._owned_filters[server_id]):
-                if inst.path == filter_path:
+                if self._match_path(inst.path, filter_path):
                     # It already exists, now check its properties
                     if inst != filter_inst:
                         server.conn.ModifyInstance(filter_inst)
-                        filter_inst = server.conn.GetInstance(filter_path)
+                        filter_inst = server.conn.GetInstance(
+                            filter_path, namespace=server.interop_ns)
```
andy-maier, line 1302: GetInstance does not have a namespace parameter.

pywbem/_subscription_manager.py
andy-maier, line 1203: We should discuss whether modifying the instance in case we identified it but its properties are not as desired, is such a good idea after all. It bears the risk that this changes something that was intentionally done that way before.

andy-maier (changes requested):
See new comments.

pywbem/_subscription_manager.py
KSchopmeyer: Agreed but I want to think it through with some examples first.

docs/changes.rst
KSchopmeyer: DONE

pywbem/_subscription_manager.py
andy-maier: DONE

pywbem/_subscription_manager.py
andy-maier: DONE

pywbem/_subscription_manager.py
andy-maier: DONE

pywbem/_subscription_manager.py
andy-maier: DONE

pywbem/_subscription_manager.py
andy-maier: I like this proposal.

pywbem/_subscription_manager.py
```diff
@@ -1253,47 +1268,92 @@ def _create_filter(self, server_id, source_namespace, query,
...
+                ownership, this_client, self._subscription_manager_id,
+                filter_id, uuid.uuid4())
         if name:
             filter_inst['Name'] = name
+
         filter_inst['SourceNamespace'] = source_namespace
         filter_inst['Query'] = query
         filter_inst['QueryLanguage'] = query_language
 
         if owned:
```
andy-maier, line 1291: The two paths for owned vs permanent can now become one path, i.e. remove the "if" and the "then" block and make the "else" block the remaining code.

pywbem/_subscription_manager.py
andy-maier: DONE

docs/changes.rst
```diff
@@ -37,6 +37,15 @@ Released: not yet
...
   #395,
 
+* Fixes bugs in WBEMSubscriptionManager such that the alternative in the
+  add_filter() and add_destinations() methods to Modify existing instances on the
+  WBEM server instead of create a new instance does not work correctly. Thus,
+  multiple filters and destinations are created with exactly  the same
+  url, query,etc. Also, we cleaned up the algorithm so that owned filters and
+  instances are correctly recovered from the WBEM server when the
+  WBEMSubscriptionManager is restarted (before this they could be returned
+  as not-owned objects)(issue #2701).
```
andy-maier, line 47: If we don't modify anymore, the change log entry needs to be adjusted accordingly.

KSchopmeyer (commented):
Somehow I became a reviewer.

pywbem/_subscription_manager.py
andy-maier: Agree

docs/changes.rst
```diff
@@ -37,6 +37,21 @@ Released: not yet
   surrounded by verbose test (if p.parse.verbose:). See pywbemcli issue
   #395,
 
+* Fixes several isues in WBEMSubscriptionManager:
```
andy-maier, line 40: Typo "isues"

pywbem/_subscription_manager.py
```diff
@@ -794,6 +810,10 @@ def add_filter(self, server_id, source_namespace, query,
                 raise ValueError(
                     _format("Filter ID contains ':': {0!A}", filter_id))
 
+        if name is not None and owned:
+            raise ValueError("The name parameter cannot be used to add "
+                             "owned filters.")
+
```
andy-maier, line 816: I think the combination of using filter_id and not owned should also be raised as an error.

docs/changes.rst

pywbem/_subscription_manager.py
KSchopmeyer: I will change this one.

docs/changes.rst
KSchopmeyer: Agreed

Issue #2704, opened 2021-06-03
pywbem_mock requires providers for CIM_IndicationHandler and CIM_Indication_Filter


KSchopmeyer:
In torder to property test the SubscriptionManager code we use pywbem_mock.  However,  these classes use the key Properties SystemName, CreattionClassName, and SystemCreationClassName. 

At least SystemName, the client can not supply without  querying the server itself. In particualar it may not be the same as the components in the server url (ex. server url is an ip address). It can only be retrieved from the server itself.

In order to build an mock compatible with servers today the server must be capable of suppluying reasonable values for at least SystemName and not just accept the client value, null value, or require that the CreateInstance for filters and Handlers include the SystemName property.

The providers should  be capable of completing some those key values that only the server would know for these classes.

PR #2705, opened 2021-06-04
Create pytest based test for SubscriptionManager

KSchopmeyer:
Adds a new group of tests for SubscriptionManager based on pytest.
Initially this is just tests that do not require a server. and includes
error tests for __init__, add_server, add_filter, and
add_listener_destination.

NOTE: There are a more tests to be added but some of them are propably tied to issue #2704.


tests/unittest/pywbem/test_subscriptionmanager.py
```diff
@@ -369,3 +371,341 @@ def test_create_not_owned_subscription(self):
...
+
+        Returns:
+            returns the server_id
+        """
+
+        # setup a WBEMConnection or pywbem_mock server
+        if connection_attrs == 'MOCKSERVER':
+            conn = setup_mock_environment()
+        else:
+            conn = WBEMConnection(**connection_attrs)
```
andy-maier, line 666: This is supposed to be a unit test. In a unit test, we should not connect to a real server. That should be done in an end2end test, where we have our server selection mechanisms.

tests/unittest/pywbem/test_subscriptionmanager.py
```diff
@@ -369,3 +371,341 @@ def test_create_not_owned_subscription(self):
...
+
+    if "server_id" in exp_result:
+        if isinstance(exp_result['server_id'], list):
+            assert set(server_ids) == set(exp_result['server_id'])
+        else:
+            assert server_id == exp_result['server_id']
+
+    if 'listener_count' in exp_result:
+        assert len(submgr.get_all_destinations(server_id)) == \
+            exp_result['listener_count']
```
andy-maier, line 711: How about verifying the number of filters and subscriptions?

tests/unittest/pywbem/test_subscriptionmanager.py
```diff
@@ -369,3 +371,341 @@ def test_create_not_owned_subscription(self):
...
+    #     an add_server to be executed. If it is a list, multiple
+    #     a server is added for each dict in the list.  This value uses the
+    #     keyword 'MOCKSERVER' to initiated the pywbem_mock environment instead
+    #     of a WBEMConnection.
+    #   * filter_attrs: dictionary of attributes for add_filter. If it is
+    #     a list, one filter is added for each dict in the list
+    #   * dest_attrs: dictionary of attributes for add_listener_destinations.
+    #     Since add_listener_destination can add lists, there is no option
+    #     to add a list for this item.
+    #   * subscription_attrs: dictionary of attributes for add_subscription    #
```
andy-maier, line 422: Add: "If it is a list, one subscription is added for each dict in the list."

andy-maier, line 422: The approach to specify the method arguments does not work well for subscriptions, which take instance paths as input. Maybe specifying the filter and destination instances via index into the list of filter_attrs / dest_attrs would work, at least for the good cases.

tests/unittest/pywbem/test_subscriptionmanager.py
```diff
@@ -298,7 +300,7 @@ def test_create_owned_subscription(self):
 
             # confirm destination instance paths match
             assert sub_mgr.get_all_destinations(server_id)
-            # TODO: ks Finish this test completely when we add other
+            # ks Finish this test completely when we add other
```
andy-maier, line 303: Nothing wrong with a TODO in test code. We only enforce no TODOs in pywbem code.

andy-maier (changes requested):
See comments

tests/unittest/pywbem/test_subscriptionmanager.py
KSchopmeyer: The only tests we do with a real WBEMConnection will be failure tests.

andy-maier: That's fine, but that should be done in end2end tests and not in unit tests.

Issue #2703, opened 2021-06-02
make check failure, notebook


KSchopmeyer:
Notebook 6.1.5 fixes open redirect vulnerability 'GHSA-c7vm-f5p4-8fqh'. See: |
| CVE-2020-26215.

andy-maier:
It seems that the notebook package only recently added its vulnerabilities to safety checking. That can be seen by the satefy IDs which are all very recent: 40380 ... 40385. That's why older versions of notebook are suddenly stated to be a fix to the respective issue. Since we are running safety with out minimum-constraints.txt file, the minimum version 4.3.1 of the notebook package is used for the check. So far for explaining why they come up now.

andy-maier:
On fixing the issue: notebook 6.1.5 which is the minimum version that fixes all of the issues runs on Python 3. only. So either we give up installing notebook on Python 2, or we ignore all thee new safety issues.
**DISCUSSION**

KSchopmeyer:
Lets make our decision on python 2.7 all toghether, not component by component.  I agree with bypassing the safety failures for notebook, especially since it is not a component of the pypi code (at least I think  it is not).

PR #2706, opened 2021-06-06
Circumvented safety issues 40380..40386 with the notebook package

andy-maier:
See commit message.

This PR circumvents the issue by ignoring the safety issues, which allows using notebook on py27. Fixing the safety issues would require using notebook 6.1.5 as a minimum version which would require us to drop support for using Jupyter Notebook on Python 2.7. There is an according discussion point in issue #2703.


andy-maier:
I am merging this PR without review to get us going again. In case we decide for a different solution, we can do that with an additional PR. See the discussion item in issue #2703.

PR #2710, opened 2021-06-08
Allow initalization defined url for FakedWBEMConnection class

KSchopmeyer:
Adds one more parameter to the __init__ method of FakedWBEMConnection
(url) to allow the user to define a different url that will be passed to
the superclass in place of http://FakedURL:5988.

This will be useful in testing, particularly subscription testing where
we want to establish multiple server_ids in the tests.


andy-maier (changes requested):
Only one comment: Add a change log entry for the extension, since this is user visible functionality.

andy-maier (approved):

Issue #2715, opened 2021-06-17
SubscriptionManager add_destinations loses owned parameter


KSchopmeyer:
The method add_destinations in SubscriptionManager loses the method parameter owned when multiple listener urls are defined and it recalls itself because it is not included in the call

PR #2716, opened 2021-06-17
Fix issue #2715, SubscriptionManager loses owned parameter

KSchopmeyer:
The subscription manager does not pass the owned parameter to recalls of
itself if there are multiple urls provided in the call.

Extends test_suscriptions slightly adds capability to remove destinations, filters, subscriptions and test result in addition to new test for creating multiple destinations. Subscription manager now up to 94% with only 15 untested lines of code.


andy-maier (approved):

andy-maier:
The rollback to 1.2 turned out to have too many dependencies on other stuff from master that I abandoned it.

PR #2721, opened 2021-06-20
Ks/#2704 subscription providers

KSchopmeyer:
Adds pywbem_mock providers for the indication subscription classes.

NOTES:
1. Better way to define install.- Right now I added a new method to pywbem_mock/_wbemconnection_mock  I think it would be better if each provider carried its own initialization, a common call) **DISCUSSION**. Note that the install for the namespace provider and these providers is effectively the same code.  It is just that we are intalling 3 providers in this one from a single file. We agreed that this is an issue to be resolved separately.  See comments below for more details.  This method is marked experimental.

### DIFF
--- a/pywbem_mock/_wbemconnection_mock.py
+++ b/pywbem_mock/_wbemconnection_mock.py
@@ -59,6 +59,9 @@ from ._utils import _uprint
 
 from ._namespaceprovider import CIMNamespaceProvider
 
+from ._subscriptionproviders import CIMIndicationSubscriptionProvider, \
+    CIMListenerDestinationProvider, CIMIndicationFilterProvider
+
 __all__ = ['FakedWBEMConnection']
 
 # Fake Server default values for parameters that apply to repo and operations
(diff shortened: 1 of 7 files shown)


pywbem_mock/_wbemconnection_mock.py
```diff
@@ -497,6 +500,67 @@ def install_namespace_provider(self, interop_namespace,
                                schema_pragma_files=schema_pragma_file,
                                verbose=verbose)
 
+    def install_subscription_providers(self, interop_namespace,
+                                       schema_pragma_file=None,
+                                       verbose=None):
+        """
```
andy-maier, line 506: Mark it experimental, since we think it will go away again.

andy-maier (changes requested):
Approved except for the one comment.

pywbem_mock/_wbemconnection_mock.py
KSchopmeyer: Marked as experimental inn the code

KSchopmeyer:
**DISCUSSION**

The python 2.7 failure is"

Collecting GitPython==2.1.1 (from -c minimum-constraints.txt (line 182))
  Could not find a version that satisfies the requirement GitPython==2.1.1 (from -c minimum-constraints.txt (line 182)) (from versions: )
No matching distribution found for GitPython==2.1.1 (from -c minimum-constraints.txt (line 182))

KSchopmeyer:
Collecting GitPython==2.1.1 (from -c minimum-constraints.txt (line 182))
  Could not find a version that satisfies the requirement GitPython==2.1.1 (from -c minimum-constraints.txt (line 182)) (from versions: )
No matching distribution found for GitPython==2.1.1 (from -c minimum-constraints.txt (line 182))
You are using pip version 19.2.3, however version 21.1.3 is available.

andy-maier:
The GitPython issue seems to have resolved itself. See also issue #2725 .

andy-maier (approved):

Issue #2712, opened 2021-06-13
Subscriptions, filters, handlers, open questions for properties


KSchopmeyer:
1. The CIM_Indication Class and profile have defined source_namespaces and deprecated source_namespace.  Should we modify the add_filter method to allow SourceNamespaces, or default to SourceNamespaces, etc. since it has been the preferred solution for many years. now.

   **Suggestion.**  Add option for SourceNamespaces in addition to sourcenamespace with test that they cannot both be provided on a add_filter method call.

2. The PersistenceType property: Should the client set the persistence type property. Note that this property is included in the CIM_ListenerDestination mof file and documented in the Indications Profile sect 7.3.23.4

3. Should we support the IndividualSubscriptionSupported property in the CIM_IndicationFilter and should the client be setting it?
   The model property documentation is:

```
If IndividualSubscriptionSupported is true, the implementation "
          "shall support direct subscription to this IndicationFilter. If "
          "IndividualSubscriptionSupported is false, the implementation shall "
          "not support direct subscription to this IndicationFilter."
```
See section 7.2.7 for more information on the IndividualSubscriptionSupported feature.

4. Should we be using CIMListenerDestinationCIMXML for the listener class or CIM_ListenerDestination since CIM_ListenerDestinationCIMXML is now deprecated and actually contributes nothing new to the values, etc. NOTE: This change (agreed to bel0w) has been moved to its own issue to be sure it is clearly noted and because it represents a possibly incompatible change to the CIM_IndicationFilter  objects created as well as a clear extension of the SubscriptionManager api.

KSchopmeyer:
Based on the the fact that the DMTF indication profile uses SourceNamesapaces we propose to move the filter creation to use SourceNamespaces.

KSchopmeyer:
OpenPegasus supports them both.  If both are provided, they are merged together.

andy-maier:
On PersistenceType: According to DSP1054 2.2.2, this ListenerDestination property basically controls whether the server cleans up failing indication deliveries - upon delivery failure the server removes the subscription and destination instances.
We agreed that this should become an option to select in the add_destination(9 etc methods of the SubscriptionManager.

andy-maier:
On the IndividualSubscriptionSupported property:
We read through DSP1054 1.2.2, and basically this property controls whether a subscription to the indication filter instance is permitted or rejected. The subscription is created via CreateInstance on the Subscription instance, so its provider is the one that needs to do this check.

[...]

PR #2723, opened 2021-07-01
Ks/#2712 add subscription properties

KSchopmeyer:

This pr depends on pr #2704, (the subscription providers)

NOTE: This does not change the SourceNamespace to SourceNamespaces because that change crosses so many other pieces in process.  Will do that as separate pr. A new issue  (issue # 2726) was created to handle that change and issue #2712 changed to reflect that alteration.

Adds PersistenceType property conditinally and changes name of ListenerDestinationClass to CIMListenerDestination.

Adds tests.for the PersistenceTpe change.

rebased this into pr #2704, user providers.  Note that we still allow CIMListenerDestinationXML as a class in the user provider

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -149,6 +149,7 @@ import re
 from socket import getfqdn
 import uuid
 import six
+from nocasedict import NocaseDict
 
 from ._server import WBEMServer
 from ._cim_obj import CIMInstance, CIMInstanceName
@@ -170,17 +171,29 @@ __all__ = ['WBEMSubscriptionManager']
 
 
 def validate_persistence_type(pt):
-    """ Validate persistence type integer with value 1,2,3 """
+    """
+    Validate persistence type parameter pt as string possible
+    values::transient" or "permanent" and convert to corresponding integer
+    value
+    Returns integer value or None for the PersistenceType property based on the
+    input string.
+    """
     if pt is None:
-        return
-    if not isinstance(pt, int):
+        return None
+    if not isinstance(pt, six.string_types):
         raise ValueError(
-            _format("The persistence_type must be an integer. {0} not "
-                    "allowed.", pt))
-    if pt < 2 or pt > 3:
+            _format("The persistence_type must be a string. Type {0} not "
+                    "allowed.", type(pt)))
+
+    persistence_type_dict = NocaseDict([('permanent', 2), ('transient', 3)])
+
+    if pt not in persistence_type_dict:
         raise ValueError(
-            _format("The persistence_type value must be between 1 and "
-                    "3. {0} not allowed.", pt))
+            _format("The persistence_type string must be one of: {0}."
+                    " The value '{1}' is invalid",
+                    ", ".join(list(persistence_type_dict.keys())), pt))
+
+    return persistence_type_dict[pt]
 
 
 class WBEMSubscriptionManager(object):
@@ -557,14 +570,22 @@ class WBEMSubscriptionManager(object):
             permanent. See :ref:`WBEMSubscriptionManager` for details about
             these ownership types.
 
-          persistence_type (:class:`py:int`):
-            Integer value representing the value of the PersistenceType
-            property.  The following values are allowed: 2, Permanent,
-            3, Transient. the default if this parameter is not included is
-            3 (Transient) for owned destinations and to let the server set the
-            value for permanent destinations.  Most WBEM servers set the value
-            to 2(Permanent) if no value is provided.
-            This client does not provide for adding OtherPersistenceType
+          persistence_type (:term:`string`):
+            Optional string where the allowed strings are "transient" and
+            "permanent" and the default is None.  The strings are used to
+            set the PersistenceType property, an integer property with the
+            values of 2 (Permanent) or 3 (Transient)
+
+            The default value is None so that the PersistenceType
+            property is not created on the destination instance for permanent
+            filters. and is created with PersistenceType 3 (Transient) for owned
+            destinations.
+
+            Most WBEM servers set the PersistenceType value to 2 (Permanent) if
+            no value is provided.
+
+            This method does not provide for adding the OtherPersistenceType
+            property.
 
         Returns:
 
@@ -590,17 +611,14 @@ class WBEMSubscriptionManager(object):
         # Here, the variable will be a single list item.
         listener_url = listener_urls
 
-        # if None and owned, set type 3
-        if persistence_type is None:
-            if owned:
-                persistence_type = 3
-            else:
-                validate_persistence_type(persistence_type)
-        else:
-            validate_persistence_type(persistence_type)
+        # Validate which returns integer for instance PropertyValue or None
+        # and if None and owned, set type 3 (transient).
+        persistence_type_value = validate_persistence_type(persistence_type)
+        if owned and persistence_type is None:
+            persistence_type_value = 3
 
         dest_inst = self._create_destination(server_id, listener_url, owned,
-                                             persistence_type)
+                                             persistence_type_value)
 
         return [dest_inst]
 
@@ -1181,7 +1199,8 @@ class WBEMSubscriptionManager(object):
                 del inst_list[i]
                 # continue loop to find any possible duplicate entries
 
-    def _create_destination(self, server_id, dest_url, owned, persistence_type):
+    def _create_destination(self, server_id, dest_url, owned,
+                            persistence_type_value):
         """
         Create a listener destination instance in the Interop namespace of a
         WBEM server and return that instance.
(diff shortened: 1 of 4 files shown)


andy-maier:
This PR is based on a different version of the change in PR #2704 . After merging #2704, this surfaces as a merge conflict. Please resolve.

pywbem/_subscription_manager.py
```diff
@@ -168,6 +169,20 @@
...
+    if pt is None:
+        return
+    if not isinstance(pt, int):
+        raise ValueError(
+            _format("The persistence_type must be an integer. {0} not "
+                    "allowed.", pt))
+    if pt < 2 or pt > 3:
+        raise ValueError(
+            _format("The persistence_type value must be between 1 and "
+                    "3. {0} not allowed.", pt))
```
andy-maier, line 183: 1. The test in the if condition rejects a value of 1, but the error message permits it.
2. It would be good to add to the error message what the values mean.

pywbem/_subscription_manager.py
```diff
@@ -541,6 +557,22 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
             permanent. See :ref:`WBEMSubscriptionManager` for details about
             these ownership types.
 
+          persistence_type (:class:`py:int` or None):
+            The PersistenceType property defines for the WBEM server
+            persistence of the destination instance as follows: 1-Other, 2-
+            Permanent, 3-Transient. Most WBEM servers set the value to
```
andy-maier, line 563: If 1-Other is not allowed, then it should not appear in this enumeration. We should explain what this parameter supports, not what the CIM schema would allow. We can still have the sentence below that explains why Other is not supported.

andy-maier: I suggest to change the format of showing the values from "2-Permanent" to "2 (Permanent)".

pywbem/_subscription_manager.py
```diff
@@ -541,6 +557,22 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
...
+            2(Permanent) if no value is provided with a new instance. pywbem
+            does not support the Other value because it does not support the
+            OtherPersistenceType property.
+
+            If persistence_type is None, the PersistenceType property is
+            created with value 3(Transient) for owned destinations; it is not
+            created for permanent destinations.
```
andy-maier, line 570: Should we then not set 3 as the default right away, instead of taking the detour through None?

pywbem/_subscription_manager.py
```diff
@@ -541,6 +557,22 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
...
+
+            If persistence_type is a valid integer value, the PersistenceType
+            property is created with that value in for both owned and permanent
+            destinations.
```
andy-maier, line 574: Too complicated. We already say earlier that this sets the PersistenceType property, and there is no need to make the statement about valid values, because invalid values raise an exception anyway. The only ifnromation left to say is that this parameter applies to both owned and permanent destinations.

pywbem/_subscription_manager.py
```diff
@@ -541,6 +557,22 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
             permanent. See :ref:`WBEMSubscriptionManager` for details about
             these ownership types.
 
+          persistence_type (:class:`py:int` or None):
+            The PersistenceType property defines for the WBEM server
```
andy-maier, line 561: Here you make a jump from the persistence_type parameter to the PersistenceType property without connecting the two. I suggest first defining what the parameter means without mentioning the property name, and then stating that it sets the property.

pywbem/_subscription_manager.py
```diff
@@ -541,6 +557,22 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True):
             permanent. See :ref:`WBEMSubscriptionManager` for details about
             these ownership types.
 
+          persistence_type (:class:`py:int` or None):
```
andy-maier, line 560: Can we not have string values "permanent" and "transient" instead of the integer values?

pywbem/_subscription_manager.py
```diff
@@ -1200,6 +1244,9 @@ def _create_destination(self, server_id, dest_url, owned):
         dest_inst['CreationClassName'] = DESTINATION_CLASSNAME
         dest_inst['SystemCreationClassName'] = SYSTEM_CREATION_CLASSNAME
         dest_inst['SystemName'] = self._systemnames[server_id]
+        # Populate this property only if value provided
+        if persistence_type:
```
andy-maier, line 1248: This condition does not set it if the value is 0, which is in conflict with the parameter description.

andy-maier (changes requested):
See comments.

In addition, the conflict that emerged after merging PR #2721 needs to be resolved.

pywbem/_subscription_manager.py
KSchopmeyer: Fixed  DONE

pywbem/_subscription_manager.py
KSchopmeyer: This could go either way since there is no real single default. If the parameter is not provided, we set to 3 for owned subscriptions and leave it as None so the Property is not created for permanent subscriptions.

Modified the text slightly to make it clearer.

pywbem/_subscription_manager.py
KSchopmeyer: No.  If the user does not supply the parameter, the default value is None. With value None, No PersistenceType property is created for permanent destinations but the PersistenceType property value is 3(Transient) for owned destinations.

Changed the text to clarify.

pywbem/_subscription_manager.py
KSchopmeyer: State in the first sentence now that it sets the PersistenceType property

pywbem/_subscription_manager.py
KSchopmeyer: Changed to test for None

DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: Rewrote the validate method to use strings and validate them.  The _create_listenerdestination still uses integers as converted in add_listener_destination.

DONE

andy-maier:
The PR has not been updated yet. Did you push your changes?

KSchopmeyer:
I pushed the cleaned up pr aabout an hour after your question.

pywbem/_subscription_manager.py
andy-maier: A resolution to this issue is not in the latest version of the code as of 7/5 (commit b7175d3).

pywbem/_subscription_manager.py
andy-maier: I did not realize the detail that the default depends on ownership.

pywbem/_subscription_manager.py
andy-maier: The latest version of the code does not have such a change, but given that we stick with the integer in the `add_listener_destinations()` method, I don't see why we would change the internal validate method to use strings.
Ok as it is now.

pywbem/_subscription_manager.py
andy-maier: A resolution to this issue is not in the latest version of the code as of 7/4 (commit b7175d3).

andy-maier:
This PR still has conflicts, and misses two of the changes you said were in.
See also the discussion in the Slack channel about rebasing.

andy-maier (changes requested):
This PR still has conflicts, and misses two of the changes you made.
See also the discussion in the Slack channel about rebasing.

pywbem/_subscription_manager.py
```diff
@@ -590,17 +611,13 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True,
...
-            if owned:
-                persistence_type = 3
-            else:
-                validate_persistence_type(persistence_type)
-        else:
-            validate_persistence_type(persistence_type)
+        # Validate and if None and owned, set type 3 (transient).
+        persistence_type_value = validate_persistence_type(persistence_type)
+        if owned and persistence_type is None:
+            persistence_type = 3
```
andy-maier, line 617: I think this should be:
```
persistence_type_value = 3
```

pywbem/_subscription_manager.py
andy-maier: In the last update to this PR, the change is in.

pywbem/_subscription_manager.py
andy-maier: In the last update to this PR, the change is in.

andy-maier (changes requested):
Two comments

pywbem/_subscription_manager.py
KSchopmeyer: You are correct.  Fixed.   DONE

KSchopmeyer:
Your issues were fixed.

pywbem/_subscription_manager.py
```diff
@@ -1181,7 +1198,8 @@ def remove_subscriptions(self, server_id, sub_paths):
                 del inst_list[i]
                 # continue loop to find any possible duplicate entries
 
-    def _create_destination(self, server_id, dest_url, owned, persistence_type):
+    def _create_destination(self, server_id, dest_url, owned,
+                            persistence_value):
```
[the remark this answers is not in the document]

PR #2730, opened 2021-07-06
Ks/#2725 add sourcenamespaces property

KSchopmeyer:
Modifies the positional property SourceNamespace to SourceNamespaces and adds a new opeiontal keyword property Sourcenamespace ad backup. 

Adds tests specifically for this change

Adds note to document incompatibility  in the docs incompatibility section about this change.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -29,6 +29,24 @@ Released: not yet
   raised in earlier versions may still be raised for other, less common cases.
   For details, see the corresponding item in the Enhancements section, below.
 
+* Changed 'SubscriptionManager.add_filter()' method to use the
+  'SourceNamespaces' property (allows multiple namespaces) of the
+  'CIM_IndicationFilter' class instead of the deprecated 'SourceNamespace'
+  property (allows only single namespace).  This changed the name of the
+  positional 'source_namespace' parameter to 'source_namespaces`. The new
+  parameter allows both string and list of strings as values.
+
+  This change brings the subscription manager in line with the incorporation of
+  the 'SourceNamespaces' property made to this CIM class definition
+  by DMTF CIM schema release 2.22.0.
+
+  An optional 'source_namespace' keyword parameter has been added to the
+  'add_filter()'method to account for any case where a WBEM Server cannot
+  handle the SourceNamespaces property. The primary incompatibility will be
+  that the instance created for CIM_Indication filter now has a property named
+  'SourceNamespaces' instead of 'SourceNamespace'. See further comments below
+  and issue #2725.
+
 **Deprecations:**
 
 **Bug fixes:**
@@ -115,6 +133,10 @@ Released: not yet
 
 * Finalized the support for SI units that was experimental so far, i.e. the
   'pywbem.siunit()' and 'pywbem.siunit_obj()' functions. (issue #2653)
+
+* Modify 'SubscriptionManager.add_filter()' to use the CIM_IndicationFilter
+  property 'SourceNamespaces' in place of the deprecated 'SourceNamespace'. (see
+  issue #2725 and the **Incompatible changes:** section above)
 
 **Cleanup:**
 
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -744,9 +744,9 @@ class WBEMSubscriptionManager(object):
                 del inst_list[i]
                 # continue loop to find any possible duplicate entries
 
-    def add_filter(self, server_id, source_namespace, query,
+    def add_filter(self, server_id, source_namespaces, query,
                    query_language=DEFAULT_QUERY_LANGUAGE, owned=True,
-                   filter_id=None, name=None):
+                   filter_id=None, name=None, source_namespace=None):
         """
         Add a :term:`dynamic indication filter` to a WBEM server, by creating
         an indication filter instance (of CIM class "CIM_IndicationFilter") in
@@ -799,8 +799,12 @@ class WBEMSubscriptionManager(object):
             The server ID of the WBEM server, returned by
             :meth:`~pywbem.WBEMSubscriptionManager.add_server`.
 
-          source_namespace (:term:`string`):
-            Source namespace of the indication filter.
+          source_namespaces (:term:`string`, list of :term:`string`, or None):
+            Source namespace or list of source namespaces  of the indication
+            filter. The values will be inserted into the CIM_IndicationFilter
+            SourceNamespaces property. If None, the SourceNamespaces property
+            will not be created. and the WBEM server will use the Interop
+            namespace as the source namespace for the indication filter.
 
           query (:term:`string`):
             Filter query in the specified query language.
@@ -837,6 +841,18 @@ class WBEMSubscriptionManager(object):
             filter instance, or `None` if the `filter_id` parameter is
             specified.
 
+          source_namespace (:term:`string):
+            Optional source namespace of the indication filter. If the
+            parameter is provided the value will be inserted into the
+            CIM_IndicationFilter SourceNamespace property. Otherwise the
+            SourceNamespace property will not be created. The `SourceNamespace`
+            property is deprecated in the DMTF CIM_IndicationFilter class in
+            favor of `SourceNamespaces`( see the `SourceNamespaces` parameter
+            above). This parameter is provided only to support calls to
+            add_filter() that use the `source_namespace` as a keyword parameter
+            or to support very old WBEM servers (prior to DMTF schema version
+            2.22) that require the `SourceNamespace` property.
+
         Returns:
 
             :class:`~pywbem.CIMInstance`: The created indication filter
@@ -870,9 +886,19 @@ class WBEMSubscriptionManager(object):
             raise ValueError("The name parameter cannot be used to add "
                              "owned filters.")
 
-        filter_inst = self._create_filter(server_id, source_namespace, query,
-                                          query_language, owned, filter_id,
-                                          name)
+        if source_namespaces is not None:
+            if not isinstance(source_namespaces, (six.string_types, list)):
+                raise TypeError(
+                    _format("source_namespaces must string or list. {0} is "
+                            "type {1}.", source_namespace,
+                            type(source_namespace)))
+
+            if isinstance(source_namespaces, six.string_types):
+                source_namespaces = [source_namespaces]
+
+        filter_inst = self._create_filter(
+            server_id, source_namespaces, query, query_language, owned,
+            filter_id, name, source_namespace=source_namespace)
 
         return filter_inst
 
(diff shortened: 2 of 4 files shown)


docs/changes.rst
```diff
@@ -116,6 +116,10 @@ Released: not yet
 * Finalized the support for SI units that was experimental so far, i.e. the
   'pywbem.siunit()' and 'pywbem.siunit_obj()' functions. (issue #2653)
 
+* Modify SubscriptionManager add_filter to use the CIM_IndicationFilter
+  property SourceNamespaces in place of the deprecated SourceNamespace. (see
+  issue #2725)
+
```
andy-maier, line 122: There should be an additional entry in the Incompatibilities section that points out how this change is incompatible.

pywbem/_subscription_manager.py
```diff
@@ -837,6 +841,18 @@ def add_filter(self, server_id, source_namespace, query,
...
+            parameter is provided the value will be inserted into the
+            CIM_IndicationFilter SourceNamespace property. Otherwise the
+            SourceNamespace property will not be created. The SourceNamespace
+            property is deprecated in the DMTF CIM_IndicationFilter class in
+            favor of SourceNamespaces( see the SourceNamespaces parameter
+            above). This parameter is provided only to support implementations
+            that use the keyword syntax with the SourceNamespace property
+            name OR to support very old WBEM servers (bprior to DMTF schema
+            version 2.22).
+
```
andy-maier, line 855: I thought we said that we would not support SourceNamespace anymore??

pywbem/_subscription_manager.py
```diff
@@ -870,9 +886,12 @@ def add_filter(self, server_id, source_namespace, query,
             raise ValueError("The name parameter cannot be used to add "
                              "owned filters.")
 
-        filter_inst = self._create_filter(server_id, source_namespace, query,
-                                          query_language, owned, filter_id,
-                                          name)
+        if isinstance(source_namespaces, six.string_types):
```
andy-maier, line 889: That is a dangerous check. What if a byte string is passed in? I suggest:
`if not isinstance(source_namespaces, Sequence):`
where Sequence first gets imported from collections.abc and upon importerror from collections.

pywbem/_subscription_manager.py
```diff
@@ -1321,6 +1342,9 @@ def _create_filter(self, server_id, source_namespace, query,
             Value for the `Name` property of the filter instance, or `None`.
             Mutually exclusive with the `filter_id` parameter.
 
+          source_namespace (:term:`string` or None):
+            A namespace or None.
+
```
andy-maier, line 1347: Same comment as above on supporting both.

andy-maier (changes requested):
See comments

pywbem/_subscription_manager.py
KSchopmeyer: After looking at it this seemed to be an easy way to keep the possibility of a requirement for the SourceNamespace property at little cost to us.

docs/changes.rst
KSchopmeyer: I know I wrote that.  System locked up yesterday so that must have gotten lost.

DONE

pywbem/_subscription_manager.py
KSchopmeyer: I am confused.  My understanding is that `sequence` may be any of at least list, tuple, string, and range;    I believe that you get down into mutable and immutable sequences before resolving this one.

string - immutable sequence
list - mutable sequence
tuple - immutable sequence

Perhaps the better solution here is to validate that the value is either str or mutable sequence (list) and reject if not. Then we know exactly  what we Have.  I impemented that.

Since we have six.string_types tests in many places to determine if we have a list or string, do we have a much bigger problem here????

pywbem/_subscription_manager.py
KSchopmeyer: This seemed easy and keeps allows the old name and property at little cost.  Seemed easy way to cover the really odd  case where someone did require SourceNamespace.

pywbem/_subscription_manager.py
andy-maier: You are right. Strings are also sequences, and even worse, I misread the current test, which uses `six.string_types` and thus covers both unicode strings and byte strings. Sorry for my confusion on this one.
WITHDRAWING COMMENT

KSchopmeyer: OK.  I spent another hour going back over the whole byte string, str, sequence, etc.  and to be honest, it is really messy.

I went ahead and did check on string_types or list for validation of the parameter in add_filter.  Thus, no byte strings or tuples.

pywbem/_subscription_manager.py
```diff
@@ -799,8 +799,12 @@ def add_filter(self, server_id, source_namespace, query,
             The server ID of the WBEM server, returned by
             :meth:`~pywbem.WBEMSubscriptionManager.add_server`.
 
-          source_namespace (:term:`string`):
-            Source namespace of the indication filter.
+          source_namespaces (:term:`string`, list of :term:`string`, or None):
+            Source namespace or list of source namesaces  of the indication
```
andy-maier, line 803: Typo: "namesaces"

pywbem/_subscription_manager.py
```diff
@@ -799,8 +799,12 @@ def add_filter(self, server_id, source_namespace, query,
...
+            filter. The values will be inserted into the CIM_IndicationFilter
+            SourceNamespaces property. If None, the SourceNamespaces property
+            will not be created. and the WBEM server will use the interop
```
andy-maier, line 806: We use mixed case for "Interop namespace"

pywbem/_subscription_manager.py
```diff
@@ -837,6 +841,18 @@ def add_filter(self, server_id, source_namespace, query,
             filter instance, or `None` if the `filter_id` parameter is
             specified.
 
+          source_namespace (:term:`string):
+            Optional source namespace of the indication filter. If the
+            parameter is provided the value will be inserted into the
+            CIM_IndicationFilter SourceNamespace property. Otherwise the
+            SourceNamespace property will not be created. The SourceNamespace
+            property is deprecated in the DMTF CIM_IndicationFilter class in
+            favor of SourceNamespaces( see the SourceNamespaces parameter
```
andy-maier, line 850: Use the function parameter name: " (see the source_namespaces parameter above)"

pywbem/_subscription_manager.py
```diff
@@ -837,6 +841,18 @@ def add_filter(self, server_id, source_namespace, query,
...
+            above). This parameter is provided only to support implementations
+            that use the keyword syntax with the SourceNamespace property
+            name OR to support very old WBEM servers (bprior to DMTF schema
```
andy-maier, line 853: Typo: "bprior"

pywbem/_subscription_manager.py
```diff
@@ -837,6 +841,18 @@ def add_filter(self, server_id, source_namespace, query,
...
 
+          source_namespace (:term:`string):
+            Optional source namespace of the indication filter. If the
+            parameter is provided the value will be inserted into the
+            CIM_IndicationFilter SourceNamespace property. Otherwise the
+            SourceNamespace property will not be created. The SourceNamespace
+            property is deprecated in the DMTF CIM_IndicationFilter class in
+            favor of SourceNamespaces( see the SourceNamespaces parameter
+            above). This parameter is provided only to support implementations
+            that use the keyword syntax with the SourceNamespace property
```
andy-maier, line 852: I don't understand the first part of this sentence: "implementations that use the keyword syntax with the SourceNamespace property".

Is this about how the "source_namespace" parameter is passed to the Python function?

If so, then this supports "users" and not "implementations"; it is a parameter and not a property;  the parameter is named differently; and the whole story is that the change is backwards incompatible for users that specified it as a positional parameter and expected the SourceNamespace property to be set. I tend to think that whole part of the sentence should be removed, and a statement about the incompatible change should be made.

pywbem/_subscription_manager.py
andy-maier: Ok, then let's keep it.

pywbem/_subscription_manager.py
andy-maier: Ok, then let's keep it.

andy-maier (changes requested):
See comments

docs/changes.rst
```diff
@@ -29,6 +29,23 @@ Released: not yet
...
   For details, see the corresponding item in the Enhancements section, below.
 
+* Changed 'SubscriptionManager.add_filter()' method to use the
+  'SourceNamespaces' property (allows multiple namespaces) of the
+  'CIM_IndicationFilter' class instead of the deprecated 'SourceNamespace'
+  property (allows only single namespace).  This changed the name of the
+  positional 'source_namespace' parameter to source_namespaces. The new
+  parameter allows both string and list of strings as values.
+
+  This change is brings the subscription manager in line with the the changes
```
andy-maier, line 39: Grammar: "is brings"; "the the"

docs/changes.rst
```diff
@@ -29,6 +29,23 @@ Released: not yet
...
+  made to this CIM class definition before DMTF CIM schema release 2.22.0.
```
andy-maier, line 40: On "the changes made [...] before DMTF CIM schema release 2.22.0.": Talking about the changes does not go well with talking about "before" a particular release. Either it was the state (or something) before, or the "changes in".

docs/changes.rst
```diff
@@ -29,6 +29,23 @@ Released: not yet
   raised in earlier versions may still be raised for other, less common cases.
   For details, see the corresponding item in the Enhancements section, below.
 
+* Changed 'SubscriptionManager.add_filter()' method to use the
+  'SourceNamespaces' property (allows multiple namespaces) of the
+  'CIM_IndicationFilter' class instead of the deprecated 'SourceNamespace'
+  property (allows only single namespace).  This changed the name of the
+  positional 'source_namespace' parameter to source_namespaces. The new
```
andy-maier, line 36: Use single quotes also around "source_namespaces", for consistency.

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: DONE

pywbem/_subscription_manager.py
KSchopmeyer: Modified text

andy-maier (changes requested):
Tha latest update resolved only some of my earlier comments. Some more to go.

docs/changes.rst
KSchopmeyer: DONE

docs/changes.rst
KSchopmeyer: Modified grammar.  However, we do not know what version exactly because none of the older schema versions are onthe DMTF web page.  The 2.22.0 in OpenPegasus shows the new property and that is my only source of information 2.22.0 itself.

docs/changes.rst
KSchopmeyer: DONE

andy-maier (approved):

PR #2732, opened 2021-07-09
Ks/#2731 restoreclass cim listener destination cimxml

KSchopmeyer:
I mistakenly used CIM_ListenerDestination as the target class for subscription manager.  Since that class is abstract, the add_filter no longer works. This restores the target class to CIM_ListenerDestinationCIMXML.


andy-maier (approved):

andy-maier:
Good catch.

This PR needs to wait until PR #2730 is final and merged, and then rebased.

andy-maier:
Rebased the PR.

PR #2734, opened 2021-07-10
Added support for ExportIndication export operation

andy-maier:
See commit message.
Now ready for merge.

### DIFF
--- a/pywbem/_cim_http.py
+++ b/pywbem/_cim_http.py
@@ -235,9 +235,10 @@ def request_exc_message(exc, conn):
     return message
 
 
-def wbem_request(conn, req_data, cimxml_headers):
+def wbem_request(conn, req_data, cimxml_headers, target_type='server'):
     """
-    Send an HTTP or HTTPS request to a WBEM server and return the response.
+    Send an HTTP or HTTPS request to a WBEM server or WBEM listener and return
+    the response.
 
     Parameters:
 
@@ -245,7 +246,8 @@ def wbem_request(conn, req_data, cimxml_headers):
         WBEM connection to be used.
 
       req_data (:term:`string`):
-        The CIM-XML formatted data to be sent as a request to the WBEM server.
+        The CIM-XML formatted data to be sent as a request to the WBEM server
+        or WBEM listener.
 
       cimxml_headers (:term:`py:iterable` of tuple(string,string)):
         Where each tuple contains: header name, header value.
@@ -256,12 +258,14 @@ def wbem_request(conn, req_data, cimxml_headers):
 
         A value of `None` is treated like an empty iterable.
 
+      target_type (:term:`string`): Target type: 'server' or 'listener'.
+
     Returns:
 
         Tuple containing:
 
-            The CIM-XML formatted response data from the WBEM server, as a
-            :term:`byte string` object.
+            The CIM-XML formatted response data from the WBEM server or
+            WBEM listener, as a :term:`byte string` object.
 
             The server response time in seconds as floating point number if
             this data was received from the server. If no data returned
@@ -278,7 +282,10 @@ def wbem_request(conn, req_data, cimxml_headers):
     if cimxml_headers is None:
         cimxml_headers = []
 
-    target = '/cimom'
+    if target_type == 'server':
+        target = '/cimom'
+    else:
+        target = ''
     target_url = '{}{}'.format(conn.url, target)
 
     # Make sure the data parameter is converted to a UTF-8 encoded byte string.
@@ -294,7 +301,7 @@ def wbem_request(conn, req_data, cimxml_headers):
     }
     req_headers.update(dict(cimxml_headers))
 
-    if conn.creds is not None:
+    if target_type == 'server' and conn.creds is not None:
         auth = '{0}:{1}'.format(conn.creds[0], conn.creds[1])
         auth64 = _ensure_unicode(base64.b64encode(
             _ensure_bytes(auth))).replace('\n', '')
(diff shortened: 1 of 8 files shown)


pywbem/_cim_operations.py
```diff
@@ -360,23 +360,28 @@ def _validate_context(context):
...
-    connection to multiple CIM namespaces on a WBEM server (when the namespace
-    is specified on subsequent operations), or as a connection to only the
-    default namespace (this allows omitting the namespace on subsequent
-    operations).
+    A client's connection to a WBEM server or WBEM listener. This is the main
+    class of the WBEM client library API.
+
+    This class is used for communication with WBEM servers and WBEM listeners,
+    because the communication mechanisms are very similar. However, a
+    particular object of this class will connect to only one target which is
```
KSchopmeyer, line 368: one target type I think

pywbem/_cim_operations.py
```diff
@@ -360,23 +360,28 @@ def _validate_context(context):
...
     As usual in HTTP, there is no persistent TCP connection; the connectedness
     provided by this class is only conceptual. That is, the creation of the
-    connection object does not cause any interaction with the WBEM server, and
-    each subsequent WBEM operation performs an independent, state-less
-    HTTP/HTTPS request. Usage of the `requests` Python package causes the
-    underlying resources such as sockets to be pooled, though.
+    connection object does not cause any interaction with the WBEM server or
+    WBEM listener, and each subsequent WBEM operation performs an independent,
+    state-less HTTP/HTTPS request. Usage of the `requests` Python package
+    causes the underlying resources such as sockets to be pooled, though.
```
KSchopmeyer, line 384: minor syntax.  change though to however and put it at start of sentence

pywbem/_cim_http.py
```diff
@@ -329,14 +336,17 @@ def wbem_request(conn, req_data, cimxml_headers):
...
 
-    # Get the optional response time header
-    svr_resp_time = resp.headers.get('WBEMServerResponseTime', None)
-    if svr_resp_time is not None:
-        try:
-            # convert to float and map from microsec to sec.
-            svr_resp_time = float(svr_resp_time) / 1000000
-        except ValueError:
-            pass
+    if target_type == 'server':
```
KSchopmeyer, line 339: Why not just include this with the listener???

pywbem/_cim_operations.py
```diff
@@ -394,7 +399,9 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
 
     After creating a :class:`~pywbem.WBEMConnection` object, various methods
     may be called on the object, which cause WBEM operations to be issued to
-    the WBEM server. See :ref:`WBEM operations` for a list of these methods.
+    the WBEM server or WBEM listener. See :ref:`WBEM operations` for a list of
```
KSchopmeyer, line 402: I don't see the export... method listed in that table.

pywbem/_cim_operations.py
```diff
@@ -489,7 +497,7 @@ def __init__(self, url, creds=None, default_namespace=None,
         Parameters:
 
           url (:term:`string`):
-            URL of the WBEM server, in the format:
+            URL of the WBEM server or WBEM listener, in the format:
```
KSchopmeyer, line 500: Should we be saying something about the port being a requirement for the WBEMListener?

pywbem/_cim_operations.py
```diff
@@ -634,12 +647,12 @@ def __init__(self, url, creds=None, default_namespace=None,
...
 
           timeout (:term:`number`):
-            Timeout in seconds, for requests sent to the server.
+            Timeout in seconds, for requests sent to the server or listener.
 
             *New in pywbem 0.8.*
 
-            If the server did not respond within the timeout duration, the
-            socket for the connection will be closed, causing a
+            If the server or listener did not respond within the timeout
```
KSchopmeyer, line 654: change dit not respond to does not respond.

pywbem/_cim_operations.py
```diff
@@ -900,7 +918,7 @@ def scheme(self):
     def host(self):
         """
         :term:`unicode string`: Normalized host and port number of the WBEM
-        server, in the format ``host:port``.
+        server or WBEM listener, in the format ``host:port``.
 
         Default port numbers (5988 for http and 5989 for https) have been
```
KSchopmeyer, line 923: These port numbers are for servers only

pywbem/_cim_operations.py
```diff
@@ -9796,6 +9977,61 @@ def DeleteQualifier(self, QualifierName, namespace=None):
             if self._operation_recorders:
                 self.operation_recorder_stage_result(None, exc)
 
+    def ExportIndication(self, NewIndication):
+        # pylint: disable=invalid-name
+        """
+        Send an indication to a WBEM listener.
+
+        This method performs the ExportIndication export operation
```
KSchopmeyer, line 9985: I think we have one too many exports here.  This method performs the ExportIndication operation ...

KSchopmeyer (commented):
Made some comments

KSchopmeyer:
NOTE: when adding exportindication to the table of operations, perhaps we should separate the table in requests and exports.

pywbem/_cim_operations.py
andy-maier: DONE

pywbem/_cim_operations.py
andy-maier: Good catch. Added the method in a second table after the first.
DONE

pywbem/_cim_operations.py
andy-maier: Yes. Added a sentence.
DONE

pywbem/_cim_operations.py
andy-maier: DONE

pywbem/_cim_operations.py
andy-maier: The `host` property will always contain `host:port`, even for listeners, since the property has the result of URL parsing, with missing port number defaulted dependent on the scheme.

For listeners, we now state that a port should be specified in the URL, and regardless of whether that is done, the `host`property will have a port.

pywbem/_cim_operations.py
andy-maier: It is an "export operation". DSP0200 uses "operation" for operations targeting a server, and "export operation" for operations originating on a server.

andy-maier:
Resolved all but two comments as proposed.

pywbem/_cim_operations.py
KSchopmeyer: No Issiue DONE

pywbem/_cim_operations.py
KSchopmeyer: Since that is the spec, I respect that.  Suggest simply note that this is as defined in DSP0200.  I think in a lot of places we include the word request for other operations which means we are wrong there I would guess.

KSchopmeyer:
I agreed to your two discents except for one comment about clarifying export operation where used.

pywbem/_cim_operations.py
andy-maier: It already says "see DSP0200".

andy-maier: Double checked with DSP0200 ans it actually uses the term "export method". Changed all occurrences of "export operation" to "export method".

Issue #2434, opened 2020-08-28
Packaging burden by breaking out pywbem code as separate libraries


tasleson:
I understand the desire to make internal code functionality into a separate library, but it comes at a cost.  For distributions like fedora that package each python library as a separate package, it requires a maintainer to create a new package, get the package through the approval process, and then maintain/test it over time.

At the moment those that are involved in pywbem packaging on fedora are considering dropping/orphaning pywbem in fedora as the packaging burden has increased (unless we can find someone to take it and all the new dependencies over).  If this occurs a few other packages in fedora which utilize pywbem will get dropped/orphaned too.

Is anyone here willing to help get these new packages specific to pywbem packaged for fedora and other distributions?

Maybe pywbem could come up with some other alternative, like git submodule?  Keep this needed functionality together and separate for reuse?  I'm not familiar with submodules and how difficult that would be do integrated releases in github for example.

Suggestions/help welcomed

Thanks

Ref. https://bugzilla.redhat.com/show_bug.cgi?id=1868128

andy-maier:
Hi Tony.

We are certainly willing to help on this one.

* nocasedict and nocaselist are new packages on Pypi. They started out in pywbem, but are independent packages for use by anyone. They are both used by the normal operation of pywbem. The options I see are: (1) They become new packages on Fedora or (2) We vendor them into pywbem. Let me know which way to proceed.
* custom-inherit is likely a mistake on our end, I need to investigate that. It probably should have been only a dependency for development of pywbem. I'll post an update on this, next week (I'm on vacation tomorrow until monday).
* yamlloader is the replacement for yamlordereddictloader, as you found out. pywbem used yamlordereddictloader only in its tests. The newer yamlloader package got an additional use in pywbem now (in pywbem/_recorder.py). That's why you don't find a dependency to yamlordereddictloader in the history of the requirements.txt file of pywbem. Given that it is the recommended replacement for yamlordereddictloader, would adding yamlloader as a package be an issue?

tasleson:
Hi @andy-maier 

We would appreciate `nocasedict` and `nocaselist` included/vendored if possible, unless you or someone else wants to pick up packaging for distros.  As for yamlloader, I will reach out to the fedora yamlordereddictloader maintainer and see if there is plans to replace it with yamlloader, which I think makes sense (assuming it's a drop in replacement).

I'll look into `custom-inherit` some more, but isn't it only required for testing?

Thanks for responding and a willingness to work with us.

[...]

PR #2737, opened 2021-07-11
[WIP] Included vendored packages: nocasedict, nocaselist, yamlloader

andy-maier:
See commit message.

**Note: This PR is ready for review, but merging needs to wait for resolution of discussion in issue #2434**

TODOs:
* The import statements in the pywbem code for the vendored packages had to be changed to get them from the `_vendor` directory. It would be nicer if there could be an import approach that is common regardless of whether the package is actually vendored or not.
* If the nocaselist and nocasedict packaging requests for Fedora get approved, remove them again from this PR.
* Use a released version of vendorize that fixes issue https://github.com/mwilliamson/python-vendorize/issues/13. Currently, a version from my fork is used.
* Use a released version of yamlloader for Python 2.7 that fixes issue https://github.com/Phynix/yamlloader/issues/24. Currently, a version from my fork is used.


KSchopmeyer (approved):

KSchopmeyer:
I approve this but what worrys me is that this is going to be a trend with a couple of vendors like Redhat driving a lot of the direction of things like python by setting their special rules.  Thus, vendorize may become more and more popular leading to larger packages and fewer uses of these vendorized packages, in turn leading to false statistics about the use of these vendorized packages.  Oh well.

andy-maier:
Added the yamlloader package also as a vendored package. See Tony's comment in issue #2434 .

andy-maier:
Using the vendorize package from my fork with a fix for its issue https://github.com/mwilliamson/python-vendorize/issues/13

andy-maier:
Since there is an outlook that the nocasedict and nocaselist Python packages can become OS-level packages, this PR has been removed from 1.3.0 again. It remains open until the packaging question is resolved.

andy-maier:
Dropped, see #2434.

PR #2738, opened 2021-07-11
Tolerated servers returning CIM_ERR_FAILED when not supporting pull ops

andy-maier:
See commit message.


KSchopmeyer:
Tested against the FSCB implementation that failed.  This worked.

Issue #2742, opened 2021-07-12
mocker/compiler create instances from abstract classes


KSchopmeyer:
All parts of pywbem environment that create instances ignore the rule that instances cannot be created from Abstract classes. The code to test and reject these attempts to do CreateInstance need to be included in a) parts of the compiler path, b) the mocker.

andy-maier:
This requires that the client side CreateInstance() method inspects the class before issuing the CreateInstance operation. 

This has several disadvantages:

* Extra operation
* Does not work with SFCB (it does not support class operations)
* Should not be a client responsibility anyway to check this, instead the server ought to reject this.

I suggest not doing this.

andy-maier:
I understand the mocker, but why in the compiler path? The compiler acts as a client against a server. If the server is a real server or a pywbem mocked server, it should reject creation of instances of abstract classes, but as a client, the compiler does not need to check this upfront.

KSchopmeyer:
The problem here is that the compiler path is actually to a local file that acts as a local server.  Should we allow the  instances to be created here to then be rejected by a server.  If this was really part of just the compiler path, the issue makes sense but this local repo acts both at a shortcut to speed things up and as a repository.  That is why I put the test in the compiler path

andy-maier:
OK, I see.

KSchopmeyer:
Actually could have argued either way since it is both a client support tool and a local server.  Arggggghhhhh

PR #2744, opened 2021-07-13
Issue #2742 - Add tests for instance create of abstract class

KSchopmeyer:
Adds validation on pywbem_mock and mof_compiler for attempts to create
an instance where the creation class has the Abstract qualifier.  Adds the validation, documents the new validation, and adds tests for both CreateInstance and compile of the instance into pywbem_mock.

This fails with CIMError CIM_ERR_FAILED.

### DIFF
--- a/pywbem/_mof_compiler.py
+++ b/pywbem/_mof_compiler.py
@@ -2394,6 +2394,14 @@ class MOFWBEMConnection(BaseRepositoryConnection):
                                 IncludeQualifiers=True)
             inst.path = CIMInstanceName.from_instance(
                 cls, inst, namespace=ns)
+
+        if "Abstract" in cls.qualifiers:
+            raise CIMError(
+                CIM_ERR_FAILED,
+                _format("CreateInstance failed. Cannot instantiate abstract "
+                        "class {0!A} in Namespace {1!A}.",
+                        inst.classname, ns))
+
         try:
             self.instances[ns].append(inst)
         except KeyError:  # namespace ns does not exist. Create it
--- a/pywbem_mock/_providerdispatcher.py
+++ b/pywbem_mock/_providerdispatcher.py
@@ -40,7 +40,7 @@ except ImportError:  # py2
 
 from pywbem import CIMInstance, CIMInstanceName, CIMClass, CIMClassName, \
     CIMParameter, CIMError, CIM_ERR_NOT_FOUND, CIM_ERR_INVALID_PARAMETER, \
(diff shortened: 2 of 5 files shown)


andy-maier:
On Python 2.7, this fails due to issue #2745 for which a solution has been merged to master.

On Python 3.9, it fails due to a pylint issue:
```
tests/unittest/pywbem_mock/test_wbemconnection_mock.py:6213:47: W0613: Unused argument 'ns' (unused-argument)
```

pywbem/_mof_compiler.py
```diff
@@ -2394,6 +2394,15 @@ def CreateInstance(self, *args, **kwargs):
                                 IncludeQualifiers=True)
             inst.path = CIMInstanceName.from_instance(
                 cls, inst, namespace=ns)
+
+        if "Abstract" in cls.qualifiers:
+            raise CIMError(
+                CIM_ERR_FAILED,
+                _format("Abstract class cannot be instantiated. Creation class "
+                        "{0!A} of new instance in Namespace {1!A} "
+                        "of the CIM repository has Abstract qualifier.",
```
andy-maier, line 2403: Can we simplify this message text, e.g. to:
"Cannot instantiate abstract class {0!A} in namespace {1!A}"

pywbem_mock/_providerdispatcher.py
```diff
@@ -213,6 +213,14 @@ def CreateInstance(self, namespace, NewInstance):
                         "exist in namespace {1!A} of the CIM repository.",
                         NewInstance.classname, namespace))
 
+        if "Abstract" in creation_class.qualifiers:
+            raise CIMError(
+                CIM_ERR_FAILED,
+                _format("Abstract class cannot be instantiated. Creation class "
+                        "{0!A} of new instance in Namespace {1!A} "
+                        "of the CIM repository has Abstract qualifier.",
+                        NewInstance.classname, namespace))
```
andy-maier, line 222: Same comment o the message text as above.

andy-maier (changes requested):
Two minor comments.

pywbem_mock/_providerdispatcher.py
KSchopmeyer: Agreed. Done.  Added CreateInstance failed as lead-in

pywbem/_mof_compiler.py
KSchopmeyer: Done.

andy-maier (approved):

Issue #2766, opened 2021-09-04
Make destination name consistent with filter name


andy-maier:
Currently, the name property of listener destination instances is always created automatically, using this structure:

`"pywbemdestination:" {ownership} ":" {client_host_name} ":" {subscription_manager_id} ":" {guid}`

There are multiple issues with this:
- Users probably expect that for permanent filters, they have control over the Name property
- the guid makes it too long and unpredictable
- the client_host_name prevents working with owned filters from different client machines

The proposal is to make the way destinations are identfied, and how the name is created, to be consistent with filters:
- Introduce a destination_id parameter to add_listener_destinations() that can be specified only with owned destinations
- Introduce a name parameter to add_listener_destinations() that can be specified only with permanent destinations
- Remove components from the auto-generated name in case of owned destinations, so that its structure becomes:

`"pywbemdestination:" {subscription_manager_id} ":" {destination_id}`

The `add_listener_destinations()` needs to check for already existing destinations with that name to keep the name unique in a server.

andy-maier:
We agreed to proceed with the naming structure as proposed.
In addition:
* We remove the `add_destinations()` method and add `add_destination()` as a replacement.
* We deprecate `add_subscriptions()` but still implement it, and add `add_subscription()`.

andy-maier:
Second thoughts on deprecating `add_subscriptions()` and adding `add_subscription()`:
`add_subscriptions()` supports a single filter. On the destination side, it does support a single destination as well as a list of destinations, and even None which defaults to all existing owned destinations. I think there is no real need to change any of that.

Issue #2765, opened 2021-09-04
Simplify auto-generated name of owned filters


andy-maier:
The current structure of auto-generated names of owned filters is:

`"pywbemfilter:" {ownership} ":" {client_host_name} ":" {subscription_manager_id} ":" {filter_id} ":" {guid}`

There are multiple issues with this when using the name in pywbemcli:
- the guid makes it too long and unpredictable
- the client_host_name prevents working with owned filters from different client machines
- the ownership is not needed because the generated name is only used for owned filters

The proposal is to remove these components from the generated name, so that it becomes:

`"pywbemfilter:" {subscription_manager_id} ":" {filter_id}`

That makes the subscription_manager_id globally unique, and filter_id unique within the same subscription_manager_id.

The `add_filter()` needs to check for already existing filters with that name to keep the name unique in a server.

PR #2767, opened 2021-09-05
Simplified auto-generated Name of owned filter instances

andy-maier:
See commit message.

Note that this PR only addresses this for filter names (see issue #2765).
A separate PR will be made for destination names (see issue #2766).

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -149,6 +149,7 @@ The :ref:`Tutorial` section contains a tutorial about the subscription manager.
 """
 
 import re
+import warnings
 from socket import getfqdn
 import uuid
 import six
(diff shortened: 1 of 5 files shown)


pywbem/_subscription_manager.py
```diff
@@ -407,14 +399,24 @@ def add_server(self, server):
...
             FILTER_CLASSNAME, namespace=server.interop_ns)
 
         for inst in filter_insts:
             if re.match(filter_name_pattern, inst.path.keybindings['Name']):
                 self._owned_filters[server_id].append(inst)
+            if re.match(filter_name_old_pattern, inst.path.keybindings['Name']):
+                warnings.warn(
+                    _format(
+                        "Ignored owned indication filter with old name format "
+                        "in WBEM server: {0}", inst.path),
```
KSchopmeyer, line 418: Add "received from WBEM server" to text. Since this warning sort of appears out of the clear blue (i.e starting the manager)

KSchopmeyer (commented):
Made one comment.

pywbem/_subscription_manager.py
andy-maier: Changed to: "Ignored owned indication filter instance in WBEM server that has the old name format: {0}"

PR #2772, opened 2021-09-10
Changes for creating listener destinations in WBEMSubscriptionManager

andy-maier:
See commit message.

Particular review points might be:
* I left the `owned` parameter in the new `add_destination()` method, because I found it too implicit that the ownership type is determined just by the choice of `name` vs `destination_id` parameter, and for consistency with `add_filter()`.
**COMMENT/KS:** I agreed with you until I went through the code and API.  Now I begin to think that the owned parameter is redundant because it offers the user an option that does not really exist (mixing ownership and xxxx_id vs name)  By definition owned must be True if xxxx_id parameter exists and False if name parameter exists.  Having the owned parameter implies that the user could do owned=True/name=fred or owned=False/destination_id=fred so we have to add tests to disallow it..  I made comment about this in review.  I understand the reasoning about wanting to show the owned/permanent as a concept in the API but I think now that the redundancy overrides this.   Note that this is not a basic issue, the code works either way.  However, it is an issue we need to resolve to get on with pywbem tools prhttps://github.com/pywbem/pywbemtools/pull/987 **DISCUSSION**
* Other than discussed in the meeting, I had second thoughts on singularizing `add_subscriptions()` and left it as it is, because it works well with the other two add functions.  **COMMENT/KS:** Agree

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -59,8 +59,22 @@ Released: not yet
 * The new simplified format of the automatically generated 'Name' property of
   owned indication filters causes existing filters with the old format to
   be ignored and a Python warning of type 'pywbem.OldNameFilterWarning' will be
-  issued. Such filter instances need to be cleaned up by the user.
-  (related to issue #2765)
+  issued. Such owned filter instances need to be either removed as owned filters
+  with a prior version of pywbem, or as permanent filters with this version of
+  pywbem. (issue #2765)
+
+* Removed the 'pywbem.WBEMSubscriptionManager.add_listener_destinations()'
+  method, because the new naming approach for listener destinations requires
+  either a name or an ID and that does not work well with supporting multiple
+  destinations in one method call. Use the new 'add_destination()' method
+  instead. (issue #2766)
+
+* The new simplified format of the automatically generated 'Name' property of
+  owned listener destinations causes existing destinations with the old format
+  to be ignored and a Python warning of type 'pywbem.OldNameDestinationWarning'
+  will be issued. Such owned destination instances need to be either removed as
+  owned destinations with a prior version of pywbem, or as permanent
+  destinations with this version of pywbem. (issue #2766)
 
 **Deprecations:**
 
@@ -226,6 +240,23 @@ Released: not yet
   The new `ToleratedSchemaIssueWarning` is expected to be used where the
   MOF compiler or code detects issues in the CIM Schema that are either
   tolerated or corrected.
+
+* Added a 'pywbem.WBEMSubscriptionManager.add_destination()' method
+  that makes the way the 'Name' property is created for listener destination
+  intances consistent with how that is now done for indication filters: There is
+  a new parameter 'name' that directly sets the 'Name' property for permanent
+  destinations, and a new parameter 'destination_id' that is used for creating
+  the 'Name' property for owned destinations. The format of the generated 'Name'
+  property has been changed from:
+  ``"pywbemdestination:" {ownership} ":" {client_host} ":" {submgr_id} ":" {guid}``
+  to:
+  ``"pywbemdestination:" {submgr_id} ":" {destination_id}``.
+  The client host was removed in order to allow different client systems to be
+  used. The ownership was removed because destinations with an auto-generated
+  Name are always owned. The GUID was removed to make the name predictable and
+  the uniqueness it attempted to guarantee is now achieved by rejecting the
+  creation of destinations with the same name. Overall, this change makes the
+  name much more suitable for use in CLI tools such as pywbemcli. (issue #2766)
 
 **Cleanup:**
 
@@ -2477,7 +2508,7 @@ This version contains all fixes up to pywbem 0.12.6.
 
 * Add tests for the `WBEMSubscriptionManager` class using the pywbem mock
   support.  This involved
-  changing the tests for the `WBEMServer` class using pywbem_mock because the the
+  changing the tests for the `WBEMServer` class using pywbem_mock because the
   WBEMSubscriptionManager class depends on the existence of the classes and
   instances that support the pywbem WbemServer class existing in the WBEM
   server.  A new file (wbemserver_mock.py) was added to the tests
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -114,7 +114,8 @@ is assumed to exist somewhere and is identified by its URL::
     server_id = sub_mgr.add_server(server)
 
     # Add a listener destination in the server:
-    dest_inst = sub_mgr.add_listener_destinations(server_id, listener_url)
+    dest_inst = sub_mgr.add_destination(
+        server_id, listener_url, owned=True, destination_id='id1')
 
     # Subscribe to a static filter of a given name:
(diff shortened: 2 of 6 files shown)


docs/changes.rst
```diff
@@ -59,8 +59,22 @@ Released: not yet
...
   owned indication filters causes existing filters with the old format to
   be ignored and a Python warning of type 'pywbem.OldNameFilterWarning' will be
-  issued. Such filter instances need to be cleaned up by the user.
-  (related to issue #2765)
+  issued. Such owned filter instances need to be either removed as owned filters
+  with a prior version of pywbem, or as permanent filters with this version of
+  pywbem. (issue #2765)
+
+* Removed the 'pywbem.WBEMSubscriptionManager.add_listener_destinations()'
+  method, because the the new naming approach for listener destinations requires
```
KSchopmeyer, line 67: work "the" is repeated

pywbem/_subscription_manager.py
```diff
@@ -389,14 +389,25 @@ def add_server(self, server):
...
         dest_insts = server.conn.EnumerateInstances(
             DESTINATION_CLASSNAME, namespace=server.interop_ns)
 
         for inst in dest_insts:
             if re.match(dest_name_pattern, inst.path.keybindings['Name']):
                 self._owned_destinations[server_id].append(inst)
+            if re.match(dest_name_old_pattern, inst.path.keybindings['Name']):
+                warnings.warn(
+                    _format(
+                        "Ignored owned listener destination instance in WBEM "
```
KSchopmeyer, line 407: I have one possible issue with this message (applies to the filter also).  The filter is not completely ignored, it appears in the list of all destinations.  The message reads as if is completely ignored.

pywbem/_subscription_manager.py
```diff
@@ -565,17 +590,38 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True,
               :term:`RFC6874`, supporting ``-`` (minus) for the delimiter
               before the zone ID string, as an additional choice to ``%25``.
 
-            Note that the port is required in listener URLs.
+            Note that the port is required in listener URL.
```
KSchopmeyer, line 593: Lets take off the words "Note that".

pywbem/_subscription_manager.py
```diff
@@ -565,17 +590,38 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True,
...
 
             See :class:`~pywbem.WBEMConnection` for examples of valid URLs,
             with the caveat that the port in server URLs is optional.
 
           owned (:class:`py:bool`):
             Defines the ownership type of the created listener destination
-            instances: If `True`, they will be owned. Otherwise, they will be
+            instance: If `True`, it will be owned. Otherwise, it will be
```
KSchopmeyer, line 600: In the line a above, why is it "ownership type". Isn't it just ownership?

pywbem/_subscription_manager.py
```diff
@@ -595,38 +641,46 @@ def add_listener_destinations(self, server_id, listener_urls, owned=True,
...
-            dest_insts = []
-            for listener_url in listener_urls:
-                new_dest_insts = self.add_listener_destinations(
-                    server_id, listener_url, owned)
-                dest_insts.extend(new_dest_insts)
-            return dest_insts
-
-        # Here, the variable will be a single list item.
-        listener_url = listener_urls
+        if owned:
```
KSchopmeyer, line 657: This is a question right now. Now that we have explicitly destination_id and name, do we still need the owned parameter for either add_destination or add_filter?  Isn't the ownership clearly defined by the existence of destination_id or name parameters? We do need it for add_subscriptions I agree.

I understand why it seems logical since that clarifies the concept of ownership clearly in the code but it really is redundant and causes an extra set of tests to be required in the code to confirm that the user correctly entered the combination of destination_id, name, and owned.

pywbem/_subscription_manager.py
```diff
@@ -870,28 +926,29 @@ def add_filter(self, server_id, source_namespaces, query,
 
         # server_id is validated in _create_...() method.
 
-        if filter_id is None and name is None:
-            raise ValueError("The filter_id and name parameters are both "
-                             "None, but exactly one of them must be specified")
-        if filter_id is not None:
+        if owned:
```
KSchopmeyer, line 929: Same question about owned parameter as for add_destination.

pywbem/_subscription_manager.py
```diff
@@ -1259,8 +1316,31 @@ def _create_destination(self, server_id, dest_url, owned,
             URL specify the target location to be used by the WBEM server.
 
           owned (:class:`py:bool`):
-            Defines whether or not the created instance is *owned* by the
-            subscription manager.
+            Defines the ownership type of the created listener destination
```
KSchopmeyer, line 1319: Shouldn't this be ownership rather than ownership type.

KSchopmeyer (commented):
This time I made several comments including one that may lead to discussion.

pywbem/_subscription_manager.py
andy-maier: The ownership is clearly an implication of the use of the ID parameter.
However, the issue is that it is not explicit. One of the Python mantras is: "explicit is better than implicit".

andy-maier:
We discussed the discussion point about the redundant parameters, and decide to leave them in as in the current version of the CR.

docs/changes.rst
andy-maier: DONE

pywbem/_subscription_manager.py
andy-maier: DONE - Changed to:
```
                        "Ignored listener destination instance when detecting "
                        "owned instances because it has the old name format: "
                        "{0}",
```

pywbem/_subscription_manager.py
andy-maier: DONE as proposed.

I used that phrase because the syntax already shows the port as required, so I thought this sentence just needs to point that rule out, but it does not represent the rule as such.

However, turning that into a rule even when it repeats the rule shown in the syntax is not a mistake, certainly.

pywbem/_subscription_manager.py
andy-maier: NO CHANGE

First, I double checked the docs and source code, and it turns out that so far, we consistently use "ownership type" or "types of ownership", but not just "ownership" (except for two places in the change log). All other occurrences of this are in the `_subscription_manager.py` module.

"Ownership" in the sense of the word only applies to owned instances, but not to permanent instances.
I would like to have a term that is neutral and does not imply just one of the two possible states. Maybe "ownership type" is not optimal for that, but just "ownership" I think is not neutral enough.

pywbem/_subscription_manager.py
andy-maier: We discussed this in our call this week and agreed to go with the explicit "owned" parameter even if it is redundant.

pywbem/_subscription_manager.py
andy-maier: see response to previous occurrence of this comment.

pywbem/_subscription_manager.py
andy-maier: see response to previous comment

PR #2784, opened 2021-10-21
Fixes issue # 2782, handling of add_destination duplicates.

KSchopmeyer:
Wating on priority prs #2795

Moves the test for duplicated name property to before the test for
duplicated URL and adds PersistenceType as test for equality between new
and existing owned destination instances.

Fixed test_subscriptionmanager.py to account for possible multiple
destination definitions in test since the code has been changed to now
allow only a single destination for each call.

NOTE: It would not bother me if we changed the code to return ValueError if the URL matched or even just to
remove the test for URL code.  I chose returning the existing instance as one of 3 generally almost equal
choices.

Added tests specifically for duplicated destination destination_id and
for returning instance if url and persistence type match.

### DIFF
--- a/pywbem/_subscription_manager.py
+++ b/pywbem/_subscription_manager.py
@@ -180,8 +180,8 @@ def validate_persistence_type(pt):
     Validate persistence type parameter pt as string possible
     values::transient" or "permanent" and convert to corresponding integer
     value
-    Returns integer value or None for the PersistenceType property based on the
-    input string.
+    Returns integer value or None for the `PersistenceType` property based on
+    the input string.
     """
     if pt is None:
         return None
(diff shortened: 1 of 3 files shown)


pywbem/_subscription_manager.py
```diff
@@ -554,7 +554,8 @@ def add_destination(self, server_id, listener_url, owned=True,
           destinations.
 
           If an owned listener destination instance for the specified listener
-          URL already exists, it is returned without creating a new instance.
+          URL and 'PersistenceType' already exists, it is returned without
```
andy-maier, line 557: Use backquotes around property names.

pywbem/_subscription_manager.py
```diff
@@ -1299,9 +1300,12 @@ def _create_destination(self, server_id, dest_url, owned, destination_id,
         In order to catch any changes the server applies, the instance is
         retrieved again using the instance path returned by instance creation.
 
-        If an instance is found in the server with the same value of the
-        `Destination` property, no new instance is created and the existing
-        is returned.
+        If the request is owned and an owned instance is found in the server
+        with the same value of the `Destination` and 'PersistenceType'
```
andy-maier, line 1304: Use backquotes around property names.

pywbem/_subscription_manager.py
```diff
@@ -1299,9 +1300,12 @@ def _create_destination(self, server_id, dest_url, owned, destination_id,
...
+        properties, no new instance is created and the existing is returned. A
+        client may detect that an existing instance was returned by comparing
+        the destination_id component of the requested 'Name' property with the
+        destination_id of the submitted 'Name' property.
```
andy-maier, line 1308: Use backquotes around property names.

andy-maier (changes requested):

pywbem/_subscription_manager.py

pywbem/_subscription_manager.py

pywbem/_subscription_manager.py

andy-maier (approved):

PR #2785, opened 2021-10-22
Add some properties to Subscription provider instances

KSchopmeyer:
First commit:

1. Added description property to each of the providers just to clarify
that these are from the Mock

2. Add a couple of Null properties just to expand the pywbemcli test.

3. Extended one property in subscription to add it NULL even if not
provided by client as a logical server based property.

These are all minor changes to the mock providers and have been tested
against pywbemcli.

The second commit:

Corrects an issue where the replacement of key properties fails if the property being tested has None as a value.  This only occurs if an instance is installed outside of the subscription manager (ex. directly compiled into the the mock environment)

### DIFF
--- a/pywbem_mock/_subscriptionproviders.py
+++ b/pywbem_mock/_subscriptionproviders.py
@@ -19,7 +19,7 @@
 #
 
 """
-This module implements user providerw for the classews required to create and
+This module implements user providerw for the classes required to create and
 manage CIM indications subscriptions including: CIM_ListenerDestinationCIMXML,
 CIM_IndicationFilter, and CIM_IndicationSubscription.  These providers are
 packaged as single file because they are all required together.
@@ -45,7 +45,8 @@ is included in a subscription.
 import six
 
 from pywbem import CIMError, CIM_ERR_NOT_SUPPORTED, \
-    CIM_ERR_INVALID_PARAMETER, CIM_ERR_FAILED, Uint16, CIMDateTime
+    CIM_ERR_INVALID_PARAMETER, CIM_ERR_FAILED, Uint16, CIMDateTime, \
+    CIMProperty
 from pywbem._utils import _format
 
 from ._instancewriteprovider import InstanceWriteProvider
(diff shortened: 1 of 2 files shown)


pywbem_mock/_subscriptionproviders.py
```diff
@@ -191,20 +205,29 @@ def fix_key_properties(self, new_instance):
...
-                    new_instance[pname].lower != test_value.lower:
+            """
+            Replace property pname with either replacement if replacement not
+            None or test_value if property does not exist, is None, or not same
+            value as test_value
+            """
+            if pname not in new_instance or new_instance[pname] is None:
+                new_instance[pname] = replacement or test_value
+            elif new_instance[pname].lower != test_value.lower:
                 new_instance[pname] = replacement or test_value
```
andy-maier, line 227: These two bodies could be combined into a single if.

pywbem_mock/_subscriptionproviders.py
```diff
@@ -191,20 +205,29 @@ def fix_key_properties(self, new_instance):
         """
         Fix the key properties that are common to many classes (SystemName,
         CreationClassName, SystemCreateionClassName) since these may be
-        either not provided with the new instance or incorrect for this
-        enviornment
+        either not provided,None, or incorrect value for this environment with
```
andy-maier, line 208: add space after comma

andy-maier (changes requested):
Two minor comments, plus rebase.

pywbem_mock/_subscriptionproviders.py
KSchopmeyer: DONE

pywbem_mock/_subscriptionproviders.py
KSchopmeyer: Done

KSchopmeyer:
In the two failure cases macos and windows python 3.5, it retrieves more-itertools 8.11.0 for python 3.5 where it should be retrieving 8.10.1.  It is like the statement:

dev-requirements.txt
27:more-itertools>=4.0.0,!=8.11.0; python_version < '3.6'

is not working.

andy-maier (approved):

PR #2790, opened 2021-11-01
Issue between Sphinx 1.85 and docutils 0.18

KSchopmeyer:
Includes commit for pr #2791,  update to safety checks.

Docutils created incompatible change with version 0.18. Sphinx fixing
this for Sphinx V 4,3.0 but not for older versions. Since we use Sphinx 1.8.5 for python < 3.5 we need to use docutils < 0.18
for these tests.

### DIFF
--- a/rtd-requirements.txt
+++ b/rtd-requirements.txt
@@ -9,7 +9,9 @@
 # Keep in sync with dev-requirements.txt
 
 Sphinx>=1.7.6,<2.0.0; python_version <= '3.4'
-Sphinx>=3.0.4,<4.0.0; python_version >= '3.5'
+Sphinx>=3.5.4, !=4.0.0; python_version >= '3.5'
+# Issue #2787 docutils v=0.18/Sphinx < python < 3.5 incompatibility
+docutils<0.18; python_version <= '3.4'
 sphinx-git>=10.1.1
 GitPython>=2.1.1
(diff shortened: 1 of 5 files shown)


rtd-requirements.txt
```diff
@@ -10,6 +10,8 @@
 
 Sphinx>=1.7.6,<2.0.0; python_version <= '3.4'
 Sphinx>=3.0.4,<4.0.0; python_version >= '3.5'
```
andy-maier, line 12: we should replace the pinning to <4.0.0 to !=4.0.0

andy-maier: Update: I now understand why the test run with minimum package levels on py39 does not fail: While it picks up Sphinx 3.0.4 which does not yet have the docutils fix, it also picks up the minimum version of docutils which is 0.13. In a real environment however, the combination of Sphinx 3.x and docutils >=0.18 could be encountered and then it would fail. So I suggest requiring Sphinx >=4.0.1 and probably also docutils >=0.18.

KSchopmeyer: I think that the issue is more than even that.  I do not know why Sphinx builds documents with any released version of Sphinx since the latest info on their site is that they will fix Sphinx to be compatible with docutils for Sphinx 4.3.0.  I suggest that we make docutils< 0.18 mandatory for all Sphinx versions until this settles down.

PR #2795, opened 2021-11-11
Set version limit on more-itertools in dev-requirements.txt

KSchopmeyer:
Fix issue #2796 that does not limit the version of more-itertools which causes test failures and fix an issue with use of >= in minimum-constraints.

This limits the version of more-itertools version 8.11.0 to python > 3.6

NOTE: I assume that the developers will correct the issue and either rerelease 8.11.0 or release an 8.11.1 since there is already a PR defined.

Priority because issue # 2796 it breaks tests.

Also fixes issue # 2792 where there is a invalid comparison error in minimum-constraints.txt.  That, however, was not a test breaking issue.

NOTE: more-tools has a PR up to fix this problem (https://github.com/more-itertools/more-itertools/pull/579) but if we assume that the fix in by creating a new version (several users have requested that they reissue 8.11.0) we will have another change to modify dev-requirements but this PR needed just to get our tests running again.

COMMENT: I marked this one for rollback but not sure it is really requredbecause:

1. This is tests only because it is all tied to pytest. The change is only in dev-requirements for the fix that causes test failure.
2. This is probably going to change again soon because more-tools should be issuing fix that would modify the change to dev-requirements.

NOTE: This also impacts pytest-easy-server since it uses pytest.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -23,6 +23,10 @@ pytest-cov>=2.7.0; python_version == '2.7' or python_version >= '3.5'
 git+https://github.com/andy-maier/coveralls-python.git@andy/add-py27#egg=coveralls; python_version == '2.7'
 coveralls>=2.1.2,<3.0.0; python_version >= '3.5'
 
+# version 8.11 requires python 3.6, See issue #2796
+more-itertools>=4.0.0,!=8.11.0; python_version < '3.6'
+more-itertools>=4.0.0; python_version >= '3.6'
+
 # Safety CI by pyup.io
 # safety 1.9.0 removed support for Python 2.7 and 3.4 (and now also enforces that)
 safety>=1.8.7,<1.9.0; python_version <= '3.4'
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -136,6 +136,8 @@ requests-mock==1.6.0
 requests-toolbelt==0.8.0
 yagot==0.5.0
 
+more-itertools==4.0.0
+
 # Easy-server packages
 easy-vault==0.7.0
 easy-server==0.8.0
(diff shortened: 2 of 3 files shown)


dev-requirements.txt
```diff
@@ -23,6 +23,9 @@ pytest-cov>=2.7.0; python_version == '2.7' or python_version >= '3.5'
 git+https://github.com/andy-maier/coveralls-python.git@andy/add-py27#egg=coveralls; python_version == '2.7'
 coveralls>=2.1.2,<3.0.0; python_version >= '3.5'
 
+# version 8.11 requires python 3.6, See issue #2796
+more-itertools<=8.10.0; python_version <= '3.5'
```
andy-maier, line 27: I think we should be more brave and just exclude more-itertools 8.11.0 instead of pinning it, in expectation of them fixing that.

In addition, I suggest using a minimum version. In other projects, I used 4.0.0 as the minimum version.

This then results in:
```
more-itertools>=4.0.0,!=8.11.0; python_version < '3.6'
more-itertools>=4.0.0; python_version >= '3.6'
```

minimum-constraints.txt
```diff
@@ -177,7 +177,7 @@ tox==2.5.0
 
 # Sphinx (no imports, invoked via sphinx-build script):
 Sphinx==1.7.6; python_version <= '3.4'
-Sphinx>= 3.5.4; python_version >= '3.5'
+Sphinx==3.5.4; python_version >= '3.5'
```
andy-maier, line 180: I suggest adding the same minimum version for more-itertools as in dev-requirements.txt:
```
more-itertools==4.0.0
```

andy-maier (changes requested):
See comments

minimum-constraints.txt
KSchopmeyer: DONE

dev-requirements.txt
KSchopmeyer: DONE

andy-maier (approved):

PR #2815, opened 2021-12-02
Added support for Python 3.10

andy-maier:


test-requirements.txt
```diff
@@ -38,7 +39,8 @@ requests-mock>=1.6.0
 requests-toolbelt>=0.8.0
 yagot>=0.5.0
 importlib-metadata<1,>=0.12; python_version <= '3.7'
-pytz>=2016.10
+# pytz before 2019.1 fails on Python 3.10 because it uses collections.Mapping
+pytz>=2019.1
```
KSchopmeyer, line 43: Why not retain the pytx>=2.16.10 with python < 3.10

KSchopmeyer (commented):
Only made one comment and approved with that comment

test-requirements.txt
andy-maier: DONE

PR #2823, opened 2021-12-21
Improved verbosity of namespace creation/deletion

andy-maier:
This is needed for the namespace creation tests in pywbemtools (PR https://github.com/pywbem/pywbemtools/pull/1114)

**DISCUSSION:**

There are two options on how to deal with this PR:

* Release pywbem 1.4.0 with this change quickly and have the namespace creation tests in pywbemtools 1.0.0
* Defer this change until after pywbem 1.4.0 and defer the namespace creation tests in pywbemtools until after pywbemtools 1.0.0
  * Then we need to decide what we do with the changes in the mock scripts in pywbemtools

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -33,6 +33,11 @@ Released: not yet
 
 **Enhancements:**
 
+* Improved verbosity of namespace creation and deletion: Added optional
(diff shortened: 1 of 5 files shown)


andy-maier:
We discussed the options in our call last tuesday and decided to quickly release pywbem 1.4.0.

PR #2837, opened 2022-01-03
Docs: Fixed index in local docs build

andy-maier:


KSchopmeyer (approved):

KSchopmeyer:
Tested on ubuntu and the index is in place.

PR #2863, opened 2022-03-13
Added package dep tools; Fixed dep versions; Fixed setuptools 61.0.0 issue

andy-maier:


andy-maier:
Merged without review since several others were on top.

PR #2876, opened 2022-03-28
Coverity corrections

tasleson:
Running coverity on version 0.15.0 which is the latest packaged version for fedora is reporting some errors.  I believe this PR addresses some potential issues that should be considered for inclusion.

### DIFF
--- a/pywbem/_cim_obj.py
+++ b/pywbem/_cim_obj.py
@@ -2052,6 +2052,7 @@ class CIMInstanceName(_CIMComparisonMixin, SlottedPickleMixin):
                 _format("WBEM URI has an invalid format for its keybindings: "
                         "{0!A}", keybindings_str))
 
+        kb_assigns = []
         if m.group(1):
             kb_assigns = [m.group(1)]
 
--- a/pywbem/_recorder.py
+++ b/pywbem/_recorder.py
@@ -618,8 +618,8 @@ class LogOperationRecorder(BaseOperationRecorder):
             # Format the 'summary' and 'paths' detail_levels
             if self.api_detail_level == 'summary':  # pylint: disable=R1705
                 if isinstance(ret, list):
-                    if ret:
-                        ret_type = type(ret[0]).__name__ if ret else ""
+                    if len(ret) > 0:
+                        ret_type = type(ret[0]).__name__
                         return _format("list of {0}; count={1}",
                                        ret_type, len(ret))
                     return "Empty"


tasleson:
> PS: The nocaselist/nocasedict packages were introduced in pywbem 1.0.0, and the latest 0.x version is 0.17.6. So in theory you should be able to upgrade pywbem in fedora to 0.17.6. Do you know of anything that would prevent that (other than work)?

I'm not sure, I'll take a look and see if we can move up to 0.17.6, thanks.

pywbem/_recorder.py
```diff
@@ -618,11 +618,9 @@ def format_result(ret, max_len):
...
             if self.api_detail_level == 'summary':  # pylint: disable=R1705
                 if isinstance(ret, list):
-                    if ret:
-                        ret_type = type(ret[0]).__name__ if ret else ""
-                        return _format("list of {0}; count={1}",
-                                       ret_type, len(ret))
-                    return "Empty"
+                    ret_type = type(ret[0]).__name__ if ret else ""
+                    return _format("list of {0}; count={1}",
+                                   ret_type, len(ret))
```
andy-maier, line 623: While the change does not harm (it returns the "list of ..." text with count 0 in case of an empty list instead of "Empty"), I would like to point out that the coverity tool's explanation: "If the expression isinstance(ret, list) evaluates to True, the expression if ret: will evaluate to False causing those lines to
to be dead." is incorrect. If "ret" is a list, "if ret" is True if the list is non-empty, and otherwise False.

pywbem/_cim_obj.py
```diff
@@ -2052,6 +2052,7 @@ def from_wbem_uri(wbem_uri):
                 _format("WBEM URI has an invalid format for its keybindings: "
                         "{0!A}", keybindings_str))
 
+        kb_assigns = []
```
andy-maier, line 2055: While this does not hurt, the matter of fact is that the regexp in WBEM_URI_KEYBINDINGS_REGEXP has a mandatory first matching group. So if that group does not match, the "if m is None:" check a few lines above would have caused an exception to be raised, so if the code gets here it is guaranteed that the first group was there. I think coverity does not analyze the regexp, so it cannot come to that conclusion.

What is really not needed because of the guaranteed presence of the first group is the "if m.group(1):". With that present though in the old code, I can understand why coverity misses the initialization.

If we take the guaranteed first group into account, the code could become:

```
        assert m.group(1)
        kb_assigns = [m.group(1)]
```

However, the initialization is also fine, and I don't see a need to change the proposed solution.

andy-maier (changes requested):
See comments

pywbem/_recorder.py
tasleson: Yes, you're correct

tasleson: I made  slight modification to your proposed change, as I think this this makes the intention more obvious ...

```python
                if isinstance(ret, list):
                    if len(ret) > 0:
                        ret_type = type(ret[0]).__name__
                        return _format("list of {0}; count={1}",
                                       ret_type, len(ret))
                    return "Empty"
```

tasleson:
@andy-maier Update?

andy-maier (approved):

andy-maier:
@tasleson Hi Tony. Sorry this took so long. We'll merge it into 1.7.0.

PR #2879, opened 2022-04-03
Fixed and improved handling of requests/urllib3 exceptions and retries

andy-maier:

This PR requires thorough review and testing in situations with broken connections.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -23,6 +23,10 @@ Released: not yet
 
 **Incompatible changes:**
 
+* Exceeding the 'WBEMConnection' timeout now results in raising
+  'pywbem.TimeoutError' in some cases where previously 'pywbem.ConnectionError'
+  was raised. (issue #2853)
+
 **Deprecations:**
 
 **Bug fixes:**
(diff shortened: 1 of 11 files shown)


pywbem/_cim_http.py
```diff
@@ -48,12 +49,19 @@
...
     HeaderParseError  # pylint: disable=redefined-builtin
 from ._utils import _ensure_unicode, _ensure_bytes, _format
+from ._warnings import RequestExceptionWarning
 
 __all__ = []
 
 _ON_RTD = os.environ.get('READTHEDOCS', None) == 'True'
 
-HTTP_CONNECT_TIMEOUT = 10        # HTTP connect timeout in seconds
+DEBUG_EXCEPTIONS = False  # Debug the behavior of requests/urllib3 exceptions
```
KSchopmeyer, line 58: Is it ever intended that a user use this to debug issues. If so may need more explanation.

KSchopmeyer:
Testes against local Pegasus, and Pegasus in container.  The only comment is that the connection timeout takes about 30 seconds now (tested with docker pause pegasus.  Not sure exactly what state docker is in when pause occurs Apparently in pauses the internal processes) but I assme it is accepting the connection and then just not passing it to Pegasus or pegasus/TCP, etc. processes are paused. 

When container started but Pegasus not started, delay is about 800 ms.

When container running delay is about 800 ms also.   NOTE: That enumerate all classes also takes about 800- ms.

One thing I am confused about in the test with container paused is that the output message always says that the read timeout=10 irrespective of what we set it to

```

File: docs/changes.rst [hunk 2]
```diff
@@ -54,13 +58,33 @@ Released: not yet
 * Fixed that the added setup.py commands (test, leaktest, installtest) were not
   displayed. They are now displayed at verbosity level 1 (using '-v').
 
+* Fixed that the sending of CIM request messages was retried in case no
+  response was received within the timeout, or in case of the 'status' and
+  'other' conditions of the "requests" package.
+  Retries are now only performed during connection setup, but not for sending
```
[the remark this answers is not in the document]
andy-maier: DONE - improved that by changing the entry to:
```
* Previously, the sending of CIM request messages was retried in case no
  response was received within the timeout. This could potentially have resulted
  in executing operations multipe times. That is an issue for non-idempotent
  operations such as instance creation/deletion or method invocation. Fixed
  that by retrying now only during connection setup, but not for the sending
  of CIM request messages. (issue #2853)
```

pywbem/_cim_http.py
andy-maier: Not intended, it is only for our use. I added a statement clarifying that.

andy-maier:
On the read timeout=10: See testcase 2 in the design description that is part of this PR. The urllib3 code uses the connect timeout (which we set to 9.999 so it is shown rounded to 10) when it performs the TLS handshake, which at some deeper level is a socket read, so the deeper level reports a read time out with the value of the connect timeout. It is not something we can change in our code.

Issue #2841, opened 2022-01-10
Operation IterEnumerateInstances Returns host name in path


KSchopmeyer:
The EnumerateInstances, . etc. does not set the host into the path elements in the returned instances.  However, the IterEnumerateInstances does set the host name into each path.

This is a difference between the operations (probably applies to Associators and References also) that we do document but in just one sentence.

**DISCUSSION:** Is this operationally correct and should we document it better?  Or should we consider modifying the code to make this completely compatible with the Pull and Original operations?  Remember that the iter... is our own pywbem operation and we could therefore justify this difference if properly documented but it does represent a difference from  what is returned in the request operations that we use

**Proposal:** ~~We are tending to think that this would be more logical if we added the host to the enumerateinstances, etc. but we should go back and see if there is any documentation in the commits as to why we added it to the iterenumerateinstances.~~  Update. Based on the information in other comments in this issue, it is proposed to make only  one change, Modify the documentation to note that we are always returning host name in the InterEnumerateInstances and IterEnumerateInstancePaths response even if the original non-pull requests were made which do not include this information in the response.  It is also noted that we actually fill in empty namespace value in the response path but probably should not have done that to more clearly reflect what the server returned, even if the server did not return this value.  However, I propose that we leave this alone (KS)

Therefore, the only change proposed is a minor one to documentation

KSchopmeyer:
I did not find anything in the commit documentation.

andy-maier:
The cases where instance paths are returned, are:
- instance path returned by enumerate instances (traditional / pull / iter)
- instance path returned by enumerate instance paths (traditional / pull / iter)

KSchopmeyer:
From the DMTF specifications:

* EnumerateInstanceNames - Per dsp0200, a list  the entity <InstanceName> is to be returned from this request.  DSP0201 defines this as class and keybindings only.  That section is shown below:
    5.3.4.10 INSTANCENAME
    946 The INSTANCENAME element defines the location of a CIM instance within a namespace (it is referred
    to in DSP0004 as a model path). It comprises a class name and key-binding information.
    947
    948
    949 If the class has a single key property, a single KEYVALUE or VALUE.REFERENCE subelement may be
    used to describe the (necessarily) unique key value without a key name. Alternatively, a single
    KEYBINDING subelement may be used instead.
    950 If the class has more than one key property, a KEYBINDING subelement shall appear for each key.
[...]

PR #2883, opened 2022-04-16
fixes issue #2841 - Clarify reason host in interenumerateinstance ret…

KSchopmeyer:
Clarifies why the host name is always included in the return objects for
iterEnumerateInstances and iterEnumerateInstancePaths. There were no
code changes.

### DIFF
--- a/pywbem/_cim_operations.py
+++ b/pywbem/_cim_operations.py
@@ -4674,8 +4674,8 @@ class WBEMConnection(object):  # pylint: disable=too-many-instance-attributes
         This method uses the corresponding pull operations if supported by the
         WBEM server or otherwise the corresponding traditional operation.
         This method is an alternative to using the pull operations directly,
-        that frees the user of having to know whether the WBEM server supports
-        pull operations.
+        that frees the user of having to know whether or not the WBEM server
+        supports pull operations.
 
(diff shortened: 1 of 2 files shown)


pywbem/_cim_operations.py
```diff
@@ -4909,8 +4910,11 @@ class are to be included in the returned instances, as follows:
 
           :term:`py:generator` iterating :class:`~pywbem.CIMInstance`:
           A generator object that iterates the resulting CIM instances.
-          These instances include an instance path that has its host and
-          namespace components set.
+          These instances include an instance path that always has its host and
+          namespace components. When this method uses the
+          EnumerateInstances operation, the host name is inserted into
+          the response from the WBEMConnection.host attribute because DMTF
+          EnumerateInstances does not return host name in the response.
```
andy-maier, line 4917: This explains how the host is always set but not why. Note that the change log announces the change would explain why.

andy-maier (changes requested):
See comment

pywbem/_cim_operations.py
KSchopmeyer: Made shorter but makes sense to explain this since even we missed it.

KSchopmeyer: DONE

PR #2891, opened 2022-05-31
Discouraged installation via 'setup.py install'

andy-maier:


docs/changes.rst
```diff
@@ -74,6 +78,9 @@ Released: not yet
   of certifi on Python 2.7 that properly declares that it requires
   Python >=3.6.
 
+* Added a note to the install section of the documentation that installation
```
KSchopmeyer, line 81: They deprecated it but we disabled it. If it is really just deprecated at this point should we be issuing a warning message right now rather than stopping the usage.   Since that is messy, would it make sense to add to note how it can be enabled if a user requires it. (i.es set the permissions of the file to executable.

KSchopmeyer: No issue. Ignore comment

KSchopmeyer: Approved.

PR #2903, opened 2022-06-20
Support for Python 3.11

andy-maier:
Ready for review.


andy-maier:
I assume that Karl's approval for Python 3.11 support in pywbemtools also applies here.

PR #2909, opened 2022-07-18
Extend inter-namespace  mock associations to be bidirectional

KSchopmeyer:
Waiting pr #2916, the latest security issues so should fail if resubmitted before that issue is fixed

Extends creation, modification, and deletion of association instances in
pywbem_mock where the instance reference defines cross namespace
associations to build a corresponding instance of the association in
each namespace.

See Commit message for details

NOTE: This pr also fixed some pylint issues that apparently popped up with new version of pylint and added test for existence of association endpoints when creating/modifying new instances in pywbem_mock
Prviously it would pass end points path definitions that were invalid.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -35,6 +35,12 @@ Released: not yet
   cause name errors in your code, if you were using them.
   (related to issue #2888)
 
+* The pywbem_mock default instance writer (pywbem_mock/_instancewriteprovider.py)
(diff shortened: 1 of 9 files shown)


andy-maier (changes requested):
Please rebase on latest master, since the PR has fixes for flake8 that are already in master.

andy-maier (approved):

Issue #2918, opened 2022-09-08
pywbem_mock namespace providers allows duplicate namespace names


KSchopmeyer:
The namespace provider allows duplicate namespace names because it only tests for duplicate path, not for duplicate of the Name property  of all instances of the CIMNamespace class.

PR #2927, opened 2022-09-27
Fix issues with namespace provider.

KSchopmeyer:
This fixes two issues:

1. Error caused by bad edit in pr# 2913
2. Fix issue #2918 - limitation where duplicate namespaces could be created.

The fix consists of modifying the test for existence of the namespace being created in the namespaceprovider rather than just the instance path of the new instance.  A method was added to baseprovider to accomplish this change and used in the namespace provider.

The test modifies and extends some of the tests for namespace provider installation

See commit for more details

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -165,6 +165,10 @@ Released: not yet
 * Added security issues 50748, 50571, 50664, 50663, 50892, 50885, 50886 to
   Makefile ignore list of new security issue August and September 2022.
 
+* Fixed issue with mock namespace provider that would acreate the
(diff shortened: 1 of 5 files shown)


andy-maier (changes requested):
Please add a change log entry.

KSchopmeyer:
Whoops.  I did not add one on purpose because the original was just a fix to a bug in another pr for this release. But then I realized that it also fixed issue #2918 but forgot to add doc.

changes.rst comment added.

andy-maier (approved):

PR #2928, opened 2022-10-02
Changed Actions schedule to 00:00

andy-maier:
No review needed.


andy-maier:
It fails due to new safety issues, which will be fixed in a subsequent PR. Merging.

PR #2929, opened 2022-10-02
Ignore security  issues and fix new issue with flake8 2022

KSchopmeyer:
2 separate commits.

* Modifies versions of importlib-metadata package because of failure with python 3.7 in flak8.  Issue #2931

* Set safety ignore for pygments.  Set to same values in pywbemtools  dev-requirements.txt and minimum-constraints.txt.

Packaged them into two commits but in the same pr because both cause github CI failure so we need both to get tests running.

Set rollback needed because they cause CI failure that would apply to previous versions of pywbem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,7 +263,8 @@ py_test_files := \
 # - For packages that are direct or indirect development or test requirements,
 #   upgrade the package version only if possible w.r.t. the supported
 #   environments and add to the ignore list otherwise.
-# Current safety ignore list, with reasons:
+# Current safety ignore list, with reasons why marked ignore rather than modify
+# the requirements files:
 # Runtime dependencies:
 # - 38100: PyYAML on py34 cannot be upgraded; no issue since PyYAML FullLoader is not used
 # - 38834: urllib3 on py34 cannot be upgraded -> remains an issue on py34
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -307,6 +307,10 @@ py_test_files := \
 # - 50664 ipwidgets - User Jupyter. Min ver 5.2.2 to 8.0.0. Sanitize descriptions
 # - 50463 ipwidgets - from 5.2.2 to 8.0.0.rc2
 # - 50792 nbconvert - from 5.0.0 to 6.5.1
+# - Oct 2022
```
andy-maier, line 310: I suggest not to have the date. It does not help.

Makefile
```diff
@@ -307,6 +307,10 @@ py_test_files := \
...
+# - 51159 cryptography  v 39.0.0 drops support for C library "LibreSSL" < 3.4
+# - 50885 pygments Lexer infinite loop in versions Pygments versions 1.5 to 2.7.3
+# - 50886 pygments regular expressions vulnerable to ReDos, versions >=1.1,<2.7.4
```
andy-maier, line 313: In the description, I suggest to have the reason why we cannot upgrade to that version, and not a repetition of the safety issue.

Makefile
```diff
@@ -307,6 +307,10 @@ py_test_files := \
 # - 50664 ipwidgets - User Jupyter. Min ver 5.2.2 to 8.0.0. Sanitize descriptions
 # - 50463 ipwidgets - from 5.2.2 to 8.0.0.rc2
 # - 50792 nbconvert - from 5.0.0 to 6.5.1
+# - Oct 2022
+# - 51159 cryptography  v 39.0.0 drops support for C library "LibreSSL" < 3.4
```
andy-maier, line 311: Should cryptography not be upgraded in the minimum-constraints and requirements files, where possible?

docs/changes.rst
```diff
@@ -162,6 +162,9 @@ Released: not yet
   instances for Open/Pull after  the request to the server is complete). (see
   issue #1801)
 
+* Added security issues 50463  50792  50748  51159 50885 50886  to Makefile
+  ignore list of new security issue August and September 2022.
```
andy-maier, line 166: This change adds less safety issues than what this change entry says.

andy-maier (changes requested):
See comments.

docs/changes.rst
KSchopmeyer: Fixed

andy-maier (approved):

PR #2953, opened 2022-12-14
Fix several security issues.

KSchopmeyer:
Fixes safety issues for py, safety, and wheel from November 2022 safety report.

This PR will not complete tests successfully because of issues fixed by pro #2954. It was rebased into that PR to test together.

This generate a conflict in version requirements between safety (>=8.0.2) and easy-vault
(>=7.1.1) on click minimum version causes failure with 2.10 minimum because
we specify  click==7.1.1 in minimum. Moved to specify click min = 8.0.2

It generates a second conflict because safety 2.2.0 requires packaging 21.0 and the definition in minimum-constrants is 19.0. Fix

NOTE: These are the same fixes made for pywbemtools in November except for the issues with easy-vault and packaging

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -32,10 +32,13 @@ more-itertools>=4.0.0; python_version >= '3.6'
 # safety 1.9.0 removed support for Python 2.7 (and now also enforces that)
 safety>=1.8.7,<1.9.0; python_version == '2.7'
 safety>=1.9.0; python_version >= '3.5'
+safety>=2.2.0; python_version >= '3.6'
 # dparse 0.5.0 has an infinite recursion issue on Python 2.7,
 #   see https://github.com/pyupio/dparse/issues/46
 dparse>=0.4.1,<0.5.0; python_version == '2.7'
 dparse>=0.5.1; python_version >= '3.5'
+# ver 0.6.2 min requirement by safety 2.2.0
+dparse>=0.6.2; python_version >= '3.6'
 
 # Tox
 tox>=2.5.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -31,12 +31,12 @@ more-itertools>=4.0.0; python_version >= '3.6'
 # Safety CI by pyup.io
 # safety 1.9.0 removed support for Python 2.7 (and now also enforces that)
 safety>=1.8.7,<1.9.0; python_version == '2.7'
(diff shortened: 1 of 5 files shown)


dev-requirements.txt
```diff
@@ -32,10 +32,13 @@ more-itertools>=4.0.0; python_version >= '3.6'
 # safety 1.9.0 removed support for Python 2.7 (and now also enforces that)
 safety>=1.8.7,<1.9.0; python_version == '2.7'
 safety>=1.9.0; python_version >= '3.5'
+safety>=2.2.0; python_version >= '3.6'
```
andy-maier, line 35: If you add this for py>=3.6, you need also to limit the previous line to py<3.6, i.e. py==3.5.

dev-requirements.txt
```diff
@@ -32,10 +32,13 @@ more-itertools>=4.0.0; python_version >= '3.6'
...
 # dparse 0.5.0 has an infinite recursion issue on Python 2.7,
 #   see https://github.com/pyupio/dparse/issues/46
 dparse>=0.4.1,<0.5.0; python_version == '2.7'
 dparse>=0.5.1; python_version >= '3.5'
+# ver 0.6.2 min requirement by safety 2.2.0
+dparse>=0.6.2; python_version >= '3.6'
```
andy-maier, line 41: If you add this for py>=3.6, you need also to limit the previous line to py<3.6, i.e. py==3.5.

minimum-constraints.txt
```diff
@@ -132,7 +134,9 @@ build==0.5.1
 Cython==0.29.22
 
 # Unit test (imports into testcases):
-packaging==19.0
+packaging==19.0; python_version <= '3.6'
+# Packaging 21.0 required by safety 2.2.0
+packaging==21.0; python_version >= '3.7'
```
andy-maier, line 139: If you increase the minimum version here, you also need to increase it in test-requirements.txt.

andy-maier (changes requested):
See comments

dev-requirements.txt
andy-maier: DONE

minimum-constraints.txt
andy-maier: DONE

dev-requirements.txt
andy-maier: DONE

Issue #1245, opened 2018-05-04
`python setup.py` failed when ran from taged release tarball


cathay4t:
* Errors:
  Got error when run `python setup.py` from release tarball from (https://github.com/pywbem/pywbem/releases/tag/0.12.2)
```
ERROR:root:Error parsing
Traceback (most recent call last):
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/core.py", line 96, in pbr
    attrs = util.cfg_to_args(path, dist.script_args)
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/util.py", line 258, in cfg_to_args
    pbr.hooks.setup_hook(config)
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/hooks/__init__.py", line 25, in setup_hook
    metadata_config.run()
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/hooks/base.py", line 27, in run
    self.hook()
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/hooks/metadata.py", line 26, in hook
    self.config['name'], self.config.get('version', None))
  File "/home/fge/Downloads/pywbem-0.12.2/.eggs/pbr-4.0.2-py3.6.egg/pbr/packaging.py", line 816, in get_version
    name=package_name))
Exception: Versioning for this project requires either an sdist tarball, or
access to an upstream git repository. It's also possible that there is a
mismatch between the package name in setup.cfg and the argument given to
pbr.version.VersionInfo. Project name pywbem was given, but was not able to be
found.
error in setup command: Error parsing
/home/fge/Downloads/pywbem-0.12.2/setup.cfg: Exception: Versioning for this
project requires either an sdist tarball, or access to an upstream git
repository. It's also possible that there is a mismatch between the package
name in setup.cfg and the argument given to pbr.version.VersionInfo. Project
name pywbem was given, but was not able to be found.
```

* Root casue
  `pbr.version.VersionInfo` failed to find out the version of current project. The `pbr` get version from
   any of below. (https://git.openstack.org/cgit/openstack-dev/pbr/tree/pbr/packaging.py)
  * Get version from git repo: `_get_version_from_git()`.
[...]

PR #2954, opened 2022-12-16
Fixed issue with Python 3.5+3.6 no longer supported in ubuntu_latest; Updated the version of setup-python

KSchopmeyer:
This PR is rebased on PR # 2953, that fixes new security issues and changes only the requirements files (i.e. minimum-constrains.txt).  The changes to fix the issues CI testing with the ubuntu version are in the file test.yaml.  Did this so we could get a running version with these two separate issues since each of them breaks the CI tests.

CI tests fail because github moved to use ubuntu-22.04 as ubuntu-latest and the setup_python github action does not support versions of python older than version 3.7 with this ubuntu version.  This means that the github actions matrix had to be updated to reflect the use of ubuntu-20.04 for tests with the older versions of ubuntu.  This is documented in pywbemtools issue #1245

Modified github actions matrix to use ubuntu 20.04 in whereever python <= 3.5 is defined.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,11 @@ jobs:
       id: select_matrix
       # Select full matrix when scheduled or when releasing, and normal matrix
       # otherwise. The matrix is defined as a JSON string.
-      # TODO: Find a way to define this with less escapes.
+      # This technique documented in:
+      #    https://stackoverflow.com/questions/65384420/how-to-make-a-github-action-matrix-element-conditional
+      # NOTE: Ubuntu 20.04 used because in some tests Ubuntu:latest fails with
+      # Python 3.5 and 3.6. See pywbemtools issue 1245 for explanation.
+      # TODO: Find a way to define this with fewer escapes.
       run: |
         if [[ "${{ github.event_name }}" == "schedule" || "${{ github.head_ref }}" =~ ^release_ ]]; then \
           echo "::set-output name=matrix::{ \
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
windows:latest python 3.5, latest fails.

Differences in build:

python 2.7 lxml:

Collecting lxml>=4.6.2
  Downloading lxml-4.9.2-cp27-cp27m-win_amd64.whl (3.8 MB)

Python 3.5 lxml:
Collecting lxml>=4.6.2
  Downloading lxml-4.9.2.tar.gz (3.7 MB)

Not sure why the difference shows up now since I cannot see anything different between Nov when this ran and now in area of lxml but I think that is where the problem starts.

KSchopmeyer:
tested on windows 11 with python 3.5.4 local and the develop process did not try to build a wheel but used:

makedevelop.txt:  Using cached lxml-4.9.2-cp35-cp35m-win_amd64.whl (3.8 MB)

as the wheel for lxml

Collecting lxml>=4.6.2
  Using cached lxml-4.9.2-cp35-cp35m-win_amd64.whl (3.8 MB)

.github/workflows/test.yml
```diff
@@ -55,19 +115,36 @@ jobs:
...
                 \"os\": \"ubuntu-latest\", \
-                \"python-version\": \"3.5\", \
+                \"python-version\": \"2.7\", \
                 \"package_level\": \"minimum\" \
               }, \
               { \
                 \"os\": \"ubuntu-latest\", \
+                \"python-version\": \"2,7\", \
+                \"package_level\": \"latest\" \
+              } \
```
andy-maier, line 128: Instead of first specifying 2.7 on line 116 and then excluding all of its (two) matrix expansions again, 2.7 should simply not be specified on line 116.

andy-maier (changes requested):
See comment

andy-maier:
I don't know why it fails on windows with py3.5/latest. We should go back the weekly tests and check at which point it succeeded last time. Note that we have already excluded py3.5/minimum on windows, we could enable that to see if it fails for the same reason.

andy-maier:
Finally, I think since the changes of both, PR 2953 and the on-top delta of this PR, are needed to get the builds to succeed again, I think we should have only this PR here, but with the commits reorganized so that there is one commit for each of the two changes.

.github/workflows/test.yml
andy-maier: DONE

andy-maier:
I resolved my review comments and rebased to master after merging the (updated) PR #2953.

andy-maier:
For unknown reasons, the lXML issue on Windows did not occur again. Merging.

Issue #2951, opened 2022-10-28
Do not retry HTTP POST upon read timeouts


andy-maier:
The current implementation retries POST operations, because most POST operations in WBEM are actually read-only. Some of them are not however, and for those this can lead to duplicate execution.

See https://github.com/pywbem/pywbem/blob/master/pywbem/_cim_operations.py#L208 which sets POST as the only method to retry.

andy-maier:
The proposal is to remove the special setup for HTTP method verbs to retry on and go with the urllib3 default list, which is here: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry.DEFAULT_ALLOWED_METHODS

andy-maier:
We discussed this and agreed to go with the proposal.

PR #2969, opened 2023-01-02
Fix issue with HTTP retries of cim-xml operations.

KSchopmeyer:
Fixes an issue in the parameters for  HTTP POST retries within urllib3 where the retry values defined in CIM_operations  allowed the retry of a CIM/XML request that such as
invokemethod, update, etc. where it is critical that the operation only be issued once.  This is because all CIM/XML operations use POST.

Changes the retry to not allow retries on POST operations.

Issue documented in #2951

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -27,6 +27,12 @@ Released: not yet
 
 **Bug fixes:**
 
+* Fix issue where we could get HTTP retries upon HTTP read timeouts from
+  the server.  Changed to not allow urllib3 to do retries on post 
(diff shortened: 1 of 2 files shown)


andy-maier (approved):
Please rebase once PR #2971 is merged.

andy-maier:
Rolling this back to 1.5.1 requires to roll back several other prerequisite PRs as well. Given that we want to release 1.6.0 soon, I suggest not to roll this one back.

andy-maier (approved):

PR #2978, opened 2023-01-14
Release 1.6.0

andy-maier:
Please approve the release of pywbem 1.6.0

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -14,16 +14,10 @@ Change log
       :revisions: 1
 
 
-pywbem 1.6.0.dev1
------------------
+pywbem 1.6.0
+------------
 
-This version contains all fixes up to version 1.5.x.
-
-Released: not yet
-
-**Incompatible changes:**
-
-**Deprecations:**
+Released: 2023-01-14
 
 **Bug fixes:**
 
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -56,12 +50,6 @@ Released: not yet
   smaller (110 mb) but the same set of models and providers as the previous
   image.
```
KSchopmeyer, line 52: I notice that the "Known issues" header is missing.  Was that on purpose.  Should we leave it there just so we have a reminder to use in future versions even if we simply change the text to "No known issues"

KSchopmeyer (commented):
One comment about the fact that the known issues is missing from changes.rst. Other than resolving that, approved.  Even that one is not a real issue, but a question.

KSchopmeyer (commented):
Couple more questions from looking at the README as presented when github is on the Release-1.6 branch.

1. The Readme badge shows 1.5 and not 1.6.
2. The list of supported python versions does not show python 3.11.

docs/changes.rst
andy-maier: That is on purpose. The start of a new release adds it again, and the release removes it. We wanted to avoid maintaining known issues as a list, so we took the short way of just having a link. However, the set of open issues changes over time, and I think that does not go well with a change log. But I have no strong opinion on this.

andy-maier:
On your comments:

1. The "pypi package" badge shows what is released on Pypi, so it shows 1.5 when 1.6 is not yet released.

2. The "python" badge shows the supported Python versions of the latest version released to Pypi, so it shows what 1.5 supports when 1.6 is not yet released.

Not sure what you mean with PKG_INFO.

docs/changes.rst
andy-maier: This release follows what we have done in all previous releases. I suggest to release it that way, and if we want to change anything, to do it for the next release.

PR #2997, opened 2023-05-01
Ks/#2995 clarify listener host

KSchopmeyer:
Clarify documentation on the host parameter for the WBEMListener class.

See commit message

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -719,11 +719,15 @@ class WBEMListener(object):
         Parameters:
 
           host (:term:`string`):
-            IP address or host name at which this listener can be reached.
+            IP address or host name to which this listener is bound (i.e.
+            at which this listener can be reached). Setting the host
+            address to IP address 0.0.0.0 defines a listener that binds to
+            all network interfaces.
 
-          http_port (:term:`string` or :term:`integer`):
+           http_port (:term:`string` or :term:`integer`):
             HTTP port at which this listener can be reached. Note that at
-            least one port (HTTP or HTTPS) must be set.
+            least one port (HTTP or HTTPS) must be set. Both the http and
+            https ports can be set.
 
             `None` means not to set up a port for HTTP.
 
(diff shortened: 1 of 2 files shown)


pywbem/_listener.py
```diff
@@ -719,16 +719,21 @@ def __init__(self, host, http_port=None, https_port=None,
...
           host (:term:`string`):
-            IP address or host name at which this listener can be reached.
-
-          http_port (:term:`string` or :term:`integer`):
+            IP address or host name to which this listener is bound (i.e. host
+            from which indications will be accepted). Setting the listener IP
+            address to  0.0.0.0 defines a listener that is not bound any
+            particular IP address. It will accept indications from any IP
+            address.  When there is any other host name/IP address, indications
+            will be accepted only from that IP address.
```
andy-maier, line 727: Here is my understanding of the meaning of the `host` parameter:

* It is an IP address of the system where the listener runs (i.e. not of the system from where the client sends the indications). Its purpose is to select one out of multiple IP addresses the listener system may have.
* If specified with a value other than 0.0.0.0, that causes the listener to receive only on that (listener-side) IP address, i.e. it only gets indication requests that target this IP address (they may originate on a different network and may originally target a different IP address when crossing NAT, but at the time they reach the network interface of the listener system, they must target that specified IP address to get through).
* If specified with the special value 0.0.0.0, that causes the listener to receive on all of its (listener-side) IP addresses on all of its network interfaces.
* All client IP addresses are always permitted to send to the listener. That has nothing to do with using or not using the special 0.0.0.0 value.

If that understanding is correct, your current text is not.

Also, if we use the terminology of "binding", we should be explicit in that it binds to a local IP address on the listener system.

andy-maier (changes requested):

KSchopmeyer:
Andy is correct in definition of host parameter of listener.  Corrected.  Tested against local environment, docker environment, wbem server in virtual box wbem server in second sysem.

pywbem/_listener.py
KSchopmeyer: Modified. I agree.  Yes a host name can be specified and there is a set of "defaults" defined at the level of sockets to tell the system to effectively bind to any network interface.  However, when a host name/IP is used, exactly what that means depends on the network interface itself and the OS.  Thus, does host name imply that it will accept only that specific IP address or all addresses in the same interface. Windows and Linux are different here and there may be differences between some of the different network interfaces even on linux (ex. wifi, loopback, etc.).

andy-maier (approved):

PR #3000, opened 2023-05-02
Improved handling of missing/extra Values items in ValueMapping

andy-maier:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -25,6 +25,10 @@ Released: not yet
 
 * Dropped support for Python 3.5. (issue #2867)
 
+* The 'pywbem.ValueMapping' class will now raise a 'pywbem.ModelError' for any
+  missing or extra items in the 'Values' qualifier, compared to the 'ValueMap'
+  qualifier. Previously, 'IndexError' was raised. (issue #2991)
+
 **Deprecations:**
 
 **Bug fixes:**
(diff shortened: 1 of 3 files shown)


KSchopmeyer (commented):
Good solution to the issue as viewed from the ValueMapping class itself.  I have only couple of questions.
1. Thus we use ValueMapping in the _server.py code but don't provide a way for the user to get access to this extension to ValueMapping if they have an issue.   Should we be using it ourselves by extending it through the server or allowing some sort of config setting?
2. Does this need any special mention in the docs other than what is in the code level documentation here?
3. I note that ppparently we do not ever test for equality of the Value and Valuemaping qualifiers Mapped values anywhere in the MOF compiler so like OpenPegasus compiler (Checked there also) this error just slides through the environment until some user hits it in our ValueMapping class..  Should we a. Write an issue on that b. Add code to at least warn about it(this one would probably slow several processes down significantly)

andy-maier:
On your comments:
1. I don't think so. If someone encounters this in an incorrect Implementation, ut can be brought up.
2. I think the docs in the class is sufficient.
3.a. What issue would you create? b. We could raise a Python warning if we tolerate it.

KSchopmeyer:
We discussed the issue above.  We agree on 1 and 2 and concluded that we will not modify the compiler.

Karl will set approved on this PR

PR #3002, opened 2023-05-07
Update pywbem config files for new safety issues

KSchopmeyer:
Updated Makefile to add a number of new safety issues that we have to ignore because older python versions we use do not support the new safety version minimums.  This does not change to use the policy file but just ignores the new issues


andy-maier (changes requested):
In addition, do you also want to make the transition to the safety policy file?

KSchopmeyer:
Withdrawing this pr since pr 3005 super cèdes it with the policy file

PR #3003, opened 2023-05-07
Limit max version of urllib3 to lt 2.0.0

KSchopmeyer:
Limit the max version of urllib3 to lt 2.0 since there are a number of issues in this change that will cause our tests to fail including

1. no longer supports python 2.7, 3.5, 3.6
2. pool module removed

as a minimum  

For the time being we will limit the versions we use to < 2.0 but have filed issue #3006 to determine the overall impact of the new version and changes we will make to support it.

Note: This will only pass the tests when run with pr # 3005, the update to safety (with the policy file)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,9 +43,14 @@ nocasedict>=1.0.1
 # urllib3 1.25.9 addressed issue 38834 reported by safety
 # urllib3 needs to be pinned to <1.25 for requests <2.22.0
 # urllib3 1.26.5 vendors six 1.16.0 which is needed on Python 3.10 to remove ImportWarning
-urllib3>=1.25.9; python_version == '2.7'
-urllib3>=1.25.9; python_version >= '3.5' and python_version <= '3.9'
-urllib3>=1.26.5; python_version >= '3.10'
+# urllib3 2.0 requires py>=3.7
+urllib3>=1.25.9,<2.0.0; python_version == '2.7'
+urllib3>=1.25.9,<2.0.0; python_version == '3.5'
+urllib3>=1.25.9,<2.0.0; python_version == '3.6'
+# TODO: Following limits use of urllib3<= 2.0 until issues about this upgrade
+# resolved. Issue #3006 failed to update urllib3 to version 2.0.0 for python >= '3.7'
+urllib3>=1.25.9,<2.0.0; python_version >= '3.7' and python_version <= '3.9'
+urllib3>=1.26.5,<2.0.0; python_version >= '3.10'
 
 # setuptools 61.0.0 breaks "setup.py install", see https://github.com/pypa/setuptools/issues/3198
 setuptools!=61.0.0; python_version >= '3.7'
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -118,8 +118,13 @@ certifi==2019.11.28
 # requests>=2.26.0 uses charset-normalizer instead of chardet; both are not used by any other package
 chardet==3.0.2; python_version <= '3.9'
 charset-normalizer==2.0.0; python_version >= '3.10'
+# urllib3 >= 2.0.0 minimum python == '3.7'
 urllib3==1.25.9; python_version == '2.7'
-urllib3==1.25.9; python_version >= '3.5' and python_version <= '3.9'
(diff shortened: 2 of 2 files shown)


andy-maier:
I think pinning it is not the right approach.

requirements.txt
```diff
@@ -43,9 +43,14 @@ nocasedict>=1.0.1
 # urllib3 1.25.9 addressed issue 38834 reported by safety
 # urllib3 needs to be pinned to <1.25 for requests <2.22.0
 # urllib3 1.26.5 vendors six 1.16.0 which is needed on Python 3.10 to remove ImportWarning
-urllib3>=1.25.9; python_version == '2.7'
-urllib3>=1.25.9; python_version >= '3.5' and python_version <= '3.9'
-urllib3>=1.26.5; python_version >= '3.10'
+# urllib3 2.0 incompatible with python 2.7, 2.5, 3.6
```
andy-maier, line 46: urllib3 2.0 is not incompatible with these versions. It simply requires py>=3.7. Let's phrase it like that.

requirements.txt
```diff
@@ -43,9 +43,14 @@ nocasedict>=1.0.1
...
+urllib3>=1.25.9,<2.0.0; python_version == '2.7'
+urllib3>=1.25.9,<2.0.0; python_version >= '3.5' and python_version <= '3.6'
```
andy-maier, line 48: Given that urllib3 2.0 requires py>=3.7, it does not need to be pinned for py<=3.6

docs/changes.rst
```diff
@@ -54,6 +54,9 @@ Released: 2023-01-14
   some operations (invoke, update) could cause data integrity issues).
   (see issue #2951)
 
+* Limited version of urllib3 used to lt 2.0 since there are incompatilities
+  in version 2.0.  (see issue #3001)
```
andy-maier, line 58: If we consider this a temporary change, I suggest not to have a change log entry.

requirements.txt
KSchopmeyer: DONE

requirements.txt
KSchopmeyer: DONE.

docs/changes.rst
andy-maier: This should still be done.

requirements.txt
```diff
@@ -43,9 +43,14 @@ nocasedict>=1.0.1
...
+
+# The following limits use of urllib <>= 2.0 until issues about this upgrade
+# resolved.
+urllib3>=1.25.9,<2.0.0; python_version >= '3.6' and python_version <= '3.7'
+urllib3>=1.26.5,<2.0.0; python_version >= '3.10'
```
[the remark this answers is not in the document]
andy-maier: I'm fine with the current content

docs/changes.rst
KSchopmeyer: Log entry removed

requirements.txt
```diff
@@ -43,9 +43,12 @@ nocasedict>=1.0.1
...
 # urllib3 needs to be pinned to <1.25 for requests <2.22.0
 # urllib3 1.26.5 vendors six 1.16.0 which is needed on Python 3.10 to remove ImportWarning
-urllib3>=1.25.9; python_version == '2.7'
-urllib3>=1.25.9; python_version >= '3.5' and python_version <= '3.9'
-urllib3>=1.26.5; python_version >= '3.10'
+# urllib3 2.0 requires py>=3.7
+urllib3>=1.25.9,<2.0.0; python_version >= '2.7' and python_version <= '3.6'
+# TODO: The following limits use of urllib <= 2.0 until issues about this upgrade
+# resolved. Issue #3006 filed to update urllib3 to version 2.0.0.
+urllib3>=1.25.9,<2.0.0; python_version >= '3.7' and python_version <= '3.9'
```
andy-maier, line 50: I don't understand the separation between the python versions 2.7-3.6 and 3.7-3.9. The Todo comment refers to the following lines, but also applies to the previous line.

Can we merge the two sets of Python versions?

andy-maier (changes requested):

andy-maier (approved):

PR #3009, opened 2023-05-28
Dropped support for Python 3.5

andy-maier:


KSchopmeyer (commented):
I reviewed this one and stepped back through most of the detailed changes.  Did not find in issues which means you must have lost a significant amount of your eye sight digging out all of those version differences, in particular for jupyter.

**DISCUSSION:** I do have one real issue here in that we did not ever mark Python 3.5 as deprecated so the removal will come as a complete surprise to the users.  I therefore marked this for discussion.

**DISCUSSION COMPLETE:**  After short discussion we concludee that we can remove python 3.5 support since 

   a. Python 3.5 itself has been deprecated for over 2 years now and 
   b: it is probable that many users would not notice our deprecation notice if we had one when the code executes. So they find out with a new version
  c: it appears to be common practice to remove support for old versions of python without marking them deprecated in a previous release

andy-maier:
The jupyter differences come from the fact that the PR was based on #3015 since it depends on those changes. After merging #3015, I rebased this PR on master.

andy-maier:
Based on the discussion completion, I'll merge this PR.

PR #3019, opened 2023-06-02
Expand documentation on the WBEMListener host parameter

KSchopmeyer:
This change clarifies the definition of the host parameter on the WBEMListener class to include the definition of how to define a listener that listens on multiple network addresses.

Note that there is still some ambiguity in that a. this does not discusss IPV6 and b) only briefly mentions that the actual behavior of a listener with a specific address may be OS or network dependent in that in some cases the test is limited to only pass requests with the specific address and in other cases, it will pass any address that is part of the specific network interface defined by the host parameter.

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -722,14 +722,15 @@ class WBEMListener(object):
             IP address or host name to which this listener is bound (i.e. at
             which this listener can be reached).  If a listener is bound to a
             particular IP address it will only receive indications addressed to
-            that IP address (or to any IP address network interface containing
-            that address depending on the OS).  The IP address 0.0.0.0 will
-            bind the listener to any configured IPV4 address on the local
-            machine, Setting the host address to an empty string (i.e. "")
-            defines a listener that binds to the listener to all configured
-            network interfaces on the local machine.
+            that IP address (or to any IP address on the network interface
+            containing that address) depending on the OS.
 
-           http_port (:term:`string` or :term:`integer`):
+            Setting the host parameter to an empty string (i.e. "") defines a
+            listener that binds to all configured network interfaces on the
+            local machine. Binding to IPV4 or IPV6 network interfaces can be
+            defined with the IP addresses "0.0.0.0" for IPV4 or "::" for IPV6.
+
+          http_port (:term:`string` or :term:`integer`):
             HTTP port at which this listener can be reached. Note that at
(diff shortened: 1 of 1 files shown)


pywbem/_listener.py
```diff
@@ -719,10 +719,15 @@ def __init__(self, host, http_port=None, https_port=None,
...
-            all network interfaces.
+            IP address or host name to which this listener is bound (i.e. at
+            which this listener can be reached).  If a listener is bound to a
+            particular IP address it will only receive indications addressed to
+            that IP address (or to any IP address network interface containing
+            that address depending on the OS).  The IP address 0.0.0.0 will
+            bind the listener to any configured IPV4 address on the local
+            machine, Setting the host address to an empty string (i.e. "")
+            defines a listener that binds to the listener to all configured
+            network interfaces on the local machine.
```
andy-maier, line 730: Grammar: "defines a listener that binds to the listener"

Also, I don't understand the distinction between 0.0.0.0 and empty string. Both seem to listen to any ip address on any network interface.

andy-maier (changes requested):

pywbem/_listener.py
KSchopmeyer: Cleaned up definition.  The "" is specific to python HTTPServer and implies that the listener will receive indications from any network interface on the system i.e. any indication addressed to the port defined.  However, by the time that gets moved down to the socket level, there is another messy level of interpretation where the address 0.0.0.0 implies that it will accept only from IPV4 ports except that on some OSs, setups, OS versions there is also the dual-stack network port that overrides the the IPV4 vs IPV6.  Apparently mostly added to allow IPV6 network interfaces to feed IPV4 NATS, etc.  

Actually, I am sorry I did anything except just the "" which is a part of the python HTTPServer host definition except that it very poorly documented.   I really do not think our user care about that level of separation especially when the actual behavior is to either block all but the selected IP address if a specific address/hostname is selected and others to make the decision at the network interface level (i.e. block all but the network interface containing the defined host name).  That is differences between some linux implementations, windows, and macos.  For our users I really think it is just accept from everybody or except from one whether that is a particular indication sender or the a particular network interface.  But I just turned it in with a bunch of that discussion in the text.  Feel free to argue back.

Note that the 0.0.0.0 is supposed to be specific to IPV4 networks but the dual-stack network interface has modified that.

andy-maier (approved):

PR #3020, opened 2023-06-02
Adds end-end test for indications

KSchopmeyer:
Adds a new end-end test for:

1. Managing indication subscriptions.
2. Receiving indications from the WBEM Server

This test is based on the OpenPegasus docker image defined for the other end-end tests.

See commit mesage for details

NOTE: This commit includes changes for the example pegasusindicaitonstest.py since basically the same code is in both the end to end test and the example and this made the two almost the same.

In order to get a consinstent pass of this test, I disabled python 2.7 and reduced python 3.7 to minimum.  We could have passed on the PACKAGE_LEVEL variable from Makefile but this establishes that the issues with quitting early are at least probably tied to python 2.7 and 3.7 We also know from earlier tests that it is probably tied to the minimum config.

I suggest we commit this and I have already filed a bug on the failures with python 2.7, etc.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -49,6 +49,11 @@ Released: not yet
   filled up with that value, and extra items in the 'Values' qualifier are
   truncated. (issue #2991)
 
+* Add end-end test for indications where an OpenPegasus server in Docker
+  container is the WBEM Server and pywbem is both the client that creates
+  indication subscriptions the indication listener. This test uses the
+  OpenPegasus container used for other end-end tests.
+
 **Cleanup:**
 
 * Replaced the safety_ignore_opts in Makefile that is used as the basis for
--- a/tests/end2endtest/test_indications.py
+++ b/tests/end2endtest/test_indications.py
@@ -1,41 +1,87 @@
-#!/usr/bin/env python
+"""
+End2end tests of indications from a wbem server. This tests:
+1. creating the subscription in the container
+2. Creating the pywbemlistener
+3. Receiving indications from the server. It uses only OpenPegasus WBEM server
+because OpenPegasus defines a provider in its test environment specifically to
+send indications based on a method invocation and we have a container
+implementation of OpenPegasus.
 
 """
-Example of handling subscriptions and indications and receiving indications
-from a particular provider in OpenPegasus . This example depends on a test
-class and method in OpenPegasus.  It creates a server and a listener, starts
-the listener and then requests the indications from the server. It waits for a
-defined time for all indications to be received and then terminates either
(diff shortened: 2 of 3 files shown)


docs/changes.rst
```diff
@@ -36,6 +36,11 @@ Released: not yet
 
 **Enhancements:**
 
+* Add end-end test for indications where an OpenPegasus server in docker
```
andy-maier, line 39: Change "docker" -> "Docker".

docs/changes.rst
```diff
@@ -36,6 +36,11 @@ Released: not yet
...
+  container is the WBEM Server and pywbem defines the client that manages
```
andy-maier, line 40: On "pywbem defines": That sounds strange to me. How about "pywbem is used as" or "pywbem is"?

docs/changes.rst
```diff
@@ -36,6 +36,11 @@ Released: not yet
...
+  indication subsecriptions and receives the indications. This test uses the
```
andy-maier, line 41: Typo: subsecriptions

tests/end2endtest/test_indications.py
```diff
@@ -0,0 +1,568 @@
...
+    port = 5000
+    log = True
+    verbose = True
+
+    # define repeat_loop and indication_count variables to rerun the test with
+    # each of these pairs. (repeat_loop, indication_count)
+    # FUTURE: Expand this to more indications when we resolve issue #3022,
+    # the container not returning all of the indications.  This appears to
+    # be much more consistent with python 3.7 and 3.7 PACKAGE_LEVEL=minimum
+    if sys.version_info[1] == 7:   # Use minimum send ct while issue #3022
```
andy-maier, line 70: Can you please be specific about the major version? If you really mean this case to be taken for 2.7 and 3.7, then say:
```
if sys.version_info[0:2] in ((2, 7), (3, 7)):
```

Second: I do not understand how this version selection matches the comment about 3.7

andy-maier (changes requested):

docs/changes.rst
KSchopmeyer: Changed to pywbem is both the client that creates indication subscriptions and the the indication listener

docs/changes.rst
KSchopmeyer: Fixed

tests/end2endtest/test_indications.py
KSchopmeyer: Fixed.

andy-maier (approved):

PR #3052, opened 2023-09-09
Modify Listener documentation and extend examples.

KSchopmeyer:
Modify the definition of the host parameter of the WBEMListener to define the concept of wild card host.

Modify the documentation of the example to clarify the same thing.

Then a lot of check and pylint issues hit us.

Add new check bypass (for == on type tests)

Modified some code to add ignores for pylint and check

Modified requirements to limit urllib3 max version because of issue in tests with pool parameter and version 2.0

Modified pylintrc to ignore the pylint issue about literal dictionaries.  Note that I changed some but it found a bunch and most did not seem to be in any critical speed path.

### DIFF
--- a/examples/listen.py
+++ b/examples/listen.py
@@ -104,6 +104,15 @@ def _main():
 
     banner = """
 WBEM listener started on host %s (HTTP port: %s, HTTPS port: %s).
+The host parameter may be:
+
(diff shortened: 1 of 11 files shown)


KSchopmeyer:
I did something wrong.  Most of the pylint and check issues had been fixed.  Closing this PR to clean up

PR #3065, opened 2023-10-13
builddoc generates many warnings.

KSchopmeyer:
See issue #3064 for more detailed information on the warning messages.

Also the full list of warnings output is included in issue #3064

This PR tries one way to eliminate the Warning messages, adding a list of ignores to the sphinx conf.py file nitpick_ignore_parameter for general classes of the warnings.

This reduces the number of warnings from about 280 to about 10 total.

There is only a single warning in __logging where the code itself was modified to eliminate the warning.

DISCUSSION: This is one way to consider eliminating the warnings, I am not saying that is is the correct way.  (see the issue).

The questions today are:

1. Why did all these warnings suddendly appear.
2. Is there a way to eliminate them by changing the autodoc definitions for the various APIs.
3. Is there another better alternative than this list?


andy-maier:
See my comment on issue #3064 - I think we have to live with the warnings until Sphinx fixes that, and turning all nitpick warnings off for most of the modules like this PR currently does, does not really help since it probably suppresses other legitimate nitpick warnings. Can we specify more of the warning message in the regexp so that only this specific nitpick warning is ignored?

KSchopmeyer:
At this point we have the regexes down to the target class level for most of the disables. The problem is that if we go any further with the regexes, we move from about 30 re

ex: r'pywbem\.MOFCompileError.*?'  where ignore the error for any method in MOFCompileError.

to back to the 280 warnings.

One alternative might be to redo the regexes to pick up common methods for any CIM.*?Error but I think that that results in more ignores since the level of commonality is actually low for the methods, even with the exceptions.

The other alternative is to process the list of errors and produce an ignore individually for each one so we would have a list of 280 ignores.
Thus instend of using the nitpick_ignore_regex we could be using the nitpick_ignore list so that in place of:

[("py:obj", r'pywbem\.MOFCompileError.*?' ), . . .]

we would have a list of 280 entries like the following for MOFCompileError, one entry for each defined error message
[('py.obj', 'pywbem.MOFCompileError.__delattr__'),
('py.obj', 'pywbem.MOFCompileError.__getattribute__'),
('py.obj', 'pywbem.MOFCompileError.__new__')'
('py.obj', 'pywbem.MOFCompileError.__reduce__'),
('py.obj', 'pywbem.MOFCompileError.__repr__'),
('py.obj', 'pywbem.MOFCompileError.__setattr__'),
('py.obj', 'pywbem.MOFCompileError.add_note'),
('py.obj', 'pywbem.MOFCompileError.with_traceback',
. . .]

Since the error messages have a defined syntax, that would be mostly a question of writing something like an AWK script to process get the type and target out of each error and produce the list of tuples.

I do not have a good answer here except that the list of 280 WARNING messages is worthless.

I think you answered my other question as to whether this is just our error or is something that changed in sphinx.

KSchopmeyer:
All sphinx WARNING: reference target not found messages are now either:

1. Disabled using the complete name of the target in the nitpick-ignore variable of docs/conf.py or the issue fixed. This means that there are about 270 entries in the table and they are all for the type py:obj except one that is for py:class. These appear to be due to sphinx issues with finding targets in superclasses
2. The remaining issues (primarily py:class and exec) were fixed.  There several formatting errors in the docstrings primarily in the pywbem_mock/_instancewriteprovider.py including, missing indented lines, incomplete reference to object, etc.
3. 1 issue resolved in _logging.py by adding a global string (default name of the log file) to the __all__ list in _logging.py.

There are no warning messages output at this point.

Issue #3097, opened 2024-01-18
Python3.12: Changed OrderedDict string


bnavigator:
There is a test failure when testing with python3.12, because they changed the string representation of `collections.OrderedDict`:

```python
# python3.11 -c "from collections import OrderedDict; x = OrderedDict({'a':'b',1:2}); print(x)"
OrderedDict([('a', 'b'), (1, 2)])
# python3.12 -c "from collections import OrderedDict; x = OrderedDict({'a':'b',1:2}); print(x)"
OrderedDict({'a': 'b', 1: 2})
```

<details>

```python
[  336s] ___ Test_LOR_PywbemArgsResults.test_invokemethod_summary[instance_wp_tuple0] ___
[  336s] 
[  336s] self = <tests.unittest.pywbem.test_recorder.Test_LOR_PywbemArgsResults object at 0x7fcb19e10380>
[  336s] capture = <LogCapture (Level 1)>
[  336s] instance_wp_tuple = (CIMInstance(classname='CIM_Foo', path=CIMInstanceName(classname='CIM_Foo', keybindings=NocaseDict({'S1': 'Ham'}), nam...ne, 'class_origin': None, 'embedded_object': None, 'is_array': False, ...}, ...}, 'pywbem_object': 'CIMInstance', ...})
[  336s] 
[  336s]     def test_invokemethod_summary(self, capture, instance_wp_tuple):
[  336s]         # pylint: disable=redefined-outer-name
[  336s]         """
[  336s]         Test invoke method log
[  336s]         """
[  336s]     
[  336s]         instance = instance_wp_tuple[0]
[  336s]         instancename = instance.path
[  336s]     
[  336s]         self.recorder_setup(detail_level='summary')
[  336s]     
[  336s]         return_val = 0
[  336s]         params = [('StringParam', 'Spotty'),
[  336s]                   ('Uint8', Uint8(1)),
[  336s]                   ('Sint8', Sint8(2))]
[  336s]     
[  336s]         self.test_recorder.stage_pywbem_args(method='InvokeMethod',
[  336s]                                              MethodName='Blah',
[  336s]                                              ObjectName=instancename,
[  336s]                                              Params=OrderedDict(params))
[  336s]     
[  336s]         self.test_recorder.stage_pywbem_result((return_val, params),
[  336s]                                                None)
[  336s]     
[  336s]         result_req = (
[  336s]             "Request:test_id InvokeMethod("
[  336s]             "MethodName='Blah', "
[...]

PR #3109, opened 2024-02-11
Ks/#3097 py312 bug

KSchopmeyer:
Fix issue documented in issue #3097, change to repr output of OrderedDict.

Requires change to test_recorder.py


KSchopmeyer:
Redoing to integreate this change and the update for python 3.12

PR #3110, opened 2024-02-12
Integrate python version 3.12.

KSchopmeyer:
Changes to integrate python 3.12 include:

1. Modify the CI test definition to include python 3.12
2. Modify dev-requirements and minimum-requirements for changes to package versions.
3. Modify setup.py to define python 3.12 support
4. Fix issue #3097, issue with OrderedDic. that causes test_recorder.py to fail
5. Set max version of build package to < 1.0 per issue #3112
6.  etc.

See commits message for details

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,11 +12,11 @@
 # Direct dependencies:
 
 # Build distribution archive
-# TODO: Remove dependency to wheel in setup.py, because build 1.1.0 removed
-#       wheel. Wheel does not provide a public API. For details, see
+# Version 1.1.0 pulled after release
+# FUTURE: Remove  wheel < 1.0 For details, see
 #       https://github.com/pypa/build/issues/750. Circumvention for now is
-#       to pin build to <1.1.0.
-build>=0.5.1,<1.1.0
+#       to pin build to <1.0. See issue #3112
+build>=0.5.1, <1.0
 
 # Cythonize
 Cython>=0.29.33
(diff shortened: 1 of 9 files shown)


andy-maier:
* 2.7 min: I assume virtualenv is too high. I use: virtualenv>=20.15.0; python_version <= '3.11'; virtualenv>=20.23.0; python_version >= '3.12'

pywbem/_cim_types.py
```diff
@@ -655,7 +661,14 @@ def get_local_utcoffset():
         A positive value indicates minutes east of UTC, and a negative
         value indicates minutes west of UTC.
         """
-        utc = datetime.utcnow()
+        # Effective Python 3.12, utcnow must define tz to avoid warning msg
+        # That is because utcnow returns a niave datetime object if no timezone
```
andy-maier, line 665: The connent suggests that utcnow() is now used with timezone, but that is not what the code does.

pywbem/_cim_types.py
```diff
@@ -655,7 +661,14 @@ def get_local_utcoffset():
...
+        # is provided to the function
+        if sys.version_info[0:2] <= (3, 12):
```
andy-maier, line 667: Can now(tz) not be used in all py versions?

pywbem/_cim_types.py
```diff
@@ -74,10 +74,16 @@
 
 from __future__ import absolute_import
 
+
 from datetime import tzinfo, datetime, timedelta
+if sys.version_info[0:2] <= (3, 12):
+    from datetime import tzinfo, datetime, timedelta timezone
```
andy-maier, line 80: Missing comma

andy-maier (changes requested):

pywbem/_cim_types.py
```diff
@@ -655,7 +661,14 @@ def get_local_utcoffset():
...
         value indicates minutes west of UTC.
         """
-        utc = datetime.utcnow()
+        # Effective Python 3.12, utcnow must define tz to avoid warning msg
+        # That is because utcnow returns a niave datetime object if no timezone
+        # is provided to the function
+        if sys.version_info[0:2] <= (3, 12):
+            utc = datetime.now(timezone.utc)
+        else:
+            utc = datetime.utcnow()
```
andy-maier, line 670: `now(tz)` returns a timezone-aware object, while `now()` returns a timezone-naive object.
This will cause the comparison `local < utc` further down to fail because they cannot be compared.

I suggest instead to replace the whole calculation in this method with:
```
return -time-ztimezone
```
See https://docs.python.org/3/library/time.html#time.timezone
Note that `time.timezone` does not consider DST, but then I'm not sure what the DST rules are for `get_local_utcoffset()`.

KSchopmeyer: it was late in the day.

andy-maier: The division by 60 only converts seconds to minutes, so it assumes that timezones are on full minutes. The old algorithm made the same assumption.

pywbem/_cim_types.py
```diff
@@ -655,11 +658,20 @@ def get_local_utcoffset():
...
+            aware_utc = datetime.now(timezone.utc)
+        else:
+            unaware_utc = datetime.utcnow()
+            aware_utc = unaware_utc.replace(tzinfo=timezone.utc)
+
         local = datetime.now()
-        if local < utc:
-            return - int(float((utc - local).seconds) / 60 + .5)
-        return int(float((local - utc).seconds) / 60 + .5)
+        local_unaware = local.replace(tcinfo=None)
```
andy-maier, line 671: Typo tcinfo

pywbem/_cim_types.py
```diff
@@ -655,11 +658,20 @@ def get_local_utcoffset():
...
+        if local_unaware < aware_utc:
```
andy-maier, line 672: This compares naive and unaware, which will result in TypeError.

andy-maier (changes requested):

pywbem/_cim_types.py
KSchopmeyer: fixed

pywbem/_cim_types.py
KSchopmeyer: fixed

KSchopmeyer:
Currently pylint  undefined var error in conftest.py with namedtuple in minumum test with python 3.6, etc.  However message flags the pylint issue as line one not the line where the namedtuple is defined or used..

KSchopmeyer:
Test fails with"Setuptools is replacing distutils. See issue #3113

NOTE: The version of pylint is different in pywbem minimum-constraints tnah in pywbemtools minimum_constraints.  I am putting together list of the differences

andy-maier:
The pylint failures on py36/37-min are not caused by the userwarning, but by:
```
tests/functiontest/conftest.py:1:0: W0612: Unused variable 'exp_result' (unused-variable)
```

andy-maier (changes requested):

dev-requirements.txt
```diff
@@ -12,11 +12,16 @@
 # Direct dependencies:
 
 # Build distribution archive
+# Version 1.1.0 pulled after release
+build>=0.5.1, !=1.1.0
```
andy-maier, line 16: Just excluding 1.1.0 does not help, as you can see in the logs. You need to pin it to <1.1.0, which is what the master branch already does.

andy-maier (changes requested):

pywbem/_cim_types.py
```diff
@@ -655,11 +659,22 @@ def get_local_utcoffset():
...
-        return int(float((local - utc).seconds) / 60 + .5)
+        if six.PY2:
+            # Effective Python 3.12, utcnow() generates deprecation warning
+            # because utcnow returns a niave datetime object. timezone does
+            # not exist for python 2.7 so we use utcnow()
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    message=".*Passing None has been deprecated.")
+                utc_unaware = datetime.utcnow()
```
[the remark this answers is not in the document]
andy-maier: resolved.

pywbem/_cim_types.py
```diff
@@ -655,11 +659,22 @@ def get_local_utcoffset():
...
-            return - int(float((utc - local).seconds) / 60 + .5)
-        return int(float((local - utc).seconds) / 60 + .5)
+        if six.PY2:
+            # Effective Python 3.12, utcnow() generates deprecation warning
+            # because utcnow returns a naive datetime object. timezone does
+            # not exist for python 2.7; use utcnow(timezone.utc).
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    message=".*Passing None has been deprecated.")
```
andy-maier, line 669: Disabling the warnings is not needed because this code is executed only on Python 2 where the warning was not issued yet.

dev-requirements.txt
KSchopmeyer: Agreed. Done

pywbem/_cim_types.py
KSchopmeyer: agreed done

andy-maier (approved):

PR #3111, opened 2024-02-25
Add new safety issues to safety policy file feb 2024

KSchopmeyer:
Adds new entries in safety for

- GitPython
- jinga2
- Jupyterlab
- new version for send to trash

Added fixes for new pylint issues so that the commit would complete.

### DIFF
--- a/.safety-policy.yml
+++ b/.safety-policy.yml
@@ -199,6 +199,15 @@ security:
             reason: Fixed cryptography version 41.0.6 requires Python>=3.7 and is used there
         62817:
             reason: Fixed prompt-toolkit version 3.0.13 requires Python>=3.6 and is used there
+        63687:
+            reason: GitPython, Python 2.7, 3.6. fixed version 3.1.35 requires Python>=3.7
+        64227:
+            reason: Jinja2, ver 3.1.3 fixes scripting vulnerability, requires Python>=3.7
+        64587:
+            reason: jupyterlab  issue in version 4.03, requires < 4.0.11 python >= 3.8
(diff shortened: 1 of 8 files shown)


.safety-policy.yml
```diff
@@ -199,6 +199,10 @@ security:
             reason: Fixed cryptography version 41.0.6 requires Python>=3.7 and is used there
         62817:
             reason: Fixed prompt-toolkit version 3.0.13 requires Python>=3.6 and is used there
+        63687:
+            reason: GitPython, Python 2.7, 3.6. fixed version 3.1.35 requires Python>=3.7
+        64227:
+            reason: jinga2 <= version 3.1.3 fixes vulnerability.
```
andy-maier, line 205: Typo "jinga2"

Second, I suggest to actually upgrade the packages where possible, e.g. Jinja2 3.1.3 for py>=3.7

andy-maier (changes requested):
See comment

.safety-policy.yml
KSchopmeyer: Package update for Python >= 3.7 and safety issue kept.

KSchopmeyer: This turned out to be more complex than thought since each time I fixed something, another issue popped up.  That include new requirements for the safety policies, more issues in requirements and finally a bad release of the build package that kept hanging around long after the developers has supposedly pulled that version.

andy-maier:
I added a commit that fixes the remaining issues.

andy-maier (approved):

PR #3115, opened 2024-03-11
Fixed use of pytest.warns(None)

andy-maier:


KSchopmeyer (approved):
I came to about same code.

PR #3121, opened 2024-03-17
Add fixes for new safety issues, 17 Mar 24.

KSchopmeyer:
Modified safety policy dev-requirements, and minimum constraints for new issues with:

1. cryptograpy - New minimum 42.0.2 for python > 3.7
2. Jupyter Notebook - New Minimum 7.07 Python > 3.8
3. jupyter-server - recommendation of minim version from Notebook safety issue.

The new safety issues are as follows:

 -> Vulnerability ID: 65183 Affected spec: >=7.0.0,<=7.0.6 ADVISORY: CVE-2024-22420 describes a vulnerability in Jupyter Notebook, where user interaction with a malicious notebook or Markdown file
enables an attacker to access and act with the same permissions as the user.
The flaw lies in the table of contents plugin. Jupyter Notebook v7.0.7
includes a patch for this issue. Users can manually disable the plugin as a
   workaround.
https://github.com/jupyterlab/jupyterlab/security/advisories/GHSA-4m77-cmpx-
   vjc4
   CVE-2024-22420
   For more information, please visit
   https://data.safetycli.com/v/65183/f17

   NOTES:
     New minimum version: 7.0.7
     7.0.2 python 3.8, 3.11
    7.0.7, first version with documented python 3.12 support. However, there is not version 7.0.7 so next version is 7.1.0. Previously the classiifier was python 3:;

-> Vulnerability found in notebook version 7.0.2
   Vulnerability ID: 65182
   Affected spec: >=7.0.0,<=7.0.6
   ADVISORY: CVE-2024-22421 is a vulnerability in Jupyter Notebook where
clicking a malicious link could expose Authorization and XSRFToken tokens to
third parties in older jupyter-server versions. Patched versions include
notebook above 7.0.7. Users are advised to upgrade jupyter-server to version
2.7.2 or newer, which includes a fix for a redirect vulnerability. No other
workaround has been identified.
https://github.com/jupyterlab/jupyterlab/secu
   rity/advisories/GHSA-44cc-43rp-5947
   CVE-2024-22421
   For more information, please visit
   https://data.safetycli.com/v/65182/f17

   Note: NEW MINIMUM 7.0.7. However, there is no 7.0.7. Next released version is 7.1.1.  Also proposes jupyter-server >= 2.7.2

-> Vulnerability found in cryptography version 41.0.6
   Vulnerability ID: 65212
   Affected spec: >=35.0.0,<42.0.2
   ADVISORY: Versions of Cryptograph starting from 35.0.0 are
[...]

Issue #3119, opened 2024-03-12
Remove max version on build package in dev-requirements.txt


KSchopmeyer:
Issue #3112 describes a problem that occurs in the CI build step when the max version of the build package is set to <1.1.0 in dev-requirements.txt where build was injecting the wheel package since setuptools also injects the wheel package.

The issue appears with the step:

    python -m build --wheel --outdir dist -C--universal .

and with the messages:

  ImportError: wheel package is not installed and bdist_wheel command is used
  
  ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel
  make: *** [Makefile:678: dist/pywbem-1.7.0.dev1-py2.py3-none-any.whl] Error 1
  Error: Process completed with exit code 2.

However it simply changes the max version of build to before the wheel injection code was removed from the build package (version < 1.03 ).

This issue documents the requirement to remove that limitation on the build version once we change the setup so that the decision on the the wheel is executed in setuptools.

PR #3122, opened 2024-03-18
Change build pkg max version limit.

KSchopmeyer:
Change the build pkg max version from < 1.0  to <= 1.0, the last version before the wheel injection code was removed from build pkg so that setuptools injects the wheel pkg as documented in issue #3119.

Based on PR #3121 - (https://github.com/pywbem/pywbem/pull/3121)


KSchopmeyer:
Close as won't do. It appears to not gain us anything until we make the changes to inject the wheel package through setuptools itself.

PR #3124, opened 2024-03-18
Split safety runs into install and all; Dependencies cleanup

andy-maier:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,11 @@ else
   endif
 endif
 
+# Run type (normal, scheduled, release)
+ifndef RUN_TYPE
+  RUN_TYPE := normal
+endif
+
 # Make variables are case sensitive and some native Windows environments have
 # ComSpec set instead of COMSPEC.
 ifndef COMSPEC
@@ -239,8 +244,9 @@ pylint_rc_file := pylintrc
 pylint_todo_opts := --disable=fixme
 pylint_no_todo_opts := --enable=fixme
 
-# Safety policy file
-safety_policy_file := .safety-policy.yml
+# Safety policy files
+safety_install_policy_file := .safety-policy-install.yml
+safety_all_policy_file := .safety-policy-all.yml
 
 # Flake8 config file
 flake8_rc_file := .flake8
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -358,7 +359,7 @@ help:
 	@echo "  check      - Run Flake8 on sources"
 	@echo "  pylint     - Run PyLint on sources"
 	@echo "  installtest - Run install tests"
-	@echo "  safety      - Run Safety on minimum_constraints.txt"
+	@echo "  safety     - Run Safety for install and all"
```
KSchopmeyer, line 362: Suggest modifying to include full file names.  This helps to clarify that safety is run on the minimum constraints files.

KSchopmeyer (commented):
I make one specific comment and have a couple of others below:

1. The word install may not be clear to others.  I assumed that this means files defined in the requirements that are used by the "make install". Also, it is not the same list as in the requirements.txt file since it includes setuptools, etc.  It obviously includes all package depencies in that are part of the pip package.
2. Should we have any documentation in the docs about this change either the *.md files or the pywbem documentation.
3. This is primarily to keep the tests running if only test/dev packages are failing.  However, I assume that we will still NOT release with any safety policy issues reporting either all or install.

Makefile
andy-maier: That would be too long for a half-line description. Do you have a proposal for something that fits into this half line?

andy-maier:
On your comments:
1. Your assumption that "install" means the files defined in the requirements that are used by the "make install" is correct. The pip/setuptools/wheel are also used during install. install = pip+dependents + pywbem+dependents.
2. I searched the documentation and we did not document safety so far. I kept it that way.
3. With this approach, fixing the issues from the safety development run has become voluntary. For the users of the released package, it does not matter whether safety development issues are solved, as long as safety install issues are solved. Having said that: We could probably find some logic that enables the return code checking of the safety development run for release checks. But I would not want to have that enabled in the weekly runs. What do you think?

KSchopmeyer:
I set review so you would read comments above.

Makefile
KSchopmeyer: I tried several things an lets forget my comment.  This is really for us.

KSchopmeyer (commented):
One more optional proposal.  Since pipdeptree has that little issue that it should be run as python ... how about putting that into Makefile with target pipdep or something.

With the comments about statement about what install means in the ...install.txt file, I approve of this one.

andy-maier:
We discussed this today, and agreed that we would use the approach from zhmcclient here as swell.

dev-requirements.txt
```diff
@@ -1,15 +1,12 @@
...
-# the minimum versions specified in minimum-constraints.txt.
+# Pip requirements file for development dependencies.
 
 
+# Include test dependencies (which includes install dependencies)
 -r test-requirements.txt
 
-# Direct dependencies:
+
+# Direct dependencies for development and indirect dependencies for development
```
KSchopmeyer, line 8: Pull these into one line for readability

andy-maier: That makes the line longer than my terminal width.

PR #3125, opened 2024-03-22
Documentation changes for urllib3 ver 2 support.

KSchopmeyer:
This change does not modify any code. It extends the description of the issues around the move to urllib3 and the corresponding incompatibilities.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -488,21 +488,6 @@ Troubleshooting
 This section describes some trouble shooting hints for the installation of
 pywbem.
 
-.. index:: pair; troubleshooting: OpenSSL
-.. index:: pair; installation fail: OpenSSL
-
-
-NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. index:: pair; installation fail: NotOpenSSLWarning
-.. index:: pair; troubleshooting: NotOpenSSLWarning
-
-This issue may be caused by the dependent package urllib3 update to version 2.0
-in pywbem version 1.7.0. In this case it is probably that the local environment
-includes a version of OpenSSL less than 2.0 Again the best options is to
-reinstall urllib3 < 2.0
-
 Installation fails with "invalid command 'bdist_wheel'"
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -524,68 +509,106 @@ To fix this, install the Python "wheel" package::
 
     pip install wheel
 
-ConnectionError raised with [SSL: UNSUPPORTED_PROTOCOL]
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. index:: pair; troubleshooting: OpenSSL
+.. index:: pair; installation fail: OpenSSL
+
+NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. index:: pair; installation fail: NotOpenSSLWarning
+.. index:: pair; troubleshooting: NotOpenSSLWarning
+
+This issue is probably caused by the dependent Python package urllib3 update to
+version 2.0 in pywbem version 1.7.0. In this case probably the local OS
+environment includes a version of OpenSSL less than 2.0 or another SSL
+implementation.
+
+See :ref:`ConnectionError raised with SSL UNSUPPORTED_PROTOCOL` for
+more information.
+
+
+.. _ConnectionError raised with SSL UNSUPPORTED_PROTOCOL:
+
+ConnectionError raised with SSL UNSUPPORTED_PROTOCOL
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. index:: pair; troubleshooting: OpenSSL
 .. index:: pair; SSL: OpenSSL
+.. index:: single LibreSSL
+.. index:: single: TLS
 .. index:: pair; UNSUPPORTED_PROTOCOL: OpenSSL
 
 On newer versions of the operating system running the pywbem client,
-communication with the WBEM server may fail with::
+communication with the WBEM server may fail with an exception similar to::
 
     pywbem.exceptions.ConnectionError: SSL error <class 'ssl.SSLError'>:
       [SSL: UNSUPPORTED_PROTOCOL] unsupported protocol (_ssl.c:1056)
 
-In pywbem version 1.7.0 the urllib3 configuration version limit was modified to allow
-urllib3 versions >= 2.0.  These versions of urllib3 limit the minimum TLS
-version to 1.2 (i.e OpenSSL version >= 1.1.1). If the version of OpenSSL is
-less than 1.1.1, this SSLError will occur with the initial request to the
-WBEM Server. Urllib3 version <= 2.0 also limits the SSL library implementations
-to just OpenSSL and possibly libreSSL.
+This error means that the WBEM server side SSL implementation does not yet support
+TLS 1.2 or higher or that the SSL library (OpenSSL or LibreSSL) used
+by pywbem does not support TLS 1.2 and the other side requires TLS >= 1.2.
 
-This also happened after an upgrade of the client OS to Debian buster
+See the Python document
+`https://peps.python.org/pep-0644/ <PEP 644 – Require OpenSSL 1.1.1 or newer>`_
+for more information on Python and OpenSSL version 1.1.1.
+
+In pywbem version 1.7.0 the urllib3 configuration version limit was modified to
+allow urllib3 package versions >= 2.0.  These new versions of urllib3 include backward
+incompatible changes including:
+
+* Support limiting the minimum TLS version to >= 1.2 (i.e OpenSSL version >= 1.1.1). If
+  the version of OpenSSL is less than 1.1.1, this SSLError will occur with the
+  initial request to the WBEM Server.
+* Limit the SSL library implementations to just OpenSSL and some versions of
+  LibreSSL.  The  version 2.0+ urllib3 implementation does not support any
+  other implementations of SSL.
+
+This also happens after an upgrade of the client OS to Debian buster
 using Python 3.7, with OpenSSL 1.1.1d.
 
-This is an error that is created by the OpenSSL library and handed back up to
-the SSL module of Python which hands it up to pywbem. The error indicates that
-OpenSSL and the WBEM server do not agree about which SSL/TLS protocol level to
-use.
+This is an error that is created by the SSL library (normally OpenSSL or
+LibreSSL) and handed back up to the SSL module of Python which hands it up to
+pywbem. The error indicates that OpenSSL and the WBEM server do not agree about
+which SSL/TLS protocol level to use.
 
 Pywbem specifies SSL parameters such that the highest SSL/TLS protocol version
-is used that both the client and server support. Thus, pywbem does not put any
-additional restrictions on top of OpenSSL.
+is used that both the client and server support. Pywbem itself does not put any
+additional TLS version restrictions on top of SSL library.
 
 Debian buster includes OpenSSL 1.1.1d and increased its security settings to
-require at least TLS 1.2 (see https://stackoverflow.com/a/53065682/1424462).
+require at least TLS 1.2 (see `https://stackoverflow.com/a/53065682/1424462`).
 
-This error means that the WBEM server side does not yet support
-TLS 1.2 or higher or that the SSL library used by pywbem does not support
-TL 1.2.
+This issue can possibly be corrected by:
 
-This issue can be corrected by:
-
-1. If the current version of urllib3 is less than 2.0, update the version
-   of urllib3 (ex. ``pip install --upgrade  --upgrade-strategy eager urllib3``). Since
-   the pywbem install does not force an eager update of packages, if a valid
-   previous version of urllib3 exists it will not be upgraded to 2.0+ in the
-   reinstallation of pywbem.
+1. If the current version of urllib3 is less than 2.0 (``pip list``), update
+   the version of urllib3 (ex. ``pip install --upgrade  --upgrade-strategy
+   eager urllib3``). to a later version. This is required because pip install
+   does not force an eager update of packages; if a valid previous version (
+   i.e. any version in the range defined by the dev-requirements.txt file) of
+   urllib3 exists it will not be upgraded to 2.0+ in the reinstallation of
+   pywbem.
 
 2. If the current version of urllib3 is greater than 2.0, the previous version
    of urllib3 (ex. version 1.26.5) can be installed (ex. ``pip install
-   urllib3<2.0``)
+   urllib3<2.0``).
 
-3. Adding TLS 1.2 support to the server side (preferred) or by lowering the
-    minimum TLS level OpenSSL requires on the client
-    side (which lowers security). The latter can be done by changing the
-    ``MinProtocol`` parameter in the OpenSSL config file on the client OS
-    (typically ``/etc/ssl/openssl.cnf`` on Linux and OS-X,
-    and ``C:\OpenSSL-Win64\openssl.cnf`` on Windows).
-    At the end of the file there is::
+3. Adding TLS 1.2 support to the WBEM server side (preferred) or lowering the
+   minimum TLS level required on the client side (which lowers security). With
+   OpenSSL the latter can be done by changing the ``MinProtocol`` parameter in
+   the OpenSSL config file on the client OS (typically ``/etc/ssl/openssl.cnf``
+   on Linux and OS-X, and ``C:\OpenSSL-Win64\openssl.cnf`` on Windows).
+
+   At the end of the file there is::
 
         [system_default_sect]
         MinProtocol = TLSv1.2
         CipherString = DEFAULT@SECLEVEL=2
+
+4. If the issue is the use of LibreSSL (ex. MacOS) as the OS level SSL
+   implementation, the issue may be that urllib3 before version 2.0.3 did not
+   support LibreSSL.  See `urllib3 issue <https://github.com/urllib3/urllib3/issues/3020>`_
+   for discussion of this issue.
 
 
 ConnectionError raised with [SSL] EC lib
@@ -650,10 +673,11 @@ directories if required including:
   since environment variables can be used to define pip command line options.
 
 * Remove the flag file that pip uses to enable the limiting behavior. See
-  `Python PEP 668`_ which would be most logical in the case of installation
-  into a container such as Docker.
+  `Python PEP 668`_ ("Marking Python base environments as “externally managed")
+  which would be most logical in the case of installation into a container such
+  as Docker.
 
-See `pywbem issue #3080 <https://github.com/pywbem/pywbem/issues/3080>`_ for
+See `pywbem issue 3080 <https://github.com/pywbem/pywbem/issues/3080>`_ for
 more information about this issue.
 
 .. _Python PEP 668: https://peps.python.org/pep-0668/
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -510,110 +510,150 @@ To fix this, install the Python "wheel" package::
     pip install wheel
 
 
+.. _`Issues with pywbem verson 1.7+, Urllib3 package version 2, and SSL`:
+
+Issues with pywbem verson 1.7+, Urllib3 package version 2, and SSL
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _`Overview of the changes and issues`:
+
+Overview of the changes and issues
+""""""""""""""""""""""""""""""""""
+
 .. index:: pair; troubleshooting: OpenSSL
-.. index:: pair; installation fail: OpenSSL
+.. index:: pair; SSL: OpenSSL
+.. index:: pair; UNSUPPORTED_PROTOCOL: OpenSSL
+
+Pywbem version 1.7 expanded the urllib3 package version requirement to include
+urllib3 version >= 2.0. Note that urllib3 version 2.0 was allowed with previous
+version of pywbem only because the package version requirements did not pin
+urllib version to < 2.0.
+
+The urlib3 package version requirements in requirements.txt are::
+
+      urllib3>=1.26.18; python_version >= '3.7'  (pywbem 1.7.0)
+
+previously::
+
+      urllib3>=1.26.18,<2.0.0; python_version >= '3.7' (pywbem 1.6.x)
+
+
+The urllib3 Python package version 2.0 made a significant number of changes
+including the following list that can be potentially non-compatible:
+
+* Urllib3 >= 2.0 limits the minimum OpenSSL version >= 1.1.1. Note that
+  Python 3.10 also limits the minimum OpenSSL version >= 1.1.1.
+
+* Urllib3 >= 2.0 changes the default minimum TLS version to TLS 1.2 (previously
+  was TLS 1.0), the default minimum defined for OpenSSL 1.1.1.
+
+* Urllib3 >= 2.0 removed support for LibreSSL, WolfSSL and other non OpenSSL
+  implementations.  Multiple operating systems including: macOS and BSD use
+  LibreSSL as the TLS/SSL implementation. LibreSSL support however, was
+  restored to urllib3 in version 2.0.3.
+
+* Urllib3 >= 2.0 changed the minimum version for Python to Python 3.7.
+
+* Urllib3 >= 2.0 removed the default set of TLS ciphers, instead now urllib3
+  uses the list of ciphers configured by the system.
+
+This means that pywbem exceptions can occur during installation, update of pywbem
+or the underlying platform because:
+
+* Level of minimum TLS protocol support depends on the Python Version and the
+  SSL implementation version.
+
+* SSL compatibility issues can happen with new pywbem installations on some
+  platforms or updates to the OS platforms or to pywbem.
+
+Pywbem uses the requests dependent package which uses the urllib3 package which
+uses the  Python openssl module to communicate with OpenSSL; neither pywbem nor the
+requests package pins the urllib3 package to versions below version 2.0. This
+is a function of the SSL implementation and urllib3 package version.
+
+Python version 3.10+ already requires OpenSSL version 1.1.1 or higher
+independent of the use of urllib3 version 2.0+ and pywbem version 1.6 allowed
+urllib3 version 2.0+ simply because it was released before that urllib3 version
+was released.
+
+Because of this pywbem change to allow urllib3 >= 2.0, Warnings, Exceptions,
+etc. can occur in the process of:
+
+* pywbem package installation
+* pywbem package update ( ie. updating urllib3)
+* Update to use a different version of Python, ex. Python 3.6 to a newer Python version
+* OS update which changes the SSL implementation version
+* Pywbem attempting to communicate with a WBEM server where the server may
+  not implement the new TLS version requirement.
+
+The following documents present more information on this change to urllib3 and
+the issues and possible corrections:
+
+* PYTHONPEP644_
+
+* `Urllib3 migration guide <https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html>`_
+
+* `macOS issues with Urllib3 - urllib3 issue 3024 <https://github.com/urllib3/urllib3/issues/3020>`_
+
+* `urllib3 issue 2168, Drop support for OpenSS\<1.1.1 <https://github.com/urllib3/urllib3/issues/2168>`_
+
+The current version and implementation  of the SSL library can be determined
+as follows::
+
+  $ python -c "import ssl; print(ssl.OPENSSL_VERSION)"
+  OpenSSL 1.1.1  7 Feb 2023
+
+or::
+
+  $ python -c "import ssl; print(ssl.OPENSSL_VERSION)"
+  LibreSSL 2.8.3
+
+OpenSSL version can be directly determined from OpenSSL as follows::
+
+    $ openssl version
+    OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
+
+The current urllib3 installed version can be determined with pip::
+
+  $ pip list | grep urllib3
+  urllib3                       2.2.1
+
+The server TLS version can be verified with the OpenSSL CLI tool::
+
+    openssl s_client -connect localhost:15989 -XXXX
+
+    where: XXXX is tls1_1, tls1_2
+    and a certificate is returned if the corresponding TLS version is valid.
+
+The following subsections define specific issues that can occur with this
+change and proposed solutions.
+
+* :ref:`NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+`
+* :ref:`ConnectionError raised with [SSL] EC lib`
+* :ref:`ConnectionError raised with [SSL: UNSUPPORTED_PROTOCOL]`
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Currently failing because of new safety issues in cryptography package.

docs/appendix.rst
```diff
@@ -524,69 +509,104 @@ To fix this, install the Python "wheel" package::
...
+TLS 1.2 or higher or that the SSL library (OpenSSL or LibreSSL) used
+by pywbem does not support TLS 1.2 and the other side requires TLS GE 1.2.
 
-This also happened after an upgrade of the client OS to Debian buster
+See the Python document
+`https://peps.python.org/pep-0644/ <PEP 644 – Require OpenSSL 1.1.1 or newer>`_
+for more information on Python and OpenSSL version 1.1.1.
+
+In pywbem version 1.7.0 the urllib3 configuration version limit was modified to
+allow urllib3 package versions GE 2.0.  These new versions of urllib3 include backward
```
andy-maier, line 557: suggest to use ">=" instead of "GE".

docs/appendix.rst
```diff
@@ -524,69 +509,104 @@ To fix this, install the Python "wheel" package::
...
-This issue can be corrected by:
+This issue can possibly be corrected by:
 
 1. If the current version of urllib3 is less than 2.0, update the version
-   of urllib3 (ex. ``pip install --upgrade  --upgrade-strategy eager urllib3``). Since
-   the pywbem install does not force an eager update of packages, if a valid
-   previous version of urllib3 exists it will not be upgraded to 2.0+ in the
-   reinstallation of pywbem.
+   of urllib3 (ex. ``pip install --upgrade  --upgrade-strategy eager urllib3``).
+   Since the pywbem install does not force an eager update of packages, if a
```
andy-maier, line 586: This sentence would leave me as a user a bit unsure about whether it states something I need to act upon. What it really means is to give the reason why urllib needs to be upgraded explicitly. I suggest to rephrase the sentence to clarify that it gives the reason for that, e.g. by starting it with "This is needed because..."

docs/appendix.rst
```diff
@@ -524,69 +509,104 @@ To fix this, install the Python "wheel" package::
...
+
+   At the end of the file there is::
 
         [system_default_sect]
         MinProtocol = TLSv1.2
         CipherString = DEFAULT@SECLEVEL=2
 
+4. If the issue is the use of LibreSSL as the OS level SSL implementation, the
+   issue may be the LibreSSL version or the urllib3 version as versions of
+   urllib3 less than 2.0.3 did not support libreSSL.
```
andy-maier, line 608: I needed to read this a few times to understand what it means.

Suggestion: "If LibreSSL is used as the OS level SSL implementation, the issue may be that urllib3 before version 2.0.3 did not support LibreSSL."

If there is an additional requirement on LibreSSL versions to be used, we should state that separately (and with the explicit versions supported).

docs/changes.rst
```diff
@@ -47,8 +48,8 @@ Released: not yet
     NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the
            'ssl' module is compiled with 'LibreSSL 2.8.3'
 
-  Also, the default set of cyphers available has changed to use the list of
-  ciphers configured by the system rather than ciphers defined by urllib3.
+  Also, the default set of cyphers available in OpenSSL 1.1.1 has changed to use the 
+  list of ciphers configured by the system rather than ciphers defined by urllib3.
```
andy-maier, line 52: Can we please shorten this change log entry? It does not need to be a repetition of the trouble shooting section.

andy-maier (changes requested):

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: I added a section based on the ImportError urlliber ... issue.

andy-maier:
The safety issues and Cython build issues have been fixed. Please resolve the remaining comment and rebase on latest master.

docs/changes.rst
KSchopmeyer: done

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+by pywbem does not support TLS 1.2 and the other side requires TLS >= 1.2.
+
+See the Python document
+`Python PEP 644 – Require OpenSSL 1.1.1 or newer <https://peps.python.org/pep-0644/>`_
+for more information on Python and OpenSSL version 1.1.1.
+
+In 'pywbem' package version 1.7.0 the urllib3 package configuration version
+limit was modified to allow urllib3 package versions >= 2.0.  These new
+versions of the 'urllib3' package include backward incompatible changes
+including:
```
andy-maier, line 559: In pywbem 1.6.0, urllib3 was not pinned, which means that urllib3 2.0 can be used with it. So this is not something that pywbem 1.7.0 introduced, it is rather tied to the release of urllib3 2.0.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
 
-This also happened after an upgrade of the client OS to Debian buster
+* Support limiting the minimum TLS version to >= 1.2 (i.e OpenSSL version >= 1.1.1).
```
andy-maier, line 561: "Support" to me suggests that it is not mandatory, i.e. OpenSSL < 1.1.1 can be used with urllib3 >= 2.0. However, the next sentence below says urllib3 2.0 requires OpenSSL >= 1.1.1. 

Also, I find it somehow confusing to start talking about the consequences (TLS version) first, before the reader gets the information why (OpenSSL version), but I guess I can live with that.

Finally, the next list item says that some versions of LibreSSL are also suported by urllib3 2.0, but it misses to say which and also what the consequences on TLS support are.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+  If the version of OpenSSL is less than 1.1.1, this SSLError the error
+  :ref:`ImportError urllib3 v2.0 only supports OpenSSL 1.1.1+` will occur with
+  the initial request to the WBEM Server.
+* Limit the SSL library implementations to just OpenSSL and some versions of
+  LibreSSL.  The  'urllib3' version 2.0+  implementation initial release does
```
andy-maier, line 566: On "The 'urllib3' version 2.0+ implementation initial release": So this is a statement about only 2.0 then? If so, the "+" and "initial" can be removed.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+  not support any other implementations of SSL.
+
+This issue also happens after an upgrade of the client OS to Debian buster
 using Python 3.7, with OpenSSL 1.1.1d.
```
andy-maier, line 570: Why do we call out a specific Python version and a specific OS, when the reason is that OpenSSL 1.1.1d is used. Does that not fall into the category "OpenSSL >= 1.1.1"?

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+This is an error that is created by the SSL library (normally OpenSSL or
+LibreSSL) and handed back up to the SSL module of Python which hands it up to
+pywbem. The error indicates that OpenSSL and the WBEM server do not agree about
+which SSL/TLS protocol level to use.
 
 Pywbem specifies SSL parameters such that the highest SSL/TLS protocol version
-is used that both the client and server support. Thus, pywbem does not put any
-additional restrictions on top of OpenSSL.
+is used that both the client and server support. Pywbem itself does not put any
+additional TLS version restrictions on top of SSL library.
```
andy-maier, line 579: insert "the" before "SSL library".

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
 
 Debian buster includes OpenSSL 1.1.1d and increased its security settings to
-require at least TLS 1.2 (see https://stackoverflow.com/a/53065682/1424462).
-
-This error means that the WBEM server side does not yet support
-TLS 1.2 or higher or that the SSL library used by pywbem does not support
-TL 1.2.
+require at least TLS 1.2 (see `https://stackoverflow.com/a/53065682/1424462`).
```
andy-maier, line 582: Again, IMO we can remove the statement about Debian buster.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+   eager urllib3``). to a later version. This is required because pip install
+   does not force an eager update of packages; if a valid previous version (
+   i.e. any version in the range defined by the dev-requirements.txt file) of
+   urllib3 exists it will not be upgraded to 2.0+ in the reinstallation of
+   pywbem.
 
 2. If the current version of urllib3 is greater than 2.0, the previous version
    of urllib3 (ex. version 1.26.5) can be installed (ex. ``pip install
-   urllib3<2.0``)
+   urllib3<2.0``).
```
andy-maier, line 596: In the requirements.txt file in the current master branch, we are pinning urllib3 to <2.0 on Python <= 3.6.

1. Why are we doing that - can the pinning not be removed?
2. We should make the requirements.txt file and this statement here consistent.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+In 'pywbem' package version 1.7.0 the urllib3 package configuration version
+limit was modified to allow urllib3 package versions >= 2.0.  These new
+versions of the 'urllib3' package include backward incompatible changes
+including:
 
-This also happened after an upgrade of the client OS to Debian buster
+* Support limiting the minimum TLS version to >= 1.2 (i.e OpenSSL version >= 1.1.1).
+  If the version of OpenSSL is less than 1.1.1, this SSLError the error
+  :ref:`ImportError urllib3 v2.0 only supports OpenSSL 1.1.1+` will occur with
```
andy-maier, line 563: This now mentions a second exception in the description of the first exception. The trouble shooting section is organized by symptoms that can occur, so I suggest to move this into its own section and not mention it here.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
-   the pywbem install does not force an eager update of packages, if a valid
-   previous version of urllib3 exists it will not be upgraded to 2.0+ in the
-   reinstallation of pywbem.
+1. If the current version of urllib3 is less than 2.0 (``pip list``), update
+   the version of urllib3 (ex. ``pip install --upgrade  --upgrade-strategy
+   eager urllib3``). to a later version. This is required because pip install
+   does not force an eager update of packages; if a valid previous version (
+   i.e. any version in the range defined by the dev-requirements.txt file) of
+   urllib3 exists it will not be upgraded to 2.0+ in the reinstallation of
+   pywbem.
```
andy-maier, line 592: Why does the upgrade of urllib3 solve a TLS version mismatch? Is it not rather the upgrade of the SSL library?

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
-    side (which lowers security). The latter can be done by changing the
-    ``MinProtocol`` parameter in the OpenSSL config file on the client OS
-    (typically ``/etc/ssl/openssl.cnf`` on Linux and OS-X,
-    and ``C:\OpenSSL-Win64\openssl.cnf`` on Windows).
-    At the end of the file there is::
+3. Adding TLS 1.2 support to the WBEM server side (preferred) or lowering the
+   minimum TLS level required on the client side (which lowers security). With
+   OpenSSL the latter can be done by changing the ``MinProtocol`` parameter in
+   the OpenSSL config file on the client OS (typically ``/etc/ssl/openssl.cnf``
+   on Linux and OS-X, and ``C:\OpenSSL-Win64\openssl.cnf`` on Windows).
```
andy-maier, line 602: Can the minimum TLS version be lowered that way with OpenSSL >= 1.1.1? The previous text suggests that is not the case.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+   At the end of the file set the following::
 
         [system_default_sect]
         MinProtocol = TLSv1.2
         CipherString = DEFAULT@SECLEVEL=2
 
+4. If the issue is the use of LibreSSL (ex. MacOS) as the OS level SSL
+   implementation, the issue may be that urllib3 before version 2.0.3 did not
+   support LibreSSL.  See  `urllib3 issue 2020: <https://github.com/urllib3/urllib3/issues/3020>`_
```
andy-maier, line 612: Remove the colon after 2020.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+.. index:: LibreSSL macOS
+.. index:: single: TLS
+.. index:: pair; ImportError urllib3: OpenSSL
+
+The 'urllib3' Python package version 2.0 removed support for LibreSSL and
+some other SSL implementations and requires OpenSSL 1.1.1 or higher.  LibreSSL
+support was restored to 'urllib3' in version 2.0.3.
+
+The 'pywbem' package uses the 'requests' package which uses 'urllib3';
+neither 'pywbem' nor 'requests' pins 'urllib3' to stay below version 2.0.
```
andy-maier, line 633: The pywbem master branch does pin urllib3 to <2.0 on Python >=3.6.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+Therefore, upgrading Python packages, with a Python that does not support
+OpenSSL 1.1.1 or higher or other SSL implementations, causes the following
+exception raised by urllib3:
```
andy-maier, line 637: According to the earlier text, this happens only in urllib3 2.0.0 - 2.0.2.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+    ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the ‘ssl’ module is compiled with LibreSSL 2.8.3.
+    See: https://github.com/urllib3/urllib3/issues/2168
+
+This can happen for example on macOS with the system Python of
+macOS as the basis for a Python virtual environment and installing pywbem
+into that virtual environment, which typically installs the latest available
+versions of dependent packages, and thus may install urllib3 with a version 2.0
+or later.
```
andy-maier, line 646: This paragraph is mostly a duplication of the paragraph "Therefore, upgrading Python packages,...".

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+  $ /usr/bin/python3 -c "import ssl; print(ssl.OPENSSL_VERSION)"
+  LibreSSL 2.8.3
+
+Note that Python since version 3.10 requires OpenSSL version 1.1.1 or higher
+(see PEP-644).
+
+At least up to macOS Ventura, Apple compiles the system Python with LibreSSL.
+As long as that does not change, you cannot use the system Python of macOS with
+urllib3>=2.0; also not as a basis for Python virtual environments.
```
andy-maier, line 666: Apparently, urllib3 2.0.3 re-established the support for LibreSSL.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+
+* Pin the urllib3 package to stay below version 2.0 when on Python 3.7 or
+  higher, by specifying in the package dependencies, e.g. in the
+  requirements.txt file::
+
+    urllib3>=1.26.5,<2.0; python_version >= '3.7'
+
+The minimum version of urllib3 should be at least what the
+minimum-constraints.txt file of the 'pywbem' project specifies as a minimum,
+for the 'pywbem' version.
```
andy-maier, line 683: For users of pywbem, it does not matter what we have in minimum-constraints.txt. What matters is what pywbem has in its package metadata, which comes from requirements.txt. I suggest not to talk about the requirements files at all, because it is simply the pywbem package requirements that matter, and pip will surface an error if it cannot match the requirements of the pywbem user with the requirements of pywbem.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+  either be downloaded from https://www.python.org/downloads/macos/ or
+  installed using a third party package installer for macOS, such as Homebrew.
+
+* Pin the urllib3 package to stay below version 2.0 when on Python 3.7 or
+  higher, by specifying in the package dependencies, e.g. in the
+  requirements.txt file::
+
+    urllib3>=1.26.5,<2.0; python_version >= '3.7'
+
+The minimum version of urllib3 should be at least what the
```
andy-maier, line 681: This paragraph belongs to the second option, so I suggest to indent it by two spaces.

docs/appendix.rst
```diff
@@ -524,76 +509,190 @@ To fix this, install the Python "wheel" package::
...
+minimum-constraints.txt file of the 'pywbem' project specifies as a minimum,
+for the 'pywbem' version.
+
+Note that pinning a dependent package prevents installing security
```
andy-maier, line 685: This paragraph belongs to the second option, so I suggest to indent it by two spaces.

docs/changes.rst
```diff
@@ -32,25 +32,12 @@ Released: not yet
...
-    or
-
-    NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the
-           'ssl' module is compiled with 'LibreSSL 2.8.3'
-
-  Also, the default set of cyphers available has changed to use the list of
-  ciphers configured by the system rather than ciphers defined by urllib3.
-  See issue #1302 and the pywbem documentation
-  `trouble shooting section <https://pywbem.readthedocs.io/en/stable/appendix.html#troubleshooting>`_
+* Update to support python urllib3 package version 2.0+ (for python versions 3.7+)
```
andy-maier, line 35: Change "python" -> "Python"

docs/changes.rst
```diff
@@ -32,25 +32,12 @@ Released: not yet
...
+  opens the SSL library implementations (ex. OpenSSL, libreSSL)
+  that requuire TLS >=1.2 (OpenSSL versions >= 1.1.1). If an older
```
andy-maier, line 37: typo "requuire"

docs/changes.rst
```diff
@@ -32,25 +32,12 @@ Released: not yet
...
-
-    NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the
-           'ssl' module is compiled with 'LibreSSL 2.8.3'
-
-  Also, the default set of cyphers available has changed to use the list of
-  ciphers configured by the system rather than ciphers defined by urllib3.
-  See issue #1302 and the pywbem documentation
-  `trouble shooting section <https://pywbem.readthedocs.io/en/stable/appendix.html#troubleshooting>`_
+* Update to support python urllib3 package version 2.0+ (for python versions 3.7+)
+  opens the SSL library implementations (ex. OpenSSL, libreSSL)
```
andy-maier, line 36: On "opens": Is something missing in that sentence? It sounds strange this way.

docs/changes.rst
andy-maier, line 35: pywbem 1.6.0 did not pin urllib3 to <2.0 on any Python version, so pywbem 1.7.0 did not really open anything up. It was the release of urllib3 2.0 that introduced changes on the requirements for the SSL library.

andy-maier (changes requested):

kloczek:
Just humble question: as it is now added urllib3 2.x support is it not would be good to start cleaning code and remove python 2.x and python<=3.7 support dropping as well use of `six`? 🤔 
If yes maybe after next version (1.6.4?) would be good to start working on pywbem 2.x announcing that 1.6.x  will be backported only critical changes? 🤔 

PS. just checked and `pyupgrade --py38-plus` generates almost 1MB patch and ruff probably would be able to generate even more cleanups .. so it is a lot of things which could be dropped making whole code much cleaner.

docs/changes.rst
KSchopmeyer: Rewrote

KSchopmeyer:
Suggestion.  Would you move or copy the comment about 2.7 support to an issue.  It will get lost here in a particular pr.

We have always known that at some point we will have to drop this support but there are still users and we are trying to keep commonality as long as possible

KSchopmeyer:
Note: Completely rewrote the troubleshooting section on urllib3.

docs/appendix.rst
KSchopmeyer: Fixed as part of rewrite

docs/appendix.rst
KSchopmeyer: rewrote

docs/appendix.rst
KSchopmeyer: Good question.  I need to review that one.

docs/appendix.rst
KSchopmeyer: modified as part of rewrite.

docs/appendix.rst
KSchopmeyer: indented both paragraphs

docs/changes.rst
KSchopmeyer: No lower case python left in appendix

docs/changes.rst
KSchopmeyer: Fixed

docs/changes.rst
KSchopmeyer: Yes there was.  Rewritten

docs/appendix.rst
KSchopmeyer: NOTE: There is also one big TODO in the troubleshooting dealing with OpenSSL 1.1.1 and Python 3.10. I have not yet gone through the code to sort out the use of the lib/ssl package vs the pyOpenSSL package by urllib3

kloczek:
> Suggestion [kloczek](https://github.com/kloczek). Would you move or copy the comment about 2.7 support to an issue. It will get lost here in a particular pr.

Yep. Done #3138

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
 
     pip install wheel
 
+
+this option should not be the preferred one.
```
andy-maier, line 513: I think this line should be removed?

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+* Urllib3 >= 2.0 limits the minimum OpenSSL version >= 1.1.1.
+
+* Urllib3 >= 2.0 changes the default minimum TLS version to TLS 1.2 (previously was TLS 1.0)
+
+* Urllib3 >= 2.0 removed support for LibreSSL, WolfSSL and other SSL
+  implementations.  Multiple operating systems including: macOS and BSD use
+  LibreSSL as the TLS/SSL implementation. LibreSSL support however, was
+  restored to urllib3 in version 2.0.3.
+
+* Urllib3 >= 2.0 changed the minimum version for urllib3 version 2+ to Python 3.7.
```
andy-maier, line 548: I suggest to change:
"the minimum version for urllib3 version 2+ to Python 3.7."
to:
"the minimum version for Python to Python 3.7.".

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+
+* Urllib3 >= 2.0 removed the default set of TLS ciphers, instead now urllib3
+  uses the list of ciphers configured by the system.
+
+Pywbem uses the requests dependent package which uses the urllib3 package which
+uses the pyopenssl to communicate with OpenSSL;
```
andy-maier, line 554: What is "pyopenssl"?

There is no Python module with that name. There is a Pypi package named "pyOpenSSL" but that is not installed with pywbem.

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+.. index:: pair; UNSUPPORTED_PROTOCOL: OpenSSL
+
+Pywbem version 1.7 expanded the urllib3 package version requirement to include
+urllib version >= 2.0.
+
+The urlib3 package version requirements are in requirements.txt and are::
+
+      urllib3>=1.26.18; python_version >= '3.7'  (pywbem 1.7.0)
+
+      previously:
```
andy-maier, line 532: Please outdent "previously:"

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+uses the pyopenssl to communicate with OpenSSL;
+neither pywbem nor requests pins the urllib3 package to versions below version 2.0.
+This is a function of the SSL implementation and pyopenssl.
+
+# TODO: This one bothers me.  It is the lib/ssl module which is a TLS/SSL wrapper
+# for socket objects that implemented PEP 644,  however, we never had any issues
+# with new minimum TLS version until 3.1.12 so the path through requests was
+# not impacted or we would have had to add urllib 2.0 in earlier tests.  It appears
+# that for the paths we use, it is requests, urllib3, pyOpenSSL packages and
+# OpenSSL version that are the issue.  I need to do another code review.
```
andy-maier, line 563: I'm not sure I understand this commented out text. Is it not sufficient to state that Python 3.10 requires OpenSSL 1.1.1? That is in sync with what urllib3 2.0 requires.

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+the issues and possible corrections:
+
+* `Python PEP 644 – Require OpenSSL 1.1.1 or newer <https://peps.python.org/pep-0644/>`_
+  which defines the future direction of SSL in the Python world.
+
+* `Urllib3 migration guide <https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html>`_
+
+* `macOS issues with Urllib3 - urllib3 issue 3024 <https://github.com/urllib3/urllib3/issues/3020>`_
+
+* `urllib3 issue 2168, Drop support for OpenSS\<1.1.1 <https://github.com/urllib3/urllib3/issues/2168>`_
```
andy-maier, line 586: Fix "OpenSS" -> "OpenSSL"

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+
+The current version and implementation  of the SSL library can be determined
+as follows::
+
+  $ python -c "import ssl; print(ssl.OPENSSL_VERSION)"
+  OpenSSL 1.1.1  7 Feb 2023
+
+  or
```
andy-maier, line 594: Outdent "or"

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+
+  $ /usr/bin/python3 -c "import ssl; print(ssl.OPENSSL_VERSION)"
+  LibreSSL 2.8.3
```
andy-maier, line 597: Can we stay consistent with how Python is invoked, i.e. use just "python"?

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+
+NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. index:: pair; installation fail: NotOpenSSLWarning
+.. index:: pair; troubleshooting: NotOpenSSLWarning
+
+This issue may be caused by the urllib3 package update to version 2.0
+in pywbem version 1.7.0. In this case it is probably that the local environment
+includes a version of OpenSSL less than 2.0 The best options is to
```
andy-maier, line 627: "a version of OpenSSL less than 2.0"
Should that not be "a version of OpenSSL less than 1.1.1"?

docs/appendix.rst
```diff
@@ -524,8 +509,145 @@ To fix this, install the Python "wheel" package::
...
+
+.. _`Issues with pywbem verson 1.7+, Urllib3 package version 2, and SSL`:
+
+Issues with pywbem verson 1.7+, Urllib3 package version 2, and SSL
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. index:: pair; troubleshooting: OpenSSL
+.. index:: pair; SSL: OpenSSL
+.. index:: pair; UNSUPPORTED_PROTOCOL: OpenSSL
+
```
andy-maier, line 524: The text between here and the next heading explains background but is not a trouble shooting item. I realized only after re-reading it that the heading level changes for the next few trouble shooting headings. Should we add a heading here, e.g.:
```
Overview
""""""""
```

docs/changes.rst
```diff
@@ -32,27 +32,16 @@ Released: not yet
 * Installation of this package using "setup.py" is no longer supported.
   Use "pip" instead.
 
-* Update to support urllib3 package version 2.0+ (for python versions 3.7+)
+* Update to support python urllib3 dependent package version 2.0+ (for python versions 3.7+)
```
andy-maier, line 35: pywbem 1.6.0 did not yet pin urllib3 to <2.0. So the change in 1.7.0 w.r.t. urllib3 is to **introduce** a pinning of urllib3 to <2.0 for Python <=3.6. This change log entry does not correctly reflect that change.

Beyond that, the pinning of urllib3 to <2.0 in the current master branch really means that we are not supporting the use of Python<=3.6 with OpenSSL 1.1.1 (which seems to be a possible combination) or with the supported LibreSSL versions. I don't understand what would cause that restriction, and therefore I think that pinning of urllib3 to <2.0 on Python <=3.6 is incorrect and should be removed. Instead, we should exclude OpenSSL versions 2.0.0 - 2.0.2 since they do not support LibreSSL, at least on macos. So the requirements.txt file should become:
```
urllib3>=1.26.18; sys_platform != 'darwin'
urllib3>=1.26.18,!=2.0.0,!=2.0.1,!=2.0.2; sys_platform == 'darwin'
```
and the minimum-constraints.txt file:
```
urllib3==1.26.18
```
In that case, this change log entry would need to be adjusted.

Let's discuss this.

andy-maier (changes requested):

andy-maier:
Karl, I have marked my older comments from 5 days ago as resolved, but there are some newer comments from 2 days ago, so I have set the label back to "update needed".

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: Done

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: Agreed and done.  That helped it.

docs/appendix.rst
KSchopmeyer: That is my booboo. There arctually multiple interface libraries to the SSL implementation. the OpenSSL library in Python and th pyOpenSSL package which interfaces to OpenSSL

docs/appendix.rst
KSchopmeyer: Removed the comment. Note that pyopenssl was actually used as the OpenSSL interface until the Python OpenSSL module was added to Python and that interface is an alternate (not recommended) in the urllib3 package.  Noted that the TLS 1.2 requirement comes from either Python 3.8 and/or urllib3 v2

The confusion is that this is a lot of parts/version all flying in loose formation with minimal communication.

docs/appendix.rst
KSchopmeyer: DONE

docs/changes.rst
KSchopmeyer: Actually we did not really support urllib3 2.0+ because we released 1.6 without testing against that version. 

We pinned urllib3 to < 2.0 because the specifically stated that they do not support python 2.7, 3.5 and 3.6 with this new version.

We also made some changes although they were really in the just cim_http module and tests because of some exception issues. 

In any case, I agree that the whole statement is not clear since the change is really to limit the urllib3 version for Python <= 3.6.  The secondary effect was for us to specifically test with urllib3 version 2.0+ and determine

docs/appendix.rst
andy-maier: I fixed that in a small commit in this PR.

andy-maier (approved):

PR #3129, opened 2024-03-27
Reduced safety ignores to the minimum needed

andy-maier:


KSchopmeyer (commented):
My only comment is what are the "the minimum needed" criteria.  Since this passes the safety tests, I will approve

andy-maier:
It means they were ignores that were not needed in order to pass the safety check. They came in for older Python versions that we dropped support for meanwhile.

PR #3130, opened 2024-03-27
Fixed dependabot issues

andy-maier:


KSchopmeyer:
I am happy with the one change in the second commit.  Will approve when the split code is committed.

PR #3137, opened 2024-04-06
Ks/#3136 indication callback queue

KSchopmeyer:
Adds a queue between indication reception and executing the callback for each indication received.

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -128,6 +128,12 @@ WBEMListener class
 ------------------
 """
 
+# FUTURE: we should add the following:
+#  1. Max size for callback queue.  We do not want to kill the client
+#     Queue has a Full state so we can use that and stop the server from
+#     sending if the queue is full.
+#
+
 import sys
 import os
 import errno
@@ -138,6 +144,11 @@ import ssl
 import threading
 import atexit
 import getpass
+# Python 2.7 uses name Queue
+try:
+    import queue
+except ImportError:
+    import Queue as queue
 try:
     import termios
 except ImportError:
@@ -500,8 +511,8 @@ class ListenerRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
                             "but {0!A}", indication_inst))
                 return
             # server.listener created in WBEMListener.start function
-            self.server.listener.deliver_indication(indication_inst,
-                                                    self.client_address[0])
+            self.server.listener.handle_indication(indication_inst,
+                                                   self.client_address[0])
 
             self.send_success_response(msgid, methodname)
 
@@ -698,6 +709,29 @@ class ListenerRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
                        __version__, self.server_version, self.sys_version)
 
 
+class StoppableThread(threading.Thread):
+    """
+    Thread subclass with a stop() method. The thread itself must check
+    regularly for the stopped() event ( i.e. if stop_event.is_set()) and
+    exit the thread code.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Init with arguments for the thread
+        """
+        super(StoppableThread, self).__init__(*args, **kwargs)
+        self.stop_event = threading.Event()
+
+    def stop(self):
+        """Set the thread stop_event. To tell thread to stop"""
+        self.stop_event.set()
+
+    def stopped(self):
+        """Test for thread stop event set"""
+        return self.stop_event.is_set()
+
+
 # pylint: disable=too-many-instance-attributes
 class WBEMListener(object):
     """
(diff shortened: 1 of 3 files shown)


pywbem/_listener.py
```diff
@@ -1146,10 +1164,54 @@ def stop(self):
...
+        return
+
+    def handle_indication(self, indication, host):
+        """
+        Entry point from the listener with a single indication.
+        Either deliver to callback or put in interprocess queue
+        """
+        # Test code to allow testing with bypass of the queued return.
+        # Set this to False to test with direct callback.
+        queued_delivery = True
```
andy-maier, line 1203: I'm not sure I can relate this comment to the code in case of False.
In addition, I think we should have only one way implemented, for simplicity.

pywbem/_listener.py
```diff
@@ -814,6 +820,15 @@ def __init__(self, host, http_port=None, https_port=None,
...
         self._callbacks = []  # Registered callback functions
 
+        # set up queued callback queue and thread.
+        self.rcvd_indication_queue = queue.Queue()
+        self.callback_thread = threading.Thread(
+            target=self.deliver_indications_from_queue,
+            args=(self.rcvd_indication_queue, ))
+        self.callback_thread.daemon = True
+        self.delivery_stop = False
+        self._logger.debug("callback queue created")
```
andy-maier, line 830: I don't think we need this log entry.

pywbem/_listener.py
```diff
@@ -1146,10 +1164,54 @@ def stop(self):
             self._https_thread = None
             self.logger.info("Stopped threaded HTTPS server")
 
-    def deliver_indication(self, indication, host):
+        self.logger.info("Stopping queued delivery")
+        self.delivery_stop = True
+        self.logger.info("Stopped threaded Deliver queue")
```
andy-maier, line 1169: Where is the stop of the queue thread?

KSchopmeyer: The stop is used to stopthe loop in deliver_indications_from_queue(...).

The better question is where is the thread kill.  Answer, same place as the kill for http and https threads, i.e. when the the whole python program stops.  This should be fixed together with the other threads issue since the change will be common to all of them.

tests/unittest/pywbem/test_indicationlistener.py
```diff
@@ -688,11 +709,12 @@ def test_WBEMListener_send_indications(send_count):
             try:
                 response = post_bsl(url, headers=headers, data=payload)
             except requests.exceptions.RequestException as exc:
-                msg = ("Testcase sending indication #{} raised {}: {}".
-                       format(i, exc.__class__.__name__, exc))
+                msg = ("Testcase sending indication #{} raised {}: {}",
+                       i, exc.__class__.__name__, exc)
```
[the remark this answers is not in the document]
KSchopmeyer: DONE

tests/unittest/pywbem/test_indicationlistener.py
```diff
@@ -709,6 +731,7 @@ def test_WBEMListener_send_indications(send_count):
         # Make sure the listener thread has processed all indications
         sleep(1)
 
+        # TODO: Shouldn't this really be after the stop?
```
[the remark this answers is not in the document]
KSchopmeyer: Move done.

TODO: consistent log and display

pywbem/_listener.py
KSchopmeyer: Removed the direct callback and yes, it was wrong

tests/unittest/pywbem/test_indicationlistener.py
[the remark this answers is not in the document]
KSchopmeyer: WHOOPS.  We really want this output at the original level since it occurs for each test defined by the input parameters.  Ignore the proposal

KSchopmeyer:
NOTE: Currently there are probably way to many logging displays in the code but I am still trying to sort out the timing of the queue callback, etc and the logging helps.

tests/end2endtest/test_indications.py
```diff
@@ -71,9 +71,9 @@ def test_indications(wbem_connection):  # pylint: disable=redefined-outer-name
     # be much more consistent with python 3.7 and 3.7 PACKAGE_LEVEL=minimum
 
     if sys.version_info[0:2] in ((2, 7), (3, 7)):
-        skip("Skipping test_indications for python 2.7")
+        skip("Skipping test_indications for python 2.7, 3.7")
```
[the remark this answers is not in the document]
KSchopmeyer: ADDED comment for 3.7. Do not know exactly what reason was fro 2.7 but I think it was that it did not work on windows.

andy-maier (approved):

PR #3140, opened 2024-04-08
Increased actions/github-script to v7 to run with node 20

andy-maier:
No review needed.


andy-maier:
I verified that the warnings about node 16 have disappeared.
Merging the PR.

PR #3173, opened 2024-04-28
Try: Andy/fix macos py3637 macos12

andy-maier:


andy-maier:
No need to try this, as it was run that way n the weekly test run a week ago.

PR #3174, opened 2024-05-02
Migrated to pyproject.toml

andy-maier:

**Note:** This PR is on top of PR #3154 - merge only after that on

WIP TODOs:
* Support for "test" extra
* Support for Cythonizing


pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+# listed explicitly. Otherwise, building the wheel archive raises "Warning:
+# Package 'pywbem._vendor.nocaselist' is absent from the `packages`
+# configuration." and complains about an ambiguous configuration.
+packages = [
+    "pywbem",
+    "pywbem._vendor",
+    "pywbem._vendor.nocasedict",
+    "pywbem._vendor.nocaselist",
+    "pywbem_mock"
+]
```
abravalheri, line 21: You probably also can try:

```suggestion
packages = {find = {include = ["pywbem*"]}}
```

Ref: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages

or equivalently:

```toml
[tool.setuptools.packages.find]
include = ["pywbem*"]
```

abravalheri (commented):
> Support for Cythonizing

I believe that still depends on complementary config in `setup.py` (i.e. keep all the static configs in `pyproject.toml` and adding only the dynamic bits or Cython configs in `setup.py`).

pyproject.toml
andy-maier: I have changed the "packages" to:
```
[tool.setuptools.packages.find]
include = ["pywbem", "pywbem_mock"]
```
without using the `pywbem*` wildcard, because that seemed too dangerous to me with the flat layout we currently use. I think we'll move to the src layout at some point.

pyproject.toml
```diff
@@ -0,0 +1,71 @@
...
+
+[tool.setuptools]
+platforms = ["any"]
+script-files = [
+    "mof_compiler", "mof_compiler.bat"
+]
+zip-safe = true
+
+[tool.setuptools.packages.find]
+include = ["pywbem", "pywbem_mock"]
```
abravalheri, line 22: I suspect that you need the `*` here anyway, otherwise the "ambiguity" error is likely to happen

```suggestion
include = ["pywbem*"]
```

pyproject.toml
abravalheri: I think that you still need the card here https://github.com/pywbem/pywbem/pull/3174#discussion_r1587619278 to avoid errors.

pyproject.toml
andy-maier: Yes, the ambiguity error did happen with `include = ["pywbem", "pywbem_mock"]`.

I tried `include = ["pywbem.*", "pywbem_mock.*"]` and that also made the ambiguity error go away. I prefer that over `pywbem*`.

abravalheri: Still risk I am afraid 😅.

These patterns in setuptools are very literal... For example, `"pywbem"` will not match `pywbem._vendor.nocasedict`. The same way `"pywbem.*"` will not match `"pywbem"`.

You probably want to have a look on the `.whl` file generated with `include = ["pywbem.*", "pywbem_mock.*"]` to see if some files are not missing.

andy-maier: Good point with the literal matching. I am now using:
`include = ["pywbem", "pywbem.*", "pywbem_mock", "pywbem_mock.*"]`
and will check the resulting wheel file.

DEVELOP.md
```diff
@@ -313,6 +299,12 @@ local clone of the ``pywbem/pywbem`` Git repo.
 
         .. _`list of open issues`: https://github.com/pywbem/pywbem/issues
 
+4.  Add an initial Git tag for the new release stream and push it to the
+    remote repo:
+
+        git tag ${MNU}a0
```
KSchopmeyer, line 305: I would guess that the a0 above needs explanation.

Makefile
```diff
@@ -165,7 +165,11 @@ coverage_html_dir := coverage_html
 # Note: The package version is defined in pywbem/_version.py.
 # Note: Errors in getting the version (e.g. if wheel package is not installed) are
 # detected in _check_version. We avoid confusion by suppressing such errors here.
-package_version := $(shell $(PYTHON_CMD) setup.py --version 2>$(DEV_NULL))
+package_version := $(shell $(PYTHON_CMD) -m setuptools_scm 2>$(DEV_NULL))
```
KSchopmeyer, line 168: shouldn't line 165 "Note: The package version ... " be removed.

DEVELOP.md
andy-maier: Agreed. Added "This is necessary because the version is calculated based on the most recent tag in the commit history."

Makefile
andy-maier: I replaced the sentence with: "Note: The package version is automatically calculated by setuptools_scm based on the most recent tag in the commit history, increasing the least significant version indicator by 1."

andy-maier:
Rebased the PR on master, after merging PR #3154

andy-maier:
We discussed the cythonize support, and decided to merge this PR. If and when we get feedback from the announcement in the general channel, we can still add cythonize support via setup.py.

PR #3178, opened 2024-05-02
Excluded tests/schema subtree from source archive

andy-maier:


docs/changes.rst
```diff
@@ -47,6 +47,10 @@ Released: not yet
 * Added '.pytest_cache' and '.ruff_cache' directories to be removed in
   'make clean'.
 
+* Excluded 'tets/schema' from the source distribution archive. This is where
```
KSchopmeyer, line 50: change tets/schema to tests/schema

KSchopmeyer (commented):
Should probably have an entry in the incompatible changes list since it is a surprise if for some reason users were actually using that schema

one other comment

docs/changes.rst
andy-maier: DONE

andy-maier:
On your comment about incompatible changes: I clarified in the change log that it is not an incompatible change, because the source distribution archive on Pypi never had the schema files, only a locally created one had.

PR #3180, opened 2024-05-09
Upgraded vendored nocasedict/nocaselist to v2 to get rid of six

andy-maier:


docs/changes.rst
```diff
@@ -84,6 +84,11 @@ Released: not yet
   archives on Pypi never had that directory, but a locally produced
   source distribution archives had that directory. (issue #3177)
 
+* Increased the minimum versions of the vendored 'nocasedict' and 'nocaselist'
+  packages to 2.0.x in order to get rid of the dependency to the 'six' package.
```
KSchopmeyer, line 88: dependency on rather than dependency to

pywbem/_vendor/nocaselist/_nocaselist.py
```diff
@@ -247,7 +261,9 @@ def __mul__(self, number):
 
         Invoked using ``ncl * number``.
         """
-        if not isinstance(number, _INTEGER_TYPES):
+        # Despite using type hints, passing a non-int object does not raise
```
KSchopmeyer, line 264: Isn't this comment simply because the type hints are not processed in python but in tests tools? The comment makes it sound like a failure somewhere whereas it is a required test.

pywbem/_vendor/nocasedict/_nocasedict.py
```diff
@@ -716,11 +616,10 @@ def update(self, *args, **kwargs):
                         stacklevel=_stacklevel_above_nocasedict())
             except AttributeError:
                 # Expecting an iterable
-                try:
-                    # Try whether KeyableByMixin() was used
-                    key_attr = self.nocasedict_KeyableByMixin_key_attr
-                except AttributeError:
-                    key_attr = None
+
+                # Try whether KeyableByMixin() was used
```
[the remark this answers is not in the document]
andy-maier: It is just the new version that is now vendored. If you want to follow all the changes in nocaselist, you need to consult its commit/change history.

pywbem/_vendor/nocaselist/_nocaselist.py
andy-maier: Again, we just picked up a newer version

docs/changes.rst
andy-maier: DONE

PR #3181, opened 2024-05-10
Fix TODO-OLD-PYTHON flagged changes for python 3.8.

KSchopmeyer:
Fix a number of issues documented as TODO-OLD-PYTHON.

**DISCUSSION:** There is one TODO-PYTHON still in the code, the use of x key in dictionary in the CIMInt code and used in functiontests to supply Uint32 data to the InvokeMethod function tests.  I would like the change for that issue to be considered separately, I want to think that one out more but the others are ready for review.


andy-maier (approved):

andy-maier:
All the errors are safety issues, which were fixed in PR #3183 - merging

PR #3187, opened 2024-05-20
Modify documentation on cim types for CIMInt.

KSchopmeyer:
This pr proposed to keep the x parameter and document it rather than modify the tests.  This covers anybody who might have been using x in the past also.  As far as I can tell, the position-only argument appears to be driven generally by implementations in C.

Modify the documentation on the CIMInt type definition to publicly support both the positional and  keyword definition of the input argument 'x' that defines the value of the CIMInt.

This differs from the definition of int() where the definition of the value argument is only positional for Python >= 3.7.  This provides compatibility for pre python 3.7 apps that use the CIM integer types Uint*, Sint*.

One other minor fix where there were still several uses of long in the documentation in pywbem. These were removed since effective with Python 3.8, all of these types are now int.

### DIFF
--- a/pywbem/_cim_types.py
+++ b/pywbem/_cim_types.py
@@ -817,7 +817,10 @@ class CIMDateTime(_CIMComparisonMixin, CIMType):
 class CIMInt(CIMType, int):
     """
     Base type for CIM integer data types. Derived from :class:`~pywbem.CIMType`
-    and :class:`py3:int` (for Python 3) or :class:`py2:long` (for Python 2).
+    and :class:`py3:int`. The input parameters are consistent with
+    :class:`py3:int` except that argument `x` can also be passed as a keyword
+    input parameter to maintain a consistent interface despite changes to the
+    Python :class:`py3:int` ('x' was changed to positional only in python 3.8).
 
     This class has a concept of a valid range for the represented integer,
     based upon the capability of the CIM data type as defined in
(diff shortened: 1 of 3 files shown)


pywbem/_cim_types.py
```diff
@@ -817,26 +817,37 @@ def __hash__(self):
 class CIMInt(CIMType, int):
     """
     Base type for CIM integer data types. Derived from :class:`~pywbem.CIMType`
-    and :class:`py3:int` (for Python 3) or :class:`py2:long` (for Python 2).
+    and :class:`py3:int`. The input parameters consistent with :class:`py3:int`
+    except that `x` may be either a positional or keyword input parameter.
```
andy-maier, line 821: Since we now support Python>=3.8, int()  does not have the x parameter anymore, so we need to ontroduce it instead of referring to it.

How about changing this to:
```
except that the value can also be passed
as a keyword argument `x`.
```

pywbem/_cim_types.py
```diff
@@ -817,26 +817,37 @@ def __hash__(self):
...
+    :data:`~pywbem.config.ENFORCE_INTEGER_RANGE`.
 
     Two objects of subclasses of this base class compare equal if their numeric
     values compare equal. Objects of this class are immutable and
     :term:`hashable`, with the hash value being based on its numeric value.
 
+    Parameters:
+          x (:term:`integer` or :term:`string`):
+            A number or a string that can be converted into an integer number.
+            The parameter 'x' can be either a positional or keyword argument
```
andy-maier, line 841: Any parameter can be passed as a keyword arg or positional arg (except if the `*` is used to require passing as keyword arg, or the `/` is used to require passing as positional, both of which we do not use so far).

If we mention that explicitly, I think we need to give a hint as to why we do that, eg. by adding "for compatibility with Python<=3.7."

andy-maier (changes requested):

andy-maier:
I think if we need the passing as kwarg, then we should also document it officially. It also provides backwards compatibility.

pywbem/_cim_types.py
KSchopmeyer: Good catch

pywbem/_cim_types.py
KSchopmeyer: Or more like ‘to maintain the interface constantly despite the changes to the Python int() interface’.

andy-maier: That's better.

andy-maier (approved):

PR #3196, opened 2024-06-05
Migrated from pywbem_os_setup scripts to bindep

andy-maier:


docs/changes.rst
```diff
@@ -34,6 +34,8 @@ Released: not yet
   distribution of pywbem, because of the transition from setup.py to
   pyproject.toml.
 
+* Dev/Test: Removed the pywbem_os_setup.py/bat scripts that install binary
+  packages for development and test.
```
KSchopmeyer, line 38: add. Using binddep to installbinary packages.

andy-maier:
Discussion point is should we add this given that at the moment bindep is just a checker but not an installer?
Karl argues to leave it in.

PR #3203, opened 2024-07-09
Remove max version limit on build package.

KSchopmeyer:

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,11 +5,7 @@
 # that are needed for some reason (must be consistent with minimum-constraints-develop.txt)
 
 # Build distribution archive
-# Version 1.1.0 pulled after release
-# FUTURE: Remove  wheel < 1.0 For details, see
-#       https://github.com/pypa/build/issues/750. Circumvention for now is
-#       to pin build to <1.0. See issue #3112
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -743,7 +743,7 @@ $(bdist_file): pyproject.toml MANIFEST.in $(dist_dependent_files) $(moftab_files
 $(bdistc_file): setup.py pyproject.toml MANIFEST.in $(dist_dependent_files) $(moftab_files) $(done_dir)/vendor_$(pymn)_$(PACKAGE_LEVEL).done
 	@echo "Makefile: Creating the cythonized wheel distribution archive: $(bdistc_file)"
 	-$(call RM_FUNC,MANIFEST)
-	-$(call RMDIR_FUNC,build $(package_name).egg-info-INFO .eggs)
+	-$(call RMDIR_FUNC,build -vv $(package_name).egg-info-INFO .eggs)
```
andy-maier, line 746: I don't think the `-vv` should stay in.

andy-maier (changes requested):

Makefile
KSchopmeyer: (KS) Fixed

andy-maier (approved):

PR #3208, opened 2024-07-27
Add capability for listener to define a case where the received indication queue is full.

KSchopmeyer:
Defines a new exception and exception handling so that user can define a queue full argument on listener initialization and the listener will create the exception and pass it to main thread  if the received indication queue reaches this limit.  If this condition occurs, the whole listener is effectively stopped and must be recreated because we have stopped receiving indications from the wbem server

* Defines an argument on the indication listener init  (max_ind_queue_size) that will define the QUEUE_FULL condition if the number of indications in the listener  indication queue exceeds that limit. This is an optional exception and is only enabled when a listener init argument defines a integer that represents the maximum number of integers allowed in the queue.

* Adds an exception definition for indications received indication queue full and adds code to generate that exception and pass it back from the indication receive thread to the listener main thread.

Since causing this exception interrupts the indication flow (the queue will accept no more indications until the count is reduced, this normally means that the whole listener has failed except for callbacks to process indications in the queue. The indication that was to go into the queue is dropped, and the response to this indication is a code representing failure for this indication.

Adds tests for received indication full condition

Modified some of the function documentation.

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -21,6 +21,14 @@ The :class:`~pywbem.WBEMListener` class provides a thread-based WBEM listener
 service that can receive CIM indications from multiple WBEM servers and that
 calls registered callback functions to deliver the received indications.
 
+The listener receives indications from an indication sender, validates the
+formtting of the indication and queues each received indication into an
+interprocess queue. A seperate thread monitors this queue and calls the
+callback functions for each indication in the queue.
+
+This insures that timing in the callback functions does not interfere with the
+reception of indications from the indication sender.
+
 Examples
 --------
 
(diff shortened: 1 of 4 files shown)


pywbem/_listener.py
```diff
@@ -798,6 +890,20 @@ def __init__(self, host, http_port=None, https_port=None,
 
             `None` means not to use a private key file. Setting up a port
             for HTTPS requires specifying a private key file.
+
+          max_ind_queue_size (:term:`integer`):
+            A positive integer which defines the maximum number of the received
+            indications that can be in the received indication queue.  If the
+            queue of received indications reaches this size, the queue is
+            blocked and the QueueFull exception is executed. No more
```
andy-maier, line 898: Use the actual exception name with a link, instead of "QueueFull".

pywbem/_listener.py
```diff
@@ -990,34 +1109,46 @@ def logger(self):
         """
         return self._logger
 
-    def stop_callback_thread(self):
+    @property
+    def max_ind_queue_size(self):
+        """
+        Return the maximum queue size. If this value is not zero, an
```
andy-maier, line 1115: Add that it is an int.

andy-maier, line 1115: change "zero" -> "0"

pywbem/_listener.py
```diff
@@ -990,34 +1109,46 @@ def logger(self):
...
-        self.logger.info("Stopping queued delivery")
-        self.callback_thread.stop()
-        self.logger.info("Joining queued delivery")
-        self.callback_thread.join()
-        self.logger.info("Join finished threaded Deliver queue")
-        self.callback_thread = None
-        self.rcvd_indication_queue = None
+        return self.ind_delivery_queue.empty()
+
+    def qsize(self):
```
andy-maier, line 1128: Suggest to name this "queue_size()" since we use the full word "queue" on the other methods.

pywbem/_listener.py
```diff
@@ -1186,9 +1330,43 @@ def password_prompt():
...
     def stop(self):
         """
-        Stop the WBEM listener threads and callback thread, if they are running.
+        Stop the WBEM listeners, the WBEM listener threads and callback thread,
+        if they are running.
+        """
+        self.stop_servers()
+        self.stop_indication_delivery()
+
+    def stop_indication_delivery(self, force=False):
```
andy-maier, line 1339: When would a user ever stop just the servers or just the delivery?

andy-maier (changes requested):

pywbem/_listener.py
KSchopmeyer: Done

pywbem/_listener.py
KSchopmeyer: Done

pywbem/_listener.py
KSchopmeyer: Done

pywbem/_listener.py
KSchopmeyer: No, they are both stopped.  However, the indication delivery is a separate activity and may also be stopped if there is an exception in the the start of the connections since it was started before the connections are started.

pywbem/_listener.py
andy-maier: DONE

pywbem/_listener.py
andy-maier: ok

andy-maier (approved):

PR #3233, opened 2024-10-14
Upgraded Python 3.13 to final GA version

andy-maier:


KSchopmeyer (commented):
Should we not also update the README to include 3.13. Note that setup.py shows 3.12 as last version.

andy-maier:
On the Python versions badge: The versions are not defined in the README file:
```
[![Supported Python](https://img.shields.io/pypi/pyversions/pywbem.svg?color=brightgreen)](https://pypi.python.org/pypi/pywbem/)
```
shields.io gets them from Pypi, and I assume that shields.io interprets the Trove classifiers for Python in the package on Pypi. There, the version that supports 3.13 is not yet released. I assume that we will see 3.13 once pywbem with 3.13 support is released.

Issue #3237, opened 2024-10-28
pydantic fails to install on Python 3.13


andy-maier:
Reason being that safety-schemas 0.0.7 requires pydantic~=2.5.0 which pins it to <2.6.0. To fix the pydantic install issue on Pyhton 3.13, pydantic>=2.8.0 is needed.

Circumvention is to exclude safety-schemas 0.0.7.

See also https://github.com/pyupio/safety/issues/620

PR #3238, opened 2024-10-28
1.7: Fixed pydantic install issue on Python 3.13

andy-maier:
Rolls back PR #3237 into 1.7.3.


andy-maier:
Closing the PR because we don't add Pyhton 3.13 support to pywbem 1.7.x.

PR #3243, opened 2024-12-19
Migrated to use towncrier for change logs

andy-maier:
You can display the change log for the upcoming version 1.8.0 by running:
```
make develop
towncrier build --draft
```
with this branch checked out.


docs/development.rst
```diff
@@ -539,6 +539,99 @@ For further discussion of good and bad practices regarding commits, see:
 .. _How to Get Your Change Into the Linux Kernel: https://www.kernel.org/doc/Documentation/SubmittingPatches
 
 
+.. _`Making a change`:
+
+Making a change
```
KSchopmeyer, line 544: suggest that we change this to "Creating and submitting a change to pywbem" . This title was so short it was hard to understand.  Same for pywbemtools.

docs/development.rst
```diff
@@ -539,6 +539,99 @@ For further discussion of good and bad practices regarding commits, see:
...
+---------------
+
```
KSchopmeyer, line 546: Add the following:   All changes to pywbem are made through github with PRs created on topic branches and merged with the current master after successful group review.

docs/development.rst
```diff
@@ -539,6 +539,99 @@ For further discussion of good and bad practices regarding commits, see:
...
+You can review how your changes will show up in the final change log for
+the upcoming release by running:
+
+.. code-block:: sh
+    towncrier build --draft
+
+Always make sure that your pushed branch has either just one commit, or if you
+do multiple things, one commit for each logical change. What is not OK is to
+keep the possibly multiple commits it took you to get to the final result for
```
KSchopmeyer, line 631: Modify to change "possibly mulitiple" to  "submit for review a PR with multiple " ..

docs/development.rst
```diff
@@ -539,6 +539,99 @@ For further discussion of good and bad practices regarding commits, see:
...
+
+Making a change
+---------------
+
+To make a change, create a topic branch. You can assume that you are the only
+one using that branch, so force-pushes to that branch and rebasing that branch
+is fine.
+
+When you are ready to push your change, describe the change for users of the
+package in a change fragment file. To create a change fragment file, execute:
```
KSchopmeyer, line 552: should we define further wha change fragment file is an the link to towncrier.

KSchopmeyer (changes requested):
We have both DEVELOP.md and docs/development.rst.   Would it make sense to just point one at the other rather than duplicatng the descriptions.  I see where in the long run this will take more work but will be more consistent eventually and hopefully lead to fewer minor changes that have to be made during the push/review process. 

Response by Andy: The content is not duplicated between the two files. Having said that, it does make sense to integrate DEVELP.md into docs/development.rst (in a separate PR).

docs/development.rst
andy-maier: DONE as proposed.

docs/development.rst
andy-maier: DONE as proposed.

docs/development.rst
andy-maier: DONE. Added short explanation and link to towncrier.

andy-maier:
I resolved all the review comments.

PR #3259, opened 2025-03-02
Added support for generating AUTHORS.md

andy-maier:


.mailmap
```diff
@@ -0,0 +1,15 @@
...
+# For details, see: https://git-scm.com/docs/git-check-mailmap
+#
+# Format for replacing name and email based upon commit email:
+#    Proper Name <proper@email.xx> <commit@email.xx>
+#
+Andreas Maier <andreas.r.maier@gmx.de> <andreas.r.maier@gmx.de>
+Andreas Maier <andreas.r.maier@gmx.de> <andy-maier@users.noreply.github.com>
+Andreas Maier <andreas.r.maier@gmx.de> <maiera@de.ibm.com>
+Andreas Maier <andreas.r.maier@gmx.de> <MAIERA@de.ibm.com>
+Karl Schopmeyer <karl.schopmeyer@gmail.com> <k.schopmeyer@swbell.net>
```
KSchopmeyer, line 11: You can drop the k.schopeyer@swbell.net address.

KSchopmeyer (commented):
One comment, an old email address for me and you suibmit

.mailmap
andy-maier: That is the source address for the conversion to karl.schopmeyer@gmail.com, and it does appear in the git log.
The mailmap file defines which email addresses (the second one) should be translated to another email address (the first one). This causes multiple entries in the AUTHORS.md file to be converged into a single one.

KSchopmeyer: No real problem. Actually I still have that account but ATT has been trying for 5 years to get it to deliver mail so god only knows where any mail to that address goes.  And they won't close it with the outstanding issue.

KSchopmeyer:
Approved

PR #3269, opened 2025-03-12
Fixed testing of installed version of pywbem

andy-maier:


KSchopmeyer (commented):
Should there be a changes entry ?  Otherwise accepted

andy-maier:
Added a change log entry.

PR #3276, opened 2025-03-30
Fixed new flake8 issue

andy-maier:


andy-maier:
The error was caused by coveralls.io being in maintenance. Merged.

PR #3277, opened 2025-03-30
1.7: Fixed new flake8 issue

andy-maier:
Rolls back PR #3276 into 1.7.


andy-maier:
The error was caused by coveralls.io being in maintenance. Merged.

PR #3329, opened 2025-10-15
Send error response when indication queue is full

andy-maier:
The PR is ready for review.

I have tested the queue full behavior with the `examples/listen.py` script for running the listener, updated as per PR #3330, and with the `examples/send_indication.sh` script for sending the indications.


andy-maier:
The results of the meeting have been implemented.

PR #3339, opened 2025-10-30
Improved log messages in WBEMListener class

andy-maier:

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -581,7 +581,7 @@ class ListenerRequestHandler(BaseHTTPRequestHandler):
             try:
                 # pylint: disable=protected-access
                 listener._handle_indication(
-                    indication_inst, self.client_address[0])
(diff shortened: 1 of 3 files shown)


pywbem/_listener.py
```diff
@@ -1572,30 +1568,30 @@ def _handle_indication(self, indication, host):
...
-                "ignoring indication")
+            self.logger.error(
+                "Indication queue not set up - ignoring indication")
             return
 
         try:
             # Non-blocking put is required to raise queue.Full exception if full
-            self._ind_queue.put((indication, host), block=False)
+            queue_item = (indication, host, msgid)
+            self._ind_queue.put(queue_item, block=False)
```
KSchopmeyer: good place for one comment about this piece of logic on next line since the logic is sort of inverted.  I.e. Put log indicating that queue is no longer full.

PR #3340, opened 2025-10-30
Fixed support for IPv6 in the WBEMListener class

andy-maier:

### DIFF
--- a/pywbem/_listener.py
+++ b/pywbem/_listener.py
@@ -219,6 +219,7 @@ except ImportError:
     # termios is only supported on some operating systems
     termios = None
 import socketserver
+import socket
 import http.client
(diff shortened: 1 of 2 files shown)


pywbem/_listener.py
```diff
@@ -896,10 +921,10 @@ def __init__(self, host, http_port=None, https_port=None,
             Network wildcard addressing enables receiving indications from
```
KSchopmeyer: question: Does this work with MAC os

KSchopmeyer (commented):
I left one question.  I thought that you said that the wild card address does not work with MAC os.  If so, shouldn't we document that?  Otherwise approved

pywbem/_listener.py
andy-maier: yes, I verified it.

